Issue #16, opened 2017-07-20
Raise errors if there are no @outputs inside a @fold scope


obi1kenobi:
Not sure if this is already the case -- verify and fix if needed.

obi1kenobi:
Closed by #18.

PR #18, opened 2017-07-20
Raise an error when @fold contains no @outputs (#16) 

benlongo:
Covers #16. It seems that this already raises an error.

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -532,7 +532,7 @@ def _validate_fold_has_outputs(fold_data, outputs):
     for output in outputs.values():
         if output['fold'] is fold_data:
             return True
-    
+
     raise GraphQLCompilationError(u'Each @fold scope must contain at least one field '
                                   u'marked @output. Encountered a @fold with no outputs '
                                   u'at query location: {}'.format(fold_data['root']))
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -533,8 +533,9 @@ def _validate_fold_has_outputs(fold_data, outputs):
         if output['fold'] is fold_data:
             return True
     
-    raise GraphQLCompilationError(u'A @fold must contain at least one @output!'
-                                  u'Location: {}'.format(fold_data['root']))
+    raise GraphQLCompilationError(u'Each @fold scope must contain at least one field '
+                                  u'marked @output. Encountered a @fold with no outputs '
+                                  u'at query location: {}'.format(fold_data['root']))
 
 
 
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -538,7 +538,6 @@ def _validate_fold_has_outputs(fold_data, outputs):
                                   u'at query location: {}'.format(fold_data['root']))
 
 
-
 def _compile_fragment_ast(schema, current_schema_type, ast, location, context):
     """Return a list of basic blocks corresponding to the inline fragment at this AST node.
 
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -254,8 +254,7 @@ def _compile_property_ast(schema, current_schema_type, ast, location, context, l
                           directives present on the current AST node *only*
     """
     _validate_property_directives(local_directives)
-    fold = context.get('fold', None)
-    is_in_fold = fold is not None
+    is_in_fold = context.get('fold', None) is not None
 
     # step P-2: process property-only directives
     tag_directive = local_directives.get('tag', None)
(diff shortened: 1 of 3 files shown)


obi1kenobi:
Hi @benlongo, thank you for the PR!

Unfortunately, I think the raised error is not the correct one -- I think the fact that the *entire query* is missing `@output` fields is causing an error, and not the `@fold` scope specifically. I think you'll be able to confirm this with the following example:
```
{
    Animal {
        uuid @output(out_name: "uuid")
        out_Animal_ParentOf @fold {
            name
        }
    }
}
```

benlongo:
Hello @obi1kenobi 
Your test case does indeed work, my mistake. 
As a side note, is it desirable that the tests themselves are captured within the coverage report?

obi1kenobi:
If you'd like to add your example query to the `test_no_output_fields` test near line 697, I'd be happy to merge it -- since it's another nice example of a query that has no output fields.

I'd also love to help you address #16. The fix should be part of the front-end of the compiler (`compiler_frontend.py`), in two parts. I'd suggest that we modify `_compile_property_ast` to add the names of outputs to the fold scope (if in one). Then, `_compile_vertex_ast` on exit from a `@fold` scope (near [line 502 of that file](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/compiler_frontend.py#L502) ) could check that the fold scope being exited has a non-empty list of outputs that were selected in that scope. This should work since `@fold` cannot exist within a `@fold` scope.

Let me know what you think!

Regarding the coverage report, I didn't mean to include the tests. I just haven't gotten around to removing them yet -- feel free to open an issue / PR if you'd like.

benlongo:
Thanks for the direction! 
For storing the outputs in the fold context, should we modify `context['fold']` to contain something like `'outputs': []` and build that within `_compile_property_ast`? I see that `context['outputs'][output_name]` contains `'fold'` [(around 294) ](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/compiler_frontend.py#L294) so it could be possible to just look through that and determine if there exists a fold with no outputs. 

I'll make another PR to ignore the test files for the coverage report.

obi1kenobi:
Both approaches work: the first one is easier, but the second one is probably cleaner since it'll have no data denormalization (output fold data is only stored once) and may prevent bugs in the future. Let's go with that idea -- it shouldn't be too bad.

How about a function call on exit from the `@fold` scope near [line 502](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/compiler_frontend.py#L502), to a function similar to:
```
def _validate_fold_has_outputs(fold_data, outputs):
    # At least one output in the outputs list must point to the fold_data,
    # or the scope corresponding to fold_data had no @outputs and is illegal.
    for output in outputs:
        if output['fold'] is fold_data:
            return True
```

Alternatively, if you don't want to use `is`, you could compare the `location` key of the fold -- it should be unique for each fold and supports `==`.

What do you think?

benlongo:
I agree the second approach is cleaner, I had a go at it.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -525,6 +526,18 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
 
 
+def _validate_fold_has_outputs(fold_data, outputs):
+    # At least one output in the outputs list must point to the fold_data,
+    # or the scope corresponding to fold_data had no @outputs and is illegal.
+    for output in outputs.values():
+        if output['fold'] is fold_data:
+            return True
+    
+    raise GraphQLCompilationError(u'A @fold must contain at least one @output!'
```
obi1kenobi: Nitpick: I'd expand the explanation here, in order to help out users. Maybe something like:
```
Each @fold scope must contain at least one field marked @output. Encountered a @fold with no outputs at query location: {}
```

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -525,6 +526,19 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
+    for output in outputs.values():
+        if output['fold'] is fold_data:
+            return True
+    
+    raise GraphQLCompilationError(u'Each @fold scope must contain at least one field '
+                                  u'marked @output. Encountered a @fold with no outputs '
+                                  u'at query location: {}'.format(fold_data['root']))
+
+
+
```
obi1kenobi: Hmm it seems that I forgot to add `flake8` to the project -- my bad! Anyway, the 3 consecutive blank lines don't pass flake8:
```
$ flake8 ./graphql_compiler/
./graphql_compiler/compiler/compiler_frontend.py:541:1: E303 too many blank lines (3)
```

benlongo:
I didn't even bother to look for pep, figured the combination of pydocstyle graphql_compiler/

isort --check-only --verbose --recursive graphql_compiler/

pylint graphql_compiler/

bandit -r graphql_compiler/

PR #19, opened 2017-07-20
Omit the test files from the coverage report

benlongo:
Currently the test files are included in the coverage report, this tells `pytest-cov` to omit them via `.coveragerc`. It should drop coverage by around 3%.


.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+# We don't want to include the tests themselves in the coverage report
+omit = graphql_compiler/tests/*
```
obi1kenobi: Nitpick: can you add a new line at the end of the file, so my editor isn't sad when I open this file?

obi1kenobi:
Great -- thank you for the help!

As an aside, there's another coverage-related thing I've been meaning to do but haven't gotten around to it. If you look at the coverage report, there are many uncovered lines to the effect of `raise AssertionError("this should never happen")`. In fact, that's all AssertionError is used for.

If you're interested, I could use some help making the coverage ignore those lines, since they are not reachable anyway so they are pointless coverage misses.

PR #22, opened 2017-07-21
Remove assertion code from test coverage.

benlongo:
This removes test coverage for `__repr__` methods, `AssertionError` and `NotImplementedError`. More options for the config file can be found [here](https://coverage.readthedocs.io/en/latest/config.html). 

I added a couple tests for the output directive relating to string validation.


obi1kenobi:
Great stuff, thank you!

PR #23, opened 2017-07-24
Add a GraphQL pretty-printer.

obi1kenobi:
The default GraphQL formatted is a bit human-unfriendly:
- It uses 2-space indents, which makes it difficult to work with in a 4-space Python environment.
- Since directives have no required parameter order, `@filter` gets its parameters printed backwards. That makes the operator the last parameter, and means that `@filter` is in hard-to-read postfix notation.

This fixes both those complaints.

### DIFF
--- a/graphql_compiler/__init__.py
+++ b/graphql_compiler/__init__.py
@@ -3,6 +3,7 @@
 from .compiler import CompilationResult, OutputMetadata  # noqa
 from .compiler import compile_graphql_to_gremlin, compile_graphql_to_match  # noqa
 from .query_formatting import insert_arguments_into_query  # noqa
+from .query_formatting.graphql_formatting import pretty_print_graphql  # noqa
 
 from .exceptions import GraphQLError  # noqa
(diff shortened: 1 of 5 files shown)


graphql_compiler/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .compiler import CompilationResult, OutputMetadata  # noqa
 from .compiler import compile_graphql_to_gremlin, compile_graphql_to_match  # noqa
 from .query_formatting import insert_arguments_into_query  # noqa
+from .query_formatting.graphql_formatting import pretty_print_graphql  # noqa
```
ColCarroll: if you want to save yourself a file, you could `import .query_formatting.graphql_formatting as tool` here, and move the logic from `tool.py` into `__main__` there.  A little nasty, but does save a file!

ColCarroll:
I think you can make coveralls not block for tiny decreases -- it is hard to test main methods!

benlongo:
@ColCarroll we can tell the coverage to ignore `__main__` methods - should I put in a pr?

PR #24, opened 2017-07-25
Omit __main__ methods from test coverage.

benlongo:
This adresses @ColCarroll's comment in #23 about testing `__main__` methods.


ColCarroll:
This looks great!  Might add a note about what that line does if you've got a moment.  (for future reference, here's the docs on this line: http://coverage.readthedocs.io/en/latest/config.html#syntax )

PR #27, opened 2017-07-26
Print directive arguments in the order that they are defined. 

benlongo:
Currently we are using regexs to modify the order in which `@filter` arguments appear. This accomplishes the same goal but is more general by printing directive arguments in the order that they are defined within the schema. Some hackery was required to get the desired indent behavior by forcefully setting a function in the graphql-core printer module, but this also allows a variable number of spaces. 

Using `dedent` on the spacing tests is much more readable. 
Consequentially I had to change the order of the arguments in the test helper `get_schema()` to reflect the `OrderedDict`s.

### DIFF
--- a/graphql_compiler/query_formatting/graphql_formatting.py
+++ b/graphql_compiler/query_formatting/graphql_formatting.py
@@ -17,14 +17,13 @@ def pretty_print_graphql(query, spaces=4):
     # opposed to copying the entire class.
     import graphql
 
-    def _indent_override(spaces):
-        def indent(maybe_str):
-            if maybe_str:
-                return maybe_str.replace('\n', '\n' + ' ' * spaces)
-            return maybe_str
-        return indent
-    graphql.language.printer.indent = _indent_override(spaces)
-    
+    def indent_override(maybe_str):
+        if maybe_str:
+            return maybe_str.replace('\n', '\n' + ' ' * spaces)
+        return maybe_str
+
+    graphql.language.printer.indent = indent_override
+
     output = visit(parse(query), CustomPrintingVisitor())
     return output
 
--- a/graphql_compiler/query_formatting/graphql_formatting.py
+++ b/graphql_compiler/query_formatting/graphql_formatting.py
@@ -1,5 +1,4 @@
 # Copyright 2017 Kensho Technologies, Inc.
-import graphql
 from graphql import parse
 from graphql.language.printer import PrintingVisitor, wrap, join
 from graphql.language.visitor import visit
@@ -8,53 +7,55 @@ from ..schema import DIRECTIVES
 
 def pretty_print_graphql(query, spaces=4):
     """Take a GraphQL query, pretty print it, and return it."""
-    # Use four spaces for indentation, to make it easier up to edit in Python
-    # source files.
-    old_indent = graphql.language.printer.indent
+    # The graphql-core printer module has an indent function outside of the
+    # PrintingVisitor class which has two spaces hard coded. Ideally, it would
+    # be a member of the class and we could simply override it in our custom
+    # printer. To work around this we can set the function inside the module
+    # forcefully. This is undesireable as it has a global effect and relies on
+    # the implementation details of the print_ast() function, but has the
+    # advantage of only modifying the leave_Directive() method on the class as
+    # opposed to copying the entire class.
+    import graphql
+
+    def _indent_override(spaces):
+        def indent(maybe_str):
+            if maybe_str:
+                return maybe_str.replace('\n', '\n' + ' ' * spaces)
+            return maybe_str
+        return indent
     graphql.language.printer.indent = _indent_override(spaces)
-
-    output = visit(parse(query), CustomPrinter(spaces))
-
-    graphql.language.printer.indent = old_indent
-
+    
+    output = visit(parse(query), CustomPrintingVisitor())
     return output
 
 
-def _indent_override(spaces):
-    def indent(maybe_str):
-        if maybe_str:
-            return maybe_str.replace('\n', '\n' + ' ' * spaces)
-        return maybe_str
-    return indent
-
-
-class CustomPrinter(PrintingVisitor):
-    def __init__(self, spaces=2):
-        self._spaces = spaces
-
-    # @filter directives are much easier to read if the operation comes before
-    # the values. The default AST printer outputs the values
-    # before the operation, so the filtering operation ends up written in
-    # postfix order.
+class CustomPrintingVisitor(PrintingVisitor):
+    # Directives are easier to read if they are in the order in which we defined
+    # them in the schema. For example, @filter directives are much easier to
+    # read if the operation comes before the values. The arguments of the
+    # directives specified in the schema are defined as OrderedDicts which
+    # allows us to sort the provided arguments to match.
     def leave_Directive(self, node, *args):
         args = node.arguments
 
-        def _arg_order(arg, directive):
+        def _arg_index(arg, directive):
             defined_arg_names = directive.args.keys()
             # Taking [0] is ok here because the graphql parser checks for the
             # existence of ':' in directive arguments
-            arg_name = arg.split(':')[0]
+            arg_name = arg.split(':', 1)[0]
             try:
                 return defined_arg_names.index(arg_name)
             except ValueError:
                 # This argument name isn't defined in our schema but we should
-                # still pretty print it
+                # still pretty print it anyway
                 return -1
 
         try:
             directive = next(d for d in DIRECTIVES if d.name == node.name)
-            args = list(sorted(args, key=lambda a: _arg_order(a, directive)))
+            args = list(sorted(args, key=lambda a: _arg_index(a, directive)))
         except StopIteration:
+            # The directive wasn't defined in our schema, use whatever order
+            # the args appeared in within the query
             pass
 
         return '@' + node.name + wrap('(', join(args, ', '), ')')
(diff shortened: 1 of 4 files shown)


graphql_compiler/tests/test_helpers.py
```diff
@@ -89,7 +89,7 @@ def get_schema():
 
         directive @recurse(depth: Int!) on FIELD
 
-        directive @filter(value: [String!]!, op_name: String!) on FIELD | INLINE_FRAGMENT
+        directive @filter(op_name: String!, value: [String!]!) on FIELD | INLINE_FRAGMENT
```
obi1kenobi: good catch

graphql_compiler/tests/test_graphql_pretty_print.py
```diff
@@ -12,47 +13,133 @@ def test_graphql_pretty_print_indentation(self):
...
             }
         }'''
 
-        expected_output = '''{
-    Animal @filter(op_name: "name_or_alias", value: ["$name"]) {
-        uuid @filter(op_name: "<=", value: ["$max_uuid"])
-        out_Entity_Related {
-            ... on Species {
-                name @output(out_name: "related_species")
+        expected_output = dedent('''\
```
obi1kenobi: I didn't know about `dedent` -- very cool!

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -1,63 +1,60 @@
...
+    # the implementation details of the print_ast() function, but has the
+    # advantage of only modifying the leave_Directive() method on the class as
+    # opposed to copying the entire class.
+    import graphql
+
+    def indent_override(maybe_str):
+        if maybe_str:
+            return maybe_str.replace('\n', '\n' + ' ' * spaces)
+        return maybe_str
+    graphql.language.printer.indent = indent_override
```
obi1kenobi: Not a fan of this. As you point out, if the underlying GraphQL library changes, this leaves us with entirely non-functional code. This risk doesn't seem to be justified based on what we're getting in return.

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -1,63 +1,60 @@
...
+        args = node.arguments
+
+        def _arg_index(arg, directive):
+            defined_arg_names = directive.args.keys()
+            # Taking [0] is ok here because the graphql parser checks for the
+            # existence of ':' in directive arguments
+            arg_name = arg.split(':', 1)[0]
+            try:
+                return defined_arg_names.index(arg_name)
+            except ValueError:
```
obi1kenobi: I prefer explicit checks rather than the cause-an-exception-and-catch-it approach.

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -1,63 +1,60 @@
...
+                # This argument name isn't defined in our schema but we should
+                # still pretty print it anyway
+                return -1
+
+        try:
+            directive = next(d for d in DIRECTIVES if d.name == node.name)
+            args = list(sorted(args, key=lambda a: _arg_index(a, directive)))
```
obi1kenobi: The python docs say that `When iterating over an ordered dictionary, the items are returned in the order their keys were first added.` It seems that you should be able to avoid the custom key function and associated complexity, and instead:
- for each argument that the directive supports
- check if the argument is there
- if so, add it to the list

Importantly, since you're going in the directive's order, everything comes out ordered on the other side, and you avoid any exceptions being raised.

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -1,63 +1,60 @@
...
+            arg_name = arg.split(':', 1)[0]
+            try:
+                return defined_arg_names.index(arg_name)
+            except ValueError:
+                # This argument name isn't defined in our schema but we should
+                # still pretty print it anyway
+                return -1
+
+        try:
+            directive = next(d for d in DIRECTIVES if d.name == node.name)
```
obi1kenobi: Here as well, I'm not a fan of causing an exception. 

Why not create a dict of directive name to directive object, and then check if the current directive name exists in that map? It seems to me that approach communicates the intent a bit better, and is easier to read.

graphql_compiler/query_formatting/graphql_formatting.py
benlongo: Agreed, that is much cleaner. Thanks!

graphql_compiler/query_formatting/graphql_formatting.py
benlongo: I am not a fan either, I'll revert back to `fix_indentation_depth`.

graphql_compiler/query_formatting/graphql_formatting.py
benlongo: The map is certainly much cleaner, good idea. Where should it live? For now I've left it as `DIRECTIVES_BY_NAME` in the formatting file. Alternatively, we could put it in `schema.py` or inside the `CustomPrintingVisitor` (which could also use a better name).

graphql_compiler/query_formatting/graphql_formatting.py
benlongo: I've always felt conflicted about this myself, but when writing python have aimed to follow the 'easier to ask for forgiveness than permission' guideline. I agree that in this case and many others it reduces the clarity of semantics and intent.

benlongo:
Thanks for the thoughtful review!
I've reverted back to using `fix_indentation_depth` and taken a first pass at cleaning up `leave_Directive`.

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -21,43 +57,7 @@ def fix_indentation_depth(query):
...
-    )
-
-    replacement_pattern = (
-        r'@filter(op_name: \2, value: [\1])'
-    )
-
-    return re.sub(filter_directive_pattern, replacement_pattern, query)
-
-
-def pretty_print_graphql(query, use_four_spaces=True):
```
obi1kenobi: Why not use and amend this existing function, since it already has detailed comments?

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -21,43 +57,7 @@ def fix_indentation_depth(query):
             raise AssertionError(u'Indentation was not a multiple of two: '
                                  u'{}'.format(consecutive_spaces))
 
-        final_lines.append(('  ' * consecutive_spaces) + line[consecutive_spaces:])
```
obi1kenobi: This whitespace change seems unnecessary.

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -1,7 +1,43 @@
...
+class CustomPrintingVisitor(PrintingVisitor):
+    # Directives are easier to read if their arguments appear in the order in
+    # which we defined them in the schema. For example, @filter directives are
+    # much easier to read if the operation comes before the values. The
+    # arguments of the directives specified in the schema are defined as
+    # OrderedDicts which allows us to sort the provided arguments to match.
+    def leave_Directive(self, node, *args):
+        """Call when exiting a directive node in the ast."""
+        args = node.arguments
+        arg_names = [a.split(':', 1)[0] for a in args]
```
obi1kenobi: Can you put back the comment you had about why `[0]` is okay to call here?

graphql_compiler/query_formatting/graphql_formatting.py
```diff
@@ -1,7 +1,43 @@
...
+
+        directive = DIRECTIVES_BY_NAME.get(node.name)
+        if directive:
+            sorted_args = []
+            for defined_arg in directive.args.keys():
+                if defined_arg in arg_names:
+                    arg = args.pop(arg_names.index(defined_arg))
```
obi1kenobi: `args` comes from `node.arguments` and is being mutated here. It's generally a bad idea to mutate a function's arguments. How about either making a copy, or simply using a `set` to remember which arguments were already processed?

graphql_compiler/tests/test_graphql_pretty_print.py
```diff
@@ -12,47 +13,133 @@ def test_graphql_pretty_print_indentation(self):
...
+                        name @output(out_name: "related_species")
+                    }
+                }
+            }
+        }
+        ''')
+
+        self.assertEquals(expected_output, pretty_print_graphql(bad_query))
+
+    def test_args_not_in_schema(self):
```
obi1kenobi: great test!

graphql_compiler/tests/test_graphql_pretty_print.py
```diff
@@ -12,47 +13,133 @@ def test_graphql_pretty_print_indentation(self):
...
+        bad_query = '''{
+            Animal @filter(value: ["$name"], unkown_arg: "value", op_name: "name_or_alias") {
```
obi1kenobi: typo: `unkown`

benlongo:
I made a copy of `node.arguments` and added back all of the comments. Thanks again!

obi1kenobi:
Once again, thank you for your excellent contribution! It's people like you that make open-source projects successful :)

In case you're interested in exploring more of the codebase and making some more pull requests, I've marked a few issues with a "contributor-friendly" label. They should be reasonably self-explanatory and not too difficult to implement -- please let me know if you're interested or have any questions.

Issue #9, opened 2017-07-19
Verify that passed GraphQL schema has all the necessary types and directives


obi1kenobi:
When compiling, we accept a GraphQL schema parameter. We should validate that the schema passed this way includes all the required scalar types and has all the required directives.

PR #29, opened 2017-08-01
Verify that the schema contains necessary directives (#9). 

benlongo:
`graphql-core` already validates that the schema contains the directives used in the query via the [`known_directives`](https://github.com/graphql-python/graphql-core/blob/master/graphql/validation/rules/known_directives.py) validation rule. I've added a test to verify this behavior, but I may have misinterpreted the issue (#9): do we wish to verify that the schema contains *all* the directives, or solely the ones used within the query?

### DIFF
--- a/graphql_compiler/tests/test_ir_generation_errors.py
+++ b/graphql_compiler/tests/test_ir_generation_errors.py
@@ -2,6 +2,9 @@
 import string
 import unittest
 
+from graphql import parse
+from graphql.utils.build_ast_schema import build_ast_schema
+
 from ..compiler.compiler_frontend import graphql_to_ir
 from ..exceptions import GraphQLCompilationError, GraphQLParsingError, GraphQLValidationError
 from .test_helpers import get_schema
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Hey @benlongo -- thanks for grabbing this issue, and apologies that I didn't write it more clearly.

If a user is using the compiler to compile GraphQL queries, we expect that their schema contains all directives the compiler needs to do its job, regardless of whether a particular query uses them or not. In addition to GraphQL's existing "any directive used in a query must have been present in the schema" rule, we want a rule that says "the directives present in the schema are exactly the directives the compiler wants".

So that would mean that the `incomplete_schema` in your test should raise errors for any passed query, since the schema doesn't contain the `@output` directive and therefore the user is using the compiler incorrectly.

Does that make sense?

benlongo:
Yeah that makes sense. I suspected this was the desired behavior but didn't want to proceed further until it was confirmed. I'll take a first pass at it later today, thanks!

obi1kenobi:
Thank you!

obi1kenobi:
Hey @benlongo, are you still working on this PR or would you like me to take it over?

benlongo:
Hey @obi1kenobi,
Sorry I have not had the time to finish this up. There is a slight annoyance in determining if the directives defined in the schema are a superset of the ones we require because their equality methods check the descriptions. There is likely an elegant workaround that doesn't dig into the graphql internals. 
I will probably be unable to work on this for the next couple of weeks so you are welcome to take it over.

obi1kenobi:
@jssandh2 feel free to take this over.

PR #30, opened 2017-08-15
Fix a minor bug in the GraphQL formatter.

obi1kenobi:
Here's the error I was getting.
```
File /.../graphql_compiler/query_formatting/graphql_formatting.py", in leave_Directive
    arg = args.pop(arg_names.index(defined_arg))
IndexError: pop index out of range
```

I wasn't able to come up with a deterministic test case that reproduces the problem, but I know the underlying cause. The problem happens because the `args` and `arg_names` lists were expected to be parallel, but popping could happen in any order. The error above is caused when the lists cease to be parallel, e.g. as follows:

```
********* before first pop, popping at index 0

['op_name: "="', 'value: ["$parameter"]']
['op_name', 'value']

********* before second pop, popping at index 1

['value: ["$parameter"]']
['op_name', 'value']

********* IndexError since there was no index
```

### DIFF
--- a/graphql_compiler/query_formatting/graphql_formatting.py
+++ b/graphql_compiler/query_formatting/graphql_formatting.py
@@ -30,22 +30,37 @@ class CustomPrintingVisitor(PrintingVisitor):
     # OrderedDicts which allows us to sort the provided arguments to match.
     def leave_Directive(self, node, *args):
         """Call when exiting a directive node in the ast."""
-        # Make a copy of the arguemnts so that we can safely pop
-        args = list(node.arguments)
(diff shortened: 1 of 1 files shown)


obi1kenobi:
@benlongo if you have a moment, since I'm modifying your code, I'd appreciate your feedback as well. Also, thank you for the thorough unit tests you wrote -- they were very helpful in making sure I didn't add more bugs by accident :)

benlongo:
What a curious little bug - the changes look good, I probably should've gone with the set approach to begin with 👍

obi1kenobi:
Hindsight is 20/20 after all :) Thanks for reviewing so quickly!

PR #31, opened 2017-09-18
Make python3 compatible

ColCarroll:
This should lead the compiler bravely into the future, while cowardly and responsibly keeping one foot in the past.

If you were going to give a hard reading to a few parts:

-- I did *not* update any of the very helpful docstrings or error messages about expecting `basestring`/`unicode`.  If you have some sample ways you'd like me to update that, I'm happy to!
-- Python2 vs Python3 bytes are a little funny in a way that isn't handled nicely by `six`.  In those cases I used the pattern:
```
if not isinstance(value, six.string_types):
    if isinstance(value, bytes):  # should only happen in py3
        value = value.decode('utf-8')
    else:
        # existing logic
```

### DIFF
--- a/graphql_compiler/compiler/helpers.py
+++ b/graphql_compiler/compiler/helpers.py
@@ -57,9 +57,9 @@ def is_graphql_type(graphql_type):
 
 
 def ensure_unicode_string(value):
-    """Ensure the value is a basestring, and return it as unicode."""
+    """Ensure the value is a string, and return it as unicode."""
     if not isinstance(value, six.string_types):
-        raise TypeError(u'Expected basestring value, got: {}'.format(value))
+        raise TypeError(u'Expected string value, got: {}'.format(value))
     return six.text_type(value)
 
 
@@ -99,7 +99,7 @@ def validate_safe_string(value):
     legal_strings_with_special_chars = frozenset({'@rid', '@class', '@this', '%'})
 
     if not isinstance(value, six.string_types):
-        raise TypeError(u'Expected basestring value, got: {} {}'.format(
+        raise TypeError(u'Expected string value, got: {} {}'.format(
             type(value).__name__, value))
 
     if not value:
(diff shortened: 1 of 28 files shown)


graphql_compiler/compiler/helpers.py
```diff
@@ -196,16 +198,12 @@ def revisit(self):
...
         return (mark_name, self.field)
 
-    def __unicode__(self):
-        """Return a human-readable unicode representation of the Location object."""
-        return u'Location({}, {}, {})'.format(self.query_path, self.field, self.visit_counter)
-
     def __str__(self):
         """Return a human-readable str representation of the Location object."""
-        return self.__unicode__().encode('utf-8')
+        return u'Location({}, {}, {})'.format(self.query_path, self.field, self.visit_counter)
```
obi1kenobi: Isn't this technically incorrect for py2? In py2, `__str__` should return bytes and not unicode, I think.

obi1kenobi: Or does the magic `python_2_unicode_compatible` solve this?

ColCarroll: Yah, magic solution.  I did not know this was a problem until I saw you take care of it, and was surprised to find there's a nice solution:

> A class decorator that takes a class defining a __str__ method. On Python 3, the decorator does nothing. On Python 2, it aliases the __str__ method to __unicode__ and creates a new __str__ method that returns the result of __unicode__() encoded with UTF-8.

https://pythonhosted.org/six/#six.python_2_unicode_compatible

ColCarroll:
Yeah -- it is a 10kb install, and I think I only used 
- `six.iter*`, 
- `six.renames.xrange`, 
- `six.text_type`, 
- `six.string_types`, 
- `six.python_2_unicode_compatible`, 
- `six.add_metaclass`

I'd vote to leave `six` in, since dropping py2 support in the next <2yrs seems reasonable, and it will make for an easy `grep`!

obi1kenobi:
Yeah, let's leave `six` in -- it makes no sense to reimplement that many things. If it was just `iteritems()` vs `items()` I would have suggested to just use the `items()` version and ignore the difference, but this is too many things :)

What do you think about the docstrings?

ColCarroll:
Got too excited about finding out how big `six` was that I forgot about docstrings.  What about "string"?

ColCarroll:
(that is what, for example, `gensim` uses, sometimes specifying differences in handling between 2 and 3.  See https://radimrehurek.com/gensim/corpora/textcorpus.html)

PR #34, opened 2017-09-22
Refactor and move helper functions out of main files.

obi1kenobi:
Should make the code a bit easier to read and less cluttered.

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -64,6 +64,11 @@ import six
 
 from . import blocks, expressions
 from ..exceptions import GraphQLCompilationError, GraphQLParsingError, GraphQLValidationError
+from .context_helpers import (has_encountered_output_source, is_in_fold_scope, is_in_optional_scope,
+                              validate_context_for_visiting_vertex_field)
+from .directive_helpers import (get_directives, validate_property_directives,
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -299,12 +245,12 @@ def _compile_property_ast(schema, current_schema_type, ast, location, context, l
         validate_safe_string(output_name)
 
         graphql_type = strip_non_null_from_type(current_schema_type)
-        if is_in_fold:
+        if is_in_fold_scope(context):
```
ColCarroll: just pointing out that this is _slightly_ different from the original logic

obi1kenobi: Good catch. I did this on purpose since this is technically more correct, the original code was a bit of a premature optimization that I was worried would come back to bite at some point in the future.

ColCarroll: yah, in general this is all great -- seems like a good amount of abstraction, but not too much abstraction.  Just don't like, make a huge database request to check `is_in_fold_scope` :D

obi1kenobi: That'd be ironic, checking the database 100 times so that it can create the perfect query that avoids checking the database 100 times... :)

PR #35, opened 2017-09-25
Add Fold/Unfold blocks to IR. Refactor IR sanity checks.

obi1kenobi:
I'm working on adding support for type coercions and `@filter` inside vertex fields marked `@fold`. I made the `filtering_in_fold_scopes` feature branch for that feature, but adding the full feature will take quite a lot of work:
- we'll need new `Fold` and `Unfold` basic blocks
- we'll need to make the compiler front-end emit such blocks
- we'll need to change how Gremlin and MATCH code is emitted to account for the new representation of `@fold`

To make the review easier, I'm splitting up the work into smaller pieces. This PR does the first of the three tasks above, as well as some cleanup that will make future work easier. 

I set the base branch for the PR to `filtering_in_fold_scopes` since I don't want to merge a half-finished feature into `master`.


graphql_compiler/compiler/blocks.py
```diff
@@ -403,22 +420,30 @@ class OutputSource(BasicBlock):
...
+        super(Fold, self).__init__()
+        self.root_location = root_location
         self.validate()
 
     def validate(self):
-        """Ensure that the OutputSource block is valid.
+        """Ensure the Fold block is valid."""
+        if not isinstance(self.root_location, Location):
+            raise ValueError(u'Expected a Location for root_location, got: '
+                             u'{}'.format(self.root_location))
```
ColCarroll: is this supposed to be `type(self.root_location)`?

graphql_compiler/compiler/blocks.py
```diff
@@ -403,22 +420,30 @@ class OutputSource(BasicBlock):
     See the comment on the @output_source directive in schema.py on why this is necessary.
     """
 
-    def __init__(self):
-        """Create a new OutputSource block."""
-        super(OutputSource, self).__init__()
+    def validate():
```
ColCarroll: still needs a `self`

graphql_compiler/compiler/blocks.py
```diff
@@ -403,22 +420,30 @@ class OutputSource(BasicBlock):
...
-    def to_gremlin(self):
-        """Return the unicode representation of this BasicBlock.
+class Unfold(MarkerBlock):
+    """A marker for the end of a @fold context."""
 
-        The correct Gremlin representation of OutputSource blocks is an empty string.
-        Their effect is applied during code generation and optimization passes.
-        """
-        return u''
+    def validate():
```
ColCarroll: still needs a `self`

graphql_compiler/compiler/context_helpers.py
```diff
@@ -1,3 +1,4 @@
+# Copyright 2017 Kensho Technologies, Inc.
```
ColCarroll: i think you could add a travis check for this -- some bash 1-liner with `find . -iname='*.py'` and `head -n1` ?

obi1kenobi:
I promise to cover the new lines in the next PR, since this one doesn't use them.

graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -0,0 +1,162 @@
+# Copyright 2017 Kensho Technologies, Inc.
+"""Functions that ensure the IR generated by the front-end satisfies all invariants."""
```
ColCarroll: general suggestion to maybe add 

```
def check_assertion(condition, error_message):
    if not condition:
        raise AssertionError(error_message)
```

though I see you have a lot of corner cases, and this is pretty readable, so definitely just a suggestion!

obi1kenobi: Yeah I've thought about it. I think I'll hold off for now, because I'm not a huge fan of passing complex expressions to functions:
```
check_assertion(this is that or 'foo' in my_dict, 'expectations were not confirmed')
```

graphql_compiler/compiler/blocks.py
obi1kenobi: I'll add the type as well, but it's nice to know the concrete value for the sake of debugging.

graphql_compiler/compiler/context_helpers.py
obi1kenobi: Good idea, I'll try to do it in a separate PR.

jssandh2:
Tests are failing. But, discussed with @obi1kenobi offline. This, LGTM. And, we can take a look at the new tests once this is in.

obi1kenobi:
Not tests, just coverage -- which makes sense since the new IR blocks aren't used yet.

jssandh2:
Also - I don't have write privileges, by the looks of it.

PR #36, opened 2017-09-26
Fold/Unfold blocks emitted by the front-end and processed by the back-end

obi1kenobi:
- Move all the base classes for compiler entities into their own file.
- Add a new `FoldScopeLocation` object that uniquely defines where a `@fold` scope is relative to the rest of the query.
- Use the `FoldScopeLocation` to slice out the IR blocks corresponding to the contents of the `@fold` scope. For now, those contents are only the Fold/Unfold blocks, but this is where filtering and type coercions will go.
- Lots of minor refactoring and cleanup.

Left to do:
- Allow filtering and type coercions inside `@fold` scopes.
- Handle back-end query emission appropriately given that outputs may be in a `@fold` scope, based on their `FoldScopeLocation`.
- Lots of tests for filtering and type coercion inside `@fold` scopes.

### DIFF
--- a/graphql_compiler/compiler/match_query.py
+++ b/graphql_compiler/compiler/match_query.py
@@ -3,8 +3,8 @@
 
 from collections import namedtuple
 
-from .blocks import (Backtrack, CoerceType, ConstructResult, Filter, MarkLocation, OutputSource,
-                     QueryRoot, Recurse, Traverse)
+from .blocks import (Backtrack, CoerceType, ConstructResult, Filter, Fold, MarkLocation,
+                     OutputSource, QueryRoot, Recurse, Traverse, Unfold)
 
 
 ###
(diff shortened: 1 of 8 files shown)


graphql_compiler/compiler/match_query.py
```diff
@@ -132,6 +134,49 @@ def _split_match_steps_into_match_traversals(match_steps):
...
+            current_folded_blocks.append(block)
+        else:
+            remaining_ir_blocks.append(block)
+
+        if isinstance(block, Unfold):
+            in_fold = False
+            first_folded_block = current_folded_blocks[0]
+            last_folded_block = current_folded_blocks[-1]
+
+            if not isinstance(first_folded_block, Fold):
```
jssandh2: Shouldn't this not happen because of the way the scoping for the fold works ?

obi1kenobi: Indeed it should not. Whenever you see `raise AssertionError`, that's my way of saying "this should never ever happen."

obi1kenobi:
@lodrion is `base_location` better compared to `root_location`?

lodrion:
Yes, base_location is probably better. fold_root_location is good as well.
On the other hand my opinion may be influenced by the fact that I've seen
the other "root" mentioned in this diff somewhere (probably because of hte
bulk moves)

On Tue, Sep 26, 2017 at 10:46 PM, Predrag Gruevski <notifications@github.com

obi1kenobi:
Ok I'll rename it to `base_location` before I merge tomorrow.

PR #37, opened 2017-09-27
Allow "name_or_alias" filtering operation on interface types.

obi1kenobi:
The compiler was unnecessarily strict with checking the validity of `name_or_alias` targets, and ended up disallowing its use on interface types that contained the necessary fields. This PR relaxes that restriction and updates the schema and tests to reflect that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Current development version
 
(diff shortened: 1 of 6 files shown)


jssandh2:
Tests pass. LGTM.

PR #38, opened 2017-09-28
Finished back-end implementation of the new `@fold` representation.

obi1kenobi:
Also fixes an unfortunate bug with `@filter` directives -- if more than two were present at the same location, some of the filters would get dropped.

I added lots of test cases for the bug and for the new functionality.

The final remaining thing is updating the documentation to state that filtering and type coercions are now allowed within `@fold` scopes.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Current development version
 
+- ** BREAKING **: Requires OrientDB 2.2.28+, since it depends on two OrientDB bugs being fixed: [bug 1](https://github.com/orientechnologies/orientdb/issues/7225) [bug 2](https://github.com/orientechnologies/orientdb/issues/7754)
```
jssandh2: Do you mean : `**BREAKING**` ? (If you want the italics, that is).

graphql_compiler/compiler/emit_match.py
```diff
@@ -127,18 +155,32 @@ def emit_code_from_ir(match_query):
...
+    # Represent and add the LET clauses for any @fold scopes that might be part of the query.
+    fold_data = [
+        _represent_fold(fold_location, fold_ir_blocks)
+        for fold_location, fold_ir_blocks in six.iteritems(match_query.folds)
+    ]
+    if fold_data:
+        query_data.append(u' LET ')
+        query_data.append(fold_data[0])
+        for fold_clause in fold_data[1:]:
+            query_data.append(u', ')
```
jssandh2: Wait - Why are we appending this to every string (except for the first one) in the folded blocks ?

graphql_compiler/compiler/expressions.py
```diff
@@ -204,6 +200,10 @@ def __init__(self, field_name):
         self.field_name = field_name
         self.validate()
 
+    def get_local_object_gremlin_name(self):
```
jssandh2: Can this please be made a constant ? It seems static.

graphql_compiler/compiler/helpers.py
```diff
@@ -271,6 +275,12 @@ def __init__(self, base_location, relative_position):
         self.base_location = base_location
         self.relative_position = relative_position
 
+    def get_location_name(self):
+        """Return a unique name for the FoldScopeLocation."""
+        edge_direction, edge_name = self.relative_position
+        return (self.base_location.get_location_name()[0] + u'___' +
```
jssandh2: Let's make this string concatenation a little safer some time in the future. Perhaps we can have a helper method to concatenate strings, and we can _hardcode_ unicode strings like : `u'___'` and `u'__'`, since it's easy to make mistakes here.

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -1,24 +1,26 @@
...
+from .blocks import ConstructResult, Filter, Fold, Unfold
 from .expressions import (BinaryComposition, ContextField, ContextFieldExistence, FalseLiteral,
                           NullLiteral, TrueLiteral)
 from .helpers import validate_safe_string
 
 
 def merge_consecutive_filter_clauses(ir_blocks):
     """Merge consecutive Filter(x), Filter(y) blocks into Filter(x && y) block."""
+    if not ir_blocks:
+        return ir_blocks
```
jssandh2: This seems interesting. This _won't_ just explicitly return `None`, but rather `[]` _or_ `None` (depending on whether a non-existent `ir_blocks` array is even possible).

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -1,24 +1,26 @@
...
+
     new_ir_blocks = [ir_blocks[0]]
 
-    for previous_block, current_block in pairwise(ir_blocks):
-        if isinstance(previous_block, Filter) and isinstance(current_block, Filter):
+    for block in ir_blocks[1:]:
+        last_block = new_ir_blocks[-1]
+        if isinstance(last_block, Filter) and isinstance(block, Filter):
             new_ir_blocks[-1] = Filter(
```
jssandh2: I'm sure you're aware, but Python has shallow copies, and if `Filter(..)` here changes the pointer that `new_ir_blocks[-1]` points to, then we lose whatever there is. Are we sure we're overwriting here ?

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -159,3 +161,46 @@ def visitor_fn(expression):
...
+        - remaining_ir_blocks: list of IR blocks that were not part of a Fold-Unfold section.
+    """
+    folds = dict()
+    remaining_ir_blocks = []
+    current_folded_blocks = []
+    in_fold_location = None
+
+    for block in ir_blocks:
+        if isinstance(block, Fold):
+            if in_fold_location is not None:
```
jssandh2: I can only assume that the first `ir_block` can never be a fold ?

graphql_compiler/compiler/emit_match.py
obi1kenobi: Because the different clauses in the `LET` section have to be comma-separated. We append the first clause, and then we append a comma before every subsequent clause.

graphql_compiler/compiler/expressions.py
obi1kenobi: It is a constant -- it is so constant, in fact, that you can't even get it without calling the function that hides it. I really don't want to pull it out of here, because classes that inherit from this one have no business reading that constant.

graphql_compiler/compiler/helpers.py
obi1kenobi: This method is the only possible way to get the name of the location, and everything calls it. Since you can't construct it any other way, I really think it's about as safe as it can ever be.

CHANGELOG.md
obi1kenobi: Hmm I guess this broke the formatting -- I'll fix it, good catch.

graphql_compiler/compiler/ir_lowering_common.py
obi1kenobi: Right -- but `ir_blocks` can't ever be `None`. I can add an assertion for this, though it seems unnecessary since:
- `ir_blocks` is generated by the compiler front-end which is super well-tested;
- if it were `None`, this would fail in an obvious way in 10 other places after this.

graphql_compiler/compiler/ir_lowering_common.py
obi1kenobi: `CompilerEntity` and all its subclasses (including these ones) are meant to be immutable, though Python doesn't have great ways of enforcing this. All the compiler's code creates new blocks and new expressions instead of editing them.

graphql_compiler/compiler/ir_lowering_common.py
obi1kenobi: The first block cannot be a `Fold`, though this is not what the assertion checks. The assertion checks that you don't have a `Fold` within a `Fold`.

graphql_compiler/compiler/expressions.py
jssandh2: If they shouldn't, then make it private here without exposing any function at all. But, this is fine too.

graphql_compiler/compiler/ir_lowering_common.py
jssandh2: Good to know that `CompilerEntity` is enforcing immutability.

graphql_compiler/compiler/expressions.py
obi1kenobi: Resolved offline.

graphql_compiler/compiler/ir_lowering_common.py
obi1kenobi: It's not enforced (grumble grumble Python), but this is the intent of the design.

PR #39, opened 2017-10-05
Update the docs. Filtering and type coercions are allowed in folds.

obi1kenobi:


jssandh2:
Can `coveralls` be smart enough not to attempt to test files that are : `test_*.py` _or_ `*.md` ?
LGTM

obi1kenobi:
I think we've made it omit `graphql_compiler/tests/*` but I don't know about the rest. I think this is an accounting bug (it might be comparing against `master`) rather than "testing" markdown files.

PR #47, opened 2017-10-23
Add UnaryTransformation Expression type, and Literal int values.

obi1kenobi:
The new expression type and support for literal integers are necessary to express checks against the number of edges of a given type that exist in a given location.

No tests here -- they are all at the level of testing the new filter type, and implicitly cover this new code as well. This is why I'm opening the PR to a feature branch (`edge_degree_filtering`) -- I'll merge that into `master` when it's completely ready and has acceptable coverage.

### DIFF
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -518,7 +518,7 @@ class ContextFieldExistence(Expression):
 def _validate_operator_name(operator, supported_operators):
     """Ensure the named operator is valid and supported."""
     if not isinstance(operator, six.text_type):
-        raise TypeError(u'Expected unicode operator, got: {} {}'.format(
+        raise TypeError(u'Expected operator as unicode string, got: {} {}'.format(
             type(operator).__name__, operator))
 
     if operator not in supported_operators:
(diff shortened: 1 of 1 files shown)


graphql_compiler/compiler/expressions.py
```diff
@@ -508,6 +515,82 @@ def to_gremlin(self):
         raise AssertionError(u'ContextFieldExistence.to_gremlin() was called: {}'.format(self))
 
 
+def _validate_operator_name(operator, supported_operators):
+    """Ensure the named operator is valid and supported."""
+    if not isinstance(operator, six.text_type):
+        raise TypeError(u'Expected unicode operator, got: {} {}'.format(
```
lodrion: you are testing for text_type, the error is "expected unicode operator"

obi1kenobi: `text_type` is `unicode` in python 2, and `str` in python 3 since `str` there is a list of unicode code points. Any recommendations for how to rephrase the error to make this more clear?

lodrion: Expected operator as unicode string?

obi1kenobi: Thanks, will change to that before merging.

PR #48, opened 2017-10-23
Allow @filter directives that apply to the enclosing scope.

obi1kenobi:
An example of such a filtering operator is the upcoming edge degree operator.

The changes are as follows:
- Instead of passing `local_directives` everywhere, we now only pass the directives that must be unique (i.e. appear 0 or 1 times at any given AST node).
- `@filter` directives are processed in exactly one place, so we only load them there. We go through a new helper that cleanly encapsulates the "which scope does this directive apply to" logic.

### DIFF
--- a/graphql_compiler/compiler/directive_helpers.py
+++ b/graphql_compiler/compiler/directive_helpers.py
@@ -66,12 +66,16 @@ def get_local_filter_directives(ast, inner_vertex_fields):
     Returns:
         list of filter directive objects
     """
-    ast_directives_obj = ast.directives or []  # ast.directives can be None
-    result = [
-        directive_obj
-        for directive_obj in ast_directives_obj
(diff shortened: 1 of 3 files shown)


graphql_compiler/compiler/directive_helpers.py
```diff
@@ -49,6 +49,42 @@ def get_directives(ast):
...
+        if directive_obj.name.value == 'filter'
+    ]
+
+    if inner_vertex_fields:  # allow the argument to be None
+        for inner_ast in inner_vertex_fields:
+            for directive_obj in inner_ast.directives:
+                if directive_obj.name.value != 'filter':
+                    continue
+
+                if is_filter_with_outer_scope_vertex_field_operator(directive_obj):
```
jssandh2: Drop these in the first for loop.

PR #49, opened 2017-10-24
Update filter handler functions to use FilterOperationInfo objects.

obi1kenobi:
Since we now have a lot of different filter operators, it has become clear that the old filter function signature can use some streamlining. This PR adds a new `FilterOperationInfo` object that holds all non-contextual data that is relevant to a filtering operation, such as the directive object itself, or the type and field name being filtered on.

This new abstraction:
- simplifies the function signature of filtering functions to just 3 arguments, while getting rid of all unused variables; and
- simplifies the reasoning for which filters are valid in which situations.

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -72,8 +72,8 @@ from .directive_helpers import (get_local_filter_directives, get_unique_directiv
                                 validate_vertex_field_directive_interactions)
 from .filters import process_filter_directive
 from .helpers import (FoldScopeLocation, Location, get_ast_field_name, get_field_type_from_schema,
-                      get_uniquely_named_objects_by_name, is_vertex_field_name,
-                      strip_non_null_from_type, validate_safe_string)
+                      get_uniquely_named_objects_by_name, get_vertex_field_type,
+                      is_vertex_field_name, strip_non_null_from_type, validate_safe_string)
 
 
 # The OutputMetadata will have the following types for its members:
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -519,7 +509,8 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
     fields = _get_fields(ast)
     vertex_fields, property_fields = fields
     fragment = _get_inline_fragment(ast)
-    local_filters_directives = get_local_filter_directives(ast, vertex_fields)
+    filter_operations = get_local_filter_directives(
+        ast, current_schema_type, vertex_fields)
```
jssandh2: `current_schema_type` is guaranteed to be the `root` of the context ?

graphql_compiler/compiler/directive_helpers.py
```diff
@@ -59,34 +62,59 @@ def get_local_filter_directives(ast, inner_vertex_fields):
...
+                    # that is a compilation error.
+                    if not is_vertex_field_type(current_schema_type):
+                        raise GraphQLCompilationError(
+                            u'Found disallowed filter on a property field: {} {} '
+                            u'{}'.format(directive_obj, current_schema_type, filtered_field_name))
+                    elif isinstance(ast, InlineFragment):
+                        raise GraphQLCompilationError(
+                            u'Found disallowed filter on a type coercion: {} '
+                            u'{}'.format(directive_obj, current_schema_type))
+                    else:
```
jssandh2: I like this. :)

graphql_compiler/compiler/helpers.py
```diff
@@ -41,6 +54,24 @@ def get_field_type_from_schema(schema_type, field_name):
...
+    # According to the schema, the vertex field itself is of type GraphQLList, and this is
+    # what get_field_type_from_schema returns. We care about what the type *inside* the list is,
+    # i.e., the type on the other side of the edge (hence .of_type).
+    # Validation guarantees that the field must exist in the schema.
+    if not is_vertex_field_name(vertex_field_name):
+        raise AssertionError(u'Trying to load the vertex field type of a non-vertex field: '
+                             u'{} {}'.format(current_schema_type, vertex_field_name))
+
+    raw_field_type = get_field_type_from_schema(current_schema_type, vertex_field_name)
+    if not isinstance(strip_non_null_from_type(raw_field_type), GraphQLList):
```
jssandh2: Why are we stripping _only_ Non-Null Types ? Is it possible for there to be a type : `[None, None]`, for example ?

graphql_compiler/compiler/compiler_frontend.py
obi1kenobi: Just to clarify terminology, `context` is a mutable dict that carries all the parsing state of the frontend (like tagged values or outputs), and is not involved here.

`current_schema_type` is guaranteed to be the GraphQL type of the part of the GraphQL query to which `ast` points. So in that sense, if I understood the question correctly, the answer is yes.

graphql_compiler/compiler/helpers.py
obi1kenobi: The GraphQL type system by default assumes that types are nullable. We don't currently use the non-null type annotations anywhere in the compiler, so we just blindly strip them everywhere.

Primer on GraphQL types
- `[String]`: nullable list of nullable strings
- `[String!]`: nullable list of non-nullable strings
- `[String]!`: non-nullable list of nullable strings
- `[String!]!`: non-nullable list of non-nullable strings

graphql_compiler/compiler/compiler_frontend.py
jssandh2: You did. Since the whole idea of correctly dealing with `@filter` directives depends on determining the scope based on where the sub-AST is rooted, I just wanted to make sure we are respecting the wishes of the previous PR.

PR #50, opened 2017-10-24
Implement the new "has_edge_degree" filter.

obi1kenobi:
~80 lines of implementation, ~500 lines of tests. Three main test ideas, tested both at the IR level and the full compiled query level:
- Test the new filter operator by itself.
- Test it together with `@optional` to pin down the edge cases there.
- Test it together with `@fold` since folding behaves sort of like an `@optional`.


jssandh2:
Assuming the tests work : I only checked the block generation.

PR #51, opened 2017-10-24
Add docs for the new "has_edge_degree" filter operation.

obi1kenobi:
@MichaelaShtilmanMinkin I'd appreciate your 👀  on this since you already helped write much of the existing documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ It's modeled after Python's `json.tool`, reading from stdin and writing to stdou
      * [in_collection](#in_collection)
      * [has_substring](#has_substring)
      * [contains](#contains)
+     * [has_edge_degree](#has_edge_degree)
   * [Type coercions](#type-coercions)
   * [Meta fields](#meta-fields)
   * [The GraphQL schema](#the-graphql-schema)
@@ -577,6 +578,9 @@ containing the animal's name in a column named `name`.
 This returns one row for every `Animal` which has a color contained in a list of colors,
 containing the `Animal`'s name and color in columns named `animal_name` and `color`, respectively.
 
+#### Constraints and Rules
+- Must be on a property field that is not of list type.
+
 ### has_substring
 #### Example Use
 ```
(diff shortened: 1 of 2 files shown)


jssandh2:
Looks good to me, but I'm not the `doc` expert. Will wait on @MichaelaShtilmanMinkin.

README.md
```diff
@@ -605,6 +612,38 @@ This returns one row for every `Animal` whose list of aliases contains the value
...
+#### Constraints and Rules
+- Must be on a property field of list type.
+
+### has_edge_degree
+#### Example Use
+```
+{
+    Animal {
+        name @output(out_name: "animal_name")
+
```
MichaelaShtilmanMinkin: Is there a reason for the newline here? I don't see it anywhere else in the docs under a similar context.

README.md
```diff
@@ -605,6 +612,38 @@ This returns one row for every `Animal` whose list of aliases contains the value
...
+        out_Animal_ParentOf @filter(op_name: "has_edge_degree", value: ["$child_count"]) @optional {
+            uuid
```
MichaelaShtilmanMinkin: Why do we just have `uuid` here?

README.md
obi1kenobi: It's not necessary, it just visually separates the output from the meat of the query with the complicated filter. Would you prefer it removed?

README.md
obi1kenobi: Because we have to have *something* between the curly braces, to satisfy the GraphQL parser. The `uuid` has no effect, since it's not tagged, output, or filtered on.

README.md
MichaelaShtilmanMinkin: I mostly care about consistency throughout the docs. It is a pretty complicated filter so maybe it's good to keep it.

README.md
MichaelaShtilmanMinkin: Yes please! It seems a little out of place.

obi1kenobi: Sounds good, I'll add a note. Thanks!

PR #52, opened 2017-10-25
Merge edge degree filtering feature into master

obi1kenobi:
This feature is ready to be merged into `master`. It is working, tested, and documented.

Unless there are further concerns, I will release this shortly as `v1.3.0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,10 +637,11 @@ the GraphQL syntax rule that requires at least one field to exist within any `{}
 Since this field is not marked with any directive, it has no effect on the query.
 
 *N.B.:* Please note the `@optional` directive on the vertex field being filtered above.
-This directive is necessary if the `$child_count` parameter can be set to `0`,
-as without it, the query semantics require `out_Animal_ParentOf` to simultaneously exist
-(because of the absence of `@optional`) and not exist (because of the `@filter` with `0` count),
-which will of course produce an empty result set.
+If in your use case you expect to set `$child_count` to 0, you must also mark that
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -605,6 +612,42 @@ This returns one row for every `Animal` whose list of aliases contains the value
...
+Each row contains the matching `Animal`'s name, in a column named `animal_name`.
+
+The `uuid` field within the `out_Animal_ParentOf` vertex field is added simply to satisfy
+the GraphQL syntax rule that requires at least one field to exist within any `{}`.
+Since this field is not marked with any directive, it has no effect on the query.
+
+*N.B.:* Please note the `@optional` directive on the vertex field being filtered above.
+This directive is necessary if the `$child_count` parameter can be set to `0`,
+as without it, the query semantics require `out_Animal_ParentOf` to simultaneously exist
+(because of the absence of `@optional`) and not exist (because of the `@filter` with `0` count),
```
lodrion: the actual problem with this is the reverse when out_Animal_ParentOf does not exist because it is not optional but should because child_count == 0

obi1kenobi: That's exactly the case I was trying to explain -- the user set `$child_count` to 0 but didn't mark it optional, so the database will return no results. Ideas on how to make it less confusing?

lodrion: Yeah. That's not how I read it.
I think you should do more exposition there:
If you expect to pass 0 as the parameter you have to label this edge as optional, because absence of optional implicitly requires at least one of the edges to exist.

obi1kenobi: Updated!

PR #53, opened 2017-10-26
Move common test input/output data to a common place.

obi1kenobi:
Most of the IR test and compilation tests use the same GraphQL input queries, and output the same metadata. Since I'm about to add a whole lot more metadata, it might be a good idea to put all the common test data in one place. That way, I can add the new metadata once, instead of once per file.

I'm only doing this for one test for now, just to get everyone's input on the idea. If everyone is in favor of this, I would:
- apply it to all tests, and then
- change the `check_test_data` methods to take a `CommonTestData` object instead of unwrapping the data item before passing it in.

### DIFF
--- a/graphql_compiler/tests/test_compiler.py
+++ b/graphql_compiler/tests/test_compiler.py
@@ -6,36 +6,33 @@ from graphql import GraphQLID, GraphQLInt, GraphQLList, GraphQLString
 import six
 
 from ..compiler import OutputMetadata, compile_graphql_to_gremlin, compile_graphql_to_match
-from ..exceptions import GraphQLCompilationError
-from ..schema import GraphQLDate, GraphQLDateTime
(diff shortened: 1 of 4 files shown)


obi1kenobi:
I removed one unnecessary test, and deduplicated the input data in all others.

If you are curious, I didn't actually do this by hand. Here's a small selection of write-only (i.e. unreadable) regex patterns that proved very helpful:
```
(?s)'''[^{][^@]+?expected_output_metadata =

def test_([a-z_0-9]+?)(\(self\):\n)([ ]+)([^ t#])

def test_([a-z_0-9]+?)(\(self\):\n)((?:[ ]+# .*\n)*)([ ]+)([^ #t])

def test_([a-z_0-9]+?)(\(self\):\n)((?:[ ]+# .*\n)*)([ ]+)graphql_input

(?s)[ ]+graphql_input = '''{[^']+'''[\n]+
```

PR #54, opened 2017-11-02
Use an explicit namedtuple for the frontend output.

obi1kenobi:
This helps us prepare for the additional dependency metadata the frontend will soon generate. We also streamline the compilation process by deduplicating the common GraphQL to compiled query code flow.


graphql_compiler/tests/test_ir_generation.py
```diff
@@ -25,7 +25,10 @@ def check_test_data(test_case, test_data, expected_blocks, expected_location_typ
 
     compilation_results = graphql_to_ir(test_case.schema, test_data.graphql_input,
                                         type_equivalence_hints=schema_based_type_equivalence_hints)
-    received_blocks, output_metadata, input_metadata, location_types = compilation_results
+    received_blocks = compilation_results.ir_blocks
```
lodrion: Since you now have a namedtuple, I would not create local variables. This would make your tests be more explicit

PR #55, opened 2018-02-28
Add workaround for OrientDB parsing bug with "while:" and "class:" clauses

obi1kenobi:
OrientDB bug in question: https://github.com/orientechnologies/orientdb/issues/8129

This workaround is confirmed to avoid the issue.

### DIFF
--- a/graphql_compiler/compiler/workarounds/orientdb_class_with_while.py
+++ b/graphql_compiler/compiler/workarounds/orientdb_class_with_while.py
@@ -1,0 +1,64 @@
+# Copyright 2017 Kensho Technologies, Inc.
+"""Workarounds for OrientDB bug disallowing "class:" clauses together with "while:" clauses.
+
+For details, see:
+https://github.com/orientechnologies/orientdb/issues/8129
+"""
+import funcy
+
+from ..blocks import Filter, Recurse
+from ..expressions import BinaryComposition, Literal, LocalField
+
+
+def _coerce_block_to_filter_block(coerce_type_block, where_block):
+    """Create an "INSTANCEOF" Filter block from a CoerceType block."""
+    coerce_type_target = coerce_type_block.target_class
+    if len(coerce_type_target) != 1:
+        raise AssertionError(u'Unexpected "coerce_type_target" for MATCH query: '
+                             u'{}'.format(coerce_type_target))
+    coerce_type_target = funcy.first(coerce_type_target)
+
+    # INSTANCEOF requires the target class to be passed in as a string,
+    # so we make the target class a string literal.
+    new_predicate = BinaryComposition(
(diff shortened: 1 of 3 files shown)


graphql_compiler/compiler/workarounds/orientdb_class_with_while.py
```diff
@@ -0,0 +1,64 @@
...
+from ..expressions import BinaryComposition, Literal, LocalField
+
+
+def _coerce_block_to_filter_block(coerce_type_block, where_block):
+    """Create an "INSTANCEOF" Filter block from a CoerceType block."""
+    coerce_type_target = coerce_type_block.target_class
+    if len(coerce_type_target) != 1:
+        raise AssertionError(u'Unexpected "coerce_type_target" for MATCH query: '
+                             u'{}'.format(coerce_type_target))
+    coerce_type_target = funcy.first(coerce_type_target)
```
eyurtsev: The length check above is strict. You can index[0] element here to ensure that you don't end up with a Maybe data type

graphql_compiler/compiler/workarounds/orientdb_class_with_while.py
```diff
@@ -0,0 +1,64 @@
...
+    # Instead of "class:", we use "INSTANCEOF" in the "where:" clause to get correct behavior.
+    # However, we don't want to switch all coercions to this format, since the "class:" clause
+    # provides valuable info to the MATCH query scheduler about how to schedule efficiently.
+    new_match_traversals = []
+
+    for current_traversal in match_query.match_traversals:
+        new_traversal = []
+
+        for match_step in current_traversal:
+            new_match_step = match_step
```
eyurtsev: Perhaps add an else block and move this declaration into an else block. 

This will make it easier to track the places where `new_match_step` is declared

graphql_compiler/compiler/workarounds/orientdb_class_with_while.py
lodrion: +1 it would also read cleaner

lodrion: +1

obi1kenobi: `coerce_type_target` is unfortunately a `set` so `[0]` isn't an option. The length check ensures that the `Maybe` is always a `Some` (or in python3 terms, the `Optional[str]` is actually just a `str`).

graphql_compiler/compiler/workarounds/orientdb_class_with_while.py
obi1kenobi: I'd rather not risk having undeclared variables that `NameError` -- this is why I declare the thing the bottom of the block is going to use at the top of the block. Also, this is the style that the rest of the compiler follows, so I'd rather stay consistent.

Issue #45, opened 2017-10-20
Allow traversing from within a @fold scope


obi1kenobi:
Allow queries like:
```
{
    Species {
        in_Animal_OfSpecies @fold {
            out_Entity_Related {
                name @output(out_name: "related_list")
            }
        }
    }
}
```

We currently don't support traversals from inside a `@fold` scope.

obi1kenobi:
Closed in #60.

PR #59, opened 2018-04-03
 Allow traversal within a @fold scope (#45)

amartyashankha:
Added `MATCH` generation support for `graphql` queries with traversal within a `@fold` scope. Also added test cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Current development version
 
+## v1.3.1
+
+- Add a workaround for a bug in OrientDB related to `@recurse` with type coercions [#55](https://github.com/kensho-technologies/graphql-compiler/pull/55)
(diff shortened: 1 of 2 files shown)


graphql_compiler/compiler/directive_helpers.py
```diff
@@ -179,3 +179,29 @@ def validate_vertex_field_directive_interactions(location, directives):
...
+
+def validate_vertex_field_directive_in_context(location, directives, context):
+    """Ensure that the specified vertex field directives are allowed in the current context."""
+    fold_directive = directives.get('fold', None)
+    optional_directive = directives.get('optional', None)
+    output_source_directive = directives.get('output_source', None)
+    recurse_directive = directives.get('recurse', None)
+
+    fold_context = 'fold' in context
+    # optional_context = 'optional' in context
```
jssandh2: The comments are unnecessary.

PR #60, opened 2018-04-03
 Allow traversal within a @fold scope (Issue #45) -- Supercedes #59

amartyashankha:
Added `MATCH` generation support for `graphql` queries with traversal within a `@fold` scope. Also added test cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ If a given `Animal` has no children, its `child_names` list is empty.
 #### Constraints and Rules
 - `@fold` can only be applied to vertex fields, except the root vertex field.
 - May not exist at the same vertex field as `@recurse`, `@optional`, or `@output_source`.
-- Only one vertex field may be expanded within a scope marked `@fold`. Note that this applies for all nested scopes within.
+- At most one vertex field may be expanded in a scope within a `@fold` marked scope. Note that this applies for all nested scopes (see example below).
 - `@tag`, `@recurse`, `@optional`, `@output_source` and `@fold` may not be used anywhere within a scope marked `@fold`.
 - Use of type coercions or `@filter` at or within the vertex field marked `@fold` is allowed.
   Only data that satisfies the given type coercions and filters is returned by the `@fold`.
@@ -239,6 +239,34 @@ If a given `Animal` has no children, its `child_names` list is empty.
   it may optimize it away. See the
   [Optional `type_equivalence_hints` compilation parameter](#optional-type_equivalence_hints-parameter)
   section for more details.
+
+#### Example
+```
+    out_Animal_ParentOf @fold {
+        name @output ...
+        in_Animal_ParentOf {
+            in_Animal_PeerOf {
+                ...
+            }
+            out_Animal_RelatedTo {
+                ...
+            }
+        }
+    }
+```
+This fragment is not allowed because two vertex fields are expanded inside `in_Animal_ParentOf`, which is within a `@fold` scope. It is also invalid to `output` at any level except the innermost one.
+```
+    out_Animal_ParentOf @fold {
+        in_Animal_ParentOf {
+            in_Animal_PeerOf {
+                out_Animal_RelatedTo {
+                    name @output ...
+                }
+            }
+        }
+    }
+```
+This is a valid use of `@fold`.
 
 ### @tag
 
--- a/README.md
+++ b/README.md
@@ -231,8 +231,8 @@ If a given `Animal` has no children, its `child_names` list is empty.
 #### Constraints and Rules
 - `@fold` can only be applied to vertex fields, except the root vertex field.
 - May not exist at the same vertex field as `@recurse`, `@optional`, or `@output_source`.
-- Expanding vertex fields is not allowed within a scope marked `@fold`.
-- `@tag` and `@fold` may not be used within a scope marked `@fold`.
+- Only one vertex field may be expanded within a scope marked `@fold`. Note that this applies for all nested scopes within.
+- `@tag`, `@recurse`, `@optional`, `@output_source` and `@fold` may not be used at any level within a scope marked `@fold`.
 - Use of type coercions or `@filter` at or within the vertex field marked `@fold` is allowed.
   Only data that satisfies the given type coercions and filters is returned by the `@fold`.
 - If the compiler is able to prove that the type coercion in the `@fold` scope is actually a no-op,
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ If a given `Animal` has no children, its `child_names` list is empty.
 - `@fold` can only be applied to vertex fields, except the root vertex field.
 - May not exist at the same vertex field as `@recurse`, `@optional`, or `@output_source`.
 - Only one vertex field may be expanded within a scope marked `@fold`. Note that this applies for all nested scopes within.
-- `@tag`, `@recurse`, `@optional`, `@output_source` and `@fold` may not be used at any level within a scope marked `@fold`.
+- `@tag`, `@recurse`, `@optional`, `@output_source` and `@fold` may not be used anywhere within a scope marked `@fold`.
 - Use of type coercions or `@filter` at or within the vertex field marked `@fold` is allowed.
   Only data that satisfies the given type coercions and filters is returned by the `@fold`.
 - If the compiler is able to prove that the type coercion in the `@fold` scope is actually a no-op,
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -69,7 +69,8 @@ from .context_helpers import (has_encountered_output_source, is_in_fold_scope, i
 from .directive_helpers import (get_local_filter_directives, get_unique_directives,
                                 validate_property_directives, validate_root_vertex_directives,
                                 validate_vertex_directives,
-                                validate_vertex_field_directive_interactions)
+                                validate_vertex_field_directive_interactions,
+                                validate_vertex_field_directive_in_context)
 from .filters import process_filter_directive
 from .helpers import (FoldScopeLocation, Location, get_ast_field_name, get_field_type_from_schema,
                       get_uniquely_named_objects_by_name, get_vertex_field_type,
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -231,8 +231,8 @@ If a given `Animal` has no children, its `child_names` list is empty.
 #### Constraints and Rules
 - `@fold` can only be applied to vertex fields, except the root vertex field.
 - May not exist at the same vertex field as `@recurse`, `@optional`, or `@output_source`.
-- Expanding vertex fields is not allowed within a scope marked `@fold`.
-- `@tag` and `@fold` may not be used within a scope marked `@fold`.
+- Only one vertex field may be expanded within a scope marked `@fold`. Note that this applies for all nested scopes within.
```
obi1kenobi: I think this wording will confuse people -- writing precise documentation is hard. The limitation is that "no scope within a scope marked `@fold` may contain more than one expanded vertex field", but ideally we'd also have an example of what is okay and what is not.

README.md
```diff
@@ -231,8 +231,8 @@ If a given `Animal` has no children, its `child_names` list is empty.
...
+- `@tag`, `@recurse`, `@optional`, `@output_source` and `@fold` may not be used at any level within a scope marked `@fold`.
```
obi1kenobi: `level` is not defined anywhere in the documentation -- let's try to stick to the existing language and terms for the sake of not confusing people. How about simply "anywhere within a scope marked @fold"?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -417,6 +421,12 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
             _validate_fold_has_outputs(context['fold'], context['outputs'])
             basic_blocks.append(blocks.Unfold())
             del context['fold']
+            try:
+                del context['fold_last_level']
+            except KeyError:
```
obi1kenobi: Instead of attempting an operation and recovering if it fails, let's explicitly check the expected precondition (`'fold_last_level' in context`) and raise an error if necessary. This is overall a safer pattern since we explicitly name what we expect to exist, instead of relying on the "correct" thing raising the "correct" error type.

graphql_compiler/compiler/context_helpers.py
```diff
@@ -4,6 +4,11 @@
 from ..exceptions import GraphQLCompilationError
 
 
+def is_in_fold_last_level_scope(context):
```
obi1kenobi: how about `innermost_scope` instead of `last_level_scope` everywhere? I think it's more precise since `level` is never formally defined anywhere

graphql_compiler/compiler/directive_helpers.py
```diff
@@ -179,3 +179,29 @@ def validate_vertex_field_directive_interactions(location, directives):
...
+    """Ensure that the specified vertex field directives are allowed in the current context."""
+    fold_directive = directives.get('fold', None)
+    optional_directive = directives.get('optional', None)
+    output_source_directive = directives.get('output_source', None)
+    recurse_directive = directives.get('recurse', None)
+
+    fold_context = 'fold' in context
+    # optional_context = 'optional' in context
+    # output_context = 'output' in context
+    # recurse_context = 'recurse' in context
```
obi1kenobi: commented out code?

graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -149,6 +149,9 @@ def _sanity_check_every_location_is_marked(ir_blocks):
         elif isinstance(block, (QueryRoot, Traverse, Recurse)):
             found_start_block = True
             mark_location_blocks = 0
+        # Don't need to mark locations after a fold block
+        elif isinstance(block, Fold):
+            break
```
obi1kenobi: This will stop processing _all_ future blocks, which is probably not what we want.

graphql_compiler/compiler/emit_match.py
```diff
@@ -123,10 +125,17 @@ def _represent_fold(fold_location, fold_ir_blocks):
...
-        final_string += u'[' + block.predicate.to_match() + u']'
+        if isinstance(block, Filter):
+            final_string += u'[' + block.predicate.to_match() + u']'
+        elif isinstance(block, Traverse):
+            template_data = {
+                'direction': block.direction,
+                'edge_name': block.edge_name,
+            }
+            final_string += traverse_edge_template % template_data
+        elif not isinstance(block, Backtrack):
```
obi1kenobi: We probably shouldn't have `Backtrack` blocks here at all, this can probably be just an `else`

graphql_compiler/tests/test_compiler.py
```diff
@@ -22,17 +38,11 @@ def check_test_data(test_case, test_data, expected_match, expected_gremlin):
...
-    compare_input_metadata(test_case, test_data.expected_input_metadata, result.input_metadata)
-
-    result = compile_graphql_to_gremlin(test_case.schema, test_data.graphql_input,
-                                        type_equivalence_hints=schema_based_type_equivalence_hints)
-    compare_gremlin(test_case, expected_gremlin, result.query)
-    test_case.assertEqual(test_data.expected_output_metadata, result.output_metadata)
-    compare_input_metadata(test_case, test_data.expected_input_metadata, result.input_metadata)
+    check_test_data_match(test_case, test_data, expected_match,
+                          schema_based_type_equivalence_hints)
+    if expected_gremlin:
```
obi1kenobi: Let's not special case like this if possible -- OrientDB 3.x promises faster Gremlin execution, and it's probably not that hard to write the Gremlin side of this functionality once the IR changes are in.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -196,16 +196,28 @@ def test_fold_directive_constraints(self):
...
+        traversal_inside_fold_block_after_output = '''{
+            Animal {
+                out_Animal_ParentOf @fold {
+                    uuid @output(out_name: "uuid")
+                    out_Animal_FedAt {
+                        uuid
+                    }
+                }
+            }
+        }'''
```
obi1kenobi: can you also add a "tree-like" traversal attempt, where all outputs are in the innermost scopes but there are multiple such scopes in the `@fold`?

graphql_compiler/compiler/directive_helpers.py
jssandh2: It's outdated, @obi1kenobi - I already flagged this.

graphql_compiler/compiler/emit_match.py
amartyashankha: That's what I thought, but `extract_folds_from_ir_blocks` does not remove the `Backtrack` blocks, and these end up in `fold_ir_blocks`. Should I modify `extract_folds_from_ir_blocks`?

obi1kenobi: No -- `extract_folds_from_ir_blocks` is a common (backend-language-agnostic) lowering step, so it needs to keep `Backtrack` blocks. I'm not sure if they are needed in Gremlin, but for MATCH we know they aren't necessary within a `@fold`.

We don't want too much complexity all in one place and one function. We should have a separate lowering step for extracted folded blocks where the `Backtrack` blocks are removed. This can be a common step if Gremlin needs it too, or a MATCH-specific one if not.

graphql_compiler/tests/test_ir_generation_errors.py
amartyashankha: Done!

README.md
```diff
@@ -231,15 +231,43 @@ If a given `Animal` has no children, its `child_names` list is empty.
...
   Only data that satisfies the given type coercions and filters is returned by the `@fold`.
 - If the compiler is able to prove that the type coercion in the `@fold` scope is actually a no-op,
   it may optimize it away. See the
   [Optional `type_equivalence_hints` compilation parameter](#optional-type_equivalence_hints-parameter)
   section for more details.
 
+#### Example
+```
+    out_Animal_ParentOf @fold {
+        name @output ...
```
obi1kenobi: Please make the examples syntactically valid GraphQL. A confused user that consults the documentation will be even more confused at the unfamiliar "new" syntax being used here.

README.md
```diff
@@ -231,15 +231,43 @@ If a given `Animal` has no children, its `child_names` list is empty.
...
+            in_Animal_PeerOf {
+                ...
+            }
+            out_Animal_RelatedTo {
+                ...
+            }
+        }
+    }
+```
+This fragment is not allowed because two vertex fields are expanded inside `in_Animal_ParentOf`, which is within a `@fold` scope. It is also invalid to `output` at any level except the innermost one.
```
obi1kenobi: Try to keep line length below 100 whenever possible.

The `It is also invalid to `output` at any level except the innermost one.` limitation is only specified in the example, and not in the constraints and rules section -- most people will miss it.

graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -133,8 +133,18 @@ def _sanity_check_every_location_is_marked(ir_blocks):
...
     mark_location_blocks = 0
+
     for block in ir_blocks:
+        # Don't need to mark locations after a fold block
+        if isinstance(block, Fold):
+            found_fold_block = True
+        if found_fold_block:
+            if isinstance(block, Unfold):
+                found_fold_block = False
+            continue
```
obi1kenobi: Why the `continue`? This is not obvious.

graphql_compiler/tests/test_compiler.py
```diff
@@ -10,7 +10,23 @@
 from .test_helpers import compare_gremlin, compare_input_metadata, compare_match, get_schema
 
 
-def check_test_data(test_case, test_data, expected_match, expected_gremlin):
+def check_test_data_match(test_case, test_data, expected, schema_based_type_equivalence_hints):
```
obi1kenobi: I don't think we need the two separate functions, since we want to keep feature parity.

README.md
```diff
@@ -231,15 +231,43 @@ If a given `Animal` has no children, its `child_names` list is empty.
 #### Constraints and Rules
 - `@fold` can only be applied to vertex fields, except the root vertex field.
 - May not exist at the same vertex field as `@recurse`, `@optional`, or `@output_source`.
-- Expanding vertex fields is not allowed within a scope marked `@fold`.
-- `@tag` and `@fold` may not be used within a scope marked `@fold`.
+- At most one vertex field may be expanded in a scope within a `@fold` marked scope. Note that this applies for all nested scopes (see example below).
```
obi1kenobi: This is a bit unclear -- how about moving around the word order:
`In the scope marked `@fold` and any scope nested within, at most one vertex field may be expanded. (Example below)`

I think this will make the "Note that" sentence unnecessary.

PR #62, opened 2018-04-06
Added gremlin support for traversal within `@folds`.

amartyashankha:


graphql_compiler/compiler/ir_lowering_gremlin.py
```diff
@@ -148,8 +148,9 @@ def validate(self):
             raise TypeError(u'Expected FoldScopeLocation fold_scope_location, got: {} {}'.format(
                 type(self.fold_scope_location), self.fold_scope_location))
 
+        allowed_block_types = (GremlinFoldedFilter, Traverse, Backtrack)
         for block in self.folded_ir_blocks:
-            if not isinstance(block, GremlinFoldedFilter):
+            if not isinstance(block, allowed_block_types):
                 raise AssertionError(u'Found non-GremlinFoldedFilter in folded_ir_blocks: {} {} '
```
obi1kenobi: this error message is misleading now -- the allowed block types are more than just `GremlinFoldedFilter`

graphql_compiler/compiler/ir_lowering_gremlin.py
```diff
@@ -252,6 +254,31 @@ def to_gremlin(self):
         return u'findAll{{entry -> {}}}'.format(self.predicate.to_gremlin())
 
 
+class GremlinFoldedTraverse(Traverse):
+    """A Gremlin-specific Traverse block to be used only within @fold scopes."""
+
+    def __init__(self, traverse_block):
```
obi1kenobi: This would perhaps be better suited to a classmethod factory function `from_traverse` or something like that, otherwise it's a confusing pattern -- the constructor of a subclass of `Traverse` requires a `Traverse` object to initialize itself.

graphql_compiler/tests/test_compiler.py
```diff
@@ -1719,8 +1718,26 @@ def test_fold_and_traverse(self):
...
+            .as('Animal___1')
+            .transform{it, m -> new com.orientechnologies.orient.core.record.impl.ODocument([
+                animal_name: m.Animal___1.name,
+                sibling_and_self_names_list: (
+                    (m.Animal___1.in_Animal_ParentOf == null) ? [] : (
+                        m.Animal___1.in_Animal_ParentOf
+                            .collect{
+                                entry -> entry.outV.next()
+                            }
+                            .collectMany{entry -> entry.out_Animal_ParentOf.collect{
```
obi1kenobi: Be consistent -- either the closure goes on its own line, or doesn't. Here you have two nested scopes but the line continuation on 1733 is only indented once.

graphql_compiler/compiler/ir_lowering_gremlin.py
```diff
@@ -148,11 +148,13 @@ def validate(self):
...
                 type(self.fold_scope_location), self.fold_scope_location))
 
-        allowed_block_types = (GremlinFoldedFilter, Traverse, Backtrack)
+        allowed_block_types = (GremlinFoldedFilter, GremlinFoldedTraverse, Backtrack)
         for block in self.folded_ir_blocks:
             if not isinstance(block, allowed_block_types):
-                raise AssertionError(u'Found non-GremlinFoldedFilter in folded_ir_blocks: {} {} '
-                                     u'{}'.format(type(block), block, self.folded_ir_blocks))
+                raise AssertionError(u'Found invalid block of type {} in folded_ir_blocks: {} '
+                                     u'Allowed types are {}. '
```
obi1kenobi: trailing whitespace at the end of the string

graphql_compiler/compiler/ir_lowering_gremlin.py
```diff
@@ -148,11 +148,13 @@ def validate(self):
...
+                                     .format(type(block), self.folded_ir_blocks,
+                                             allowed_block_types))
```
obi1kenobi: try to avoid formatting into these "jagged" formations -- you can move the error message string to a new line if need be:
```
raise AssertionError(
    u'error message '
    u'here: {}'.format(stuff))
```

graphql_compiler/compiler/ir_lowering_gremlin.py
```diff
@@ -257,11 +259,11 @@ def to_gremlin(self):
...
-        """Initialize a normal Traverse block using as a copy of the given block."""
+    @classmethod
+    def from_traverse(cls, traverse_block):
+        """Create a GremlinFoldedTraverse block as a copy of the given Traverse block."""
         if isinstance(traverse_block, Traverse):
-            super(GremlinFoldedTraverse, self).__init__(
-                traverse_block.direction, traverse_block.edge_name)
+            return cls(traverse_block.direction, traverse_block.edge_name)
         else:
             raise RuntimeError(u'Tried to initialize an instance of GremlinFoldedTraverse '
```
obi1kenobi: make this an AssertionError now, it should never happen

graphql_compiler/tests/test_compiler.py
```diff
@@ -1729,8 +1729,10 @@ def test_fold_and_traverse(self):
                             .collect{
                                 entry -> entry.outV.next()
                             }
-                            .collectMany{entry -> entry.out_Animal_ParentOf.collect{
-                                edge -> edge.inV.next()}
+                            .collectMany{entry -> entry.out_Animal_ParentOf
```
obi1kenobi: There are many more instances of this formatting that need fixing, please do it everywhere.

PR #63, opened 2018-04-09
Added test cases for `@fold` traversal

amartyashankha:
Tests for `@fold` directive constraints. Test cases for type coercions and filters within a fold traversal.

### DIFF
--- a/graphql_compiler/tests/test_compiler.py
+++ b/graphql_compiler/tests/test_compiler.py
@@ -1742,6 +1742,57 @@ FROM (
 
         check_test_data(self, test_data, expected_match, expected_gremlin)
 
+    def test_fold_and_deep_traverse(self):
+        test_data = test_input_data.fold_and_deep_traverse()
+
+        expected_match = '''
+            SELECT
+                Animal___1.name AS `animal_name`,
+                $Animal___1___in_Animal_ParentOf.name AS `sibling_and_self_names_list`
+            FROM (
+                MATCH {{
+                    class: Animal,
+                    as: Animal___1
+                }}
+                RETURN $matches
+            ) LET
+                $Animal___1___in_Animal_ParentOf =
+                    Animal___1.in("Animal_ParentOf")
+                              .out("Animal_ParentOf")
+                              .in("Animal_ParentOf")
+                              .asList()
+        '''
+        expected_gremlin = '''
+            g.V('@class', 'Animal')
+            .as('Animal___1')
+            .transform{it, m -> new com.orientechnologies.orient.core.record.impl.ODocument([
+                animal_name: m.Animal___1.name,
+                sibling_and_self_names_list: (
+                    (m.Animal___1.in_Animal_ParentOf == null) ? [] : (
+                        m.Animal___1.in_Animal_ParentOf.
+                            collect{
+                                entry -> entry.outV.next()
+                            }
+                            .collectMany{
+                                entry -> entry.out_Animal_ParentOf
+                                    .collect{
+                                        edge -> edge.inV.next()
+                                    }
+                            }
+                            .collectMany{
+                                entry -> entry.in_Animal_ParentOf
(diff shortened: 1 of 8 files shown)


graphql_compiler/tests/test_compiler.py
```diff
@@ -1742,6 +1742,51 @@ def test_fold_and_traverse(self):
...
+                              .asList()
+        '''
+        expected_gremlin = '''
+            g.V('@class', 'Animal')
+            .as('Animal___1')
+            .transform{it, m -> new com.orientechnologies.orient.core.record.impl.ODocument([
+                animal_name: m.Animal___1.name,
+                sibling_and_self_names_list: (
+                    (m.Animal___1.in_Animal_ParentOf == null) ? [] : (
+                        m.Animal___1.in_Animal_ParentOf.
```
obi1kenobi: For consistency, this `.` needs to go on the next line

graphql_compiler/tests/test_input_data.py
```diff
@@ -962,13 +987,15 @@ def fold_and_traverse():
...
     graphql_input = '''{
         Animal {
             name @output(out_name: "animal_name")
-            in_Animal_ParentOf {
-                out_Animal_ParentOf @fold {
-                    name @output(out_name: "sibling_and_self_names_list")
+            in_Animal_ParentOf @fold {
+                out_Animal_ParentOf {
+                    in_Animal_ParentOf {
+                        name @output(out_name: "sibling_and_self_names_list")
```
obi1kenobi: This output name is misleading, and the in -> out -> in on the same edge is a bit weird too. For the sake of people diving into the codebase for the first time and reading the tests, maybe change the last traversal to be across a different edge? Then you can name the output something intuitive.

graphql_compiler/tests/test_input_data.py
```diff
@@ -1295,6 +1322,35 @@ def coercion_on_interface_within_fold_scope():
...
 
+def coercion_on_interface_within_fold_traversal():
+    graphql_input = '''{
+        Animal {
+            name @output(out_name: "animal_name")
+            in_Animal_ParentOf @fold {
+                out_Entity_Related {
+                    ... on Animal {
+                        out_Animal_ParentOf {
+                            name @output(out_name: "related_animals")
```
obi1kenobi: The output name is misleading again -- this hurts the comprehension of the test case since the output suggests one thing, and the structure of the GraphQL suggests another.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -210,9 +210,11 @@ def test_fold_directive_constraints(self):
         traversal_inside_fold_block_after_output = '''{
             Animal {
                 out_Animal_ParentOf @fold {
-                    uuid @output(out_name: "uuid")
-                    out_Animal_FedAt {
-                        uuid
+                    in_Animal_ParentOf @fold {
```
obi1kenobi: I think this second `@fold` is a typo. This testcase (per the variable name) should be testing the impossibility of traversing after an output within a folded block, but is already invalid because of the fold-within-fold.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -227,11 +229,69 @@ def test_fold_directive_constraints(self):
...
+            Animal {
+                out_Animal_ParentOf @fold {
+                    out_Animal_FedAt @output_source {
+                        uuid @output(out_name: "uuid")
+                    }
+                }
+            }
+        }'''
+
+        multiple_vertex_fields_within_fold = '''{
```
obi1kenobi: Maybe one more variant of this testcase would be nice, where the `@fold` initially goes with one inner traverse, and then forks into two different ones within it. I'm confident that test will pass too, but it would be nice to have it covered explicitly.

PR #66, opened 2018-04-12
Allowing traversal inside `@optional` scope

amartyashankha:
The changes in `blocks.py` allow for gremlin support.

### DIFF
--- a/graphql_compiler/compiler/blocks.py
+++ b/graphql_compiler/compiler/blocks.py
@@ -267,6 +267,10 @@ class Traverse(BasicBlock):
                         edge_name=self.edge_name,
                         edge_quoted=safe_quoted_string(self.edge_name)))
         elif self.in_optional_context:
+            # During a traversal, the pipeline element may be null.
+            # The following code returns null when the current pipeline entity is null
+            # (an optional edge did not exist at some earlier traverse).
+            # Otherwise it performs a normal traversal (previous optional edge did exist).
             return (u'ifThenElse{{it == null}}'
                     u'{{null}}{{it.{direction}({edge_quoted})}}'.format(
                         direction=self.direction,
--- a/graphql_compiler/compiler/blocks.py
+++ b/graphql_compiler/compiler/blocks.py
@@ -211,7 +211,7 @@ class MarkLocation(BasicBlock):
 class Traverse(BasicBlock):
     """A block that encodes a traversal across an edge, in either direction."""
 
-    def __init__(self, direction, edge_name, optional=False):
+    def __init__(self, direction, edge_name, optional=False, in_optional_context=False):
         """Create a new Traverse block in the given direction and across the given edge.
 
         Args:
@@ -227,6 +227,7 @@ class Traverse(BasicBlock):
         self.direction = direction
         self.edge_name = edge_name
         self.optional = optional
+        self.in_optional_context = in_optional_context
         self.validate()
 
     def validate(self):
@@ -264,6 +265,11 @@ class Traverse(BasicBlock):
                     u'{{null}}{{it.{direction}({edge_quoted})}}'.format(
                         direction=self.direction,
                         edge_name=self.edge_name,
+                        edge_quoted=safe_quoted_string(self.edge_name)))
+        elif self.in_optional_context:
+            return (u'ifThenElse{{it == null}}'
+                    u'{{null}}{{it.{direction}({edge_quoted})}}'.format(
+                        direction=self.direction,
                         edge_quoted=safe_quoted_string(self.edge_name)))
         else:
             return u'{direction}({edge})'.format(
--- a/graphql_compiler/compiler/blocks.py
+++ b/graphql_compiler/compiler/blocks.py
@@ -244,6 +244,10 @@ class Traverse(BasicBlock):
 
         validate_safe_string(self.edge_name)
 
+    def get_field_name(self):
+        """Return the field name corresponding to the edge being traversed."""
+        return u'{direction}_{edge}'.format(direction=self.direction, edge=self.edge_name)
+
     def to_gremlin(self):
         """Return a unicode object with the Gremlin representation of this block."""
         self.validate()
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -382,6 +382,7 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         in_topmost_optional_block = False
 
         edge_traversal_is_optional = optional_directive is not None
+        in_optional_context = 'optional' in context and not edge_traversal_is_optional
         if edge_traversal_is_optional:
             # Entering an optional block!
             # Make sure there's a tag right before it for the optional Backtrack to jump back to.
@@ -411,7 +412,8 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
             basic_blocks.append(blocks.Recurse(edge_direction, edge_name, recurse_depth))
         else:
             basic_blocks.append(blocks.Traverse(edge_direction, edge_name,
-                                                optional=edge_traversal_is_optional))
+                                                optional=edge_traversal_is_optional,
+                                                in_optional_context=in_optional_context))
 
         inner_basic_blocks = _compile_ast_node_to_ir(schema, field_schema_type, field_ast,
                                                      inner_location, context)
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -383,6 +383,7 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
 
         edge_traversal_is_optional = optional_directive is not None
         in_optional_context = 'optional' in context and not edge_traversal_is_optional
+
         if edge_traversal_is_optional:
             # Entering an optional block!
             # Make sure there's a tag right before it for the optional Backtrack to jump back to.
@@ -684,13 +685,16 @@ def _compile_output_step(outputs):
 
         expression = None
         if fold_scope_location:
-            if optional:
-                raise AssertionError(u'Unreachable state reached, optional in fold: '
-                                     u'{}'.format(output_context))
-
             _, field_name = location.get_location_name()
             expression = expressions.FoldedOutputContextField(
                 fold_scope_location, field_name, graphql_type)
+
+            if optional:
+                existence_check = expressions.ContextFieldExistence(
+                    fold_scope_location.base_location
+                )
+                expression = expressions.TernaryConditional(
+                    existence_check, expression, expressions.NullLiteral)
         else:
             expression = expressions.OutputContextField(location, graphql_type)
 
--- a/graphql_compiler/compiler/emit_match.py
+++ b/graphql_compiler/compiler/emit_match.py
@@ -160,7 +160,7 @@ def _construct_output_to_match(output_block):
 # Public API #
 ##############
 
-def emit_code_from_ir(match_query):
+def emit_code_from_match_query(match_query):
     """Return a MATCH query string from a list of IR blocks."""
     query_data = deque([u'MATCH '])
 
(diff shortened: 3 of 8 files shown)


graphql_compiler/compiler/blocks.py
```diff
@@ -243,6 +244,10 @@ def validate(self):
 
         validate_safe_string(self.edge_name)
 
+    def get_field_name(self):
+        """Return the field name corresponding to the edge being traversed."""
+        return u'{direction}_{edge}'.format(direction=self.direction, edge=self.edge_name)
```
jssandh2: Why not just: 
```python
return u'{}_{}'.format(self.direction, self.edge_name)
```
?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -8,19 +8,30 @@
...
 from .ir_sanity_checks import sanity_check_ir_blocks_from_frontend
-from .match_query import MatchStep, convert_to_match_query
+from .match_query import MatchQuery, MatchStep, convert_to_match_query
 from .workarounds import orientdb_class_with_while, orientdb_eval_scheduling
 
 
+###
+# A CompoundMatchQuery is a representation of several MatchQuery objects containing
+#   - match_queries: a list MatchQuery objects
+CompoundMatchQuery = namedtuple('CompoundMatchQuery', ('match_queries'))
```
jssandh2: Seems unnecessary, why not just have `convert_optional_traversal_to_compound_match_query` return `list[MatchQuery]` ?

graphql_compiler/compiler/emit_match.py
```diff
@@ -198,3 +198,51 @@ def emit_code_from_ir(match_query):
...
+    #
+    #    $result = UNIONALL($optional__0, $optional__1, . . . )
+
+    match_queries = compound_match_query.match_queries
+    if len(match_queries) == 1:
+        query_string = emit_code_from_match_query(match_queries[0])
+
+    elif len(match_queries) > 1:
+        optional_variable_base_name = 'optional__'
+        union_variable_name = 'result'
```
jssandh2: Unnecessary. Just create `query_string` up to `LET `.

graphql_compiler/compiler/ir_lowering_match.py
amartyashankha: You can fight this one out with @obi1kenobi  ;)

graphql_compiler/compiler/blocks.py
```diff
@@ -243,6 +244,10 @@ def validate(self):
 
         validate_safe_string(self.edge_name)
 
+    def get_field_name(self):
+        """Return the field name corresponding to the edge being traversed."""
+        return u'{}_{}'.format(self.direction, self.edge_name)
```
obi1kenobi: @jssandh2 named parameters are slightly nicer because it's harder to flip the order by accident, and it's more obvious when it happens. This is probably fine too, but if someone has named them already then let's leave it be in the future.

graphql_compiler/compiler/blocks.py
```diff
@@ -265,6 +270,15 @@ def to_gremlin(self):
                         direction=self.direction,
                         edge_name=self.edge_name,
                         edge_quoted=safe_quoted_string(self.edge_name)))
+        elif self.in_optional_context:
+            # During a traversal, the pipeline element may be null.
+            # The following code returns null when the current pipeline entity is null
+            # (an optional edge did not exist at some earlier traverse).
+            # Otherwise it performs a normal traversal (previous optional edge did exist).
```
obi1kenobi: Appreciate the comment! Can we make sure there's a test that hits this branch?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -382,6 +382,8 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         in_topmost_optional_block = False
 
         edge_traversal_is_optional = optional_directive is not None
+        in_optional_context = 'optional' in context and not edge_traversal_is_optional
```
obi1kenobi: Is this correct? What if we have an optional traversal that is already inside of an optional context -- won't this end up being `False` in that case? That seems questionable.

In either case, let's cover it with a test.

Also, I'm not sure what the desired semantics of `in_optional_context` are -- is it meant to only be True when within an existing optional context, but not when opening the initial optional context? A comment + maybe a variable name tweak might be useful.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -681,21 +684,27 @@ def _compile_output_step(outputs):
...
-                                     u'{}'.format(output_context))
-
             _, field_name = location.get_location_name()
             expression = expressions.FoldedOutputContextField(
                 fold_scope_location, field_name, graphql_type)
+
+            if optional:
+                existence_check = expressions.ContextFieldExistence(
+                    fold_scope_location.base_location
+                )
```
obi1kenobi: pull up this `)` to the previous line

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -681,21 +684,27 @@ def _compile_output_step(outputs):
         fold_scope_location = output_context['fold']
 
         expression = None
+        existence_check = None
         if fold_scope_location:
-            if optional:
-                raise AssertionError(u'Unreachable state reached, optional in fold: '
```
obi1kenobi: We're still not allowing `@optional` inside a `@fold`, no? I'm confused about why this is deleted.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -681,21 +684,27 @@ def _compile_output_step(outputs):
...
             if optional:
                 existence_check = expressions.ContextFieldExistence(location.at_vertex())
-                expression = expressions.TernaryConditional(
-                    existence_check, expression, expressions.NullLiteral)
+
+        # pylint: disable=redefined-variable-type
+        if existence_check:
+            expression = expressions.TernaryConditional(
+                existence_check, expression, expressions.NullLiteral)
+        # pylint: enable=redefined-variable-type
```
obi1kenobi: Overall, let's try to pull as much of this new code out of this one massive function and into helper functions whenever reasonable and possible -- the function is getting pretty unwieldy.

graphql_compiler/compiler/emit_match.py
```diff
@@ -198,3 +198,51 @@ def emit_code_from_ir(match_query):
     query_data.appendleft(_construct_output_to_match(match_query.output_block))
 
     return u' '.join(query_data)
+
+
+def emit_code_from_ir(compound_match_query):
+    """Return a MATCH query string from a list of MATCH queries (compound match query)."""
```
obi1kenobi: `a list of MATCH queries (compound match query)` is confusing. Just reference the `CompoundMatchQuery` namedtuple, and people can look up what that is when they are curious.

graphql_compiler/compiler/compiler_frontend.py
amartyashankha: But we might be allowing `@fold` within `@optional`?
The other case should raise other errors.

obi1kenobi: 👍 Let's just make sure to test these situations thoroughly.

graphql_compiler/compiler/compiler_frontend.py
amartyashankha: Were you referring to `_compile_output_step`? It seems to be doing only one thing?

obi1kenobi: No, the weird folding Github does made me think this was still part of `_compile_vertex_ast` or one of those big functions. No worries, ignore this.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -42,10 +42,15 @@
 
     step V-4. Recurse into any vertex field children of the current AST node:
               (see _compile_vertex_ast())
-        - before recursing into each vertex, process any @optional directive
-          present on the child AST node;
-        - after returning from each vertex, return to the marked query location using
-          the appropriate manner, depending on whether @optional was present on the child or not.
+          - before recursing into each vertex:
```
obi1kenobi: All indentation here is two spaces deeper than the corresponding bullet points in V-3.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -382,6 +387,12 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         in_topmost_optional_block = False
 
         edge_traversal_is_optional = optional_directive is not None
+
+        # This is true for any vertex expanded within an @optional scope.
+        within_optional_scope = 'optional' in context and not edge_traversal_is_optional
+        # Currently @optional is not allowed within @optional.
```
obi1kenobi: It's nice to keep comments right above the thing they are referring to. This is because right now, someone could accidentally put something on line 395 and make it confusing what the comment refers to.

I'd move lines 393-394 to right above 392.

graphql_compiler/compiler/emit_match.py
```diff
@@ -198,3 +198,64 @@ def emit_code_from_ir(match_query):
...
+    #    $result = UNIONALL($optional__0, $optional__1, . . . )
+
+    match_queries = compound_match_query.match_queries
+    if len(match_queries) == 1:
+        query_string = emit_code_from_single_match_query(match_queries[0])
+    elif len(match_queries) > 1:
+        query_string = emit_code_from_multiple_match_queries(match_queries)
+    else:
+        raise AssertionError(u'Received CompoundMatchQuery with an empty list of MatchQueries'
+                             u': {}'.format(match_queries))
```
obi1kenobi: Move the `: ` part to the previous line, and make this line be `u'{}'.format(match_queries))`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +390,139 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                field_name = step.root_block.get_field_name()
+                new_predicate = filter_local_field_existence(field_name)
+                old_filter = new_traverse[-1].where_block
+                if old_filter:
+                    new_predicate = BinaryComposition(u'&&', old_filter.predicate, new_predicate)
+                new_traverse[-1] = new_traverse[-1]._replace(where_block=Filter(new_predicate))
+                return new_traverse
+        raise AssertionError(u"No optional traverse found in: {}".format(traverse))
+
+    def make_mandatory_traverse(traverse):
```
obi1kenobi: Can you move these four helper functions outside of this function? I'd rather not worry about accidentally leaking / shadowing variables across nested function scopes.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +390,139 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                new_traverse.append(step)
+            else:
+                field_name = step.root_block.get_field_name()
+                new_predicate = filter_local_field_existence(field_name)
+                old_filter = new_traverse[-1].where_block
+                if old_filter:
+                    new_predicate = BinaryComposition(u'&&', old_filter.predicate, new_predicate)
+                new_traverse[-1] = new_traverse[-1]._replace(where_block=Filter(new_predicate))
+                return new_traverse
+        raise AssertionError(u"No optional traverse found in: {}".format(traverse))
```
obi1kenobi: Kensho style is to only use `"` in docstrings, and always use `'` quotes for strings.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +390,139 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                elif isinstance(expression, TernaryConditional):
+                    location_name, _ = expression.if_true.location.get_location_name()
+                    if location_name in current_locations:
+                        if location_name in current_non_optional_locations:
+                            new_output_fields[output_name] = expression.if_true
+                        else:
+                            new_output_fields[output_name] = expression
+                else:
+                    raise AssertionError(u'Invalid expression of type {} in output block'
+                                         u': {}'.format(type(expression).__name__, output_block))
```
obi1kenobi: move the `: ` to the previous line, so this line starts with `u'{}'`

PR #68, opened 2018-04-18
Optional traversal fixes

amartyashankha:


graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -523,6 +527,76 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
         return CompoundMatchQuery(match_queries=match_queries)
 
 
+def collect_filters_to_first_location_instance(compound_match_query):
+    """Collapse all filters applied to a particular location into a single `as_block`.
+
+    Specifically, for any location `L`, find all filters present on `L`, and apply
+    the conjunction of all these filters to the first occurence of `L` (within its MatchQuery).
+    """
+    def predicate_list_to_where_block(predicate_list):
```
obi1kenobi: let's avoid having nested functions in general whenever possible, and use private (underscored) helper functions instead

graphql_compiler/compiler/blocks.py
```diff
@@ -207,6 +207,10 @@ def to_gremlin(self):
         mark_name, _ = self.location.get_location_name()
         return u'as({})'.format(safe_quoted_string(mark_name))
 
+    def __hash__(self):
```
obi1kenobi: It's a bit weird to have this one specific subclass of BasicBlock be hashable, and nothing else. Let's discuss this in person and see if we can avoid this.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -390,6 +390,53 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
     return new_folded_ir_blocks
 
 
+def filter_local_field_existence(field_name):
```
obi1kenobi: since all four of these functions are just private (file-local) helper functions, make their names start with underscores

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -523,6 +527,76 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
         return CompoundMatchQuery(match_queries=match_queries)
 
 
+def collect_filters_to_first_location_instance(compound_match_query):
+    """Collapse all filters applied to a particular location into a single `as_block`.
+
+    Specifically, for any location `L`, find all filters present on `L`, and apply
```
obi1kenobi: If multi-line docstring is needed, always add `Args:` and `Returns:` sections

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -523,6 +527,76 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+
+    new_match_queries = []
+    # Each MatchQuery is processed independently
+    for match_query in compound_match_query.match_queries:
+        location_to_predicates = {}
+        # Construct a dictionary mapping locations --> a list of predicates
+        # applied to the corresponding location (in `where_blocks`)
+        for match_traversal in match_query.match_traversals:
+            for match_step in match_traversal:
+                curr_filter = match_step.where_block
```
obi1kenobi: avoid abbreviations like `curr`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -523,6 +527,76 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+            return predicate_list[0]
+        else:
+            return BinaryComposition(u'&&',
+                                     predicate_list_to_where_block(predicate_list[1:]),
+                                     predicate_list[0])
+
+    new_match_queries = []
+    # Each MatchQuery is processed independently
+    for match_query in compound_match_query.match_queries:
+        location_to_predicates = {}
```
obi1kenobi: this seems to be keyed on `MarkLocation` and not on `Location` -- if you make it use `Location` then `MarkLocation` won't have to be hashable

PR #69, opened 2018-04-23
Optional traversal fixes

amartyashankha:
Added support for `@recurse` and `@tag` within `@optional`.
Allowing expansion of multiple vertex fields within `@optional`.


graphql_compiler/compiler/blocks.py
```diff
@@ -412,3 +423,11 @@ class Unfold(MarkerBlock):
     def validate(self):
         """Unfold blocks are always valid in isolation."""
         pass
+
+
+class EndOptional(MarkerBlock):
+    """A marker for the end of a @fold context."""
```
obi1kenobi: `@fold` is copy-pasta

graphql_compiler/compiler/blocks.py
```diff
@@ -412,3 +423,11 @@ class Unfold(MarkerBlock):
...
+
+    def validate(self):
+        """Unfold blocks are always valid in isolation."""
```
obi1kenobi: more copy-pasta

graphql_compiler/compiler/blocks.py
```diff
@@ -412,3 +423,11 @@ class Unfold(MarkerBlock):
     def validate(self):
         """Unfold blocks are always valid in isolation."""
         pass
+
+
+class EndOptional(MarkerBlock):
```
obi1kenobi: Since there's no `StartOptional` marker block, describe how optionals are considered started for the purposes of this marker.

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -204,3 +205,55 @@ def extract_folds_from_ir_blocks(ir_blocks):
                 remaining_ir_blocks.append(block)
 
     return folds, remaining_ir_blocks
+
+
+def extract_location_to_optional_from_ir_blocks(ir_blocks):
```
obi1kenobi: I don't get how the name of this function corresponds to the docstring. I don't see anything related to `@optional` in there

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -204,3 +205,55 @@ def extract_folds_from_ir_blocks(ir_blocks):
...
+            encountered_traverse_within_optional = False
+        elif isinstance(block, MarkLocation):
+            location_to_optional[block.location] = in_optional_location
+
+        if not isinstance(block, EndOptional):
+            remaining_ir_blocks.append(block)
+
+        previous_block = block
+
+    return optional_locations, location_to_optional, remaining_ir_blocks
```
obi1kenobi: the docstring says a tuple of two things is returned

graphql_compiler/tests/test_input_data.py
```diff
@@ -1637,3 +1637,257 @@ def traverse_and_optional_and_traverse():
...
+        type_equivalence_hints=None)
+
+
+def simple_optional_recurse():
+    graphql_input = '''{
+        Animal {
+            name @output(out_name: "name")
+            out_Animal_ParentOf @optional {
+                name @output(out_name: "parent_name")
+                in_Animal_ParentOf @recurse(depth:3) {
```
obi1kenobi: space after `:`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+def _filter_local_field_existence(field_name):
+    """Return an Expression that is True iff `field_name` does not exist."""
+    local_field = LocalField(field_name)
+    local_field_size = UnaryTransformation(u'size', local_field)
+    field_null_check = BinaryComposition(u'=', local_field, NullLiteral)
+    field_size_check = BinaryComposition(u'=', local_field_size, ZeroLiteral)
+    return BinaryComposition(u'||', field_null_check, field_size_check)
+
+
+def _prune_traverse_using_omitted_locations(traverse,
```
obi1kenobi: avoid stacking function parameters and function call arguments vertically like this, and put them side by side on consecutive lines instead

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                                            omitted_locations,
+                                            optional_locations,
+                                            location_to_optional):
+    """Return a prefix of the given traverse, excluding any blocks after an omitted optional."""
+    new_traverse = []
+    for step in traverse:
+        if isinstance(step.root_block, Traverse) and step.root_block.optional:
+            in_optional_location = location_to_optional.get(
```
obi1kenobi: this is pretty hard to read -- extract `step.as_block.location` to a variable and then use the variable instead

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                step.as_block.location, None)
+            if in_optional_location in omitted_locations:
+                # Add filter to indicate that the omitted edge(s) shoud not exist
+                field_name = step.root_block.get_field_name()
+                new_predicate = _filter_local_field_existence(field_name)
+                old_filter = new_traverse[-1].where_block
+                if old_filter:
```
obi1kenobi: `if old_filter is None` presumably?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                    new_predicate = BinaryComposition(u'&&', old_filter.predicate, new_predicate)
+                new_traverse[-1] = new_traverse[-1]._replace(
+                    where_block=Filter(new_predicate))
+                break
```
obi1kenobi: Why `break`? It seems like this depends on some invariant that is not obvious to me, and should have an explicit check and associated `AssertionError` if not true.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+
+def convert_optional_traversals_to_compound_match_query(
+        match_query, optional_locations, location_to_optional):
     """Return 2^n distinct MatchQuery objects in a CompoundMatchQuery.
 
     Args:
         match_query: MatchQuery object potentially containing n `@optional` scopes
-            which expand vertex fields
+                     which expand vertex fields
+        optional_locations: List of locations with @optional that expand vertex fields within.
```
obi1kenobi: List -> list

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+        location_to_optional: Dict mappingall locations within optional scopes
```
obi1kenobi: Dict -> dict

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                new_traverse.append(step._replace(root_block=new_root_block))
+            else:
+                new_traverse.append(step)
+        else:
+            new_traverse.append(step)
+
+    return new_traverse
+
+
+def convert_optional_traversals_to_compound_match_query(
```
obi1kenobi: Try to improve the docstring. There is very little detail and context as to what is going on here, and some grammar issues in the args.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
-
-    def make_mandatory_traverse(traverse):
-        """Return a new traversal, without the optional condition."""
-        new_traverse = []
-        for step in traverse:
-            if not isinstance(step.root_block, Traverse) or not step.root_block.optional:
-                new_traverse.append(step)
+    optional_location_subsets = itertools.chain(
+        *[itertools.combinations(optional_locations, x)
+            for x in range(0, len(optional_locations)+1)]
```
obi1kenobi: The formatting here is off. Also, there is a lot of complexity in this expression -- let's break it up a few lines with well-named variables.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+    )
+    compound_match_traversals = []
+    for omitted_locations in reversed(list(optional_location_subsets)):
+        new_match_traversals = []
+        for traverse in match_query.match_traversals:
+            location = traverse[0].as_block.location
+            if location_to_optional.get(location, None) not in omitted_locations:
+                new_traverse = _prune_traverse_using_omitted_locations(
+                    traverse,
```
obi1kenobi: avoid vertical stacks like this

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -445,71 +447,81 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
-        if is_optional_traverse(traversal):
-            current_posibilities = [
-                remove_optional_traverse(traversal),
-                make_mandatory_traverse(traversal)
-            ]
-        else:
-            current_posibilities = [traversal]
-        match_traversals_possibilities.append(current_posibilities)
-    compound_match_traversals = list(itertools.product(*match_traversals_possibilities))
+                continue
```
obi1kenobi: explain why `continue` is the right thing to do, this is not obvious

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +590,167 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+                                 _predicate_list_to_where_block(predicate_list[1:]),
+                                 predicate_list[0])
+
+
+def collect_filters_to_first_location_instance(compound_match_query):
+    """Collate all filters for a particular location to the first instance of the location."""
+    new_match_queries = []
+    # Each MatchQuery is processed independently
+    for match_query in compound_match_query.match_queries:
+        location_to_predicates = {}
```
obi1kenobi: This function has too much complexity, and is very hard to read. Break it up into smaller functions, and document them more thoroughly.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +590,167 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        if location_name not in current_locations:
+            return TrueLiteral
+        else:
+            return expression
+    elif isinstance(expression, TernaryConditional):
+        if expression.predicate == TrueLiteral:
+            return expression.if_true
+        else:
+            return expression
+    elif isinstance(expression, Filter):
```
obi1kenobi: `Filter` is a `Block`, and not an `Expression`. This looks like a type confusion bug.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +590,167 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    else:
+        new_match_queries = []
+        for match_query in compound_match_query.match_queries:
+            match_traversals = match_query.match_traversals
+            current_locations = set()
+
+            for traversal in match_traversals:
+                for step in traversal:
+                    if step.as_block is not None:
+                        location_name, _ = step.as_block.location.get_location_name()
```
obi1kenobi: Any time you get 5 indentation levels deeper than where the function started, there is too much complexity and the function should be refactored and broken up. This is especially the case if 3 of the indentation levels are `for` loops.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -726,6 +899,8 @@ def lower_ir(ir_blocks, location_types, type_equivalence_hints=None):
     sanity_check_ir_blocks_from_frontend(ir_blocks)
 
     # These lowering / optimization passes work on IR blocks.
+    optional_locations, location_to_optional, ir_blocks \
```
obi1kenobi: `\` should be avoided pretty much always. Instead, assign the result value to an intermediate variable, then destructure it on the next line.

graphql_compiler/compiler/ir_lowering_match.py
amartyashankha: The `if` condition is entered only if it corresponds to one of the omitted optionals.
Everything following is within that optional, and can be disregarded.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
         return CompoundMatchQuery(match_queries=match_queries)
 
 
+def _construct_location_to_predicate_list(match_query):
+    """Return a dict mapping location -> list of filter predicates applied at that location."""
```
obi1kenobi: `filter predicates` is somewhat confusing and ill-defined. When you separate out the predicate expression from a `Filter` block, calling it a `predicate` is a bit questionable since it's not clear to what and where it applies.

I am also confused about why a list is necessary, when there is a lowering pass that merges filters that appear at the same location. Are we duplicating code here? That pass is very similar the code around line 624 here.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    if len(predicate_list) == 1:
+        return predicate_list[0]
+    else:
+        return BinaryComposition(u'&&',
+                                 _predicate_list_to_where_block(predicate_list[1:]),
+                                 predicate_list[0])
+
+
+def _collect_filters_to_first_location_in_match_traversal(match_traversal, location_to_predicates):
+    """Compose all filters for a specific location into its first occurence in given traversal."""
```
obi1kenobi: This docstring is confusing to me, I don't understand what's going on. Can you write a full docstring, with args/returns sections and everything?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+            coerce_type_block=match_step.coerce_type_block,
+            where_block=where_block,
+            as_block=match_step.as_block
+        )
+        new_match_traversal.append(new_match_step)
+    return new_match_traversal
+
+
+def collect_filters_to_first_location_instance(compound_match_query):
+    """Collate all filters for a particular location to the first instance of the location."""
```
obi1kenobi: I don't understand this docstring either. Can you write a full docstring here too? Also, do you really mean `collate` or is that a typo?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    new_match_queries = []
+    # Each MatchQuery is processed independently
```
obi1kenobi: Make this a proper sentence, put a period in the end.

Bonus points if you explain why the independence is the case, since that'd be the real value of the comment -- the for-loop already suggests that each match query is processed independently.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    for match_query in compound_match_query.match_queries:
+        location_to_predicates = _construct_location_to_predicate_list(match_query)
+
+        new_match_traversals = []
+        for match_traversal in match_query.match_traversals:
+            new_match_traversal = _collect_filters_to_first_location_in_match_traversal(
+                match_traversal, location_to_predicates)
+            new_match_traversals.append(new_match_traversal)
+        new_match_queries.append(
```
obi1kenobi: add a blank line above this, for readability

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+                match_traversals=new_match_traversals,
+                folds=match_query.folds,
+                output_block=match_query.output_block
+            )
+        )
+
+    return CompoundMatchQuery(match_queries=new_match_queries)
+
+
+def _update_tagged_expression(expression, current_locations):
```
obi1kenobi: `_update_tagged_expression` is an extremely vague name for what this function does. Can you try to find a better one?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    """Replace non-existent tag Expressons with True, and simplify the result."""
```
obi1kenobi: Full docstring here as well, please. This logic is far from obvious, and I bet nobody else would get how this function fits into the big picture here.

A few other specific concerns:
- There is no such thing as a `tag expression`, so this is confusing. All tag info has been stripped by the time we have IR, and tags are now `ContextField` expressions.
- This function doesn't seem to take any arguments that would specify which "tags" (presumably ContextFields?) do or don't exist, so it's not clear how it works.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+                return TrueLiteral
+            if expression.left == TrueLiteral:
+                return expression.right
+            if expression.right == TrueLiteral:
+                return expression.left
+            else:
+                return expression
+        else:
+            return expression
+        location_name, _ = context_field.location.get_location_name()
```
obi1kenobi: This is quite confusing. There are a ton of `return` statements above this, including one in an `else` clause right above, and yet here we have more code. It's confusing and far from obvious that two of the if-elif cases actually fall through to here.

A few suggestions:
- Break up the `if/elif/else` cascade into clearly disjoint cases, then cut off the biggest subtrees and make them helper functions.
- Make helpers for "BinaryComposition with a ContextField as the left/right expression", "BinaryComposition with &&", "BinaryComposition with ||", "TernaryConditional".
- Add checks and AssertionErrors for expressions that you don't expect to see, and therefore don't handle. This includes not just `BetweenClause` but also things like `OutputContextField`, `FoldedOutputContextField` etc.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+            raise AssertionError(u'Found BetweenClause with ContextFields as lower/upper bounds. '
+                                 u'THis should never happen: {}'.format(expression))
+    else:
+        return expression
+
+
+def _construct_update_tags_visitor_fn(current_locations):
+    """Return an Expression updater using the given `current_locations`."""
+    def visitor_fn(expression):
+        return _update_tagged_expression(expression, current_locations)
```
obi1kenobi: Given that `current_locations` doesn't change, but `expression` does, `_update_tagged_expression` should take these arguments in the opposite order.

With that, this entire function simply becomes `functools.partial(_update_tagged_expression, current_locations)` and there is no need for a separate function like this.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    return visitor_fn
+
+
+def _get_current_locations_from_match_traversals(match_traversals):
```
obi1kenobi: The docstring is great -- why don't we adopt its naming everywhere and refer to `present_locations` rather than `current_locations` everywhere?

Bonus points if you write a full docstring and explain how come locations can be absent.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    """Return a set of locations present in the given match traversals."""
```
obi1kenobi: `a set` -> `the set`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        new_match_traversals.append(new_match_traversal)
+    return new_match_traversals
+
+
+def lower_tagged_expressions_in_compound_match_query(compound_match_query):
+    """Lower Expressons involving non-existent ContextFields (tags)."""
+    if len(compound_match_query.match_queries) == 0:
+        raise AssertionError(u'Received CompoundMatchQuery with an empty list of MatchQueries.')
+    elif len(compound_match_query.match_queries) == 1:
+        # All ContextFields exist if there is only one MatchQuery
```
obi1kenobi: period at the end of the sentence

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+                new_step = step._replace(where_block=new_filter)
+            else:
+                new_step = step
+            new_match_traversal.append(new_step)
+        new_match_traversals.append(new_match_traversal)
+    return new_match_traversals
+
+
+def lower_tagged_expressions_in_compound_match_query(compound_match_query):
+    """Lower Expressons involving non-existent ContextFields (tags)."""
```
obi1kenobi: Try not to refer to tags in files that deal with IR, since the IR does not contain such a concept. This includes in function names, in comments, and in docstrings.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -578,6 +593,194 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    for traversal in match_traversals:
+        for step in traversal:
+            if step.as_block is not None:
+                location_name, _ = step.as_block.location.get_location_name()
+                current_locations.add(location_name)
+    return current_locations
+
+
+def _lower_filters_in_match_traversals(match_traversals, visitor_fn):
+    """Return new match traversals, replacing filters involving non-existent tags with True."""
```
obi1kenobi: `replacing filters involving non-existent tags with True` is somewhat misleading -- if the only filter is one with a `ContextField`, it will not be replaced with `True` but eliminated altogether.

Please write a full docstring and explain these subtleties.

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -204,3 +206,63 @@ def extract_folds_from_ir_blocks(ir_blocks):
...
+
+def extract_location_to_optional_from_ir_blocks(ir_blocks):
+    """Construct a mapping from locations within @optional to their correspoding optional Traverse.
+
+    Args:
+        ir_blocks: list of IR blocks to extract optional data from
+
+    Returns:
+        tuple (optional_locations, location_to_optional, new_ir_blocks):
+        - optional_locations: List of @optional locations that expand vertex fields
```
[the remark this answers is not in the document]
amartyashankha: Not quite ... `location_to_optional.values` may contain optional locations that do not expand vertex fields. It doesn't need to, but this makes generating the dict easier.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+
+
+def _simplify_ternary_conditional(expression):
+    """Return the `if_true` clause if the predicate of the TernaryConditional is a TrueLiteral.
+
+    Args:
+        expression: TernaryConditional to be simplified.
+
+    Returns:
+        simplified TernaryConditional, if the predicate is True,
```
obi1kenobi: this is misleading -- if the predicate was `TrueLiteral`, the return isn't a `TernaryConditional` at all

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        raise AssertionError(u'Found unexpected expression of type {}. This should never happen: '
+                             u'{}'.format(type(expression).__name__, expression))
+    elif isinstance(expression, no_op_blocks):
+        return expression
+    else:
+        raise AssertionError(u'Found unexpected expression of type {}. This should never happen: '
+                             u'{}'.format(type(expression).__name__, expression))
+
+
+def _construct_update_context_field_visitor_fn(present_locations):
```
obi1kenobi: I think I had a comment on this in a previous iteration that didn't get addressed. I would do two things here:
- make `_update_context_field_expression` take the `present_locations` as the first argument, since it's the one that is less likely to change
- then, this entire function is replaceable by just `partial(_update_context_field_expression, present_locations)` so you can just put that code into the place where this function is invoked

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        present_locations: set of all locations present in the given match traversals
+    """
+    present_locations = set()
+
+    for match_traversal in match_traversals:
+        for step in match_traversal:
+            if step.as_block is not None:
+                location_name, _ = step.as_block.location.get_location_name()
+                present_locations.add(location_name)
+    return present_locations
```
obi1kenobi: add a blank line before this

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    TernaryConditionals, where the predicate is lowerd to a TrueLiteral,
+    are replaced by their if_true predicate.
+    The `visitor_fn` implements these behaviors (see `_update_context_field_expression`).
+
+    Args:
+        match_traversals: list of match traversal enities to be lowered
+        visitor_fn: visit_and_update function for lowering expressions in given match traversal
+
+    Returns:
+        new_match_traversals: new list of match_traversals, with all filter expressions lowered
```
obi1kenobi: no need to name the return argument

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+            if step.where_block is not None:
+                new_filter = step.where_block.visit_and_update_expressions(visitor_fn)
+                if new_filter.predicate == TrueLiteral:
+                    new_filter = None
+                new_step = step._replace(where_block=new_filter)
+            else:
+                new_step = step
+            new_match_traversal.append(new_step)
+        new_match_traversals.append(new_match_traversal)
+    return new_match_traversals
```
obi1kenobi: add some blank lines to make this method easier to read

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+
+
+def lower_context_field_expressions_in_compound_match_query(compound_match_query):
+    """Lower Expressons involving non-existent ContextFields."""
+    if len(compound_match_query.match_queries) == 0:
+        raise AssertionError(u'Received CompoundMatchQuery with an empty list of MatchQueries.')
+    elif len(compound_match_query.match_queries) == 1:
+        # All ContextFields exist if there is only one MatchQuery.
```
obi1kenobi: Elaborate? I think I know why, but I don't think I'll remember my reasoning in six months.

graphql_compiler/tests/test_compiler.py
```diff
@@ -2868,3 +2868,907 @@ def test_traverse_and_optional_and_traverse(self):
...
+                        {{
+                            class: Animal,
+                            as: Animal___1
+                        }}
+                        RETURN $matches
+                    )
+            ),
+            $optional__1 = (
+                SELECT Animal___1.name AS `animal_name`,
+                        Animal__in_Animal_ParentOf___1.name AS `parent_name`,
```
obi1kenobi: bad indentation, and also doesn't follow the style of other MATCH queries in this file

graphql_compiler/tests/test_compiler.py
```diff
@@ -2868,3 +2868,907 @@ def test_traverse_and_optional_and_traverse(self):
...
+
+    def test_optional_traversal_and_optional_without_traversal(self):
+        test_data = test_input_data.optional_traversal_and_optional_without_traversal()
+
+        expected_match = '''
+            SELECT EXPAND($result)
+            LET
+            $optional__0 = (
+                SELECT Animal___1.name AS `animal_name`,
+                        if(
```
obi1kenobi: more bad indentation

graphql_compiler/tests/test_compiler.py
```diff
@@ -2868,3 +2868,907 @@ def test_traverse_and_optional_and_traverse(self):
...
+                        }}.in('Animal_ParentOf') {{
+                            optional: true,
+                            as: Animal__in_Animal_ParentOf___1
+                        }} ,
+                        {{
+                            class: Animal,
+                            as: Animal___1
+                        }}
+                        RETURN $matches
+                    )
```
obi1kenobi: Note how here there's a double de-indent -- that means that the indentation is messed up somewhere, since all of the paren + de-indents should come in one at a time. This issue appears several more times in this file.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -451,127 +455,215 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+    optional_location_subsets = list(itertools.chain(*optional_location_combinations_list))
+
+    compound_match_traversals = []
+    for omitted_locations in reversed(optional_location_subsets):
+        new_match_traversals = []
+        for match_traversal in match_query.match_traversals:
+            location = match_traversal[0].as_block.location
+            location_in_dict = location in location_to_optional
+
+            if not location_in_dict or location_to_optional[location] not in omitted_locations:
```
obi1kenobi: This seems to be checking that `location` exists within `location_to_optional`, then checking the value grabbed from that dict. Since `location_to_optional` is not allowed to have `None` values, why not write as:

```
optional_location = location_to_optional.get(location, None)
if optional_location is None or optional_location not in omitted_locations:
    ...
``

File: graphql_compiler/compiler/ir_lowering_match.py [hunk 46]
```diff
@@ -451,127 +455,215 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+    the resulting match traversals may have sections of the query omitted.
+    These locations will not be included in the returned `present_locations`.
+    All of the above locations that are not optional traverse locations
+    will be included in present_non_optional_locations.
+
+    Args:
+        match_traversals: one possible list of match traversals generated from a query
+            containing @optional traversal(s)
+
+    Returns:
```
obi1kenobi: You need to state that the return type is a tuple `(present_locations, present_non_optional_locations)`.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -451,127 +455,215 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+        present_locations: set of all locations present in the given match traversals
+        present_non_optiona_locations: set of all locations present in the given match traversals
```
obi1kenobi: typo: `present_non_optiona_locations`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -451,127 +455,215 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+            that are not reached through optional traverses.
```
obi1kenobi: So is `present_non_optional_locations` a subset of `present_locations`?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -451,127 +455,215 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
-                                             .format(expression.location, current_locations))
+                                             u'present_locations: {}'
+                                             .format(expression.location, present_locations))
                     new_output_fields[output_name] = expression
                 elif isinstance(expression, TernaryConditional):
                     location_name, _ = expression.if_true.location.get_location_name()
-                    if location_name in current_locations:
-                        if location_name in current_non_optional_locations:
+                    if location_name in present_locations:
+                        if location_name in present_non_optional_locations:
```
obi1kenobi: If `present_non_optional_locations` is guaranteed to be a subset of `present_locations` as I asked above, then the `if` on line 659 isn't necessary, no?

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +628,305 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        # Apply all filters for a location to the first occurence of that location
+        if match_step.as_block.location in location_to_predicates:
+            where_block = Filter(
+                _predicate_list_to_where_block(
+                    location_to_predicates[match_step.as_block.location]
+                )
+            )
+            # Delete the location entry. No further filters needed for this location.
+            # If the same location is found in another call to this function,
+            # no filters will be added.
```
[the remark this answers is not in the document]
obi1kenobi: @amartyashankha still confused about this ^

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +676,282 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        )
+
+    return CompoundMatchQuery(match_queries=new_match_queries)
+
+
+def _update_context_field_binary_composition(expression, present_locations):
+    """Lower BinaryCompositions involving non-existent ContextFields to True.
+
+    Args:
+        expression: BinaryComposition with at least one ContextField operand
```
obi1kenobi: If you require that the `BinaryComposition` be such that it has `at least one ContextField operand`, you should verify that precondition in the function. Right now, the function is just as happy with any kind of BinaryComposition, and will always return a value.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +676,282 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+            return TrueLiteral
+
+    return expression
+
+
+def _simplify_non_context_field_binary_composition(expression):
+    """Return a simplified BinaryComposition if either operand is a TrueLiteral.
+
+    Args:
+        expression: BinaryComposition without any ContextField operand(s)
```
obi1kenobi: you should also assert this precondition

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +676,282 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        and the original TernaryConditional otherwise
+    """
+    if expression.predicate == TrueLiteral:
+        return expression.if_true
+    else:
+        return expression
+
+
+def _update_context_field_expression(present_locations, expression):
+    """Lower Expressions involving non-existent ContextFields True, and simplify result."""
```
obi1kenobi: typo: `involving non-existent ContextFields True`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +676,282 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+                             u'{}'.format(type(expression).__name__, expression))
+
+
+def _lower_filters_in_match_traversals(match_traversals, visitor_fn):
+    """Return new match traversals, lowering filters involving non-existent ContextFields.
+
+    Expressions involving non-existent ContextFields are evaluated to True.
+    BinaryCompositions, where one of the operands is lowered to a TrueLiteral,
+    are lowered appropriately based on the present operator (u'||' and u'&&' are affected).
+    TernaryConditionals, where the predicate is lowerd to a TrueLiteral,
```
obi1kenobi: typo: `lowerd`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +676,282 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+    are replaced by their if_true predicate.
+    The `visitor_fn` implements these behaviors (see `_update_context_field_expression`).
+
+    Args:
+        match_traversals: list of match traversal enities to be lowered
```
obi1kenobi: typo: `enities`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -584,6 +676,282 @@ def prune_output_blocks_in_compound_match_query(compound_match_query):
...
+        new_match_traversals.append(new_match_traversal)
+
+    return new_match_traversals
+
+
+def lower_context_field_expressions_in_compound_match_query(compound_match_query):
+    """Lower Expressons involving non-existent ContextFields."""
+    if len(compound_match_query.match_queries) == 0:
+        raise AssertionError(u'Received CompoundMatchQuery with '
+                             u'an empty list of MatchQuery objects.')
```
obi1kenobi: nit: this is claiming it's an empty list, but it only checks the size part and not the list part -- include the actual received object in the error message to guard against accidentally misleading future debuggers of this code

PR #70, opened 2018-04-24
Lowering eligible comparisons to `BETWEEN` expressions.

amartyashankha:


graphql_compiler/compiler/expressions.py
```diff
@@ -794,3 +794,58 @@ def to_gremlin(self):
...
+
+        Args:
+            field: LocalField Expression, denoting the field in consideration
+            lower_bound: lower bound constraint for given field
+            upper_bound: upper bound constraint for given field
+        Returns:
+            a new BetweenClause object
+        """
+        super(BetweenClause, self).__init__(self, field, lower_bound, upper_bound)
+        self.field = field
```
lodrion: use underscores for the private variables?

amartyashankha: I was following https://github.com/kensho-technologies/graphql-compiler/blob/optional_traversal/graphql_compiler/compiler/expressions.py#L613.
These are not necessarily private variables I think.

obi1kenobi: Yeah this isn't necessarily private, lowering steps might refer to this variable.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +436,114 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+                             u'{}'.format(expression_list))
+    elif len(expression_list) == 1:
+        return expression_list[0]
+    else:
+        remaining_conjunction = _expression_list_to_conjunction(expression_list[1:])
+        return BinaryComposition(u'&&', expression_list[0], remaining_conjunction)
+
+
+def _extract_conjuction_elements_from_expression(expression):
+    """Return a list of expressions that are connected by `&&`s in the given expression."""
```
obi1kenobi: This is not a list anymore, the docstring is misleading.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +436,114 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+            dict mapping local field names to "operator => BinaryComposition" dictionaries,
+            for each BinaryComposition operator involving the LocalField
+        remaining_expression_list:
+            list of remaining expressions that were *not*
+            BinaryCompositions on a LocalField using any of the specified operators
+    """
+    local_field_to_expressions = {}
+    remaining_expression_list = deque([])
+    for expression in expression_list:
+        if all([isinstance(expression, BinaryComposition),
```
obi1kenobi: Prefer to use tuples rather than lists for `all()`:
`if all((option_a, option_b, option_c)):`

It's also a little strange that only `expression.left` is considered. Can we make this symmetrical? Otherwise, a lowering change elsewhere could make this pass stop working.

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +436,114 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+    else:
+        yield expression
+
+
+def _construct_field_operator_expression_dict(expression_list, operators):
+    """Construct a mapping from local fields to specified operators, and corresponding expressions.
+
+    Args:
+        expression_list: list of expressions to analyze
+        operators: list of BInaryComposition operators used to select the "important" expressions
```
obi1kenobi: `BInaryComposition`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -378,6 +436,114 @@ def remove_backtrack_blocks_from_fold(folded_ir_blocks):
...
+
+    Returns:
+        local_field_to_expressions:
+            dict mapping local field names to "operator => BinaryComposition" dictionaries,
```
obi1kenobi: nit: use a `->` instead of `=>`, I think we use `->` elsewhere

graphql_compiler/tests/test_ir_generation.py
```diff
@@ -565,6 +565,126 @@ def test_between_filter_on_datetime(self):
...
 
+    def test_between_lowering_on_simple_scalar(self):
+        test_data = test_input_data.between_lowering_on_simple_scalar()
+
+        base_location = helpers.Location(('Animal',))
+
+        expected_blocks = [
+            blocks.QueryRoot({'Animal'}),
+            blocks.Filter(
+                    expressions.BinaryComposition(
```
obi1kenobi: are all of these indents double-width?

graphql_compiler/tests/test_input_data.py
```diff
@@ -319,6 +319,85 @@ def between_filter_on_datetime():
...
+    return CommonTestData(
+        graphql_input=graphql_input,
+        expected_output_metadata=expected_output_metadata,
+        expected_input_metadata=expected_input_metadata,
+        type_equivalence_hints=None)
+
+
+def no_between_lowering_on_simple_scalar():
+    # The "between" filter emits different output depending on what the compared types are.
+    # This test checks for correct code generation when the type is a simple scalar (a String).
```
obi1kenobi: I know we discussed this in person, but can you add an explanation of why we decided not to lower to `BETWEEN` in this situation? Neither of us will remember the conversation or our reasoning in 6 months' time.

PR #72, opened 2018-04-30
Added documentation for @optional traversal

amartyashankha:


README.md
```diff
@@ -27,13 +27,13 @@ It's modeled after Python's `json.tool`, reading from stdin and writing to stdou
...
   * [Definitions](#definitions)
   * [Directives](#directives)
-     * [@optional](#optional)
-     * [@output](#output)
-     * [@fold](#fold)
-     * [@tag](#tag)
-     * [@filter](#filter)
-     * [@recurse](#recurse)
-     * [@output_source](#output_source)
+     * [@optional](#-optional)
```
obi1kenobi: Why make these changes? They seem to affect documentation unrelated to the info being added.

README.md
```diff
@@ -47,6 +47,7 @@ It's modeled after Python's `json.tool`, reading from stdin and writing to stdou
   * [The GraphQL schema](#the-graphql-schema)
   * [Execution model](#execution-model)
   * [Miscellaneous](#miscellaneous)
+     * [Traversing within `@optional`](#traversing-within-optional-optional-)
```
obi1kenobi: why the trailing `-` in the link?

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
...
+}.in('Animal_ParentOf') {
+    optional: true,
+    as: Animal__in_Animal_ParentOf___1
+}.in('Animal_ParentOf') {
+    as: Animal__in_Animal_ParentOf__in_Animal_ParentOf___1
+}
+```
+
+Therefore we need a different way to represent traversals within an `@optional` scope.
+Going forwards, we will refer to two different kinds of `@optional` directives.
```
obi1kenobi: `forwards` -> `forward` I think

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
...
+        }
+    }
+}
+```
+
+We can represent a *compound* optional by taking an union (`UNIONALL`) of two distinct
+`MATCH` queries.  For instance, the GraphQL query above can be represented as follows:
+```
+LET
+
```
obi1kenobi: I think this might look better with these extra blank lines removed.

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
...
+    }.in('Animal_ParentOf') {
+        as: Animal__in_Animal_ParentOf__in_Animal_ParentOf___1
+    }
+),
+
+$final_match = UNIONALL($match1, $match2)
+```
+Notice that in the first case where the optional edge is not followed,
+we have to explicitly filter out all vertices where the edge *could have been folowed*.
+This is to eliminate duplicates between the two `MATCH` blocks.
```
obi1kenobi: `MATCH blocks` is a bit confusing. How about `MATCH selections` instead?

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
...
+(we also have `SELECT` statements within `$match{1/2}`), but it illustrates the point.
+
+#### Performance Penalty
+
+If we have many *compound* optionals in the given `GraphQL`,
+the above procedure results in the union of a large number of `MATCH` queries.
+Specifically, for `n` compound optionals, we generate 2<sup>n</sup> different `MATCH` queries.
+For each of the 2<sup>n</sup> subsets `S` of the `n` optional edges:
+- We remove the `@optional` restriction for each traversal in `S`.
+- For each traverse `t` in the complement of `S`, we entirely  discard `t`
```
obi1kenobi: double space after `entirely`

graphql_compiler/compiler/ir_lowering_match.py
```diff
@@ -9,17 +9,17 @@
 to simplify the final code generation step.
 """
 
-from collections import namedtuple
+from collections import deque, namedtuple
```
obi1kenobi: Can you split out these changes into a separate PR, since they are unrelated to the documentation?

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
...
+  along with all the vertices and directives within it, and we add a filter
+  on the previous traverse to ensure that the edge corresponding to `t` does not exist.
+
+So, essentially we get a performance penalty that growes exponentially
```
obi1kenobi: I would replace `So, essentially` with `Therefore,`

obi1kenobi:
Adding @MichaelaShtilmanMinkin since she had fantastic comments on the last large batch of documentation, and I think her perspective would be very useful here as well.

README.md
amartyashankha: The links were broken. Just `#optional` doesn't go anywhere. The correct link is `#-optional`.

README.md
amartyashankha: I believe it is because of the trailing backtick, but this is the correct link.

README.md
```diff
@@ -166,7 +167,12 @@ For each `Animal`:
 
 #### Constraints and Rules
 - `@optional` can only be applied to vertex fields, except the root vertex field.
+- It is allowed to expand vertex fields within an `@optional` scope.
+  However, doing so is currently associated with a performance penalty in `MATCH`.
+  For more detail, see: [Traversing within `@optional`](#traversing-within-optional-optional-).
 - May not exist at the same vertex field as `@recurse`, `@fold`, or `@output_source`.
+- `@optional`, `@output_source` and `@fold` may not be used anywhere within a scope
```
MichaelaShtilmanMinkin: I think there's some overlap with previous documentation in here in the line before (or maybe it's the phrasing?)

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
 
 ## Miscellaneous
 
+### Traversing within [`@optional`](#optional)
+
+OrientDB MATCH currently alows the last step in any traversal to be optional.
```
MichaelaShtilmanMinkin: `alows` -> `allows`

README.md
```diff
@@ -956,6 +961,124 @@ the opposite order:
...
+This is a nice tranlation of `@optional` traversals in a GraphQL query,
```
MichaelaShtilmanMinkin: `tranlation` -> `translation`

README.md
amartyashankha: I take it back. This only happens on my browser for some reason :/

README.md
amartyashankha: More issues with my browser. Will fix.

README.md
```diff
@@ -956,6 +961,117 @@ the opposite order:
...
+    }.in('Animal_ParentOf') {
+        as: Animal__in_Animal_ParentOf___1
+    }.in('Animal_ParentOf') {
+        as: Animal__in_Animal_ParentOf__in_Animal_ParentOf___1
+    }
+),
+$final_match = UNIONALL($match1, $match2)
+```
+In the first case where the optional edge is not followed,
+we have to explicitly filter out all vertices where the edge *could have been folowed*.
```
MichaelaShtilmanMinkin: `folowed` -> `followed` 

Please double check for typos. A fast way to do this is to copy/paste this into something that has automatic spell check, like Word.

README.md
```diff
@@ -956,6 +961,117 @@ the opposite order:
...
+This is to eliminate duplicates between the two `MATCH` selections.
+
+The previous example is not *exactly* how we implement *compound* optionals
+(we also have `SELECT` statements within `$match1` and `$match2`), but it illustrates the point.
```
MichaelaShtilmanMinkin: Perhaps change to `the point` to `the general idea`?

README.md
```diff
@@ -956,6 +961,117 @@ the opposite order:
...
+
+#### Performance Penalty
+
+If we have many *compound* optionals in the given `GraphQL`,
```
MichaelaShtilmanMinkin: Could you re-examine how you use bullets here, please? I think you can leverage having parallel structure for sentences here. e.g.

```
- For each of the 2^n subsets....
- For each traverse `t`
```

Not sure how much it would affect meaning, though... Try it out?

PR #73, opened 2018-05-02
Added missing tests in `test_compiler` and `test_ir_generation`

amartyashankha:


graphql_compiler/tests/test_compiler.py
```diff
@@ -1151,6 +1151,147 @@ def test_in_collection_op_filter_with_optional_tag(self):
 
         check_test_data(self, test_data, expected_match, expected_gremlin)
 
+    def test_contains_op_filter_with_variable(self):
+        test_data = test_input_data.contains_op_filter_with_variable()
+
+        expected_match = '''
+            SELECT Animal___1.name AS `animal_name`
```
obi1kenobi: Please make the formatting of all new MATCH and Gremlin queries match the formatting of the other test cases.

obi1kenobi:
@jssandh2 I'm not sure I understand your request? Is it the same formatting thing I commented on, or is it something else?

graphql_compiler/tests/test_ir_generation.py
```diff
@@ -1692,6 +1692,98 @@ def test_fold_after_traverse(self):
...
+            blocks.Unfold(),
+            blocks.ConstructResult({
+                'animal_name': expressions.OutputContextField(
+                    base_location.navigate_to_field('name'), GraphQLString),
+                'sibling_and_self_species_list': expressions.FoldedOutputContextField(
+                    parent_fold, 'name', GraphQLList(GraphQLString)),
+            }),
+        ]
+        expected_location_types = {
+            # The folded location was never traversed to, so it does not appear here.
```
obi1kenobi: Comments like this were written before traversing inside a fold was allowed, and they aren't really true anymore -- lines 1735-1736 are both traverses inside a fold. Let's go with `No MarkLocation blocks are output within folded scopes.` instead, here and everywhere else.

graphql_compiler/tests/test_ir_generation.py
```diff
@@ -2176,8 +2176,7 @@ def test_coercion_on_interface_within_fold_scope(self):
             }),
         ]
         expected_location_types = {
-            # The folded location was never traversed to, so it does not appear here.
-            base_location: 'Animal',
```
obi1kenobi: did you mean to put this in the comment as well?

graphql_compiler/tests/test_ir_generation.py
```diff
@@ -2208,8 +2207,7 @@ def test_coercion_on_interface_within_fold_traversal(self):
...
```
obi1kenobi: same here

obi1kenobi:
The irony when you add tests, and coverage goes down :)

PR #74, opened 2018-05-07
Add schema validation

cw6515:


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -709,6 +710,40 @@ def _preprocess_graphql_string(graphql_string):
...
+        ast: abstract syntax tree representation of a graphql query
+
+    Returns:
+        list containing schema and/or query validation errors
+    """
+    core_graphql_errors = validate(schema, ast)
+
+    # Extract name, locations, and args keys in order to compare schema directives and directives
+    # which are supported by the graphql compiler.
+    graphql_compiler_directives = [(s.name, s.locations, s.args.keys()) for s in DIRECTIVES]
```
obi1kenobi: I'm a bit concerned about `s.args.keys()` since this code is meant to be py2 and py3 at the same time. Also, GraphQL directives don't prescribe argument order (unfortunately).

My recommendation would be to use `set(s.args.keys())`.

I would also make the list comprehension multi-line just like the one on 735-738, and maybe also use `directive` as the variable of the comprehension, for the sake of consistency.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -709,6 +710,40 @@ def _preprocess_graphql_string(graphql_string):
...
+
+    # Directives provided in the schema.
+    # pylint: disable=protected-access
```
obi1kenobi: If you access it via `get_directives()` then you won't need this suppression.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -709,6 +710,40 @@ def _preprocess_graphql_string(graphql_string):
...
+    schema_directives = [
+        (directive.name, directive.locations, directive.args.keys())
+        for directive in schema._directives
+    ]
+    # pylint: enable=protected-access
+    for directive in graphql_compiler_directives:
```
obi1kenobi: Seems like we're trying to compare two sets of declared directives for equality. Thoughts on doing that explicitly?

Specifically, I'm thinking we could make the two lists of directives (expected vs actual) into sets. Then, we can compute:
```
missing_directives = expected_directives - actual_directives
unexpected_directives = actual_directives - expected_directives
```
If either of these two sets is non-empty, we can complain with an appropriate error message and say specifically which directives are the issue.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -1034,3 +1036,35 @@ def test_invalid_edge_degree_queries(self):
         for invalid_graphql in invalid_queries:
             with self.assertRaises(GraphQLCompilationError):
                 graphql_to_ir(self.schema, invalid_graphql)
+
+    def test_directives_not_in_schema(self):
```
obi1kenobi: I'd also add maybe three more tests:
- unexpected directive that the compiler doesn't know what to do with
- recognized directive but with wrong "locations" (FIELD vs INLINE_FRAGMENT vs both)
- recognized directive but with too many / too few / wrong arguments

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -709,6 +710,40 @@ def _preprocess_graphql_string(graphql_string):
     return graphql_string + '\n'
 
 
+def _validate_schema_and_ast(schema, ast):
+    """Validate the supplied graphql schema and ast.
+
+    This method wraps around graphql-core's validation to enforce a stricter requirement of the
+    schema-- all directives supported by the compiler must be declared by the schema, regardless of
```
obi1kenobi: nit: I'd add a space before the `--` as well as after

PR #76, opened 2018-05-09
Restructuring `MATCH` and `Gremlin` lowering into respective directories

amartyashankha:


graphql_compiler/compiler/common.py
```diff
@@ -1,7 +1,7 @@
 # Copyright 2017 Kensho Technologies, LLC.
 from collections import namedtuple
 
-from . import emit_gremlin, emit_match, ir_lowering_gremlin, ir_lowering_match
```
obi1kenobi: I would keep the names the same. We aren't lowering Gremlin or MATCH -- we are lowering IR for the sake of converting into Gremlin or MATCH.

graphql_compiler/compiler/match_lowering/match_between_lowering.py
```diff
@@ -0,0 +1,121 @@
+from collections import deque
```
obi1kenobi: I would prefer to not include `match` in the name of this file -- the name of the directory (package) that it's in already communicates that adequately. I think `between_lowering.py` is probably fine.

graphql_compiler/compiler/match_lowering/match_optional_traversal.py
```diff
@@ -0,0 +1,558 @@
+from collections import namedtuple
```
obi1kenobi: same here, I wouldn't include `match` in the name of this file either

graphql_compiler/tests/test_emit_output.py
```diff
@@ -9,7 +9,7 @@
                                     OutputContextField, TernaryConditional, Variable)
 from ..compiler.helpers import Location
 from ..compiler.ir_lowering_common import OutputContextVertex
-from ..compiler.ir_lowering_match import CompoundMatchQuery
+from ..compiler.match_lowering.match_optional_traversal import CompoundMatchQuery
```
obi1kenobi: `CompoundMatchQuery` should be in `utils` together with `BetweenClause`

graphql_compiler/compiler/ir_lowering_gremlin/__init__.py
```diff
@@ -0,0 +1,50 @@
+from .ir_lowering import (lower_coerce_type_block_type_data, lower_coerce_type_blocks,
```
obi1kenobi: Add the `# Copyright` statement as the first line of every new file.

obi1kenobi: You can do it in a separate PR, I can merge this one.

PR #77, opened 2018-05-10
 Added docstrings, comments and changed variable names.

amartyashankha:


graphql_compiler/compiler/ir_lowering_match/optional_traversal.py
```diff
@@ -206,27 +217,34 @@ def prune_non_existent_outputs(compound_match_query):
...
             folds = match_query.folds
 
-            present_locations_tuple = _get_present_locations_from_match_traversals(match_traversals)
+            present_locations_tuple = _get_present_locations(match_traversals)
             present_locations, present_non_optional_locations = present_locations_tuple
 
             new_output_fields = {}
             for output_name, expression in six.iteritems(output_block.fields):
                 if isinstance(expression, OutputContextField):
+                    # An OutputContextField without a TernaryConditional should not be within
```
obi1kenobi: I see how we've established that the expression is an `OutputContextField` on the line above. I don't see where we've established that it is `without a TernaryConditional` though, and I'm confused.

PR #78, opened 2018-05-10
Bug for `@fold` and `@optional` in same scope.

amartyashankha:


graphql_compiler/tests/test_compiler.py
```diff
@@ -2724,3 +2724,60 @@ def test_coercion_filters_and_multiple_outputs_within_fold_traversal(self):
...
+                )
+            ])}
+        '''
+
+        check_test_data(self, test_data, expected_match, expected_gremlin)
+
+    def test_fold_and_optional(self):
+        test_data = test_input_data.fold_and_optional()
+
+        expected_match = '''
```
obi1kenobi: did you mean to leave these empty?

graphql_compiler/tests/test_input_data.py
```diff
@@ -1502,3 +1502,59 @@ def coercion_filters_and_multiple_outputs_within_fold_traversal():
...
+
+def fold_and_optional():
+    graphql_input = '''{
+        Animal {
+            name @output(out_name: "animal_name")
+            out_Animal_ParentOf @fold {
+                name @output(out_name: "child_names_list")
+            }
+            in_Animal_ParentOf @optional {
+                    name @output(out_name: "parent_name")
```
obi1kenobi: double indent here

graphql_compiler/tests/test_input_data.py
```diff
@@ -1502,3 +1502,59 @@ def coercion_filters_and_multiple_outputs_within_fold_traversal():
...
+        }
+    }'''
+    expected_output_metadata = {
+        'animal_name': OutputMetadata(type=GraphQLString, optional=False),
+        'parent_name': OutputMetadata(type=GraphQLString, optional=True),
+        'child_names_list': OutputMetadata(
+            type=GraphQLList(GraphQLString), optional=False),
+    }
+    expected_input_metadata = {
+    }
```
obi1kenobi: if no inputs, make this `{}` on the same line

graphql_compiler/tests/test_compiler.py
amartyashankha: Yes, this query actually causes a compilation error.

obi1kenobi: I'm not sure I understand -- compilation error as in a `raise` kind of error, or just a buggy compilation?

Do both of these testcases currently fail, or just one of them?

amartyashankha: The first one produces incorrect MATCH because of revisits.
The second one raises an `AssertionError`

graphql_compiler/tests/test_input_data.py
```diff
@@ -1502,3 +1502,57 @@ def coercion_filters_and_multiple_outputs_within_fold_traversal():
...
         expected_input_metadata=expected_input_metadata,
         type_equivalence_hints=None)
+
+
+def optional_and_fold():
+    graphql_input = '''{
+        Animal {
+            name @output(out_name: "animal_name")
+            in_Animal_ParentOf @optional {
+                    name @output(out_name: "parent_name")
```
obi1kenobi: this is doubly-indented

amartyashankha: Sorry, I thought I fixed that.

graphql_compiler/tests/test_compiler.py
obi1kenobi: Can you post all this info as a comment in the PR, including the specifics on why the MATCH is wrong and what the AssertionError is? This is a bug report, and it should be as detailed as possible so that other people that run into it can correctly triage and assess it -- and this will also help with fixing it.

obi1kenobi:
This is superseded by #86 right? Should I close it?

PR #86, opened 2018-05-21
Fixing @fold-@optional interaction bug.

amartyashankha:


graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -128,6 +124,19 @@ def _sanity_check_block_pairwise_constraints(ir_blocks):
                                      u'{}'.format(ir_blocks))
 
 
+def _sanity_check_mark_location_preceding_optional_traverse(ir_blocks):
+    """Assert that optional traverse blocks are preceded by a MarkLocation."""
```
obi1kenobi: `traverse` -> `Traverse` here as well

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -225,32 +223,41 @@ def extract_optional_location_root_info(ir_blocks):
...
-                raise AssertionError(u'in_optional_location was not None at an optional traverse: '
-                                     u'{} {}'.format(current_block, ir_blocks))
-            if not isinstance(previous_block, MarkLocation):
-                raise AssertionError(u'No MarkLocation found before an optional traverse: '
-                                     u'{} {}'.format(current_block, ir_blocks))
-            in_optional_location = previous_block.location
-        elif in_optional_location is not None and isinstance(current_block, (Traverse, Recurse)):
+            if in_optional_root_location is not None:
+                raise AssertionError(u'in_optional_root_location was not None at an optional '
+                                     u'traverse: {} {}'.format(current_block, ir_blocks))
```
obi1kenobi: This is my unfortunate code, but can you make `traverse` -> `Traverse` everywhere in the errors and docstrings since it's a type?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -398,7 +404,8 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
             # Make sure there's a tag right before it for the optional Backtrack to jump back to.
```
obi1kenobi: let's make `tag` -> `marked location` in this entire comment to avoid confusion with `@tag` directives

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -357,12 +357,21 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
         # Folded scopes don't actually traverse to the location, so it's never really visited.
         context['location_types'][location] = strip_non_null_from_type(current_schema_type)
         basic_blocks.append(_mark_location(location))
+        # The marked_location_stack explicitly maintains a stack (implemented as list)
+        # of dicts (each corresponding to a MarkLocation) containing:
+        #  - location: the location within the corresponding MarkLocation object
+        #  - num_traverses: the number of Recurse and Traverse blocks created
+        #                   after the corresponding MarkLocation
+        # The following append is the initial MarkLocation for the current vertex.
+        marked_location_stack.append({'location': location, 'num_traverses': 0})
```
obi1kenobi: Let's push `namedtuples` onto the stack instead of dicts. That way we know what keys they must have, at minimal extra cost.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -629,6 +654,7 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
         'inputs': dict(),
         'location_types': dict(),
         'type_equivalence_hints': type_equivalence_hints or dict(),
+        'marked_location_stack': []
```
obi1kenobi: I know the existing code doesn't do this (and that's a problem of its own), but let's start adding a short explanation of what each thing in the `context` does, what it is intended for, and what problem it solves. You can add this as a multi-line comment before the key.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -357,12 +357,21 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
                                                      inner_location, context)
         basic_blocks.extend(inner_basic_blocks)
 
+    marked_location_stack = context['marked_location_stack']
```
obi1kenobi: Rather than grabbing the reference like this, and counting on the reader to remember that `marked_location_stack` is actually part of the context, let's try to use `context['marked_location_stack']` directly each time unless there's very good reason not to. They technically do the same thing, but we want it to be impossible for the reader to miss the fact that the context is being read or mutated.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -393,15 +402,20 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         # This will need to change if nested @optionals have to be supported.
         within_optional_scope = 'optional' in context and not edge_traversal_is_optional
 
+        # Extracting the last inserted MarkLocation
+        current_marked_location = context['marked_location_stack'][-1]
```
obi1kenobi: Let's inline this so it's obvious that the check below uses the context, and that later on, the mutation affects an element on the stack.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -357,12 +357,21 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
         # We only mark the position if we aren't in a folded scope.
         # Folded scopes don't actually traverse to the location, so it's never really visited.
         context['location_types'][location] = strip_non_null_from_type(current_schema_type)
         basic_blocks.append(_mark_location(location))
+        # The marked_location_stack explicitly maintains a stack (implemented as list)
+        # of dicts (each corresponding to a MarkLocation) containing:
+        #  - location: the location within the corresponding MarkLocation object
+        #  - num_traverses: the number of Recurse and Traverse blocks created
+        #                   after the corresponding MarkLocation
+        # The following append is the initial MarkLocation for the current vertex.
```
obi1kenobi: This is misleading, there is no `MarkLocation` block here -- only a `Location` and an int counter.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -357,12 +357,21 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
 
+    marked_location_stack = context['marked_location_stack']
+
     # step V-3: mark the graph position, and process output_source directive
     if not is_in_fold_scope(context):
         # We only mark the position if we aren't in a folded scope.
         # Folded scopes don't actually traverse to the location, so it's never really visited.
         context['location_types'][location] = strip_non_null_from_type(current_schema_type)
         basic_blocks.append(_mark_location(location))
+        # The marked_location_stack explicitly maintains a stack (implemented as list)
```
obi1kenobi: This comment is good -- let's move it to the initial definition of the context.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -457,14 +476,20 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
                 # return to a position after the optional set of blocks.
                 location = location.revisit()
                 context['location_types'][location] = strip_non_null_from_type(current_schema_type)
                 basic_blocks.append(_mark_location(location))
+                marked_location_stack.pop()
+                marked_location_stack.append({'location': location, 'num_traverses': 0})
             else:
                 basic_blocks.append(blocks.Backtrack(location))
 
+    # Pop off the initial MarkLocation for the current vertex.
```
obi1kenobi: No `MarkLocation` is involved here -- same problem as above.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -457,14 +476,20 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
+    if not is_in_fold_scope(context):
+        marked_location_stack.pop()
```
obi1kenobi: For bonus points, here's a sanity check that I think would be nice to have:
- Record the length of the stack upon entering this function.
- Then, right after this `.pop()`, check that the current stack length is equal to the initial stack length at the start of the function. Since the comment here says we're popping off the initial location we pushed earlier, the lengths must always match.

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -313,16 +324,14 @@ def visitor_fn(expression):
             # If the root_block is a Backtrack, translate its Location if necessary.
             if isinstance(new_step.root_block, Backtrack):
                 old_location = new_step.root_block.location
-                if old_location in location_translations:
-                    new_location = location_translations[old_location]
-                    new_step = new_step._replace(root_block=Backtrack(new_location))
+                new_location = location_translations.get(old_location, old_location)
+                new_step = new_step._replace(root_block=Backtrack(new_location))
```
obi1kenobi: Why make this change? If the location of the `Backtrack` doesn't change, we don't need to call `._replace()` at all -- that's just wasteful in memory.

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -313,16 +324,14 @@ def visitor_fn(expression):
...
 
             # If the as_block exists, translate its Location if necessary.
             if new_step.as_block is not None:
                 old_location = new_step.as_block.location
-                if old_location in location_translations:
-                    new_location = location_translations[old_location]
-                    new_step = new_step._replace(as_block=MarkLocation(new_location))
+                new_location = location_translations.get(old_location, old_location)
```
obi1kenobi: Same comment here as above.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -395,13 +407,16 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
         if edge_traversal_is_optional:
             # Entering an optional block!
-            # Make sure there's a tag right before it for the optional Backtrack to jump back to.
-            # Otherwise, the traversal could rewind to an old tag and might ignore
-            # entire stretches of applied filtering.
-            if not isinstance(basic_blocks[-1], blocks.MarkLocation):
+            # Make sure there's a marked location right before it for the optional Backtrack
+            # to jump back to. Otherwise, the traversal could rewind to an old marked location
+            # and might ignore entire stretches of applied filtering.
+            if context['marked_location_stack'][-1].num_traverses != 0:
```
obi1kenobi: This should technically be `> 0` no? I realize they end up doing the same thing, but it's nice to be explicit.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -82,6 +82,13 @@
                       is_vertex_field_name, strip_non_null_from_type, validate_safe_string)
 
 
+# LocationStackEntry contains the following:
+# - location: Location object correspoding to an inserted MarkLocation block
+# - num_traverses: Int counter for the number of traverses intserted after the last MarkLocation
+#                  (corresponding Location stored in `location`)
+LocationStackEntry = namedtuple('LocationStackEntry', ('location', 'num_traverses'))
```
obi1kenobi: Bonus points if you make a helper function that makes these objects after asserting that `num_traverses` is non-negative and that `location` is a `Location` object. This helper would be the way to construct such objects everywhere, instead of explicitly calling the constructor or using `._replace()`

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -624,11 +665,32 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
     # Construct the start location of the query, and the starting context object.
     location = Location((base_start_type,))
     context = {
+        # 'tags' maps to a dict containing
```
obi1kenobi: describe what the key is: `'tags' is a dict mapping the tag name to a dict containing...`

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -624,11 +665,32 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
...
+        #  - location: Location where the tag was defined
+        #  - optional: boolean representing whether the tag was defined within an @optional scope
+        #  - type: GraphQLType of the tagged value
         'tags': dict(),
+        # 'outputs' maps to a dict mapping each output name to another dict which contains
```
obi1kenobi: I would generally avoid saying `maps to` where you mean `is`, since we're already talking about dicts mapping keys to values, and we want to minimize confusion.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -624,11 +665,32 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
...
+        #  - location: Location where to output from
+        #  - optional: boolean representing whether the output was defined within an @optional scope
+        #  - type: GraphQLType of the output
+        #  - fold: FoldScopeLocation object if the current output was defined within a fold scope,
+        #          and None otherwise
         'outputs': dict(),
         'inputs': dict(),
```
obi1kenobi: Here's a description for `inputs` you can use:
```
'inputs' is a dict mapping input parameter names to their respective expected GraphQL types,
as automatically inferred by inspecting the query structure

File: graphql_compiler/compiler/compiler_frontend.py [hunk 19]
```diff
@@ -624,11 +665,32 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
...
+        # 'location_types' maps to a dict mapping each Location to its GraphQLType
+        # (schema type of the location)
         'location_types': dict(),
+        # 'type_equivalence_hints' maps to a dict mapping GraphQL types to equivalent GraphQL unions
         'type_equivalence_hints': type_equivalence_hints or dict(),
+        # The marked_location_stack explicitly maintains a stack (implemented as list)
+        # of dicts (each corresponding to a MarkLocation) containing:
```
obi1kenobi: they aren't dicts any more

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -624,11 +665,32 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
...
+        #  - location: the location within the corresponding MarkLocation object
+        #  - num_traverses: the number of Recurse and Traverse blocks created
+        #                   after the corresponding MarkLocation
+        # The following append is the Location corresponding to the initial MarkLocation
```
obi1kenobi: what following append?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -447,10 +461,12 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
-            current_location_stack_entry = context['marked_location_stack'][-1]
-            current_num_traverses = current_location_stack_entry.num_traverses
-            new_location_stack_entry = current_location_stack_entry._replace(
-                num_traverses=current_num_traverses+1)
+            old_location_stack_entry = context['marked_location_stack'][-1]
+            old_location = old_location_stack_entry.location
+            old_num_traverses = old_location_stack_entry.num_traverses
+
+            new_location_stack_entry = _construct_location_stack_entry(
+                old_location, old_num_traverses+1)
```
obi1kenobi: You'll need spaces around the `+`, and ideally use `old_location_stack_entry.location` and `old_location_stack_entry.num_traverses + 1` directly here

PR #87, opened 2018-05-22
SQL generation scripts for snapshot tests DB

amartyashankha:


graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,71 @@
...
+random.seed(0)
+
+CREATE_VERTEX = 'create vertex '
+CREATE_EDGE = 'create edge '
+SEPARATOR = '__'
+
+
+def get_uuid():
+    """Return a pseudorandom uuid."""
+    random_digits = ''.join([str(random.randint(0, 9)) for _ in range(32)])  # nosec
```
obi1kenobi: `get_uuid()` is simply `str(UUID(int=random.randint(0, 2**128 - 1)))`, no need to format the uuid manually

graphql_compiler/scripts/generate_test_sql/__init__.py
```diff
@@ -0,0 +1,14 @@
+from .animals import get_animal_generators
+from .species import get_species_generators
+
+
+if __name__ == '__main__':
+    sql_generators = [
+        get_species_generators,
+        get_animal_generators,
+    ]
+    sql_generator_strings = []
```
obi1kenobi: why not just:
```
for sql_string in chain.from_iterable(sql_generators):
    print sql_string
```

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -0,0 +1,75 @@
+import random
+
+from .species import SPECIES_LIST
+from .utils import (create_edge_statement, create_name, create_vertex_statement, get_uuid,
+                    strip_index_from_name)
+
+
+random.seed(0)
```
obi1kenobi: Let's not mutate shared global state on import. Do this in the entry point function for the module instead.

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -0,0 +1,75 @@
...
+        species_to_names[species_name] = current_animal_names
+        generator_list.extend(_get_initial_animal_generators(species_name, current_animal_names))
+
+        for _ in range(NUM_GENERATIONS):
+            while True:
+                parent_names = frozenset(random.sample(current_animal_names, NUM_PARENTS))
+                # Duplicating a set of parents could result in Animals with the same names.
+                # This would invalidate unique selection of Animals by name.
+                if parent_names not in previous_parent_sets:
+                    break
```
obi1kenobi: This `while True:` then `break` on a non-obvious condition doesn't look great. What's the reason it has to be this way?

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -0,0 +1,75 @@
...
+            previous_parent_sets.add(parent_names)
+
+            parent_indices = [index for _, index in
```
obi1kenobi: remember how we format multi-line comprehensions

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,71 @@
...
+def select_vertex_statement(vertex_type, name):
+    """Return a SQL statement to select a vertex of given type by its `name` field."""
+    template = '(select from {vertex_type} where name = \'{name}\')'
+    args = {'vertex_type': vertex_type, 'name': name}
+    return template.format(**args)
+
+
+def set_statement(field_name, field_value):
+    """Return a SQL clause (used in creating a vertex) to set a field to a value."""
+    if not isinstance(field_name, basestring):
```
obi1kenobi: this doesn't work in py3, `basestring` is not a thing

graphql_compiler/tests/test_data_tools/data/commands.sql
```diff
@@ -0,0 +1,244 @@
+create vertex Species set name = 'Nazgul', uuid = '87425473-4595-2762-9989-378641469948'
```
obi1kenobi: Is this the output from the sql generator script? Add a comment header line or two, something like:
```
# Auto-generated output from <script>, do not edit directly!
# Generated on <datetime> from git revision <git sha>
```

graphql_compiler/tests/test_data_tools/schema.sql
```diff
@@ -0,0 +1,80 @@
...
+# this will ensure that no UniquelyIdentifiable is added without a valid uuid field
+ALTER PROPERTY UniquelyIdentifiable.uuid DEFAULT ''
+INSERT INTO UniquelyIdentifiable SET uuid = ''
+CREATE INDEX UniquelyIdentifiable.uuid UNIQUE_HASH_INDEX
+###############
+
+### Entity ###
+CREATE CLASS Entity EXTENDS V, UniquelyIdentifiable ABSTRACT
+CREATE PROPERTY Entity.name String
+CREATE INDEX Entity.name NOTUNIQUE
```
obi1kenobi: Did you want names to be unique? You seem to imply it in `animals.py` on line 55. If so, make this a unique index.

graphql_compiler/tests/test_data_tools/schema.sql
```diff
@@ -0,0 +1,80 @@
...
+CREATE CLASS Animal_BornAt EXTENDS E
+CREATE PROPERTY Animal_BornAt.in LINK Event
+CREATE PROPERTY Animal_BornAt.out LINK Animal
+CREATE INDEX Animal_BornAt ON Animal_BornAt (in, out) UNIQUE_HASH_INDEX
+###############
+
+### Species ###
+CREATE CLASS Species EXTENDS Entity
+
+CREATE PROPERTY Species.limbs Date
```
obi1kenobi: Date?

graphql_compiler/scripts/generate_test_sql/__init__.py
```diff
@@ -15,6 +15,4 @@
     ]
     sql_generator_strings = []
     for generator in sql_generators:
-        sql_generator_strings.extend(generator())
-
-    print '\n'.join(sql_generator_strings)
+        print(generator())
```
obi1kenobi: Two things:
- I think this doesn't do the right thing, since the generator returns a list, and we don't want to print a list. Instead, we need to print one-element-per-line from the list.
- This should be caught by a test. Let's test that the generated file stored in the repo is the same file that running  the generation script produces.

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -36,6 +36,16 @@ def _get_initial_animal_generators(species_name, current_animal_names):
     return generator_list
 
 
+def _get_new_parents(current_animal_names, previous_parent_sets):
+    """Return a list of `NUM_PARENTS` parent names that is not present in `previous_parent_sets`."""
```
obi1kenobi: let's make `NUM_PARENTS` an input variable, no need to introduce global state into this function

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -48,24 +58,19 @@ def get_animal_generators():
...
-                # Duplicating a set of parents could result in Animals with the same names.
-                # This would invalidate unique selection of Animals by name.
-                if parent_names not in previous_parent_sets:
-                    break
-            previous_parent_sets.add(parent_names)
+            new_parent_names = _get_new_parents(current_animal_names, previous_parent_sets)
+            previous_parent_sets.add(new_parent_names)
 
             parent_indices = [
                 index for _, index in [
```
obi1kenobi: this isn't formatted well

graphql_compiler/scripts/generate_test_sql/species.py
```diff
@@ -0,0 +1,28 @@
+import random
+
+from .utils import create_vertex_statement, get_uuid
+
+
+random.seed(0)
+
+SPECIES_LIST = [
```
obi1kenobi: make this a tuple -- ordered but immutable

graphql_compiler/scripts/generate_test_sql/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+if __name__ == '__main__':
+    sql_generators = [
+        get_species_generators,
+        get_animal_generators,
+    ]
+    sql_generator_strings = []
+    for generator in sql_generators:
+        print('\n'.join(generator()))
```
obi1kenobi: You'll need to do `sys.stdout.write()` here, the linter doesn't like the `print`. Also, I thought we were going to go with printing things as they are generated instead of printing one massive string at once?

graphql_compiler/scripts/generate_test_sql/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .animals import get_animal_generators
+from .species import get_species_generators
+
+
+random.seed(0)
+
+
+if __name__ == '__main__':
+    sql_generators = [
```
obi1kenobi: Let's move all this code into a main function, including the `random.seed(0)` call.

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -0,0 +1,82 @@
...
+    generator_list = []
+    for index in range(NUM_INITIAL_ANIMALS):
+        animal_name = create_name(species_name, str(index))
+        current_animal_names.append(animal_name)
+        generator_list.append(_create_animal_statement(animal_name))
+    return generator_list
+
+
+def _get_new_parents(current_animal_names, previous_parent_sets, num_parents):
+    """Return a list of `num_parents` parent names that is not present in `previous_parent_sets`."""
```
obi1kenobi: it returns a set, not a list, no?

graphql_compiler/scripts/generate_test_sql/animals.py
```diff
@@ -0,0 +1,82 @@
...
+
+        for _ in range(NUM_GENERATIONS):
+            new_parent_names = _get_new_parents(
+                current_animal_names, previous_parent_sets, NUM_PARENTS)
+            previous_parent_sets.add(new_parent_names)
+
+            parent_indices = [
+                index
+                for _, index in [
+                    strip_index_from_name(parent_name) for parent_name in new_parent_names
```
obi1kenobi: always put the `for` on a new line, unless it's an extremely short list comprehension

graphql_compiler/scripts/generate_test_sql/species.py
```diff
@@ -0,0 +1,28 @@
+import random
+
+from .utils import create_vertex_statement, get_uuid
+
+
+random.seed(0)
```
obi1kenobi: Importing a new file without doing anything with it shouldn't reset your random seed. I suspect this isn't even doing what you wanted -- all these are imported at the beginning of the generator program, which means there are back-to-back `random.seed(0)` calls with no random generations in between. I suspect that's not what you had in mind.

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    if not isinstance(field_value, six.string_types):
+        raise AssertionError(u'Expected string field_value. Received {}'.format(field_value))
+    template = '{} = \'{}\''
+    return template.format(field_name, field_value)
+
+
+def create_vertex_statement(vertex_type, fields_dict):
+    """Return a SQL statement to create a vertex."""
+    statement = CREATE_VERTEX + vertex_type
+    set_field_clauses = [set_statement(field_name, field_value)
```
obi1kenobi: reformat this

graphql_compiler/scripts/generate_test_sql/__init__.py
```diff
@@ -1,18 +1,35 @@
...
-from .species import get_species_generators
+from .animals import get_animal_generation_commands
+from .species import get_species_generation_commands
 
 
-random.seed(0)
+def main():
+    random.seed(0)
 
+    module_path = os.path.relpath(__file__)
```
obi1kenobi: `relpath` is relative to the current working directory, which can easily change as the user is using their shell. Let's either use `abspath` or use just the module name `generate_test_sql`.

graphql_compiler/scripts/generate_test_sql/__init__.py
```diff
@@ -1,18 +1,35 @@
-import itertools
+import datetime
+import git
```
obi1kenobi: This is a 3rd party library which is probably not going to be installed -- let's use the package version instead?

graphql_compiler/tests/test_data_tools/data/commands.sql
```diff
@@ -1,8 +1,11 @@
+# Auto-generated output from `graphql_compiler/scripts/generate_test_sql/__init__.py`.
+# Do not edit directly!
+# Generated on 2018-06-21T17:03:59.164223 from git revision 9b1f4eb3afc132535fbed9aa4d47c010bd397809.
+
 create vertex Species set name = 'Nazgul', uuid = 'e3e70682-c209-4cac-629f-6fbed82c07cd'
 create vertex Species set name = 'Pteranodon', uuid = '82e2e662-f728-b4fa-4248-5e3a0a5d2f34'
 create vertex Species set name = 'Dragon', uuid = 'd4713d60-c8a7-0639-eb11-67b367a9c378'
-create vertex Species set name = 'Hippogriff', uuid = '23a7711a-8133-2876-37eb-dcd9e87a1613'
-create vertex Animal set name = 'Nazgul__0', uuid = 'e6f4590b-9a16-4106-cf6a-659eb4862b21'
+create vertex Species set name = 'Hippogriff', uuid = '23a7711a-8133-2876-37eb-dcd9e87a1613'create vertex Animal set name = 'Nazgul__0', uuid = 'e6f4590b-9a16-4106-cf6a-659eb4862b21'
```
obi1kenobi: this seems like a bug

graphql_compiler/tests/test_data_tools/data/commands.sql
```diff
@@ -241,4 +244,4 @@ create edge Animal_OfSpecies from (select from Animal where name = 'Hippogriff__
 create edge Animal_OfSpecies from (select from Animal where name = 'Hippogriff__((431)(4(210)0)1)') to (select from Species where name = 'Hippogriff')
 create edge Animal_OfSpecies from (select from Animal where name = 'Hippogriff__((4(210)0)(210)1)') to (select from Species where name = 'Hippogriff')
 create edge Animal_OfSpecies from (select from Animal where name = 'Hippogriff__((431)(4(210)2)1)') to (select from Species where name = 'Hippogriff')
-create edge Animal_OfSpecies from (select from Animal where name = 'Hippogriff__((431)(4(210)0)(4(210)2))') to (select from Species where name = 'Hippogriff')
+create edge Animal_OfSpecies from (select from Animal where name = 'Hippogriff__((431)(4(210)0)(4(210)2))') to (select from Species where name = 'Hippogriff')
```
obi1kenobi: add a blank line at the end -- this is the same bug as above

.gitignore
```diff
@@ -25,6 +25,7 @@ nosetests.xml
 
 # Virtual environment
 **/venv/
+**/env/
```
obi1kenobi: Unless there is good reason to distinguish between `venv` and `env`, can we avoid adding this?

.travis.yml
```diff
@@ -7,8 +7,10 @@ python:
 install:
   - pip install -r dev-requirements.txt
   - pip install -e .
+before_script:
+  - docker-compose up &
```
obi1kenobi: This will make tests flaky since `&` doesn't guarantee that the containers are up before proceeding. What you want is `-d` instead, which will wait until the containers are up and *then* detach.

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,68 @@
+import random
+from uuid import UUID
+
+import six
+
+
+random.seed(0)
```
obi1kenobi: `random.seed()` calls should not be hit on import. Importing a file should not alter global state other than to put the functions and variables of that file into scope.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -0,0 +1,110 @@
...
+        for row in row_dicts
+    ]
+    row_frozensets = [frozenset(row_dict.items()) for row_dict in row_dicts_using_tuples]
+    rows_multiset = Counter(row_frozensets)
+    row_counters_frozenset = frozenset(rows_multiset.items())
+
+    return row_counters_frozenset
+
+
+class OrientdbMatchQueryTests(TestCase):
```
obi1kenobi: consider `OrientDBMatchQueryTests`

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -0,0 +1,110 @@
...
+    @pytest.mark.usefixtures('graph_client')
+    def test_optional_and_deep_traverse(self):
+        test_data = test_input_data.optional_and_deep_traverse()
+
+        rows = execute_graphql(self.schema, test_data, self.graph_client)
+
+        self.assertMatchSnapshot(rows)
+
+    @pytest.mark.usefixtures('graph_client')
+    def test_fold_on_output_variable(self):
```
obi1kenobi: Don't do this in this PR, but just for context:
- Eventually we'll want these functions to be programmatically generated for each possible test input.
- We should eventually have a way to identify which queries don't actually capture any data, and amend that. Randomly generated data is fine but probably doesn't hit all of the intricacies of the test cases.

graphql_compiler/tests/test_data_tools/schema.py
```diff
@@ -0,0 +1,21 @@
...
+from os import path
+
+
+def load_schema(client):
+    """Read the schema file and apply the specified SQL updates to the client."""
+    project_root = path.dirname(path.dirname(path.abspath(__file__)))
+    file_path = path.join(project_root, 'test_data_tools/schema.sql')
+    sql_files = glob(file_path)
+    if len(sql_files) == 0:
+        raise AssertionError(u'No schema file found. Expected /home/shankhabiswas/graphql-compiler/'
```
obi1kenobi: don't include your local path in the error message

graphql_compiler/tests/test_data_tools/schema.py
```diff
@@ -0,0 +1,21 @@
...
+                             u'graphql_compiler/tests/test_data_tools/schema.sql')
+
+    with open(file_path, 'r') as update_file:
```
obi1kenobi: this seems duplicated

orientdb-server-config.xml
```diff
@@ -0,0 +1,91 @@
...
+                </parameters>
+            </socket>
+        </sockets>
+        <protocols>
+            <protocol implementation="com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary" name="binary"/>
+            <protocol implementation="com.orientechnologies.orient.server.network.protocol.http.ONetworkProtocolHttpDb" name="http"/>
+        </protocols>
+        <listeners>
+            <listener protocol="binary" socket="default" port-range="2424-2430" ip-address="0.0.0.0"/>
+            <listener protocol="http" socket="default" port-range="2480-2490" ip-address="0.0.0.0">
```
obi1kenobi: make these `0.0.0.0` into `127.0.0.1` and see if that'll work?

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -13,8 +13,9 @@
 # pylint: disable=no-member
 
 
-def execute_graphql(schema, test_data, client):
+def execute_graphql(schema, test_data, client, sample_parameters=None):
```
obi1kenobi: Let's not make this optional. Executed queries should explicitly say that they require no parameters. Making it optional is making it easy for people to forget to set this.

.travis.yml
```diff
@@ -7,8 +7,10 @@ python:
 install:
   - pip install -r dev-requirements.txt
   - pip install -e .
+before_script:
+  - docker-compose up -d
 script:
-  - flake8 graphql_compiler/
+  - flake8 --exclude **/snap_*.py graphql_compiler/
```
obi1kenobi: consider marking `snap_*.py` as excluded from coverage as well -- there's no point in tracking their coverage

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def get_uuid():
+    """Return a pseudorandom uuid."""
+    return str(UUID(int=random.randint(0, 2**128 - 1)))  # nosec
+
+
+def get_random_net_worth():
+    """Return a pseudorandom uuid."""
+    return int(1e3*random.random() * 100) / 100.0  # nosec
```
obi1kenobi: spacing between operators -- also, any particular reason why `1e3` and `100` are separate?

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def get_random_limbs():
+    """Return a pseudorandom uuid."""
```
obi1kenobi: copy-pasta

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    return random.randint(2, 10)  # nosec
+
+
+def get_random_choice(choices):
```
obi1kenobi: This function seems not that useful -- it's literally the same thing as `random.choice`

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    args = {'vertex_type': vertex_type, 'name': name}
+    return template.format(**args)
+
+
+def set_statement(field_name, field_value):
+    """Return a SQL clause (used in creating a vertex) to set a field to a value."""
+    if not isinstance(field_name, six.string_types):
+        raise AssertionError(u'Expected string field_name. Received {}'.format(field_name))
+    field_value_representation = repr(field_value)
+    if isinstance(field_value, datetime.date):
```
obi1kenobi: Do `datetime.datetime` objects need similar special handling like this? E.g. another `isoformat()` call etc.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -0,0 +1,116 @@
...
+import pytest
+import six
+from snapshottest import TestCase
+
+from .. import test_input_data
+from ... import graphql_to_match
+from ..test_helpers import get_schema
+
+
+# pylint: disable=no-member
```
obi1kenobi: why? this isn't great

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -0,0 +1,116 @@
...
+    # For this we perform the following steps
+    # - convert lists (returned from @fold scopes) to tuples to make them hashable
+    # - convert each row dict to a frozenset of its items
+    # - create a Counter (multi-set) of the row frozensets
+    # - convert the multi-set to a frozenset of its items
+    row_dicts = [row.oRecordData for row in client.command(result.query)]
+    row_dicts_using_tuples = [
+        {
+            column_name: tuple(value) if isinstance(value, list) else value
+            for (column_name, value) in row.items()
```
obi1kenobi: no need for parens around `column_name, value`

orientdb-server-config.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
obi1kenobi: Is this file necessary?

orientdb.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/sh
```
obi1kenobi: Is this file still necessary?

graphql_compiler/scripts/generate_test_sql/utils.py
```diff
@@ -16,20 +16,15 @@ def get_uuid():
...
 def get_random_net_worth():
-    """Return a pseudorandom uuid."""
-    return int(1e3*random.random() * 100) / 100.0  # nosec
+    """Return a pseudorandom net worth."""
+    return int(1e5 * random.random()) / 100.0  # nosec
 
 
 def get_random_limbs():
-    """Return a pseudorandom uuid."""
+    """Return a pseudorandom numper of limbs."""
```
obi1kenobi: typo: `numper`

PR #91, opened 2018-05-24
Add support for filtering and output of Decimal-typed fields.

### LINKED ISSUE #90: GraphQLDecimal not implemented
We will need to create a new scalar type to support this.

https://github.com/graphql-python/graphql-core/issues/24

obi1kenobi:
Resolves #90.

### DIFF
--- a/graphql_compiler/query_formatting/gremlin_formatting.py
+++ b/graphql_compiler/query_formatting/gremlin_formatting.py
@@ -1,6 +1,7 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Safely represent arguments for Gremlin-language GraphQL queries."""
 import datetime
+from decimal import Decimal, InvalidOperation
 import json
 from string import Template
 
@@ -11,7 +12,7 @@ import six
 from ..compiler import GREMLIN_LANGUAGE
 from ..compiler.helpers import strip_non_null_from_type
 from ..exceptions import GraphQLInvalidArgumentError
-from ..schema import GraphQLDate, GraphQLDateTime
+from ..schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal
 from .representations import represent_float_as_str, type_check_and_str
 
 
@@ -46,6 +47,22 @@ def _safe_gremlin_string(value):
 
     final_escaped_value = '\'' + re_escaped + '\''
     return final_escaped_value
+
+
+def _safe_gremlin_decimal(value):
+    """Represent decimal objects as Gremlin strings."""
+    if isinstance(value, Decimal):
+        # We got a decimal object, serialize it directly.
+        # The "G" suffix on a decimal number forces it to be a BigInteger/BigDecimal literal:
+        # http://docs.groovy-lang.org/next/html/documentation/core-syntax.html#_number_type_suffixes
+        return str(value) + 'G'
+    else:
+        # We didn't get a decimal object, try to make it into one and then serialize it.
+        try:
+            coerced_decimal = Decimal(value)
+        except InvalidOperation as e:
+            raise GraphQLInvalidArgumentError(e)
+        return _safe_gremlin_decimal(coerced_decimal)
 
 
 def _safe_gremlin_date_and_datetime(graphql_type, expected_python_types, value):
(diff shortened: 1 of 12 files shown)


graphql_compiler/tests/test_end_to_end.py
```diff
@@ -11,32 +12,41 @@
...
         name @output(out_name: "name")
+        net_worth @filter(op_name: ">=", value: ["$min_worth"])
     }
 }'''
 
 
 class QueryFormattingTests(unittest.TestCase):
     def test_correct_arguments(self):
         wanted_name = 'Top Cat'
+        min_worth = Decimal('123456789.0123456')
```
amartyashankha: Would it be better to have the decimal argument test separate from the string argument test?

graphql_compiler/tests/test_input_data.py
```diff
@@ -37,6 +37,47 @@ def immediate_output():
...
+        expected_output_metadata=expected_output_metadata,
+        expected_input_metadata=expected_input_metadata,
+        type_equivalence_hints=None)
+
+
+def immediate_output_with_custom_scalar_filter():
+    graphql_input = '''{
+        Animal {
+            name @output(out_name: "animal_name")
+            net_worth @filter(op_name: ">=", value: ["$min_worth"])
```
amartyashankha: Mostly irrelevant: ?aybe add a birthday filter here as well (to parallel the previous test)? Of course date filters are tested elsewhere too. So, not really necessary

graphql_compiler/query_formatting/gremlin_formatting.py
```diff
@@ -48,6 +49,22 @@ def _safe_gremlin_string(value):
...
 
+def _safe_gremlin_decimal(value):
+    """Represent decimal objects as Gremlin strings."""
+    if isinstance(value, Decimal):
+        # We got a decimal object, serialize it directly.
+        # The "G" suffix on a decimal number forces it to be a BigInteger/BigDecimal literal:
+        # http://docs.groovy-lang.org/next/html/documentation/core-syntax.html#_number_type_suffixes
+        return str(value) + 'G'
+    else:
+        # We didn't get a decimal object, try to make it into one and then serialize it.
```
amartyashankha: In what situation does this happen?

graphql_compiler/tests/test_end_to_end.py
obi1kenobi: Here we're not testing with a string argument at all -- we only test with a `Decimal` argument. For anything else, if Python knows how to turn it into a `Decimal` argument -- great, and if not we get an error.

graphql_compiler/tests/test_input_data.py
obi1kenobi: I would prefer not to, since having two filters in the same query would implicitly test other code (e.g. the lowering steps that merge consecutive filters). It's nice to keep this test small and limited to as little code as possible, so we can pinpoint issues more easily.

graphql_compiler/query_formatting/gremlin_formatting.py
obi1kenobi: If e.g. a user used a JSON-based endpoint to submit a GraphQL query, since JSON doesn't support `Decimal` natively, they may have used a `float` or `string` representation instead. If the `Decimal` constructor is able to parse the number correctly, we do that just to be easier on the user since we already have type metadata available.

graphql_compiler/tests/test_end_to_end.py
amartyashankha: No, I meant the previous version of the test was only for `name_or_alias` string argument i.e. "Top Cat".
I was just wondering if it's better to test that separately from the decimal formatting.

graphql_compiler/query_formatting/match_formatting.py
```diff
@@ -50,6 +51,20 @@ def _safe_match_date_and_datetime(graphql_type, expected_python_types, value):
     return _safe_match_string(serialized_value)
 
 
+def _safe_match_decimal(value):
+    """Represent decimal objects as MATCH strings."""
+    if isinstance(value, Decimal):
```
lodrion: this code has been repeated twice, ideally there would be a function 

```
def coerce_decimal(value):
  if isinstance(value, Decimal):
     return value
  else:
    try:
￼       return Decimal(value)
       except InvalidOperation as e:
￼     raise GraphQLInvalidArgumentError(e)
```
that gets called from both gremlin and match generator

obi1kenobi:
Coveralls seems to have a bug -- the run completed and says it successfully submitted its coverage info, but Coveralls says it never got it. The issue is between the Coveralls client and server, so ignoring it for now.

PR #95, opened 2018-06-01
Orientdb @optional bugfix

amartyashankha:


graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -177,14 +177,18 @@ def __init__(self, simple_optional_root_info):
...
             edge_field = root_info_dict['edge_field']
 
-            optional_edge_filter = _filter_orientdb_simple_optional_edge(
+            optional_edge_where_filter = _filter_orientdb_simple_optional_edge(
                 root_location.query_path, inner_location_name, edge_field)
-            where_filter_expressions.append(optional_edge_filter)
+            inner_location_name_to_where_filter[inner_location_name] = optional_edge_where_filter
+
+        # Sort expressions by inner_location_name to obtain deterministic order
+        where_filter_expressions = [inner_location_name_to_where_filter[key]
```
obi1kenobi: please reformat:
```
where_filter_expressions = [
    inner_location_name_to_where_filter[key]
    for key in sorted(inner_location_name_to_where_filter.keys())
]
```

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -60,6 +85,117 @@ def to_gremlin(self):
...
+    # When an edge does not exist at a given vertex, OrientDB represents that in one of two ways:
+    #   - the edge's field does not exist (is null) on the vertex document, or
+    #   - the edge's field does exist, but is an empty list.
+    # We check both of these possibilities.
+    if not isinstance(edge_expression, (LocalField, SelectEdgeContextField)):
+        raise AssertionError(u'Received invalid edge_expression {} of type {}.'
+                             u'Expected LocalField or SelectEdgeContextField.'
+                             .format(edge_expression, type(edge_expression).__name__))
+    if isinstance(edge_expression, LocalField):
+        if edge_expression.field_name[:3] != u'in_' and edge_expression.field_name[:4] != u'out_':
```
obi1kenobi: can we extract this into a helper function, since it's used in a few different places now?

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -60,6 +85,117 @@ def to_gremlin(self):
...
+def _filter_orientdb_simple_optional_edge(root_location_path, inner_location_name, edge_field):
+    """Return an Expression that is False for rows that don't follow the @optional specification.
+
+    OrientDB does not filter correctly within optionals. Namely, a result where the optional edge
+    DOES EXIST will be returned regardless of whether the inner filter is satisfed.
+    To mitigate this, we add a final filter to reject such results.
+    A valid result must satisfy either of the following:
+    - The location within the optional exists (the filter will have been applied in this case)
+    - The optional edge field does not exist at the root location of the optional traverse
+    So, if the inner location within the optional was never visited, it muct be the case that
```
obi1kenobi: typo: `muct`

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -60,6 +85,117 @@ def to_gremlin(self):
...
+            OR
+            (Animal__out_Animal_ParentOf___1 IS NOT null)
+        )
+
+    Args:
+        root_location_path: tuple of strings representing the path to the root (preceding location)
+                            of the optional traversal
+        inner_location_name: string representing location within the corresponding optional traverse
+        edge_field: string representing optional traverse edge field
+    Returns:
```
obi1kenobi: blank line before `Returns` -- this is the sort of thing that I hope you'd catch yourself when you read over your code one more time before opening the PR

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -60,6 +85,117 @@ def to_gremlin(self):
...
+        Expression that evaluates to False for rows that do not follow the @optional specification
+    """
+    inner_local_field = LocalField(inner_location_name)
+    inner_location_existence = BinaryComposition(u'!=', inner_local_field, NullLiteral)
+
+    edge_field_location = Location(root_location_path, field=edge_field)
```
obi1kenobi: We're passing both of these values separately as function args, then constructing a `Location` from them. Why not pass a `Location` then read the two elements from it? Seems more robust and cleaner.

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -60,6 +85,117 @@ def to_gremlin(self):
...
+            inner_location_name_to_where_filter[inner_location_name] = optional_edge_where_filter
+
+        # Sort expressions by inner_location_name to obtain deterministic order
+        where_filter_expressions = [
+            inner_location_name_to_where_filter[key]
+            for key in sorted(inner_location_name_to_where_filter.keys())
+        ]
+
+        predicate = expression_list_to_conjunction(where_filter_expressions)
+        super(SelectWhereFilter, self).__init__(predicate)
```
obi1kenobi: Let's make this take a predicate expression instead, and move the logic elsewhere. I'm not a fan of having tons of complex logic in a constructor like this.

graphql_compiler/compiler/match_query.py
```diff
@@ -141,16 +143,27 @@ def _split_match_steps_into_match_traversals(match_steps):
...
-    output_block = ir_blocks[-1]
+    where_block = ir_blocks[-1]
+    if not isinstance(where_block, SelectWhereFilter):
+        raise AssertionError(u'Expected last IR block to be SelectWhereFilter, found: '
+                             u'{} {}'.format(where_block, ir_blocks))
+    ir_except_where = ir_blocks[:-1]
+
+    output_block = ir_except_where[-1]
     if not isinstance(output_block, ConstructResult):
         raise AssertionError(u'Expected last IR block to be ConstructResult, found: '
```
obi1kenobi: This isn't the last IR block any more -- the message should be updated if we stick with this representation.

graphql_compiler/tests/test_compiler.py
```diff
@@ -443,6 +463,16 @@ def test_filter_on_optional_variable_name_or_alias(self):
                 }}
                 RETURN $matches
             )
+            WHERE
+                (
```
obi1kenobi: Elsewhere, the first `(` is on the same line as the keyword that starts it. Let's keep the formatting like that: `WHERE (` then de-indent everything below by one.

graphql_compiler/tests/test_emit_output.py
```diff
@@ -29,7 +29,8 @@ def test_simple_immediate_output(self):
             MarkLocation(base_location),
             ConstructResult({
                 'foo_name': OutputContextField(base_name_location, GraphQLString),
-            })
+            }),
+            SelectWhereFilter({}),
```
obi1kenobi: I don't think we want this block added if it isn't doing anything. Also, regardless of what the MATCH query looks like, it's strange from an IR perspective to construct results before filtering them again.

When this block appears, let's move it to right before `ConstructResult` instead of right after. That way, `ConstructResult` is always the last IR block, and that simplifies the error-checking code I commented on above.

graphql_compiler/compiler/ir_lowering_match/utils.py
amartyashankha: Sure, where else is it being used?

graphql_compiler/compiler/emit_match.py
```diff
@@ -156,6 +157,14 @@ def _construct_output_to_match(output_block):
     return u'SELECT %s FROM' % (u', '.join(selections),)
 
 
+def _construct_where_to_match(where_block):
```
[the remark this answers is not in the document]
amartyashankha: This is being handled elsewhere now. This might still be useful if the a `TrueLiteral` is explicitly used as a predicate for the `Filter`?

graphql_compiler/compiler/ir_lowering_match/utils.py
amartyashankha: nvm, found some

graphql_compiler/compiler/expressions.py
```diff
@@ -235,6 +235,53 @@ def to_gremlin(self):
...
+
+        mark_name, field_name = self.location.get_location_name()
+        validate_safe_string(mark_name)
+        validate_safe_string(field_name)
+
+        return u'%s.%s' % (mark_name, field_name)
+
+    def to_gremlin(self):
+        """Not implemented, should not be used."""
+        raise AssertionError(u'SelectEdgeContextField is only used for the WHERE statement in MATCH'
```
obi1kenobi: nit: you need a `. ` at the end of `MATCH`, otherwise the string gets joined into: `... in MATCHThis function should not ...`

graphql_compiler/compiler/ir_lowering_match/__init__.py
```diff
@@ -17,12 +19,13 @@
                                  lower_context_field_expressions, prune_non_existent_outputs)
 from ..match_query import convert_to_match_query
 from ..workarounds import orientdb_class_with_while, orientdb_eval_scheduling
-
+from .utils import _construct_where_filter_predicate
```
obi1kenobi: Underscored functions should be private and not imported (except _maybe_ in tests). Remove the underscore from the function name.

graphql_compiler/compiler/match_query.py
```diff
@@ -135,6 +136,34 @@ def _split_match_steps_into_match_traversals(match_steps):
     return output
 
 
+def _extract_global_operations(ir_blocks):
+    """Extract all global operation blocks (all blocks following GlobalOperationsStart).
```
obi1kenobi: So is `ConstructResult` considered part of the global operations and returned in `global_operation_blocks`, or is it returned in `remaining_ir_blocks`?

From the use of this function below, it seems like `ConstructResult` shouldn't be part of `ir_blocks` at all -- in which case the docstring must reflect that, and there should be an assertion that ensures that.

PR #100, opened 2018-07-09
 Add intersects filter operator

yangsong97:
Example usage:

```
graphql_input = '''{
        Animal {
            alias @filter(op_name: "intersects", value: ["$wanted"])
        }
    }'''
```

```
expected_match = '''
            SELECT
                Animal___1.name AS `animal_name`
            FROM (
                MATCH {{
                    class: Animal,
                    where: ((intersect(alias, {wanted}).asList().size() > 0)),
                    as: Animal___1
                }}
                RETURN $matches
            )
        '''
```

```
        expected_gremlin = '''
            g.V('@class', 'Animal')
            .filter{it, m -> it.alias.intersects($wanted)}
            .as('Animal___1')
            .transform{it, m -> new com.orientechnologies.orient.core.record.impl.ODocument([
                animal_name: m.Animal___1.name
            ])}
        '''
```

### DIFF
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -6,8 +6,8 @@ from ..exceptions import GraphQLCompilationError
 from ..schema import GraphQLDate, GraphQLDateTime
 from .compiler_entities import Expression
 from .helpers import (STANDARD_DATE_FORMAT, STANDARD_DATETIME_FORMAT, FoldScopeLocation, Location,
-                      ensure_unicode_string, is_graphql_type, safe_quoted_string,
-                      strip_non_null_from_type, validate_safe_string)
+                      ensure_unicode_string, is_graphql_type, is_vertex_field_name,
+                      safe_quoted_string, strip_non_null_from_type, validate_safe_string)
 
 
 # Since MATCH uses $-prefixed keywords to indicate special values,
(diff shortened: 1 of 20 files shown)


graphql_compiler/compiler/expressions.py
```diff
@@ -713,6 +714,7 @@ def to_match(self):
                 u'||': (u'OR', regular_operator_format),
                 u'&&': (u'AND', regular_operator_format),
                 u'contains': (u'CONTAINS', regular_operator_format),
+                u'intersects': (u'intersect', intersects_operator_format),
```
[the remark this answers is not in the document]
yangsong97: `intersect` here is a function, so I don't think I can capitalize it

graphql_compiler/compiler/filters.py
```diff
@@ -493,6 +493,44 @@ def _process_contains_filter_directive(filter_operation_info, context, parameter
...
+    base_field_type = strip_non_null_from_type(filtered_field_type)
+    if not isinstance(base_field_type, GraphQLList):
+        raise GraphQLCompilationError(u'Cannot apply "intersects" to non-list '
+                                      u'type {}'.format(filtered_field_type))
+
+    argument_inferred_type = strip_non_null_from_type(filtered_field_type)
+    argument_expression, non_existence_expression = _represent_argument(
+        context, parameters[0], argument_inferred_type)
+
+    filter_predicate = expressions.BinaryComposition(
```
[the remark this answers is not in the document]
yangsong97: All the other filter directives use `filter_predicate` here so I think we should leave it like that for consistency.

graphql_compiler/compiler/expressions.py
```diff
@@ -691,6 +691,7 @@ def to_match(self):
         # The MATCH versions of some operators require an inverted order of arguments.
         # pylint: disable=unused-variable
         regular_operator_format = '(%(left)s %(operator)s %(right)s)'
+        intersects_operator_format = '(%(operator)s(%(left)s, %(right)s).asList().size() > 0)'
```
obi1kenobi: Please add this below the regular/inverted formats -- those two are related to each other, whereas this one is entirely separate and should not be between them.

graphql_compiler/compiler/expressions.py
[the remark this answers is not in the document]
obi1kenobi: Yes, since it's a function, let's keep it as-is. If it were a keyword operator, we'd capitalize it to keep with traditional SQL style.

graphql_compiler/compiler/filters.py
```diff
@@ -468,7 +468,7 @@ def _process_contains_filter_directive(filter_operation_info, context, parameter
                     if the collection is optional and missing, the check will return True
 
     Returns:
-        a Filter basic block that performs the substring check
+        a Filter basic block that performs the contains check
```
obi1kenobi: thank you for finding and fixing this

graphql_compiler/tests/test_compiler.py
```diff
@@ -1308,6 +1308,120 @@ def test_in_collection_op_filter_with_optional_tag(self):
...
+                    class: Animal,
+                    where: ((intersect(alias, {wanted}).asList().size() > 0)),
+                    as: Animal___1
+                }}
+                RETURN $matches
+            )
+        '''
+        expected_gremlin = '''
+            g.V('@class', 'Animal')
+            .filter{it, m -> it.alias.intersects($wanted)}
```
obi1kenobi: Two issues here:
- The Gremlin method is `intersect` not `intersects`.
- `intersect` returns the intersection itself, which is not a boolean and produces an error on evaluation:
`java.lang.ClassCastException: java.util.LinkedHashSet cannot be cast to java.lang.Boolean`

The easiest compilation option here probably is:
```
.filter{it, m -> (!it.alias.intersects($wanted).empty)}
```

That means the Gremlin format for `intersects` would be: 
```
u'(!{left}.{operator}({right}).empty)'
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -1308,6 +1308,120 @@ def test_in_collection_op_filter_with_optional_tag(self):
...
+                }}.in('Animal_ParentOf') {{
+                    optional: true,
+                    as: Animal__in_Animal_ParentOf___1
+                }} , {{
+                    class: Animal,
+                    as: Animal___1
+                }}.out('Animal_ParentOf') {{
+                    where: ((
+                         ($matched.Animal__in_Animal_ParentOf___1 IS null)
+                         OR (intersect(alias, $matched.Animal__in_Animal_ParentOf___1.alias)
```
obi1kenobi: try to stick to the existing style as much as possible -- I think leaving the `OR` on its own line and moving the `(intersect...` to the next line would be more in line with the style of the rest of the file.

obi1kenobi:
Please make `flake8` happy -- it seems like you have trailing whitespace in some files. You can adjust your editor to remove trailing whitespace on save.

PR #103, opened 2018-07-10
Truncate repeated single step traversals in CompoundMatchQuery.

amartyashankha:
Also fixed bug in `truncate_repeated_single_step_traversals`.

### DIFF
--- a/graphql_compiler/compiler/ir_lowering_match/__init__.py
+++ b/graphql_compiler/compiler/ir_lowering_match/__init__.py
@@ -19,7 +19,7 @@ from .optional_traversal import (collect_filters_to_first_location_occurrence,
                                  lower_context_field_expressions, prune_non_existent_outputs)
 from ..match_query import convert_to_match_query
 from ..workarounds import orientdb_class_with_while, orientdb_eval_scheduling
-from .utils import construct_where_filter_predicate
+from .utils import construct_where_filter_predicate, CompoundMatchQuery
 
 ##############
 # Public API #
@@ -103,4 +103,9 @@ def lower_ir(ir_blocks, location_types, type_equivalence_hints=None):
     compound_match_query = lower_context_field_expressions(
         compound_match_query)
 
-    return compound_match_query
+    lowered_match_queries = []
+    for match_query in compound_match_query.match_queries:
+        new_match_query = truncate_repeated_single_step_traversals(match_query)
+        lowered_match_queries.append(new_match_query)
+
+    return CompoundMatchQuery(match_queries=lowered_match_queries)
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/ir_lowering_match/__init__.py
```diff
@@ -103,4 +103,9 @@ def lower_ir(ir_blocks, location_types, type_equivalence_hints=None):
     compound_match_query = lower_context_field_expressions(
         compound_match_query)
 
-    return compound_match_query
+    lowered_match_queries = []
+    for match_query in compound_match_query.match_queries:
+        new_match_query = truncate_repeated_single_step_traversals(match_query)
+        lowered_match_queries.append(new_match_query)
+
+    return CompoundMatchQuery(match_queries=lowered_match_queries)
```
obi1kenobi: Can you extract this into a function? I would like to keep as little logic as possible in this function itself, and just call out to other functions that implement the actual lowering process.

graphql_compiler/compiler/ir_lowering_match/__init__.py
```diff
@@ -103,4 +104,4 @@ def lower_ir(ir_blocks, location_types, type_equivalence_hints=None):
     compound_match_query = lower_context_field_expressions(
         compound_match_query)
 
-    return compound_match_query
+    return lower_sub_queries(compound_match_query)
```
obi1kenobi: Please preserve the structure and style of the function -- more passes will be added, and people shouldn't have to modify the return line:
```
compound_match_query = lower_sub_queries(compound_match_query)

return compound_match_query
```

graphql_compiler/compiler/ir_lowering_match/optional_traversal.py
```diff
@@ -559,3 +560,13 @@ def lower_context_field_expressions(compound_match_query):
             )
 
     return CompoundMatchQuery(match_queries=new_match_queries)
+
+
+def lower_sub_queries(compound_match_query):
```
obi1kenobi: Either this method needs to not be generic (probably preferred option), or it needs to not be in the `optional_traversal.py` file. I think it's better to not have this "generically lowering sub-queries" approach, and instead focus on one-thing-per-lowering-step functions.

So I'd suggest renaming this function to something less generic and more directly related to what it does.

amartyashankha: Wouldn't that mean writing this same loop structure multiple times for each lowering step we want to apply to the `CompoundMatchQuery`?
I thought having it be generic would allow us to add more lowering steps into the loop. In that case, this function could maybe go in `__init__.py`?

PR #108, opened 2018-07-17
(WIP) Adding more Snapshot tests

amartyashankha:
Current issue: `pip install` gets incorrect version of `snapshottest` library
Next Steps: Several of the snapshots are empty. Need to add more comprehensive data to ensure that all queries give meaningful results.

Added support for snapshots of Decimal values.
Added more items to the schema.
Added snapshot tests for all non-parameterized test cases (from `test_input_data.py`) .


scripts/copyright_line_check.sh
```diff
@@ -43,5 +43,8 @@ ensure_file_has_copyright_line './setup.py'
 
 # Check every python file in the package's source directory.
 for filename in ./graphql_compiler/**/*.py; do
-    ensure_file_has_copyright_line "$filename"
+    # Don't run copyright check for snapshot files
+    if [[ $filename != */snapshots/snap_*.py ]]; then
```
obi1kenobi: Try adding double quotes around the values you are comparing? If the filename has spaces/special chars, this will cause unexpected behavior.

graphql_compiler/tests/test_data_tools/data/commands.sql
```diff
@@ -11,238 +11,878 @@ create vertex Animal set birthday = DATE("2006-09-20 00:00:00"), color = 'orange
...
+create edge Animal_ParentOf from (select from Animal where name = 'Pteranodon__(121738)') to (select from Animal where name = 'Pteranodon__12')
+create edge Animal_ParentOf from (select from Animal where name = 'Pteranodon__(121738)') to (select from Animal where name = 'Pteranodon__17')
+create edge Animal_ParentOf from (select from Animal where name = 'Pteranodon__(121738)') to (select from Animal where name = 'Pteranodon__3')
+create edge Animal_ParentOf from (select from Animal where name = 'Pteranodon__(121738)') to (select from Animal where name = 'Pteranodon__8')
+create vertex Animal set birthday = DATE("2011-08-24 00:00:00"), color = 'magenta', name = 'Pteranodon__((((1013155)131719)0118)11219)', net_worth = 699.6, uuid = '2bdb5720-63ea-a07f-f436-498e68afe2b8'
+create edge Animal_ParentOf from (select from Animal where name = 'Pteranodon__((((1013155)131719)0118)11219)') to (select from Animal where name = 'Pteranodon__(((1013155)131719)0118)')
+create edge Animal_ParentOf from (select from Animal where name = 'Pteranodon__((((1013155)131719)0118)11219)
```
obi1kenobi: If we are already storing the provenance of the animals, thoughts on making the names not be ambiguous? This can be parsed in several ways: `<stuff>, 13, 15, 2`, `<stuff>, 13, 1, 52` etc.

graphql_compiler/tests/test_data_tools/schema.sql
```diff
@@ -44,8 +48,8 @@ CREATE PROPERTY Animal.net_worth Decimal
 CREATE INDEX Animal.net_worth NOTUNIQUE
```
obi1kenobi: If `net_worth` is supposed to be `Decimal`, then create it as `Decimal` in the generated SQL commands as well. Right now, the data is initialized as floats, which introduces some potential of spurious and frustrating errors in the tests. See the GraphQL handling of Decimal inputs for an example of the OrientDB syntax to use.

PR #109, opened 2018-07-25
Ensuring that all snapshot tests return results.

amartyashankha:


scripts/generate_test_sql/events.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2017-present Kensho Technologies, LLC.
```
obi1kenobi: This seems like a new file. If so, the year has to be the present year -- this is extremely important! Please be very careful with it.

scripts/generate_test_sql/species.py
```diff
@@ -8,18 +10,48 @@
...
 
 
+def _create_species_eats_statement(from_name, to_name):
+    """Return a SQL statement to create a Species_Eats edge."""
+    if to_name in SPECIES_LIST:
+        to_class = 'Species'
+    elif to_name in FOOD_LIST:
+        to_class = 'Food'
+    else:
+        raise AssertionError(u'Invalid name for SpeciesEats endpoint: {}'.format(to_name))
```
obi1kenobi: `Species_Eats` for consistency

PR #110, opened 2018-07-27
Ensure type coercions are preserved when hiding non-preferred locations.

obi1kenobi:
When exposing only preferred locations as start points for query processing, we have to hide all other locations that are eligible to be start points. There existed an edge case where in this process, we'd accidentally drop type coercions at such locations during the hiding process. This diff fixes that problem.

### DIFF
--- a/graphql_compiler/compiler/common.py
+++ b/graphql_compiler/compiler/common.py
@@ -91,6 +91,7 @@ def _compile_graphql_generic(language, lowering_func, query_emitter_func,
 
     lowered_ir_blocks = lowering_func(
         ir_and_metadata.ir_blocks, ir_and_metadata.location_types,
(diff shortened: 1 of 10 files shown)


obi1kenobi:
@amartyashankha we'll need to add this testcase to the snapshots at some point, but not in this PR.

@jmeulemans this PR would also have benefited from the global metadata table we were discussing. When we manage to build it, we should refactor the tests to make sure all queries infer correct metadata, and then clean up the various lowering steps that have to introspect the query instead of simply looking up the metadata they require.

obi1kenobi:
@lodrion I think we'll go a similar but different route -- I think we'll pass all the metadata together in one object. That's the refactor that @jmeulemans and I were talking about.

PR #111, opened 2018-07-27
Adding parameterized snapshot tests.

amartyashankha:


graphql_compiler/tests/test_input_data.py
```diff
@@ -192,7 +192,8 @@ def name_or_alias_filter_on_interface_type():
 def output_source_and_complex_output():
     graphql_input = '''{
         Animal {
-            name @filter(op_name: "=", value: ["$wanted"]) @output(out_name: "animal_name")
+            name @filter(op_name: "=", value: ["$wanted"])
```
obi1kenobi: Any particular reason to make this whitespace-only change? The line was under 100 chars.

graphql_compiler/tests/test_input_data.py
```diff
@@ -1060,7 +1058,7 @@ def has_edge_degree_op_filter_with_fold():
             in_Animal_OfSpecies {
                 name @output(out_name: "parent_name")
 
-                out_Animal_ParentOf @filter(op_name: "has_edge_degree", value: ["$child_count"])
+                in_Animal_ParentOf @filter(op_name: "has_edge_degree", value: ["$child_count"])
                                     @fold {
```
obi1kenobi: the indentation on `@fold` doesn't match now

graphql_compiler/tests/test_input_data.py
```diff
@@ -1000,7 +998,7 @@ def has_edge_degree_op_filter():
     graphql_input = '''{
         Animal {
             name @output(out_name: "animal_name")
-            out_Animal_ParentOf @filter(op_name: "has_edge_degree", value: ["$child_count"])
+            in_Animal_ParentOf @filter(op_name: "has_edge_degree", value: ["$child_count"])
                                 @output_source {
```
obi1kenobi: the indentation on `@output_source` doesn't match now

graphql_compiler/tests/test_input_data.py
```diff
@@ -448,7 +449,6 @@ def complex_optional_variables():
     # The operands in the @filter directives originate from an optional block.
     graphql_input = '''{
         Animal {
-            name @filter(op_name: "=", value: ["$animal_name"])
             out_Animal_ParentOf {
                 out_Animal_FedAt @optional {
                     name @tag(tag_name: "child_fed_at_event")
```
obi1kenobi: It looks like the indentation here and on the line below is a bit off

graphql_compiler/tests/test_input_data.py
```diff
@@ -448,7 +449,6 @@ def complex_optional_variables():
     # The operands in the @filter directives originate from an optional block.
     graphql_input = '''{
         Animal {
-            name @filter(op_name: "=", value: ["$animal_name"])
```
obi1kenobi: It'd be preferable to keep this if it's not too much work -- it ensures that the query is scheduled and executed in a predictable order. Otherwise, we simply don't know if this location runs first, or its `out_Animal_ParentOf` is the first one.

amartyashankha: Hmm, but if we start from a specific `Animal`, we aren't testing all possible behaviors of the `@optional` traversals and tagged values. That's why I removed the filter. It should be easy enough to put it back, but the results will be less interesting.

graphql_compiler/tests/test_input_data.py
amartyashankha: I'm not sure why this is off.

graphql_compiler/tests/test_input_data.py
amartyashankha: I thought we were keeping different directives on separate lines (fixed now).

graphql_compiler/tests/test_input_data.py
obi1kenobi: I'd put a separate test case with this different query. Having the filter there vs not having it trigger very different query planning decisions from the database, so I don't think this is quite as equivalent as you suggest.

PR #112, opened 2018-07-30
more than a single recursion in a query

cw6515:
Running GraphQL queries with two inplace recursions seems to fail; this PR adds a test that should pass, but currently does not.


graphql_compiler/tests/test_input_data.py
```diff
@@ -680,6 +680,40 @@ def simple_recurse():
...
+                    name @output(out_name: "important_event")
+                }
+            }
+            out_Animal_ParentOf @recurse(depth: 2) {
+                name @output(out_name: "ancestor_name")
+            }
+            in_Animal_ParentOf @recurse(depth: 2) {
+                name @output(out_name: "descendent_name")
+            }
+
```
obi1kenobi: Missing `}`? I don't think this parses as valid GraphQL.

graphql_compiler/tests/test_input_data.py
```diff
@@ -680,6 +680,40 @@ def simple_recurse():
...
+    }'''
+    expected_output_metadata = {
+        'animal_name': OutputMetadata(type=GraphQLString, optional=False),
+        'important_event': OutputMetadata(type=GraphQLString, optional=True)
+        'ancestor_name': OutputMetadata(type=GraphQLString, optional=False)
+        'descendent_name': OutputMetadata(type=GraphQLString, optional=False)
+    }
+    expected_input_metadata = {
+        ''animal_name': OutputMetadata(type=GraphQLString, optional=False)'
```
obi1kenobi: double `''` here and `'` at the end of the line will probably make python unhappy

graphql_compiler/tests/test_input_data.py
```diff
@@ -680,6 +680,40 @@ def simple_recurse():
...
+                name @output(out_name: "ancestor_name")
+            }
+            in_Animal_ParentOf @recurse(depth: 2) {
+                name @output(out_name: "descendent_name")
+            }
+
+    }'''
+    expected_output_metadata = {
+        'animal_name': OutputMetadata(type=GraphQLString, optional=False),
+        'important_event': OutputMetadata(type=GraphQLString, optional=True)
```
obi1kenobi: you'll probably need `,` at the end of each line here

graphql_compiler/tests/test_input_data.py
```diff
@@ -680,6 +680,41 @@ def simple_recurse():
         type_equivalence_hints=None)
 
 
+def double_recurse():
+    graphql_input = '''{
+        Animal @filter(op_name: "name_or_alias", value: ["$animal_name_or_alias"]) {
+            name @output(out_name: "animal_name")
+            out_Animal_ImportantEvent @optional {
```
obi1kenobi: Is the `@optional` required to trigger the bug? If the bug happens without it, let's remove it -- we're aiming for the minimal possible repro.

graphql_compiler/tests/test_input_data.py
```diff
@@ -680,6 +680,41 @@ def simple_recurse():
...
+            in_Animal_ParentOf @recurse(depth: 2) {
+                name @output(out_name: "descendent_name")
+            }
+    }'''
+    
+    expected_output_metadata = {
+        'animal_name': OutputMetadata(type=GraphQLString, optional=False),
+        'important_event': OutputMetadata(type=GraphQLString, optional=True),
+        'ancestor_name': OutputMetadata(type=GraphQLString, optional=False),
+        'descendent_name': OutputMetadata(type=GraphQLString, optional=False)
```
obi1kenobi: nit: I'd leave a `,` on the last element as well -- that way, adding new stuff modifies one line instead of two.

graphql_compiler/tests/test_input_data.py
```diff
@@ -680,6 +680,41 @@ def simple_recurse():
...
+                ... on Event {
+                    name @output(out_name: "important_event")
+                }
+            }
+            out_Animal_ParentOf @recurse(depth: 2) {
+                name @output(out_name: "ancestor_name")
+            }
+            in_Animal_ParentOf @recurse(depth: 2) {
+                name @output(out_name: "descendent_name")
+            }
```
obi1kenobi: is there maybe a missing `}` here? I'm not sure this parses as valid GraphQL

PR #113, opened 2018-07-30
Ensure that multiple traversals from a type-coerced location work correctly

obi1kenobi:
Added a test case that was incorrectly triggering an `AssertionError`, but now passes with the code change in this PR.

### DIFF
--- a/graphql_compiler/compiler/workarounds/orientdb_query_execution.py
+++ b/graphql_compiler/compiler/workarounds/orientdb_query_execution.py
@@ -267,8 +267,10 @@ def _expose_only_preferred_locations(match_query, location_types, coerced_locati
                     # we ensure that we again infer the same type bound.
                     eligible_location_types[current_step_location] = current_type_bound
 
-                    if current_step_location not in coerced_locations:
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/workarounds/orientdb_query_execution.py
```diff
@@ -267,8 +267,10 @@ def _expose_only_preferred_locations(match_query, location_types, coerced_locati
                     # we ensure that we again infer the same type bound.
                     eligible_location_types[current_step_location] = current_type_bound
 
-                    if current_step_location not in coerced_locations:
-                        # The type bound here is already implied by the GraphQL query structure.
+                    if (current_step_location not in coerced_locations or
+                            previous_type_bound is not None):
```
amartyashankha: Use `any` here instead of multi-line `if`?

obi1kenobi: I think @lodrion's preference might be the other way. If I'm wrong, I'll fix it in the next PR.

PR #114, opened 2018-07-30
Add schema validation

### LINKED ISSUE #9: Verify that passed GraphQL schema has all the necessary types and directives

cw6515:
Resolves #9.


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,86 @@ def _preprocess_graphql_string(graphql_string):
...
+        schema: GraphQL schema object, created using the GraphQL library
+        ast: abstract syntax tree representation of a graphql query
+
+    Returns:
+        list containing schema and/or query validation errors
+    """
+    core_graphql_errors = validate(schema, ast)
+
+    # The following directives do not appear in the core-graphql library, but are permissable.
+    whitelisted_directives = set([
```
obi1kenobi: I actually don't think we should allow any of these. Let's expect and allow only the ones defined by the compiler for now, until we have a better story of how custom directives fit into the picture.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,86 @@ def _preprocess_graphql_string(graphql_string):
...
+        tuple of the sorted input list
+    """
+    return tuple(sorted(unordered_list))
+
+
+def _validate_schema_and_ast(schema, ast):
+    """Validate the supplied graphql schema and ast.
+
+    This method wraps around graphql-core's validation to enforce a stricter requirement of the
+    schema-- all directives supported by the compiler must be declared by the schema, regardless of
```
obi1kenobi: space before `--`?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,86 @@ def _preprocess_graphql_string(graphql_string):
...
+        ('include', ('FIELD', 'FRAGMENT_SPREAD', 'INLINE_FRAGMENT'), ('if',)),
+        ('skip', ('FIELD', 'FRAGMENT_SPREAD', 'INLINE_FRAGMENT'), ('if',)),
+        ('deprecated', ('ENUM_VALUE', 'FIELD_DEFINITION'), ('reason',))
+    ])
+
+    # Extract name, locations and args keys in order to compare schema directives and directives
+    # which are supported by the graphql compiler.
+
+    # Directives provided in the GraphQL string.
+    expected_directives = set(
```
obi1kenobi: use a set comprehension?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,86 @@ def _preprocess_graphql_string(graphql_string):
...
+        (
+            directive.name,
+            _sorted_tuple(directive.locations),
```
obi1kenobi: Don't we just want these to be sets instead? `_sorted_tuple` seems to just be a workaround for order-independent equality checking, which sets provide.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,86 @@ def _preprocess_graphql_string(graphql_string):
...
+            _sorted_tuple(six.viewkeys(directive.args))
+        )
+        for directive in DIRECTIVES
+    )
+    expected_directives.update(whitelisted_directives)
+
+    # Directives provided in the core-graphql schema.
```
obi1kenobi: These comments seem confusingly reversed -- this set of directives is from the provided GraphQL string, and the other set is the set defined and required by the compiler.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -1050,3 +1052,173 @@ def test_invalid_edge_degree_queries(self):
         for invalid_graphql in invalid_queries:
             with self.assertRaises(GraphQLCompilationError):
                 graphql_to_ir(self.schema, invalid_graphql)
+
+    def test_missing_directives_in_schema(self):
+        """The schema should define all directives which are supported by the graphql compiler,
```
obi1kenobi: Please use the first line of the docstring as a one-line summary/headline, and move the rest of the explanation and detail into the body of the docstring.

graphql_compiler/compiler/compiler_frontend.py
cw6515: Some of the documentation is off, but I think we can keep the structure the same, idea being that we ensure that `parsed_schema_directives` = `expected_directives` + `unsupported_default_core-graphql_directives`

graphql_compiler/compiler/compiler_frontend.py
cw6515: yes, but we can't do a set of sets, so the tuple is kinda a hack. I could change this to a list of sets and change up  how the intersection/disjunction of directives is computed, but I don't have a strong preference either way.

graphql_compiler/compiler/compiler_frontend.py
cw6515: indeed they are :O

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -1050,3 +1052,175 @@ def test_invalid_edge_degree_queries(self):
         for invalid_graphql in invalid_queries:
             with self.assertRaises(GraphQLCompilationError):
                 graphql_to_ir(self.schema, invalid_graphql)
+
+    def test_missing_directives_in_schema(self):
+        """Ensure that validators properly identifiy missing directives in the schema.
+
+           The schema should all directives which are supported by the graphql compiler,
```
obi1kenobi: Mind de-indenting this part to the indentation level where `"""` start?

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -1050,3 +1052,175 @@ def test_invalid_edge_degree_queries(self):
...
+           whether they are not used in the query. Hence we raise an error when the following
```
obi1kenobi: `whether they are not used in the query`? Maybe `even if they might not be used in the query`?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,92 @@ def _preprocess_graphql_string(graphql_string):
...
+        ]),
+        frozenset([
+            'deprecated',
+            frozenset(['ENUM_VALUE', 'FIELD_DEFINITION']),
+            frozenset(['reason'])
+        ])
+    ])
+
+    # Extract name, locations and args keys in order to compare schema directives and directives
+    # which are supported by the graphql compiler.
```
obi1kenobi: It's a bit confusing that this comment is separated by a blank line (line 854) from the nearest piece of code.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,92 @@ def _preprocess_graphql_string(graphql_string):
...
+            frozenset(six.viewkeys(directive.args))
+        ])
+        for directive in DIRECTIVES
+    }
+
+    # Graphql-core automatically injects default directives into the schema, regardless of whether
+    # the schema supports said directives. Hence, while the directives contained in
+    # unsupported_default_directives are incompatible with the graphql-compiler, we still expect
+    # them in the parsed schema string.
+    expected_directives.update(unsupported_default_directives)
```
obi1kenobi: So the compiler doesn't actually care about whether the `unsupported_default_directives` appear in the schema or not -- if a future version of the core GraphQL library happens to remove them, the compiler won't mind. What do you think about making the code tolerate the appearance of these directives, but not require it?

So, for example, if the schema contains all the compiler-required directives, plus `deprecated` but none of the other default directives, the compiler could accept that.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -812,10 +813,92 @@ def _preprocess_graphql_string(graphql_string):
...
+    if missing_directives:
+        missing_message = (u'The following directives were missing from the '
+                           u'provided schema: {}'.format(missing_directives))
+        core_graphql_errors.append(missing_message)
+
+    # Directives that are not specified by the core graphql library.
+    extra_directives = actual_directives - expected_directives
+    if extra_directives:
+        extra_message = (u'The following directives were supplied in the given schema, but are not '
+                         u'defined in the core graphql library: {}'.format(extra_directives))
```
obi1kenobi: I think this error should say `not supported by the compiler` rather than mentioning the core GraphQL library. Also, always capitalize `graphql` -> `GraphQL`.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -1050,3 +1052,175 @@ def test_invalid_edge_degree_queries(self):
...
+            }
+        '''
+        parsed_incorrect_schema = build_ast_schema(parse(schema_with_incorrect_args))
+        query = '''{
+            Animal {
+                name @output(out_name: "animal_name")
+            }
+        }'''
+        with self.assertRaises(GraphQLValidationError):
+            graphql_to_ir(parsed_incorrect_schema, query)
```
obi1kenobi: Love the thorough tests!

graphql_compiler/compiler/compiler_frontend.py
cw6515: this comment also seems a bit unnecessary-- im removing it

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -1060,3 +1062,175 @@ def test_invalid_edge_degree_queries(self):
         for invalid_graphql in invalid_queries:
             with self.assertRaises(GraphQLCompilationError):
                 graphql_to_ir(self.schema, invalid_graphql)
+
+    def test_missing_directives_in_schema(self):
+        """Ensure that validators properly identifiy missing directives in the schema.
+
+        The schema should all directives which are supported by the graphql compiler,
```
obi1kenobi: typo: `The schema should all directives`

PR #115, opened 2018-07-30
Adding a double-recursion test case

cw6515:
Currently, having two recursions on the same scope leads to an error-- this PR provides tests that should pass given the GraphQL compiler constraints as given in the documentation.


graphql_compiler/tests/test_input_data.py
```diff
@@ -756,6 +756,104 @@ def simple_recurse():
...
+    expected_input_metadata = {}
+
+    return CommonTestData(
+        graphql_input=graphql_input,
+        expected_output_metadata=expected_output_metadata,
+        expected_input_metadata=expected_input_metadata,
+        type_equivalence_hints=None)
+
+
+def traverse_in_filter_then_recurse():
```
obi1kenobi: suggestion for a different name: `filter_then_traverse_and_recurse`?

graphql_compiler/tests/test_input_data.py
```diff
@@ -756,6 +756,104 @@ def simple_recurse():
...
+    }
+
+    return CommonTestData(
+        graphql_input=graphql_input,
+        expected_output_metadata=expected_output_metadata,
+        expected_input_metadata=expected_input_metadata,
+        type_equivalence_hints=None)
+
+
+def double_recurse():
```
obi1kenobi: How about `two_consecutive_recurses` to differentiate from a potential double nested recurse?

graphql_compiler/tests/test_compiler.py
```diff
@@ -1209,6 +1209,182 @@ def test_simple_recurse(self):
...
+        expected_match = '''
+            SELECT
+                Animal__out_Animal_ParentOf___1.name AS `ancestor_name`,
+                Animal___1.name AS `animal_name`,
+                Animal__out_Animal_ImportantEvent___1.name AS `important_event`
+            FROM (
+                MATCH {{
+                    class: Animal,
+                    where: (((name = {animal_name_or_alias})
+                          OR (alias CONTAINS {animal_name_or_alias}))),
```
obi1kenobi: Consider reformatting this to more closely match the style in the other test cases, e.g. see the `test_optional_on_union` which also has a multi-line `where`

graphql_compiler/tests/test_compiler.py
```diff
@@ -1209,6 +1209,182 @@ def test_simple_recurse(self):
...
+        expected_match = '''
+            SELECT
+                Animal__out_Animal_ParentOf___1.name AS `ancestor_name`,
+                Animal___1.name AS `animal_name`,
+                Animal__in_Animal_ParentOf___1.name AS `descendent_name`,
+                Animal__out_Animal_ImportantEvent___1.name AS `important_event`
+            FROM (
+                MATCH {{
+                    class: Animal,
+                    where: (((name = {animal_name_or_alias})
```
obi1kenobi: formatting here as well

graphql_compiler/tests/test_input_data.py
cw6515: sounds good

graphql_compiler/tests/test_input_data.py
cw6515: ditto

PR #120, opened 2018-08-19
Allowing nested @optional` tags

amartyashankha:
Most of the action happens in `OptionalTraversalTrie` (in `utils.py`).
The most recent test is not passing. This is the one with a 250 line output. I'll put it in a separate file later.

### DIFF
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -38,7 +38,6 @@ def compare_ir_blocks(test_case, expected_blocks, received_blocks):
     for i in six.moves.xrange(len(expected_blocks)):
         expected = expected_blocks[i]
         received = received_blocks[i]
-        if expected != received:
         test_case.assertEqual(expected, received,
                               msg=u'Blocks at position {} were different: {} vs {}\n\n'
                                   u'{}'.format(i, expected, received, mismatch_message))


graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -219,3 +220,104 @@ def construct_where_filter_predicate(simple_optional_root_info):
 # A CompoundMatchQuery is a representation of several MatchQuery objects containing
 #   - match_queries: a list MatchQuery objects
 CompoundMatchQuery = namedtuple('CompoundMatchQuery', ('match_queries'))
+
+
+class OptionalTraversalTrie(dict):
```
obi1kenobi: Let's discuss this in person, and figure out whether and how it might relate to the QueryMetadataTable work that has been happening in parallel.

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -222,12 +222,17 @@ def construct_where_filter_predicate(simple_optional_root_info):
 CompoundMatchQuery = namedtuple('CompoundMatchQuery', ('match_queries'))
 
 
-class OptionalTraversalTrie(dict):
-    def __init__(self):
+class OptionalTraversalTrie():
```
obi1kenobi: you have to put `object` between the parens here, otherwise in py2 this becomes an old-style object which is problematic

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -222,12 +222,17 @@ def construct_where_filter_predicate(simple_optional_root_info):
...
+    def __init__(self, complex_optional_roots):
         """Initialize empty trie of optional root Locations (elements of complex_optional_roots)."""
-        super(OptionalTraversalTrie, self).__init__()
+        self._location_to_children = {
+            optional_root_location: set()
+            for optional_root_location in complex_optional_roots
+        }
+        self._root_location = Location(())
```
obi1kenobi: Since this location is a dummy, I'd make it `None` instead of such an invalid `Location` object. Then you can define it as a class-level constant.

graphql_compiler/compiler/ir_lowering_match/utils.py
obi1kenobi: I think this class needs much more documentation -- I only understand what it does because we discussed it in front of a whiteboard. I also wouldn't count on people knowing what `complex_optional_roots` is.

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -237,50 +242,54 @@ def insert(self, optional_root_locations_path, complex_optional_roots):
...
+            else:
+                self._location_to_children[parent_location].add(optional_root_location)
+                parent_location = optional_root_location
 
-    def get_all_rooted_subtrees_as_lists(self):
+    def get_all_rooted_subtrees_as_lists(self, start_location=None):
         """Return a list of all rooted subtrees (each as a list of Location objects)."""
-        if self == {}:
-            # Empty trie only retuens a singleton list containing the null set (as an empty list).
+        if start_location is not None and start_location not in self._location_to_children.keys():
```
obi1kenobi: `start_location not in self._location_to_children` instead of using `.keys()`

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -237,50 +242,54 @@ def insert(self, optional_root_locations_path, complex_optional_roots):
...
+            raise AssertionError(u'received invalid start_location {} that was not present '
```
obi1kenobi: Capitalize the error message please.

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -237,50 +242,54 @@ def insert(self, optional_root_locations_path, complex_optional_roots):
...
+                                 u'in complex_optional_roots: {}'
```
obi1kenobi: Referring to `complex_optional_roots` in the error message, but no such variable is being checked or included in the error message. I think you mean `_location_to_children` but might be best to be more explicit about what you mean here.

graphql_compiler/compiler/ir_lowering_match/utils.py
```diff
@@ -237,50 +242,54 @@ def insert(self, optional_root_locations_path, complex_optional_roots):
...
+                                 .format(start_location, self._location_to_children.keys()))
+
+        if start_location is None:
+            start_location = self._root_location
+
+        if len(self._location_to_children[start_location]) == 0:
+            # Node with no children only returns a singleton list containing the null set.
             return [[]]
 
+        current_children = sorted(self._location_to_children[start_location], key=lambda l: repr(l))
```
obi1kenobi: please don't sort by `repr` output -- this is expensive, non-obvious and very likely unstable across compiler versions

graphql_compiler/tests/test_compiler.py
```diff
@@ -5196,107 +5197,86 @@ def test_nested_optional_and_traverse(self):
...
-                            __out_Animal_OfSpecies___1
-                    }}
-                    RETURN $matches
-                )
-            ),
-            $result = UNIONALL($optional__0, $optional__1, $optional__2)
-        '''
+        # The correct MATCH output is outrageously long, and is stored in a separate file.
+        match_output_file = './complex_nested_optionals_output.sql'
+        expected_match = open(os.path.join(os.path.dirname(__file__), match_output_file)).read()
```
obi1kenobi: Leaking the file descriptor here. Use a `with` together with the `open`

graphql_compiler/compiler/helpers.py
```diff
@@ -347,6 +349,26 @@ def __ne__(self, other):
         """Check another object for non-equality against this one."""
         return not self.__eq__(other)
 
+    def __lt__(self, other):
```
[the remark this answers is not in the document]
amartyashankha: Should this be a separate PR? I don't think we need it for nested optionals.

graphql_compiler/compiler/helpers.py
```diff
@@ -347,6 +349,26 @@ def __ne__(self, other):
...
+            return len(self.query_path) < len(other.query_path)
+
+        for self_vertex_field, other_vertex_field in zip(self.query_path, other.query_path):
+            if self_vertex_field != other_vertex_field:
+                return self_vertex_field < other_vertex_field
+
+        if self.visit_counter != other.visit_counter:
+            return self.visit_counter < other.visit_counter
+
+        if self.field is None:
```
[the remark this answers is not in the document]
amartyashankha: Should I just explicitly test that Location s objects are compared correctly?

obi1kenobi:
Yeah probably a good idea to write some test cases for comparisons. You can
make the comparison logic a separate PR, I'll review it tomorrow AM and
then update this one with that PR's stuff.

graphql_compiler/compiler/helpers.py
```diff
@@ -237,6 +237,45 @@ def get_location_name(self):
...
 
+    @abstractmethod
+    def _compare(self, other):
+        """Return True if the other object is smaller than self in the total ordering."""
+        raise NotImplementedError()
+
+    def __eq__(self, other):
+        """Return True if the other object is smaller than self in the total ordering."""
+        if isinstance(self, Location) and isinstance(other, Location):
+            return self == other
```
obi1kenobi: I think you can leave `__eq__` as an `abstractmethod`. Otherwise, I think that this line will cause infinite recursion.

graphql_compiler/compiler/helpers.py
```diff
@@ -237,6 +237,45 @@ def get_location_name(self):
         """Return a tuple of a unique name of the location, and the current field name (or None)."""
         raise NotImplementedError()
 
+    @abstractmethod
+    def _compare(self, other):
+        """Return True if the other object is smaller than self in the total ordering."""
+        raise NotImplementedError()
+
+    def __eq__(self, other):
+        """Return True if the other object is smaller than self in the total ordering."""
```
obi1kenobi: copy-pasta

graphql_compiler/compiler/helpers.py
```diff
@@ -237,6 +237,45 @@ def get_location_name(self):
         """Return a tuple of a unique name of the location, and the current field name (or None)."""
         raise NotImplementedError()
 
+    @abstractmethod
+    def _compare(self, other):
```
obi1kenobi: How about `_check_if_object_of_same_type_is_smaller` or something similarly more descriptive than `_compare`? Currently, it doesn't communicate that the other object is required to be of the same type as `self`.

graphql_compiler/compiler/helpers.py
```diff
@@ -477,3 +520,27 @@ def __ne__(self, other):
...
+            raise AssertionError(u'Expected FoldScopeLocation type for other. Received {}: {}'
+                                 .format(type(other).__name__, other))
+
+        if self.base_location != other.base_location:
+            return self.base_location < other.base_location
+
+        if len(self.fold_path) != len(other.fold_path):
+            return len(self.fold_path) < len(other.fold_path)
+
+        for self_edge_tuple, other_edge_tuple in zip(self.fold_path, other.fold_path):
```
obi1kenobi: Does `if self.fold_path != other.fold_path: return self.fold_path < other.fold_path` not work? If I remember correctly, it's a tuple-of-tuples, which should sort correctly.

obi1kenobi:
Also Github seems to think there are conflicts and that this can't be merged right now.

graphql_compiler/tests/test_helpers.py
```diff
@@ -38,6 +38,8 @@ def compare_ir_blocks(test_case, expected_blocks, received_blocks):
     for i in six.moves.xrange(len(expected_blocks)):
         expected = expected_blocks[i]
         received = received_blocks[i]
+        if expected != received:
+            __import__('pdb').set_trace()
```
obi1kenobi: You should probably remove this before we merge into `master`.

amartyashankha: Woops, fixed

obi1kenobi: I don't think it is, actually -- the `if expected != received:` line is still there and it's causing a syntax error.

amartyashankha: My bad. Hopefully actually fixed now.

PR #122, opened 2018-08-20
Pass the QueryMetadataTable out of frontend and into IR lowering.

obi1kenobi:
Baby steps toward making lowering make full use of the QueryMetadataTable, continuing in the spirit of the smallest possible PR that makes a positive change without breaking tests.

This PR simply passes the QueryMetadataTable out of the compiler frontend and into the IR lowering, updates the API contracts and documentation around that change, and updates tests so they continue to pass.

Subsequent PRs will rewrite lowering steps to make them rely on the QueryMetadataTable more directly and in more depth, and work toward keeping the QueryMetadataTable in sync with IR transformations.

### DIFF
--- a/graphql_compiler/compiler/ir_lowering_match/__init__.py
+++ b/graphql_compiler/compiler/ir_lowering_match/__init__.py
@@ -28,14 +28,14 @@ from .utils import construct_where_filter_predicate
 ##############
 
 
-def lower_ir(ir_blocks, location_types, coerced_locations, type_equivalence_hints=None):
+def lower_ir(ir_blocks, query_metadata_table, type_equivalence_hints=None):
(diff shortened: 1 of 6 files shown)


graphql_compiler/compiler/ir_lowering_match/__init__.py
```diff
@@ -57,6 +57,19 @@ def lower_ir(ir_blocks, location_types, coerced_locations, type_equivalence_hint
     """
     sanity_check_ir_blocks_from_frontend(ir_blocks)
 
+    # Construct the mapping of each location to its corresponding GraphQL type.
+    location_types = {
```
jmeulemans: Nice, I like that this MATCH metadata can be pushed down like this now.

PR #124, opened 2018-08-22
Include QueryMetadataTable in IR sanity checks.

obi1kenobi:
This PR includes QueryMetadataTable in the IR sanity checks, allowing cross-checking of the IR versus the data recorded in the metadata table. It also adds some more sanity-checks in the QueryMetadataTable itself, and strengthens one of the existing sanity-checks to account for the fact that `MarkLocation` blocks now also appear between `Fold` and `Unfold` blocks.

### DIFF
--- a/graphql_compiler/compiler/helpers.py
+++ b/graphql_compiler/compiler/helpers.py
@@ -331,7 +331,9 @@ class Location(BaseLocation):
     def is_revisited_at(self, other_location):
         """Return True if other_location is a revisit of this location, and False otherwise."""
         # Note that FoldScopeLocation objects cannot revisit Location objects, or each other.
-        return isinstance(other_location, Location) and self.query_path == other_location.query_path
+        return (isinstance(other_location, Location) and
+                self.query_path == other_location.query_path and
+                self.visit_counter < other_location.visit_counter)
 
     def __str__(self):
         """Return a human-readable str representation of the Location object."""
--- a/graphql_compiler/compiler/helpers.py
+++ b/graphql_compiler/compiler/helpers.py
@@ -328,6 +328,11 @@ class Location(BaseLocation):
         mark_name = u'__'.join(self.query_path) + u'___' + six.text_type(self.visit_counter)
         return (mark_name, self.field)
 
+    def is_revisited_at(self, other_location):
+        """Return True if other_location is a revisit of this location, and False otherwise."""
+        # Note that FoldScopeLocation objects cannot revisit Location objects, or each other.
+        return isinstance(other_location, Location) and self.query_path == other_location.query_path
+
     def __str__(self):
         """Return a human-readable str representation of the Location object."""
         return u'Location({}, {}, {})'.format(self.query_path, self.field, self.visit_counter)
--- a/graphql_compiler/compiler/ir_sanity_checks.py
+++ b/graphql_compiler/compiler/ir_sanity_checks.py
@@ -36,7 +36,16 @@ def sanity_check_ir_blocks_from_frontend(ir_blocks, query_metadata_table):
 
 def _sanity_check_registered_locations_parent_locations(query_metadata_table):
     """Assert that all registered locations' parent locations are also registered."""
-    for _, location_info in query_metadata_table.registered_locations:
+    for location, location_info in query_metadata_table.registered_locations:
+        if (location != query_metadata_table.root_location and
+                not query_metadata_table.root_location.is_revisited_at(location)):
+            # If the location is not the root location and is not a revisit of the root,
+            # then it must have a parent location.
+            if location_info.parent_location is None:
+                raise AssertionError(u'Found a location that is not the root location of the query '
+                                     u'or a revisit of the root, but does not have a parent: '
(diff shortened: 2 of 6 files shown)


graphql_compiler/compiler/helpers.py
```diff
@@ -328,6 +328,11 @@ def get_location_name(self):
         mark_name = u'__'.join(self.query_path) + u'___' + six.text_type(self.visit_counter)
         return (mark_name, self.field)
 
+    def is_revisited_at(self, other_location):
```
lodrion: I think the name is deceptive. "revisited" implies time dependency which this is not checking

amartyashankha: Agreed

graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -139,34 +176,33 @@ def _sanity_check_mark_location_preceding_optional_traverse(ir_blocks):
...
+    # Exactly one MarkLocation block is found between any block that starts an interval of blocks
+    # that all affect the same query position, and the first subsequent block that affects a
+    # different position in the query. Such intervals include the following examples:
+    # - from Fold to Unfold
+    # - from QueryRoot to Traverse/Recurse
+    # - from one Traverse to the next Traverse
+    # - from Traverse to Backtrack
     found_start_block = False
-    found_fold_block = False
     mark_location_blocks = 0
```
amartyashankha: I would call this something like `num_marklocation_blocks`.

graphql_compiler/compiler/metadata.py
```diff
@@ -47,6 +54,21 @@ def register_location(self, location, location_info):
...
                                  .format(location, old_info, location_info))
+
+        if location.field is not None:
+            raise AssertionError(u'Attempting to register a location at a field, this is '
+                                 u'not allowed: {} {}'.format(location, location_info))
+
+        if location_info.parent_location is None:
+            # Only the root location and revisits of the root location are allowed
+            # to not have a parent location.
+            is_root_location = location == self._root_location
```
jmeulemans: I believe that one of these can be done away with if you wanted, since the root location will be a revisit of itself.

graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -30,6 +30,43 @@ def sanity_check_ir_blocks_from_frontend(ir_blocks):
     _sanity_check_mark_location_preceding_optional_traverse(ir_blocks)
     _sanity_check_every_location_is_marked(ir_blocks)
     _sanity_check_coerce_type_outside_of_fold(ir_blocks)
+    _sanity_check_all_marked_locations_are_registered(ir_blocks, query_metadata_table)
+    _sanity_check_registered_locations_parent_locations(query_metadata_table)
+
+
+def _sanity_check_registered_locations_parent_locations(query_metadata_table):
```
jmeulemans: Another sanity check that occurs to me is one ensuring that only one `LocationInfo` has a `None` parent, if that's not covered elsewhere.

graphql_compiler/compiler/helpers.py
obi1kenobi: I think in this specific use case, we care that the new location is actually a *re*-visit -- the `other_location.visit_counter` has to be strictly greater than `self.visit_counter`, and I'm adding that to the code now.

I don't think it has to be an either-or though -- I'm open to refactoring this code or adding another helper method if there are use cases for it.

graphql_compiler/compiler/ir_sanity_checks.py
obi1kenobi: Renamed to `mark_location_blocks_count`, is that acceptable?

graphql_compiler/compiler/metadata.py
obi1kenobi: I'm worried about confusion here, since "re"-visit seems to imply a time dependence as @lodrion and @amartyashankha point out, and I think it has to be specifically a visit different than the first. I'd prefer to make locations not revisits of themselves for that reason.

graphql_compiler/compiler/ir_sanity_checks.py
```diff
@@ -139,34 +176,33 @@ def _sanity_check_mark_location_preceding_optional_traverse(ir_blocks):
 
 def _sanity_check_every_location_is_marked(ir_blocks):
     """Ensure that every new location is marked with a MarkLocation block."""
-    # Exactly one MarkLocation block is found between a QueryRoot / Traverse / Recurse block,
-    # and the first subsequent Traverse, Recurse, Backtrack or ConstructResult block.
+    # Exactly one MarkLocation block is found between any block that starts an interval of blocks
```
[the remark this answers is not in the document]
obi1kenobi: Great question! It started with Gremlin (the first language that we built a backend for), where marking a location does not commute with filtering or other operations like this, so you have to mark after you've pruned everything you wanted pruned. Otherwise, it was kind of a toss-up between marking before and after, so we chose to mark after since it required less code.

PR #127, opened 2018-09-06
Fix bug with no-op coerce in the metadata table

bojanserafimov:


graphql_compiler/tests/test_ir_generation.py
```diff
@@ -4022,3 +4022,46 @@ def test_complex_nested_optionals(self):
...
+                'animal_name': expressions.OutputContextField(
+                    grandkid_location.navigate_to_field('name'), GraphQLString),
+            }),
+        ]
+        expected_location_types = {
+            base_location: 'Animal',
+            kid_location: 'Animal',
+            grandkid_location: 'Animal',
+            related_location: 'Entity',
+
```
obi1kenobi: remove this unnecessary blank line

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -579,6 +578,9 @@ def _compile_fragment_ast(schema, current_schema_type, ast, location, context):
         current_schema_type.is_same_type(equivalent_union_type)
     )
 
+    if not is_same_type_as_scope:
+        query_metadata_table.record_coercion_at_location(location, coerces_to_type_obj)
```
obi1kenobi: This is fine as a temporary fix, but I'd prefer in the long run to only record coercions if there is actually a `CoerceType` block there. I think this sort of side-steps the problem rather than actually solving it -- there may be another (more complex) edge case that can trigger the same issue.

graphql_compiler/tests/test_compiler.py
```diff
@@ -5282,3 +5282,42 @@ def test_complex_nested_optionals(self):
         '''
 
         check_test_data(self, test_data, expected_match, expected_gremlin)
+
+    def test_no_op_coercion(self):
+        test_data = test_input_data.no_op_coercion()
+
+        expected_match = '''
+        SELECT Animal__out_Animal_ParentOf__out_Animal_ParentOf___1.name
```
obi1kenobi: nit: the other expected MATCH/Gremlin queries start one level indented from here, any reason not to do that here and below as well? Especially for Gremlin, you can break lines anywhere and all you need to do is simply follow what other testcases do as a convention.

graphql_compiler/tests/test_input_data.py
```diff
@@ -2431,3 +2431,37 @@ def complex_nested_optionals():
         expected_output_metadata=expected_output_metadata,
         expected_input_metadata=expected_input_metadata,
         type_equivalence_hints=None)
+
+
+def no_op_coercion():
```
obi1kenobi: nit: mind moving this test input (and its associated test case) to the place in the inputs/tests files where type coercions are being tested? Grouping similar tests together helps with refactoring and looking for similar issues.

graphql_compiler/tests/test_input_data.py
```diff
@@ -1616,6 +1616,40 @@ def no_op_coercion_inside_fold():
...
+                    out_Entity_Related {
+                        ... on Entity {
+                            name @filter(op_name: "in_collection", value: ["$entity_names"])
+                        }
+                    }
+                }
+            }
+        }
+    }'''
+    type_equivalence_hints = {}
```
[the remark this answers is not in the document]
bojanserafimov: The existing tests cover this.

graphql_compiler/tests/test_ir_generation.py
```diff
@@ -2646,6 +2646,12 @@ def test_no_op_coercion_inside_fold(self):
 
         check_test_data(self, test_data, expected_blocks, expected_location_types)
 
+    def test_no_op_coercion_noninvertible_hints(self):
+        # The type where the coercion is applied is already Entity, so the coercion is a no-op.
```
obi1kenobi: copy-pasta, also this should be unnecessary once you move the non-invertible test to `test_ir_generation_errors.py`

obi1kenobi:
I'll merge and release in v1.8.1 as soon as CI finishes.

PR #128, opened 2018-09-07
Adding SQLAlchemy backend 

jmeulemans:
This PR adds support for SQLAlchemy Core as a compilation target, allowing the compiler to target the relational DBs supported by SQLAlchemy. Currently verified backends from this list include PostgreSQL, Sqlite, and MSSQL. Oracle and MYSQL should require minimal work to incorporate as well.

This is a large diff, and still a work in progress. Remaining work for this PR includes:
- [x] Converting AssertionErrors to compiler exceptions.

- [x] Adding more exception handling, once overall design is settled upon.

- [x] Adding more documentation, especially for the overall recursive structure of the compilation.

Remaining work for this PR or a subsequent one includes:
- [x] Adding short-circuit to cycle detection in recursive queries for all backends.
- [ ] Verifying Oracle and MYSQL backends.

Directives not (yet) supported by the backend:
- Fold
- Output Source

Filter operations not supported by the backend
- name_or_alias
- interesects
- has_edge_degree

Known Limitations:
- Querying an interface is only currently possible if a table exists for that interface. It's possible in the future that the `UNION ALL` of all types implementing the interface could be taken as the query result.


graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        The SQL backend does not require special handling to get the correct semantics for such
+        output fields.
+        """
+        if not isinstance(expression, expressions.TernaryConditional):
+            return expression
+        return expression.if_true
+    return construct_result.visit_and_update_expressions(visitor_fn)
+
+
+def lower_unary_transformations(ir_blocks):
```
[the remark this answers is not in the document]
obi1kenobi: Isn't `has_edge_degree` just a check against the cardinality of the join for a particular value?

Otherwise, I think this is probably fine for the time being. I think over time we'll want to move toward metadata-driven `NotSupportedByBackend` errors: the front-end generates IR and metadata, then the backend analyzes the metadata and complains if necessary.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -0,0 +1,747 @@
...
+)
+
+3. Construct the recursive clause. Here we only recurse on the columns necessary to JOIN before
+and after the recursion, output columns are not carried along. The recursion is joined to the
+CTE of the base query, ensuring that the recursion only starts at the required starting points,
+no more.
+
+Also worth noting with the recursive clause is the __depth_internal_name, which keeps track of
+recursion depth per the compiler's semantics, and __path_internal_name, which is used for cycle
+detection. Simply using UNION here to join the anchor query to the recursive query is insufficient
```
obi1kenobi: Not sure I understand what the problem with cycles is? Shouldn't we output cycles if they exist?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -0,0 +1,747 @@
...
+        recursive_cte.animal_id,
+        animal_parentof.parentof_id,
+        -- increment the depth
+        recursive_cte.__depth_internal_name + 1 AS __depth_internal_name,
+        -- append the next element to the path
+        recursive_cte.__path_internal_name || CAST(animal_parentof.parentof_id AS VARCHAR) ||
+            ',' AS __path_internal_name,
+        -- check if a cycle is already present, or if this creates a cycle
+        CASE WHEN (recursive_cte.__cycle_detected_internal_name = 1) THEN 1
+             WHEN (recursive_cte.__path_internal_name LIKE '%' ||
```
obi1kenobi: Consider parens around the right argument to `LIKE`? Otherwise, even if this is executed correctly by the SQL database, the order of operations and therefore the semantics are not obvious to the reader.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -0,0 +1,747 @@
...
+    return _expression_to_sql(expression, filter_selectable, filter_location_info, context)
+
+
+def _expression_to_sql(expression, selectable, location_info, context):
+    """Recursively convert a compiler predicate to it's SQLAlchemy expression representation."""
+    if isinstance(expression, expressions.LocalField):
+        column_name = expression.field_name
+        column = _get_column(selectable, column_name)
+        return column
+    if isinstance(expression, expressions.Variable):
```
obi1kenobi: nit: these `if` statements are meant to be mutually exclusive, so make them `elif`s instead

graphql_compiler/compiler/emit_sql.py
```diff
@@ -0,0 +1,747 @@
...
+        variable_name = expression.variable_name
+        return bindparam(variable_name)
+    if isinstance(expression, expressions.Literal):
+        return expression.value
+    if isinstance(expression, expressions.ContextField):
+        tag_field_name = expression.location.field
```
obi1kenobi: consider making anything that isn't 1-3 lines' worth of code in the top-level `if` into its own helper function -- that will massively shrink the size of this function and improve readability

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    # Perform lowering passes over IR blocks
+    construct_result = ir_blocks.pop()
+    construct_result = lower_construct_result(construct_result)
+    block_index_to_location = get_block_index_to_location_map(ir_blocks)
+    ir_blocks = lower_unary_transformations(ir_blocks)
+    ir_blocks = lower_context_field_existence(ir_blocks)
+    ir_blocks = lower_optional_fields(
+        ir_blocks, block_index_to_location, query_path_to_location_info)
+    query_path_to_node = {}
+    query_path_to_filter = defaultdict(list)
```
obi1kenobi: This is my personal preference, but I don't like `defaultdict` because:
- accesses that look like reads sometimes end up being writes, and
- it means you'll never get `KeyError` even if you have a bug in the code and therefore should have gotten it.

Instead, I prefer to use `.setdefault()` on a regular dict.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    query_path_to_tag_fields = defaultdict(list)
+    tree_root = None
+    for index, block in enumerate(ir_blocks):
+        if isinstance(block, (blocks.Recurse, blocks.Traverse, blocks.QueryRoot)):
+            location = block_index_to_location[index]
+            location_info = query_metadata_table.get_location_info(location)
+            query_path = location.query_path
+            if tree_root is None:
+                if not isinstance(block, blocks.QueryRoot):
+                    raise AssertionError(u'Encountered QueryRoot but tree root is already set.')
```
obi1kenobi: In error messages, always include the values based on which you are raising the error -- in this case, the `block` and the `tree_root`. That will make getting to the bottom of the error much easier.

In this case, I might also include the `ir_blocks` and the `query_metadata_table` because having more data is always helpful.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+                parent_query_path = parent_location.query_path
+                parent_node = query_path_to_node[parent_query_path]
+                child_node = SqlNode(block=block, query_path=query_path)
+                parent_node.add_child_node(child_node)
+                query_path_to_node[query_path] = child_node
+        elif isinstance(block, blocks.Filter):
+            for context_field in get_tag_fields(block.predicate):
+                query_path_to_tag_fields[context_field.location.query_path].append(context_field)
+            query_path_to_filter[query_path].append((block, query_path))
+        else:
```
obi1kenobi: I'm worried that this is a bug waiting to happen. At some point, we may add a new block, for which it's not clear that `continue` is the correct decision to make here.

I'd much prefer an explicit `elif ...: continue` (bonus points for a comment explaining why continuing is the right thing there) followed by `else: raise AssertionError(...)`

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        exist, which can be expressed in SQL as an (field IS NULL OR predicate) statement,
+        which can be constructed with this particular rewrite.
+        """
+        if not isinstance(expression, expressions.BinaryComposition):
+            return expression
+        if not isinstance(expression.left, expressions.LocalField):
+            return expression
+        if not isinstance(expression.right, (expressions.Variable, expressions.ContextField)):
+            return expression
+        # Either the expression is true, or
```
obi1kenobi: looks like you started typing a comment here and then gave up :)

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        return expressions.BinaryComposition(
+            u'||',
+            # either the expression is true
+            expression,
+            # or the field referenced by the expression is NULL (doesn't exist)
```
obi1kenobi: How do we know that the `left` is the field in question, and not the `right`? Seems like there should be an assertion here.

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2018-present Kensho Technologies, LLC.
+from collections import namedtuple
+from enum import Enum
```
obi1kenobi: I'm not sure `enum` exists in py2

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -0,0 +1,110 @@
...
+from sqlalchemy.sql import functions
+
+from ... import exceptions
+
+
+# These columns are reserved for the construction of recursive queries
+DEPTH_INTERNAL_NAME = u'__depth_internal_name'
+PATH_INTERNAL_NAME = u'__path_internal_name'
+LINK_INTERNAL_NAME = u'__link_internal_name'
+CYCLE_DETECTED_INTERNAL_NAME = u'__cycle_detected_internal_name'
```
obi1kenobi: I'd prefer to construct a rule for which names are reserved compiler-wide, instead of banning *some specific* names, for *some specific* backends. We'll undoubtedly need more names, and we don't want migrating to a different backend to suddenly impose different naming constraints on the queries.

There will be some special fields (e.g. `__typename`) that start with double-underscores. How about reserving all names that start with triple-underscore, compiler-wide?

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -0,0 +1,110 @@
...
+    LINK_INTERNAL_NAME,
+    CYCLE_DETECTED_INTERNAL_NAME,
+}
+
+UNSUPPORTED_META_FIELDS = {
+    u'@class': u'__typename'
+}
+
+
+class Operator:
```
obi1kenobi: this is an old-style class in py2, which is bad

graphql_compiler/tests/test_helpers.py
```diff
@@ -150,6 +154,8 @@ def get_schema():
             uuid: ID
             out_Species_Eats: [FoodOrSpecies]
             in_Species_Eats: [Species]
+            out_Species_EatenBy: [Species]
```
obi1kenobi: isn't this the same thing as `in_Species_Eats`?

graphql_compiler/tests/test_helpers.py
```diff
@@ -127,7 +128,10 @@ def get_schema():
             uuid: ID
             out_Animal_ParentOf: [Animal]
             in_Animal_ParentOf: [Animal]
+            out_Animal_FriendsWith: [Animal]
+            in_Animal_FriendsWith: [Animal]
             out_Animal_OfSpecies: [Species]
+            out_Animal_Eats: [FoodOrSpecies]
```
obi1kenobi: from a schema perspective, this is confusing -- why would we also track what a particular animal eats when we already track what that animal's species eats?

graphql_compiler/tests/test_helpers.py
```diff
@@ -162,8 +168,19 @@ def get_schema():
             alias: [String]
             uuid: ID
             in_Species_Eats: [Species]
+            in_Animal_Eats: [Animal]
             in_Entity_Related: [Entity]
             out_Entity_Related: [Entity]
+            out_Food_OfCuisine: [Cuisine]
+        }
+        type Cuisine implements Entity {
```
obi1kenobi: please preserve formatting and spacing

graphql_compiler/compiler/ir_lowering_sql/__init__.py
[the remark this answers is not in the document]
jmeulemans: Yeah exactly, that's what it would be.

I've been thinking about this, and one approach would be to use window functions to do the counting, and then filter that on only rows with a certain count. So if the edge is from column `animal_id` to column `parentof_id`, we would have something like
`COUNT(*) OVER (PARTITION BY parentof_id) AS edge_degree`  and then in a second CTE filter out rows that have the correct target edge degree. 

With your second comment, do you mean something that identifies constructs used in the metadata (like `has_edge_degree` was used), and then using that to reject on the backend?

graphql_compiler/compiler/emit_sql.py
jmeulemans: My motivation was that without cycle detection, we will repeat a lot of the same traversals that we already have sometime previously, which I was avoiding to save the DB engine some work. This was mainly motivated by performance on some real world datasets I was experimenting with.

This is another question I have with semantics then. Let's say we're repeatedly traversing an edge with cycle A -> B -> C -> A, where the types all work out and we have a very high recursion depth. Do we output just the rows:
```
..., A
..., B
..., C
```
and stop with that traversal because the next row would be `... A`, starting a traversal we've already traversed, or do we continue to repeat, and continue to spit out 
```
..., A
..., B
..., C
..., A
..., B
..., C
...
```
for whatever our recursion depth limit allows?

graphql_compiler/compiler/ir_lowering_sql/constants.py
jmeulemans: That sounds a lot better, I can create a PR for that separately.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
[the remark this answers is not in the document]
obi1kenobi: That seems very reasonable, though exactly the sort of thing for which I'd hope to have a compiler in the loop -- I would not like to write that SQL query by hand :) We don't have to enable it in v1 of the SQL backend though, happy to save it for later.

Regarding the metadata, I'd like to extend the `QueryMetadataTable` to include information about all the locations and kinds of filters that were used (as well as lots of other info, like fold and output locations etc.). Then, before lowering or emitting SQL, the SQL backend could look over this metadata and decide whether anything in there is unsupported. This is how you'd also raise `NotSupportedByBackend` errors for `@fold` on anything but Postgres due to the lack of proper array aggregates -- the backend knows the target platform, and the frontend supplies the metadata for the backend to decide on.

graphql_compiler/compiler/emit_sql.py
obi1kenobi: Semantically, it would be the latter.

One of the reasons for this is because there have been many requests for implementing a meta-field that allows outputting the recursion depth (which we haven't implemented yet because of lack of support for this in OrientDB 2). If and when we implement this feature, one should be able to query for things like "all the cycles and cycle lengths that start and end at A, to max depth 7".

Unfortunately, recursion can definitely get quite expensive, even on the other backends. This is something the compiler can help with, when we get a chance to integrate query size estimation and automatic query paging, but in the end we have to have some amount of trust that the user knows what they are asking for. Even without cycles, the database can have very deep trees and the user can set a recursion depth that is way too high -- and the resulting query will be expensive no matter what.

I'd be open to allowing distinct-only recursion in the future as an addition, if there is a good use case for it and ideally if we are able to figure out a way to emit appropriate OrientDB code for it.

obi1kenobi:
Feel free to do this in a separate PR, but please remember to update the documentation in the README file to reflect the changes here.

graphql_compiler/compiler/ir_lowering_sql/constants.py
jmeulemans: Created with https://github.com/kensho-technologies/graphql-compiler/pull/130

graphql_compiler/compiler/emit_sql.py
jmeulemans: That makes sense, I'll update accordingly.

The SQL backend can output the depth relatively easily if we ever support that at the compiler level. 

I've also thought it would be very useful to get the path taken in a recursion. That would allow you to treat results in a BFS vs DFS manner on the client side if desired, depending on how the path output is used for sorting. As another future thing if it can be worked in OrientDB.

That sounds good WRT distinct recursions, it is very costly as you point out, and giving users ways to address that is good future work.

obi1kenobi:
@jmeulemans, thoughts on making this code available for testing purposes as a pre-release version (e.g., `2.0.0.dev1`)? Might be useful to allow downstream projects to prepare while we work toward merging this into the `master` branch and a full-fledged `2.0` release.

Let me know what you think, and if you think that'd be useful then I'll upload a pre-release build on PyPI.

jmeulemans:
That sounds great @obi1kenobi, let's go for it.

pmantica1:
I am closing this pr for now since the lowering approach is stale.

PR #130, opened 2018-09-17
 Reserve out_name namespace for compiler internal use

jmeulemans:
This will allow for compiler internal only names to be used as output without the risk of conflicting with a user supplied `out_name`. This comes up with the SQL backend with columns that are needed in intermediate results, but that aren't included in the final output.


obi1kenobi:
If you can appease `isort`, this is fine by me.

PR #131, opened 2018-09-18
Fix recursion type validation

cw6515:


graphql_compiler/tests/test_input_data.py
```diff
@@ -2465,3 +2465,30 @@ def complex_nested_optionals():
         expected_output_metadata=expected_output_metadata,
         expected_input_metadata=expected_input_metadata,
         type_equivalence_hints=None)
+
+
+def recursive_field_type_is_subtype_of_parent_field():
+    """Ensure that traversal is allowed along an edge linked to a supertype of the parent field."""
```
obi1kenobi: nit: `recursion` instead of `traversal` -- a vanilla traversal is always allowed

graphql_compiler/tests/test_helpers.py
```diff
@@ -178,6 +178,8 @@ def get_schema():
             in_Animal_ImportantEvent: [Animal]
             in_Entity_Related: [Entity]
             out_Entity_Related: [Entity]
+            out_Event_RelatedEvent: [Event]
```
obi1kenobi: These should be of union type, `EventOrBirthEvent`

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -469,6 +469,18 @@ def test_recurse_directive_constraints(self):
             }
         }''')
 
+        # Note that ImportantEvent is a union of Event | BirthEvent, hence this query attempts
+        # to recurse on Animal at depth 0, and then on an Event for depth 1 and 2.
+        recurse_on_union_edge_without_parent_type = (GraphQLCompilationError, '''{
+            Animal {
+                out_Animal_ImportantEvent @recurse(depth:2) {
```
obi1kenobi: nit: space between `:` and `2`

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -469,6 +469,18 @@ def test_recurse_directive_constraints(self):
             }
         }''')
 
+        # Note that ImportantEvent is a union of Event | BirthEvent, hence this query attempts
```
obi1kenobi: nit: `ImportantEvent` sounds like a type, but you're actually referring to the edge `out_Animal_ImportantEvent`. This can introduce unnecessary confusion, so let's avoid abbreviations as much as possible.

graphql_compiler/tests/test_input_data.py
```diff
@@ -2465,3 +2465,30 @@ def complex_nested_optionals():
         expected_output_metadata=expected_output_metadata,
         expected_input_metadata=expected_input_metadata,
         type_equivalence_hints=None)
+
+
+def recursive_field_type_is_subtype_of_parent_field():
```
obi1kenobi: When you add a test input in `test_input_data.py`, you must not only add a test case for it in `test_compiler.py` but also add one in `test_ir_generation.py`. I suspect tests will fail on this.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -456,8 +491,19 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
             basic_blocks.append(fold_block)
             context['fold'] = inner_location
         elif recurse_directive:
+            # Get the valid fields of the outer scope, and the fields used in the inner scope.
+            outer_scope_type_map = schema.get_type_map().get(current_schema_type.name)
+            outer_scope_possible_properties = list(outer_scope_type_map.fields)
+            inner_scope_used_properties = [
```
obi1kenobi: I'm not sure I understand why the outer scope and inner scope properties are important here -- aren't we just doing a type check? If the types work out, the properties must work out too, and if the types don't work out, we don't allow duck-typing based on the properties.

graphql_compiler/tests/test_input_data.py
cw6515: yup both are in this PR :)

graphql_compiler/tests/test_helpers.py
cw6515: To my understanding, we want a structure such as the following:
subclass {
    out_superclass_edge_to_superclass_type{
        common property @output...
   }
}

Though I suppose this is a broader case than:
subclass {
    out_superclass_edge_to_superclass_or_subclass_type{
        common property @output...
   }
}

graphql_compiler/compiler/compiler_frontend.py
cw6515: good point, I think I swapped my supertypes and subtypes (i.e. this would imply that if you start on a subtype and then recurse on a supertype, then you could end up plucking a supertype property which the subtype doesn't have. Which makes no sense...

graphql_compiler/tests/test_input_data.py
obi1kenobi: I think you forgot to fix this ^ in this instance -- you fixed the other one though.

graphql_compiler/tests/test_input_data.py
```diff
@@ -2465,3 +2465,30 @@ def complex_nested_optionals():
...
+    """Ensure that traversal is allowed along an edge linked to a supertype of the parent field."""
+    graphql_input = '''{
+        BirthEvent {
+            out_Event_RelatedEvent @recurse(depth:2) {
```
obi1kenobi: nit: add space between `:` and `2`

graphql_compiler/compiler/helpers.py
```diff
@@ -226,7 +226,7 @@ def invert_dict(invertible_dict):
             raise TypeError(u'Expected an invertible dict, but value at key {} has type {}'.format(
                 k, type(v).__name__))
         if v in inverted:
-            raise TypeError(u'Expected an invertible dict, but keys'
+            raise TypeError(u'Expected an invertible dict, but keys '
```
obi1kenobi: thanks!

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -312,20 +312,40 @@ def _get_recurse_directive_depth(field_name, field_directives):
...
+    """Perform type checks on the enclosing type and the recursed type for a recurse directive.
+
+    Args:
+        current_schema_type: GraphQLType, the schema type at the current location
+        field_schema_type: GraphQLType, the schema type at the inner scope
+        context: dict, various per-compilation data (e.g. declared tags, whether the current block
+                 is optional, etc.). May be mutated in-place in this function!
+    """
+    # Get the set of all allowed types in the current scope.
+    equivalent_types_to_field_scope = [
```
obi1kenobi: I'm not a fan of this list being of type `GraphQL type object or None`, can we avoid adding to it if the hints / inverse hints dict doesn't have the key?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -312,20 +312,40 @@ def _get_recurse_directive_depth(field_name, field_directives):
...
+        context['type_equivalence_hints'].get(field_schema_type),
+        context['type_equivalence_hints_inverse'].get(field_schema_type)
+    ]
+
+    allowed_current_types = set([field_schema_type])
```
obi1kenobi: use a set literal

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -312,20 +312,40 @@ def _get_recurse_directive_depth(field_name, field_directives):
...
         field_schema_type in current_schema_type.interfaces
     )
 
-    if not any((has_union_type, is_same_type, is_implemented_interface)):
+    if not any((current_scope_is_allowed, is_implemented_interface)):
         raise GraphQLCompilationError(u'Edges expanded with a @recurse directive must either '
-                                      u'be of union type, or be of the same type as their '
-                                      u'enclosing scope, or be of an interface type that is '
+                                      u'be of the same type as their enclosing scope, a supertype '
+                                      u'of the enclosing scope which expands on a union of '
```
obi1kenobi: I'd remove `which expands on a union of both scopes` since that's an implementation detail and might cause confusion.

graphql_compiler/tests/test_helpers.py
obi1kenobi: I don't think `out_superclass_edge_to_superclass_or_subclass_type` is valid -- you can't recurse into a subclass type, because the starting point of the recursion isn't of that type (by definition).

PR #133, opened 2018-09-18
Make the README file appear on the PyPI project page.

obi1kenobi:
H/t @ColCarroll, inspired by https://github.com/ColCarroll/callisto/blob/master/setup.py


setup.py
```diff
@@ -40,33 +40,40 @@ def find_name():
...
-          'Programming Language :: Python :: 3',
-          'Programming Language :: Python :: 3.5',
-          'Programming Language :: Python :: 3.6',
-      ],
-      keywords='graphql database compiler orientdb',
-      python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*',
-      )
+def find_long_description():
+    """Return the content of the README.md file."""
+    return read_file('../README.md')
```
jmeulemans: It looks like we run the `make_new_release.sh` from the top-level of the repo, so wouldn't this be `./README.md`?

obi1kenobi: I don't think so -- `read_file()` attempts to read a "package" file from inside the `graphql_compiler` directory of the repo ([line 19](https://github.com/kensho-technologies/graphql-compiler/pull/133/files#diff-2eeaed663bd0d25b7e608891384b7298R19)), so we have to go up a directory from there. That's why `find_name` and `find_version` can read `__init__.py` directly.

jmeulemans: Oh yes you're right, I missed that. Looks good then.

ColCarroll:
Needs a new release before it shows up, right?

obi1kenobi:
Right, it won't be visible on pypi until the next release.

PR #137, opened 2018-10-01
Use metadata to decide when to mark locations prior to optional traversals

obi1kenobi:
This is a bit of cleanup that required the recently-merged metadata changes. Instead of maintaining an explicit location stack in the compiler frontend, we can now rely on the metadata we already collect to let us know if an additional `MarkLocation` block needs to be emitted before an optional traversal.

The net result of this PR is the removal of a field from the compiler frontend's `context` variable, and a nice (and large!) simplification of the most complex method in the compiler frontend.

cc @amartyashankha, we finally got to a good enough spot with the metadata to remove this code :)

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -393,15 +393,8 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
                                                      inner_location, context)
         basic_blocks.extend(inner_basic_blocks)
 
-    # The length of the stack should be the same before exiting this function
-    initial_marked_location_stack_size = len(context['marked_location_stack'])
-
     # step V-3: mark the graph position, and process output_source directive
     basic_blocks.append(_mark_location(location))
-    if not is_in_fold_scope(context):
-        # The following append is the Location corresponding to the initial MarkLocation
-        # for the current vertex and the `num_traverses` counter set to 0.
-        context['marked_location_stack'].append(_construct_location_stack_entry(location, 0))
 
     output_source = _process_output_source_directive(schema, current_schema_type, ast,
                                                      location, context, unique_local_directives)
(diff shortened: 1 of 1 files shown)


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -442,13 +435,16 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
             # Make sure there's a marked location right before it for the optional Backtrack
             # to jump back to. Otherwise, the traversal could rewind to an old marked location
             # and might ignore entire stretches of applied filtering.
-            if context['marked_location_stack'][-1].num_traverses > 0:
+            # The only way there might not be a marked location here is if the current location
+            # has existing child locations that are not folds.
```
bojanserafimov: Can you document this assumption? We don't have a good way to deal with these, but a convention like this might be ok for now: # Assumption: If the current location doesn't have existing child locations that are not folds, then there is no mark location here.

This way it's clerer that something is assumed, you can name your assumptions with short names if you want to, or if we do have a better way in the future we can easily find all of them.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -438,17 +431,24 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         within_optional_scope = 'optional' in context and not edge_traversal_is_optional
 
         if edge_traversal_is_optional:
-            # Entering an optional block!
-            # Make sure there's a marked location right before it for the optional Backtrack
-            # to jump back to. Otherwise, the traversal could rewind to an old marked location
-            # and might ignore entire stretches of applied filtering.
-            if context['marked_location_stack'][-1].num_traverses > 0:
+            # Invariant: There must always be a marked location corresponding to the query position
```
jmeulemans: I see that this invariant is enforced in `ir_sanity_checks.py`, but admittedly it took me awhile to realize that. Up to you but it might be helpful to others to mention that this invariant has a sanity check in place.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -442,13 +435,16 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
...
             # and might ignore entire stretches of applied filtering.
-            if context['marked_location_stack'][-1].num_traverses > 0:
+            # The only way there might not be a marked location here is if the current location
+            # has existing child locations that are not folds.
+            non_fold_child_locations = {
+                child_location
+                for child_location in query_metadata_table.get_child_locations(location)
+                if not isinstance(child_location, FoldScopeLocation)
+            }
+            if non_fold_child_locations:
```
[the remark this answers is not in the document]
obi1kenobi: The reason I decided against the `any` short-circuit was because the code ended up looking a bit too complex for my taste. I wanted to give the intermediate value a name, for the sake of the reader's intuition, even if that meant slightly slower execution.

graphql_compiler/compiler/compiler_frontend.py
obi1kenobi: Good point, I'll mention that in this comment.

PR #139, opened 2018-10-03
Fix bad equality checking in Traverse and Recurse blocks, fix tests.

obi1kenobi:
`Traverse` and `Recurse` blocks weren't including all of their fields in equality checks, so some of the test data for IR generation was incomplete and implicitly wrong. I don't think generated queries were affected, since the compiler's behavior was correct.

I also removed a stale comment from before we supported nested `@optional` directives, and tweaked a piece of code that assumed that situation would never occur.

### DIFF
--- a/graphql_compiler/compiler/blocks.py
+++ b/graphql_compiler/compiler/blocks.py
@@ -225,7 +225,8 @@ class Traverse(BasicBlock):
         Returns:
             new Traverse object
         """
-        super(Traverse, self).__init__(direction, edge_name, optional=optional)
+        super(Traverse, self).__init__(
+            direction, edge_name, optional=optional, within_optional_scope=within_optional_scope)
         self.direction = direction
         self.edge_name = edge_name
         self.optional = optional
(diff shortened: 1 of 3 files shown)


graphql_compiler/compiler/blocks.py
```diff
@@ -225,7 +225,8 @@ def __init__(self, direction, edge_name, optional=False, within_optional_scope=F
         Returns:
             new Traverse object
         """
-        super(Traverse, self).__init__(direction, edge_name, optional=optional)
+        super(Traverse, self).__init__(
+            direction, edge_name, optional=optional, within_optional_scope=within_optional_scope)
```
bojanserafimov: Maybe make these args mandatory

obi1kenobi: Yeah I thought about it. Maybe not right now -- I fixed the equality check so the tests will catch problems in those args, and when we move to py3-only, we can prevent them from being passed positionally, so maybe that will be enough.

PR #141, opened 2018-10-05
Amend metadata table for "explain" support.

gurer-kensho:
* QueryMetadataTable.get_explain_infos() returns information required for explain analysis.
It is in the form of a list of ExplainInfo objects.

* ExplainFilterInfo provides:
  - op_name: Operation is useful to estimate complexity, i.e. == is faster than between.
  - field_name: For checking if the field has index or not.
  - args: If there are $args, we can check the size of user provided input.

* ExplainRecurseInfo provides the recursion depth for now.

### DIFF
--- a/graphql_compiler/compiler/filters.py
+++ b/graphql_compiler/compiler/filters.py
@@ -11,6 +11,7 @@ from .helpers import (
     get_uniquely_named_objects_by_name, is_vertex_field_name, is_vertex_field_type,
     strip_non_null_from_type, validate_safe_string
 )
+from .metadata import ExplainFilterInfo
 
 
 def scalar_leaf_only(operator):
@@ -593,15 +594,15 @@ def is_filter_with_outer_scope_vertex_field_operator(directive):
     return op_name in OUTER_SCOPE_VERTEX_FIELD_OPERATORS
 
 
-def process_filter_directive(filter_operation_info, context):
+def process_filter_directive(filter_operation_info, location, context):
     """Return a Filter basic block that corresponds to the filter operation in the directive.
 
     Args:
         filter_operation_info: FilterOperationInfo object, containing the directive and field info
                                of the field where the filter is to be applied.
+        location: Location where this filter is used.
         context: dict, various per-compilation data (e.g. declared tags, whether the current block
                  is optional, etc.). May be mutated in-place in this function!
-        directive: GraphQL @filter directive object, obtained from the AST node
 
     Returns:
         a Filter basic block that performs the requested filtering operation
@@ -640,5 +641,10 @@ def process_filter_directive(filter_operation_info, context):
             op_name not in INNER_SCOPE_VERTEX_FIELD_OPERATORS):
         raise GraphQLCompilationError(u'The filter with op_name "{}" must be applied on a field. '
                                       u'It may not be applied on a type coercion.'.format(op_name))
+
+    context['metadata'].record_explain_info(location,
+                                            ExplainFilterInfo(field_name=filter_operation_info.field_name,
+                                                              op_name=op_name,
+                                                              args=operator_params))
 
     return process_func(filter_operation_info, context, operator_params)
--- a/graphql_compiler/compiler/metadata.py
+++ b/graphql_compiler/compiler/metadata.py
@@ -1,10 +1,10 @@
 # Copyright 2018-present Kensho Technologies, LLC.
 """Utilities for recording, inspecting, and manipulating metadata collected during compilation."""
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 
 import six
 
-from .helpers import Location
+from .helpers import Location, FoldScopeLocation
 
 
 LocationInfo = namedtuple(
@@ -20,6 +20,23 @@ LocationInfo = namedtuple(
         'recursive_scopes_depth',  # int, how many nested recursion scopes this location is in
         'is_within_fold',          # bool, True if this location is within a fold scope;
                                    #       fold scopes are not allowed to nest within each other.
+    )
+)
+
+
+ExplainFilterInfo = namedtuple(
+    'FilterExplainInfo',
+    (
+        'field_name',
+        'op_name',
+        'args',
+    )
+)
+
+ExplainRecurseInfo = namedtuple(
+    'RecurseExplainInfo',
+    (
+        'depth',
     )
 )
 
@@ -44,6 +61,8 @@ class QueryMetadataTable(object):
         self._inputs = dict()                # dict, input name -> input info namedtuple
         self._outputs = dict()               # dict, output name -> output info namedtuple
         self._tags = dict()                  # dict, tag name -> tag info namedtuple
+
+        self._explain_infos = defaultdict(list)
 
         # dict, revisiting Location -> revisit origin, i.e. the first Location with that query path
         self._revisit_origins = dict()
@@ -131,6 +150,18 @@ class QueryMetadataTable(object):
             raise AssertionError(u'Attempted to get the location info of an unregistered location: '
                                  u'{}'.format(location))
         return location_info
+
+    def record_explain_info(self, location, explain_info):
+        """Record estimation information about the location."""
+        if isinstance(location, FoldScopeLocation):
+            # NOTE(gurer): ignore filters inside the fold for now
+            return
+        record_location = location.at_vertex()
+        self._explain_infos[record_location].append(explain_info)
+
+    def get_explain_infos(self, location):
+        """Get estimation information of the location."""
+        return self._explain_infos[location]
 
     def get_child_locations(self, location):
         """Yield an iterable of child locations for a given Location/FoldScopeLocation object."""
--- a/graphql_compiler/tests/test_explain_info.py
+++ b/graphql_compiler/tests/test_explain_info.py
@@ -1,0 +1,57 @@
+# Copyright 2018-present Kensho Technologies, LLC.
+
+from . import test_input_data
+from ..compiler.compiler_frontend import OutputMetadata, graphql_to_ir
+from ..compiler.helpers import Location
+from ..compiler.metadata import ExplainFilterInfo, ExplainRecurseInfo
+from .test_helpers import get_schema
+
+
+def check_explain_info(graphql_test, expected):
+    """Verify query produces expected explain infos."""
+    schema = get_schema()
+    ir_and_metadata = graphql_to_ir(schema, graphql_test().graphql_input)
+    meta = ir_and_metadata.query_metadata_table
+    for loc, eis in expected:
+        assert meta.get_explain_infos(loc) == eis
+    assert len(expected) == len(meta._explain_infos)
+
+
+def test_filter():
+    check_explain_info(test_input_data.traverse_filter_and_output,
+                       [
+                           (Location(('Animal', 'out_Animal_ParentOf'), None, 1),
+                            [ExplainFilterInfo(field_name='out_Animal_ParentOf',
+                                               op_name='name_or_alias',
+                                               args=['$wanted'])]),
+                       ])
+
+
+def test_filters():
+    check_explain_info(test_input_data.complex_optional_traversal_variables,
+                       [
+                           (Location(('Animal',), None, 1),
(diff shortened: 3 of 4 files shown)


graphql_compiler/compiler/metadata.py
```diff
@@ -132,6 +151,18 @@ def get_location_info(self, location):
...
+        """Record estimation information about the location."""
+        if isinstance(location, FoldScopeLocation):
+            # NOTE(gurer): ignore filters inside the fold for now
+            return
+        record_location = location.at_vertex()
+        self._explain_infos[record_location].append(explain_info)
+
+    def get_explain_infos(self, location):
+        """Get estimation information of the location."""
+        return self._explain_infos[location]
```
jmeulemans: A custom error message might be nice here instead of failing on a `KeyError`.

graphql_compiler/compiler/metadata.py
```diff
@@ -24,6 +24,23 @@
 )
 
 
+FilterInfo = namedtuple(
```
jmeulemans: These are explicitly for supporting `EXPLAIN`, but the usefulness here seems more general. What are your (and @obi1kenobi's) thoughts on adding this information as general metadata on the `LocationInfo` class? 

This would get rid of the overlap between `RecurseInfo.depth` and `LocationInfo. recursive_scopes_depth`, and the `EXPLAIN` operation is just an interpretation of metadata we know about  each location in a query.

graphql_compiler/compiler/metadata.py
```diff
@@ -132,6 +151,18 @@ def get_location_info(self, location):
                                  u'{}'.format(location))
         return location_info
 
+    def record_explain_info(self, location, explain_info):
+        """Record estimation information about the location."""
+        if isinstance(location, FoldScopeLocation):
+            # NOTE(gurer): ignore filters inside the fold for now
+            return
+        record_location = location.at_vertex()
+        self._explain_infos[record_location].append(explain_info)
```
jmeulemans: It might be useful to draw a distinction here between explain infos we only expect 1 of (recurse), and explain infos we expect multiple of (filters). This could get weird if we mistakenly append 2 `RecurseInfo` tuples. 

This might end up being more ergonomic. Right now to ask a question about a `RecurseInfo` at the current location, we have to get all explain infos and then take the first things that's a `RecurseInfo`, vs storing this separately to enforce the uniqueness property.

graphql_compiler/compiler/metadata.py
gurer-kensho: Filter info is optional for most locations. It is easier to get [] from callers perspective than to check for exception or none.

graphql_compiler/compiler/metadata.py
gurer-kensho: I have separated FilterInfo and RecurseInfo into their own record/get calls. My original intention was to keep any external users happy as any change in the LocationInfo breaks them down. Not everybody is interested with every type of information too.

graphql_compiler/compiler/metadata.py
gurer-kensho: I have separated them. It made more sense. You can directly get RecurseInfo as you mentioned. Recording it twice gives an assertion. You can loop over filters without checking.

graphql_compiler/compiler/metadata.py
```diff
@@ -166,14 +168,12 @@ def get_filter_infos(self, location):
 
     def record_recurse_info(self, location, recurse_info):
         """Record recursion information about the location."""
-        if location in self._recurse_infos:
-            raise AssertionError(u'Cannot register RecurseInfo twice at same location: '
-                                 u'{}'.format(location))
-        self._recurse_infos[location] = recurse_info
+        record_location = location.at_vertex()
+        self._recurse_infos.setdefault(record_location, []).append(recurse_info)
```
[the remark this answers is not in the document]
obi1kenobi: Yes, this was my suggestion -- I personally dislike `defaultdict` and prefer not to use it since it makes things that look like reads actually be writes. I think `setdefault()` is obviously a write and not that much more difficult to type out, and I don't like the idea of `foo[bar]` modifying my `foo`.

graphql_compiler/compiler/metadata.py
```diff
@@ -45,6 +64,9 @@ def __init__(self, root_location, root_location_info):
         self._outputs = dict()               # dict, output name -> output info namedtuple
         self._tags = dict()                  # dict, tag name -> tag info namedtuple
 
+        self._filter_infos = dict()          # Location -> FilterInfo array
+        self._recurse_infos = dict()         # Location -> RecurseInfo array
```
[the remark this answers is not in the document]
obi1kenobi: I think it's possible to have multiple `RecurseInfo` objects at the same location. Take the following query, for example:
```
{
    Foo {
        out_Foo_Foo @recurse(depth: 2) {
            name @output(out_name: "out_foos")
        }
        in_Foo_Foo @recurse(depth: 2) {
            name @output(out_name: "in_foos")
        }
    }
}
```

I think the `Foo___1` location is going to have both `RecurseInfo` objects associated with it. Based on that, I think the current approach is fine.

graphql_compiler/compiler/metadata.py
obi1kenobi: I wasn't too worried about breaking `LocationInfo` with an additive-only change, but my preference was also for doing things separately from `LocationInfo` just to keep that namedtuple immutable. This way, it gets created once and is never altered again, without any loss of expressive power and with minimal cost to ergonomics. If this feels un-ergonomic, we could easily write a helper method that takes a location and returns the combination of (Filter/Recurse/Location)Info data in whatever form we feel like is easiest to use.

I also think this metadata has uses far beyond `EXPLAIN` -- for example, we can use this same metadata together with a stream of writes for cache invalidation, or to evaluate whether to allow certain GraphQL queries based on a metadata-driven policy, etc.

Also, note that `RecurseInfo.depth` and `LocationInfo.recursive_scopes_depth` are not the same depth -- the first is the max depth of that recursion, whereas the second is the number of distinct `@recurse` directives that the current location is within (which has bearing on how the query is scheduled and executed).

graphql_compiler/compiler/metadata.py
```diff
@@ -132,6 +154,27 @@ def get_location_info(self, location):
...
+        record_location = location.at_vertex()
+        self._filter_infos.setdefault(record_location, []).append(filter_info)
+
+    def get_filter_infos(self, location):
+        """Get information about filters at the location."""
+        return self._filter_infos.get(location, [])
+
+    def record_recurse_info(self, location, recurse_info):
+        """Record recursion information about the location."""
+        record_location = location.at_vertex()
```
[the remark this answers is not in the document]
obi1kenobi: I think this behavior is correct, because different visit counters signify different stages of the evaluation of the data at a given location, and I'd prefer to capture that info accurately in metadata.

graphql_compiler/compiler/metadata.py
```diff
@@ -132,6 +154,27 @@ def get_location_info(self, location):
                                  u'{}'.format(location))
         return location_info
 
+    def record_filter_info(self, location, filter_info):
+        """Record filter information about the location."""
+        if isinstance(location, FoldScopeLocation):
+            # NOTE(gurer): ignore filters inside the fold for now
```
obi1kenobi: @gurer-kensho if you'd like a suggestion for a following PR, you could simply add an `at_vertex()` method on `FoldScopeLocation` and then delete this check. That method doesn't exist because there wasn't a use for it, and not because it doesn't make sense to have it -- `FoldScopeLocation` objects can point to fields just as much as vanilla `Location` objects can.

graphql_compiler/tests/test_explain_info.py
```diff
@@ -0,0 +1,95 @@
...
+                        [FilterInfo(field_name='name',
+                                    op_name='=',
+                                    args=['%parent_fed_at_event']),
+                         FilterInfo(field_name='event_date',
+                                    op_name='between',
+                                    args=['%other_child_fed_at', '%parent_fed_at'])]),
+                   ],
+                   [
+                   ])
+
```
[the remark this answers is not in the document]
obi1kenobi: Agreed, probably would be good to check a few combinations of multiple filters on the same field / on fields and edges etc.

graphql_compiler/tests/test_explain_info.py
```diff
@@ -0,0 +1,95 @@
...
+            recurse = meta.get_recurse_infos(location)
+            self.assertEqual(recurse, expected_recurses.get(location, []))
+            if recurse:
+                del expected_recurses[location]
+        # Any expected infos missing?
+        self.assertEqual(0, len(expected_filters))
+        self.assertEqual(0, len(expected_recurses))
+
+    def test_filter(self):
+        self.check(test_input_data.traverse_filter_and_output,
```
obi1kenobi: These are a bit hard to read, consider assigning the expected values to variables and reducing the number of distinct indentation levels in the same expression.

graphql_compiler/tests/test_explain_info.py
```diff
@@ -0,0 +1,95 @@
...
+                       (Location(('Animal', 'out_Animal_ParentOf'), None, 1),
+                        [FilterInfo(field_name='out_Animal_ParentOf',
+                                    op_name='name_or_alias',
+                                    args=['$wanted'])]),
+                   ],
+                   [
+                   ])
+
+    def test_filters(self):
+        self.check(test_input_data.complex_optional_traversal_variables,
```
obi1kenobi: It's convention elsewhere in the codebase to call each test function that uses `test_input_data` based on the name of the `test_input_data` function it uses. So if using `test_input_data.foo`, name the test function `def test_foo(self):` -- this makes it easier to search. I'd move the explanation of why this test case in particular is chosen to a docstring or comment instead of the function name.

graphql_compiler/tests/test_explain_info.py
```diff
@@ -0,0 +1,95 @@
...
+                   [
+                       (Location(('Animal',), None, 1),
+                        [RecurseInfo(edge_direction='out', edge_name='Animal_ParentOf', depth=1)]),
+                   ])
+
+    def test_two_recurses(self):
+        self.check(test_input_data.two_consecutive_recurses,
+                   [
+                       (Location(('Animal',), None, 1),
+                        [FilterInfo(field_name='Animal',
```
obi1kenobi: I'd be particularly interested in a test case that includes the `has_edge_degree` filter -- both what location it appears as assigned to, and what `field_name` carries in that case. We might have to tweak how data is passed around / what data `FilterInfo` carries based on that.

This is my only major concern with this PR at the moment.

graphql_compiler/compiler/metadata.py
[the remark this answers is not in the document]
jmeulemans: I see, I thought the `recurse` was being associated with the (Foo, out_Foo_Foo) location.

graphql_compiler/compiler/filters.py
```diff
@@ -641,4 +642,14 @@ def process_filter_directive(filter_operation_info, context):
         raise GraphQLCompilationError(u'The filter with op_name "{}" must be applied on a field. '
                                       u'It may not be applied on a type coercion.'.format(op_name))
 
+    context['metadata'].record_filter_info(
+        location,
+        FilterInfo(
+            fields=(filter_operation_info.field_name,) if op_name != 'name_or_alias' else
```
obi1kenobi: Extract the `fields` value into a variable? This one statement is getting pretty complicated now.

Bonus points if you make some constants for the filter names.

graphql_compiler/compiler/metadata.py
obi1kenobi: We decided on doing this in a subsequent PR, right?

graphql_compiler/compiler/filters.py
gurer-kensho: Extracted it. I would prefer to leave filter name constants to another patch because if I do that, I can change the rest of the code to use them too. That could simplify/cleanup some things which is nice. But it is conceptually somewhat separate from what this patch is doing. So I don't necessarily want to mix both.

PR #143, opened 2018-12-12
Add scaffolding for SQL backend without implementation.

jmeulemans:
This PR steps back and takes a more bite-sized approach to add the new SQL backend. The first step here sets up the main entry-points to compilation without implementing them.

The main change related to these entry points is an interface change to allow backend specific metadata to optionally be passed to the query emit step. For the SQL backend, this metadata contains information related to the tables present in the DB.


setup.py
```diff
@@ -61,6 +61,7 @@ def find_long_description():
         'graphql-core==1.1',
         'pytz>=2016.10',
         'six>=1.10.0',
+        'sqlalchemy>=1.2.1',
```
obi1kenobi: How do you feel about a `<1.3` or `<2` bound in addition to this? `graphql-core` is pinned to an exact version since it's very important, and I'd imagine that a major (and maybe even minor) `sqlalchemy` version bump might break things as well.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -0,0 +1,7 @@
+# Copyright 2018-present Kensho Technologies, LLC.
+"""Transforms a tree representation of an SQL query into an executable SQLAlchemy query."""
```
obi1kenobi: Imperative?

graphql_compiler/__init__.py
```diff
@@ -53,12 +58,48 @@ def graphql_to_match(schema, graphql_query, parameters, type_equivalence_hints=N
         query=insert_arguments_into_query(compilation_result, parameters))
 
 
+def graphql_to_sql(schema, graphql_query, parameters, compiler_metadata,
```
obi1kenobi: Do you think in the long run (~v2.0) we'll want to include `type_equivalence_hints` into `compiler_metadata`? That way we could have a single compiler metadata object, and ideally make it the first argument so it can be partial-ed away easily. This is because that metadata and the schema don't really change, but the GraphQL query and parameters do -- so users could create a partial function `def compile_graphql(graphql_query: str, parameters: Dict[str, Any]) -> CompilationResult` for easier use.

In either case, I wouldn't change anything like that _now_, just thinking about the future.

setup.py
jmeulemans: Great idea, I'll pin to <1.3 to be safe.

graphql_compiler/__init__.py
jmeulemans: Yeah I think that's a great idea, the intent of my metadata object and the type equivalence hints are largely the same and unifying would make sense. That would seem a little less ergonomic to have metadata supplied first, but I like the intent. Maybe we would expect people to use `funcy.rpartial` since that's a project dependency?

graphql_compiler/__init__.py
```diff
@@ -53,12 +58,48 @@ def graphql_to_match(schema, graphql_query, parameters, type_equivalence_hints=N
...
+    raise NotImplementedError(u'Compiling GraphQL to SQL is not yet supported.')
+
+
 def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints=None):
     """Compile the GraphQL input using the schema into a Gremlin query and associated metadata.
 
     Args:
         schema: GraphQL schema object describing the schema of the graph to be queried
-        graphql_string: the GraphQL query to compile to Gremlin, as a string
+        graphql_query: the GraphQL query to compile to Gremlin, as a string
```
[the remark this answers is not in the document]
jmeulemans: No problem! Have to give pycharm the credit on this one

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+                                Recursive expansion of type equivalence hints is not performed,
+                                and only type-level correctness of this argument is enforced.
+                                See README.md for more details on everything this parameter does.
+                                *****
+                                Be very careful with this option, as bad input here will
+                                lead to incorrect output queries being generated.
+                                *****
+
+    Returns:
+        Tree representation of IR blocks for recursive traversal by SQL backend.
```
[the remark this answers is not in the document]
jmeulemans: Done

graphql_compiler/compiler/emit_sql.py
jmeulemans: Done

PR #144, opened 2018-12-20
 Adding integration test harness for backend compatibility/validation

jmeulemans:
The intent here is to make it straightforward to say that this compiler supports targeting database X, and that the results are guaranteed to be the same across all supported backend targets.

This is especially important with differences between typical SQL query semantics and GraphQL query semantics. 

One notable exception here is Oracle DB. It was excluded for now only because it requires a Docker login and some extra work to supply that login to the Travis build process.


docker-compose.yml
```diff
@@ -8,3 +8,33 @@ services:
...
+    environment:
+      MYSQL_ROOT_PASSWORD: root
+  mssql:
+    image: mcr.microsoft.com/mssql/server:2017-latest
+    restart: always
+    ports:
+      - "127.0.0.1:1433:1433"
+    environment:
+      ACCEPT_EULA: "yes"
+      MSSQL_SA_PASSWORD: Root-secure1 # password requirements are more stringent for MSSQL image
```
obi1kenobi: nit: maybe two spaces before `#`?

Pipfile
```diff
@@ -6,10 +6,14 @@ verify_ssl = true
 [dev-packages]
 bandit = "==1.4.0"
 coveralls = "==1.2.0"
+cython = "==0.29.2"
```
obi1kenobi: Does `cython` have to be included explicitly, instead of getting picked up as a dependency? I don't see `import cython` anywhere in our code, so I'd prefer not to include and pin it explicitly if possible.

graphql_compiler/tests/conftest.py
```diff
@@ -35,6 +50,24 @@ def init_graph():
...
+    request.cls.graph_client = init_snapshot_graph_client
+
+
+@pytest.fixture(scope='class')
+def integration_graph_client(request, init_integration_graph_client):
+    """Get a client for an initialized db, with all test data imported."""
+    request.cls.graph_client = init_integration_graph_client
+
+
+@pytest.fixture(scope="class")
```
obi1kenobi: nit: use single-quotes around `class` for consistency?

graphql_compiler/tests/conftest.py
```diff
@@ -35,6 +50,24 @@ def init_graph():
...
+def sql_integration_data(request):
+    """Generate integration data for SQL backends."""
+    sql_test_backends = init_sql_integration_test_backends()
+    generate_sql_integration_data(sql_test_backends)
+    request.cls.sql_test_backends = sql_test_backends
+    # yield the fixture to allow testing class to run
+    yield
```
obi1kenobi: Oh interesting, I didn't realize pytest fixtures supported this.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -0,0 +1,57 @@
...
+    """Compiles and runs a SQL query against the supplied SQL backend."""
+    # TODO: un-mock the SQL compilation once the SQL backend can run queries.
+    mock_compilation_result = CompilationResult(
+        query=text('SELECT name AS animal_name FROM animal'),
+        language=SQL_LANGUAGE,
+        input_metadata={},
+        output_metadata={'animal_name': OutputMetadata(GraphQLString, False)}
+    )
+
+    def mock_sql_compilation(schema, graphql_query, parameters, compiler_metadata):
```
obi1kenobi: missing docstring here

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -0,0 +1,57 @@
...
+        return mock_compilation_result
+
+    compilation_result = mock_sql_compilation(schema, graphql_query, parameters, None)
+    query = compilation_result.query
+    results = [
+        dict(result) for result
```
obi1kenobi: I'd move the `for` to the next line, to keep the `for` and `in` on the same line since they are part of the same clause.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -0,0 +1,77 @@
...
+        for results in six.itervalues(backend_results):
+            self.assertResultsEqual(expected_results, results)
+
+    @classmethod
+    def compile_and_run_query(cls, graphql_query, parameters):
+        """Compiles and runs the graphql query with the supplied parameters against all backends.
+
+        Args:
+            graphql_query: str, GraphQL query string to run against every backend.
+            parameters: Dict, input parameters to the query.
```
obi1kenobi: `Dict[str, Any]`, right?

graphql_compiler/tests/test_backend.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2018-present Kensho Technologies, LLC.
+class TestBackend(object):
```
obi1kenobi: Do we need a class here? Can these just be module constants instead, and we can refer to them via the module name like `test_backend.POSTGRES`?

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -16,3 +37,60 @@ def generate_data(client):
                     continue
 
                 client.command(sanitized_command)
+
+
+@retry(tries=20, timeout=1)  # pylint: disable=no-value-for-parameter
```
obi1kenobi: `tries=20` seems very high, tests might take a long time to fail

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -16,3 +37,60 @@ def generate_data(client):
...
+        engine = create_engine(connection_string)
+        # MYSQL and MARIADB do not have a default DB so a DB must be created
+        if backend_name in {TestBackend.MYSQL, TestBackend.MARIADB}:
+            # safely create the DB
+            engine.execute(text('DROP DATABASE IF EXISTS animals; CREATE DATABASE animals'))
+            # update the connection string and engine to connect to this DB specifically
+            connection_string = connection_string + u'/animals'
+            engine = create_engine(connection_string)
+        connection = engine.connect()
+        transaction = connection.begin()
```
obi1kenobi: This seems a bit strange, starting a transaction and passing it around like that. I think it'd be better practice if the component that starts the transaction also makes sure that it ends, either with a commit or an abort. It's also probably a good idea to have isolation between tests by running each test in a separate transaction.

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
```diff
@@ -0,0 +1,5 @@
+# Copyright 2018-present Kensho Technologies, LLC.
+create vertex Animal set name = 'Animal 1', uuid = 'cfc6e625-8594-0927-468f-f53d864a7a50'
+create vertex Animal set name = 'Animal 2', uuid = 'cfc6e625-8594-0927-468f-f53d864a7a51'
+create vertex Animal set name = 'Animal 3', uuid = 'cfc6e625-8594-0927-468f-f53d864a7a52'
+create vertex Animal set name = 'Animal 4', uuid = 'cfc6e625-8594-0927-468f-f53d864a7a53'
```
obi1kenobi: Is the plan in a future PR to make the integration test data converge with the snapshot test data? Otherwise, we'd be maintaining two separate input data sets. And if they converge, then the snapshot tests start working for all databases too.

Pipfile
jmeulemans: This has been an issue with pymssql not picking up the dependency. I'll double check here that it's still an issue though.

obi1kenobi: Yeah I asked around and it seems like a few other people have seen it as well. Let's leave it in.

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
jmeulemans: That would be great, maintaining two is not ideal. The value I see in this current approach is that the results can be reasoned about. The dataset is small and tractable since we generate it ourselves.

On the snapshot side, a failure between backends is going to be somewhat difficult to debug if there were a discrepancy because the resultant dataset can be hard to reason about, given that it was randomly generated. To unify then, perhaps we could revisit the snapshot dataset, and create a non-random version?

graphql_compiler/tests/test_data_tools/data_tool.py
jmeulemans: From local testing if I see about 15s to run the test from against cold start of the containers, so this gives us a little bit of overhead. Is 20s to failure too long? We give the OrientDB backend that long to startup as well.

graphql_compiler/tests/test_backend.py
jmeulemans: Yeah that sounds good.

graphql_compiler/tests/integration_tests/test_backends_integration.py
jmeulemans: Yeah will update

graphql_compiler/tests/conftest.py
jmeulemans: I didn't either, this is my first encounter with fixtures. It's pretty slick.

graphql_compiler/tests/conftest.py
jmeulemans: Sure will fix.

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
obi1kenobi: I think a hand-curated dataset would be hard to maintain as we add features to the compiler, and may become a bottleneck to development. I think of these tests as the last line of defense -- if something fails at this level, it indicates a missing unit test. Since such failures shouldn't happen often (or, ideally, at all), I'm not that worried about having to start from a complex randomly-generated test case and try to work out a minimal repro, and I'd prefer to make it easier to develop new features instead.

What do you think?

graphql_compiler/tests/test_data_tools/data_tool.py
obi1kenobi: I guess it's okay if that's how long it takes. We might want to revisit in the future and see if we can make tests a bit faster -- a shorter feedback loop would speed up development.

graphql_compiler/tests/test_data_tools/data_tool.py
jmeulemans: Good call, I'll leave creating engines and connections in the module scope, and start transactions at the function scope.

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
jmeulemans: With the SQL backend, I was planning on using hand-curated data at the unit-test level. An easy way to test was based on using Sqlite, compiling and running a query to get expected results back. From there it seemed natural to generalize and test across other relational DBs + OrientDB while I was at it. Especially since we don't do this sort of testing yet beyond reg tests with the snapshots. 

I do agree that features to the compiler, especially for MATCH should not be gated based on having hand-curated data, but I expect most SQL features will get it regardless unless there's an alternative way to test the SQL backend. Could the scope of tests like this then be limited to the features both backends support?

graphql_compiler/tests/integration_tests/integration_test_helpers.py
jmeulemans: Okay will do.

graphql_compiler/tests/test_data_tools/data_tool.py
jmeulemans: I forgot here, the outer transaction was so that after the integration test is over, the data is cleared by rolling back the transaction. Instead there is now a nested transaction for each test function that is run, which is rolled back after the test.

As far as what you mean by passing around a transaction, this is managed by the test fixture that calls this, and that fixture is also responsible for calling rollback. Should it be more explicit than that? That would be the `sql_integration_data` function, which calls `init` here and then `teardown`, which then begin and rollback the transactions, respectively.

graphql_compiler/tests/test_data_tools/data_tool.py
jmeulemans: So just for reference here, I timed out 10 integration tests and once the fixture is loaded, they each take about 100 ms. The fixture loading is about 1s. So even with 50 integration tests, we're looking at < 10s all told.

graphql_compiler/tests/test_data_tools/data_tool.py
obi1kenobi: > I forgot here, the outer transaction was so that after the integration test is over, the data is cleared by rolling back the transaction. Instead there is now a nested transaction for each test function that is run, which is rolled back after the test.

I'm not sure I understand this. Is the intention to run each test in its own nested transaction, within the single transaction that wraps all tests? What's the benefit of that versus not nesting transactions and relying on the fixture to clean up after itself anyway? I am concerned about both the complexity as well as potential cross-test issues here, such as the database deciding that the outer transaction must abort (for any reason, as it is allowed to do so) and preventing all remaining tests from running at all.

As for what becomes part of the `SqlTestBackend`, I wasn't clear on why we're sending the connection string, and also the engine created from said connection string, as well as a connection from said engine. Shouldn't just one of them be enough, depending on how much of the test infrastructure we want tests to share?

Happy to chat over video call if that would be faster and easier here.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -0,0 +1,77 @@
...
+            results = compile_and_run_sql_query(
+                cls.schema, graphql_query, parameters, sql_test_backend)
+            backend_to_results[backend_name] = results
+        for backend_name in MATCH_BACKENDS:
+            results = compile_and_run_match_query(
+                cls.schema, graphql_query, parameters, cls.graph_client)
+            backend_to_results[backend_name] = results
+        return backend_to_results
+
+    @pytest.mark.usefixtures('integration_graph_client', 'sql_integration_data')
```
obi1kenobi: OrientDB is a possible backend here, right? If so, should the fixtures be renamed to indicate that? For example, right now, `sql_integration_data` sounds like it's doing something SQL-specific and exclusive of OrientDB.

obi1kenobi: I guess I'm just confused about which fixture provides what.

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
obi1kenobi: Might be best to sync up on this sometime tomorrow and figure out the best way forward, as I'd really like to unblock this ASAP.

jmeulemans: I figured out a unit-test method, I'll make a separate PR for that setup.

graphql_compiler/tests/conftest.py
```diff
@@ -35,6 +51,30 @@ def init_graph():
...
+@pytest.fixture(scope='class')
+def integration_graph_client(request, init_integration_graph_client):
+    """Get a client for an initialized db, with all test data imported."""
+    request.cls.graph_client = init_integration_graph_client
+
+
+@pytest.fixture(scope='class')
+def sql_integration_data(request):
+    """Generate integration data for SQL backends."""
+    # initialize each SQL backend, and open a transaction on each one.
```
obi1kenobi: this comment seems stale

graphql_compiler/tests/conftest.py
```diff
@@ -35,6 +51,30 @@ def init_graph():
...
+    sql_test_backends = init_sql_integration_test_backends()
+    # write fixture test data within the transaction
```
obi1kenobi: this too, I don't see a transaction

graphql_compiler/tests/conftest.py
```diff
@@ -35,6 +51,30 @@ def init_graph():
...
+    generate_sql_integration_data(sql_test_backends)
+    # make sql engines accessible within the test class
+    request.cls.sql_backend_name_to_engine = {
+        backend_name: sql_test_backend.engine for
```
obi1kenobi: put the `for` on its own line

graphql_compiler/tests/conftest.py
```diff
@@ -35,6 +51,30 @@ def init_graph():
...
+        backend_name, sql_test_backend in six.iteritems(sql_test_backends)}
```
obi1kenobi: `}` on its own line too

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -0,0 +1,98 @@
...
+
+from ..test_helpers import get_schema
+from .integration_backend_config import MATCH_BACKENDS, SQL_BACKENDS
+from .integration_test_helpers import (
+    compile_and_run_match_query, compile_and_run_sql_query, sort_db_results
+)
+
+
+# Store the test parametrization for running against all backends. Individual tests can customize
+# the list of backends to test against with the full @parametrized.expand([...]) decorator
```
obi1kenobi: `@parametrized.expand([...])` doesn't seem to be used anywhere

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def assertResultsEqual(self, graphql_query, parameters, backend_name, expected_results):
+        """Assert that two lists of DB results are equal, independent of order."""
+        backend_results = self.compile_and_run_query(graphql_query, parameters, backend_name)
+        try:
+            self.assertListEqual(sort_db_results(expected_results),
+                                 sort_db_results(backend_results))
+        except AssertionError as error:
+            # intercept and modify error message to indicate which backend(s) failed
+            error.args = (u'Failure for backend {}: '.format(backend_name) + error.args[0],)
```
obi1kenobi: If already formatting, why not format all the way instead of doing `.format()` followed by a string concatenation? Also, note that you're discarding `error.args[1:]` entirely here, which is unexpected.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -0,0 +1,98 @@
...
+            raise error
```
obi1kenobi: Maybe `raise` instead of `raise error` so the stack trace is preserved?

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -16,3 +36,68 @@ def generate_data(client):
...
+        if backend_name not in EXPLICIT_DB_BACKENDS:
+            continue
+        # explicitly release engine resources, specifically to disconnect from active DB
+        # some backends including Postgres do no not allow an in use DB to be dropped
+        sql_test_backend.engine.dispose()
+        # connect to base server, not explicit DB, so DB can be dropped
+        engine = create_engine(sql_test_backend.base_connection_string)
+        # set execution options to AUTOCOMMIT so that the DB drop is not performed in a transaction
+        # as this is not allowed on some SQL backends
+        (engine.execution_options(isolation_level="AUTOCOMMIT")
```
obi1kenobi: How about extracting the `text('...')` into a variable so that you can fit all of this on the same line and don't need the wrapping `()` around the entire thing?

PR #148, opened 2018-12-25
Add python 3.7 to the Travis testing matrix

obi1kenobi:


.travis.yml
```diff
@@ -28,6 +28,11 @@ matrix:
     python: "3.6"
     script:
       - py.test --cov=graphql_compiler graphql_compiler/tests
+  - name: "Python 3.7 unit tests"
+    python: '3.7'
+    sudo: true
```
jmeulemans: Why does 3.7 require sudo when the others don't? I see that used to be an issue, but this repo at least seems to think the issue has since been resolved: https://github.com/deadsnakes/travis-ci-python3.7-example

PR #154, opened 2019-01-08
Make the "make_new_release.sh" script move to the git root directory

obi1kenobi:
This is in line with the change to the `copyright_line_check.sh` script.

### DIFF
--- a/scripts/make_new_release.sh
+++ b/scripts/make_new_release.sh
@@ -7,12 +7,8 @@ set -euo pipefail
 # Enable recursive globbing, and make globs that do not match return null values.
 shopt -s globstar nullglob
 
-# Make sure the current working directory is the root directory.
-if [ ! -f "./requirements.txt" ] || [ ! -f "./CHANGELOG.md" ]; then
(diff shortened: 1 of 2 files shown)


scripts/make_new_release.sh
```diff
@@ -7,12 +7,8 @@ set -euo pipefail
...
 shopt -s globstar nullglob
 
-# Make sure the current working directory is the root directory.
-if [ ! -f "./requirements.txt" ] || [ ! -f "./CHANGELOG.md" ]; then
-    echo -e 'Please run this script from the root directory of the repo:\n'
-    echo -e '    ./scripts/make_new_release.sh\n'
-    exit 1
-fi
+# Make sure the current working directory for this script is the root directory.
+cd "$(git rev-parse --show-toplevel)"
```
jbylund: Maybe consider putting `cd $(git -C $(dirname $(realpath "${0}") ) rev-parse --show-toplevel )`

obi1kenobi: Fun fact: `realpath` does not exist by default on OSX. Going to remove that bit for now since the workaround is more work than it's worth at the moment.

PR #158, opened 2019-01-17
Add a "__count" property that captures the size of a fold scope

obi1kenobi:
This PR solves the problem of wanting to filter on, or output, the number of elements in a fold scope. The previous recommended solution was to output an unnecessary field in the fold scope, and do the counting on the client side. This is an unfortunate and complex workaround, and has poor performance since it requires the transfer of potentially a large amount of irrelevant data that is simply going to be counted.

There are three possible approaches for implementing this counting functionality:
- adding a custom `__count` field that is only usable from within a `@fold` (the chosen approach);
- adding a new directive that exposes this functionality, or
- modifying the existing `@fold` directive to support this functionality.

Modifying the existing `@fold` directive is not an appealing choice, since it would somehow have to both support outputting data (what the `@output` directive does) and filtering on the count (what the `@filter` directive is meant for). This would introduce code duplication, and cause confusion among users since there would now be several unrelated ways to output and filter data -- you'd have to know which way to use for what.

Adding a new directive is not an appealing choice, because in addition to the same issues as above, we'd also have to define how this new directive interacts with the existing set of directives, leading to a potentially complex and confusing set of rules.

Additionally, both of these approaches would have to deal with the fact that a user might want to specify several different filters for the count.

This is why we choose to add this new `__count` meta-field that is only usable within a `@fold` scope. This allows users to use the existing `@filter` and `@output` directives that they already know. Specifying several different filters here is handled in the same way it works for any other field -- just specify multiple `@filter` directives on the field in question.

This new field has to be added to the GraphQL schema in order for queries that use it to pass validation. This PR provides a new `insert_meta_fields_into_existing_schema()` function that will append the field to all types and interfaces in an existing schema, as well as providing the meta field definitions in a way that schema builders can easily import and include at schema build time.

Since adding support for this field in the Gremlin backend would require some non-trivial refactoring, in the interest of time the Gremlin backend will not initially support this feature. I will open an issue to add support for it at a future point.

The following steps are required before this PR is ready to be merged:
- [x] functions for inserting the new meta field into a GraphQL schema
- [x] modifications to the compiler front-end to generate correct IR
- [x] modifications to the MATCH backend so that it generates correct MATCH output queries
- [x] updates to the compiler documentation explaining how to use this new feature

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -305,7 +305,7 @@ def _compile_property_ast(schema, current_schema_type, ast, location,
 
         graphql_type = strip_non_null_from_type(current_schema_type)
         if is_in_fold_scope(context):
-            # Fold outputs are only allowed at the last level of traversal
+            # Fold outputs are only allowed at the last level of traversal.
             set_fold_innermost_scope(context)
 
             if location.field != COUNT_META_FIELD_NAME:
@@ -673,6 +673,9 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
     for filter_operation_info in filter_operations:
         filter_block = process_filter_directive(filter_operation_info, location, context)
         if isinstance(location, FoldScopeLocation) and location.field == COUNT_META_FIELD_NAME:
+            # Filtering on the fold count field is only allowed at the innermost scope of a fold.
+            set_fold_innermost_scope(context)
+
             # This Filter is going in the global operations section of the query, so it cannot
             # use LocalField expressions since there is no "local" location to use.
             # Rewrite it so that all references of data at a location instead use ContextFields.
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -676,6 +676,11 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
             visitor_fn = expressions.make_replacement_visitor(expected_field, replacement_field)
             filter_block = filter_block.visit_and_update_expressions(visitor_fn)
 
+            visitor_fn = expressions.make_type_replacement_visitor(
+                expressions.ContextField,
+                lambda context_field: expressions.GlobalContextField(context_field.location))
+            filter_block = filter_block.visit_and_update_expressions(visitor_fn)
+
             context['fold_has_count_filter'] = True
             context['global_filters'].append(filter_block)
         else:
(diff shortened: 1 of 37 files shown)


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -502,15 +516,14 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         basic_blocks.extend(inner_basic_blocks)
 
         if edge_traversal_is_folded:
-            _validate_fold_has_outputs(context['fold'], context['outputs'])
+            _validate_fold_has_outputs_or_count_filter(
+                context['fold'], 'fold_has_count_filter' in context, context['outputs'])
```
lodrion: constant for the string

obi1kenobi:
@lodrion and @jmeulemans, all the subtasks are finished and I think this is ready for final review.

graphql_compiler/schema.py
```diff
@@ -275,3 +277,61 @@ def _parse_datetime_value(value):
...
+EXTENDED_META_FIELD_DEFINITIONS = OrderedDict((
+    (COUNT_META_FIELD_NAME, GraphQLField(GraphQLInt)),
+))
+
+
+def is_meta_field(field_name):
+    """Return True if the field is considered a meta field in the schema, and False otherwise."""
+
+
+def insert_meta_fields_into_existing_schema(graphql_schema):
```
jmeulemans: Without this being called, the `__count` property if queried will result in a `String does not validate: [GraphQLError('Cannot query field "__count" on type "Foo".',)]` error. There seems to be a little bit of a disconnect between that error and the fact that `__count` is a meta-field that needs to be injected into the schema, and 
So the two things I'm wondering about here are: 
1. What are your thoughts of injecting by default into the schema (maybe through a compiler supplied `build_compiler_schema_ast` to drop-in replace the graphql core `build_schema_ast`?
2. Should the error message around missing meta fields specifically be improved to indicate that this might be the root cause of the issue, ie `This error could be caused because the schema was not augmented with insert_meta_fields_into_existing_schema` or something along those lines to help with the solution?

obi1kenobi: This is a great catch with great suggestions.

- I'm a bit wary of "magically" injecting into the schema by default, without the user being aware that their schema is being modified. I think it's reasonable to make the user use either the provided `insert_meta_fields_into_existing_schema()` function, or the `EXTENDED_META_FIELD_DEFINITIONS` constant when building the schema, since both of those amount to one "edit" to the traditional workflow of building the schema, the same as replacing `build_schema_ast()` with a custom `build_compiler_schema_ast()`.

- I'm definitely open to trying to make the error message around missing meta fields steer the user in the right direction. I'm going to add a bit more documentation around inserting the meta fields into the schema. Would you mind if I open a `user friendliness` issue for catching the "missing meta field" error and reraising it with some more useful info, since that might not be trivial to do and might take some work? I'd prefer not to block this PR because of that, since it's not immediately clear to me how much work doing that would be, and because that's a "you didn't follow the documentation, here's what you missed" sort of error improvement that I think is important but not a blocker.

PR #161, opened 2019-01-18
Adding SELECT...FROM SQL functionality

jmeulemans:
This is ready for a high-level first pass, I do still need to double check details, and likely revisit and add more testing.


graphql_compiler/compiler/emit_sql.py
```diff
@@ -1,7 +1,115 @@
...
+
+def _get_output_columns(node, context):
+    """Get the output columns required by the query.
+
+    Args:
+        node: SqlNode, the current node.
+        context: CompilationContext, global compilation state and metadata.
+
+    Returns:
+        List[column], list of SqlAlchemy columns to output for this query.
```
obi1kenobi: is `column` the name of the type? Should it be `Column` instead?

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
-                                Be very careful with this option, as bad input here will
-                                lead to incorrect output queries being generated.
-                                *****
+        type_equivalence_hints: optional dict of GraphQL interface or type -> GraphQL union. Unused.
 
     Returns:
         tree representation of IR blocks for recursive traversal by SQL backend.
     """
-    raise NotImplementedError(u'SQL IR lowering is not yet implemented.')
+    construct_result = ir_blocks.pop()
```
obi1kenobi: Assert that this is indeed the `ConstructResult` block. If that ever changes, this implicit assumption will be a nightmare to track down.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
+from .. import blocks
+from ..ir_lowering_sql.constants import SqlOutput
+from ..metadata import LocationInfo
+from ... import exceptions
+
 ##############
 # Public API #
 ##############
 
+_SKIPPABLE_BLOCKS = (
```
obi1kenobi: nit: These are _types_ of blocks, not blocks themselves -- how about `_SKIPPABLE_BLOCK_TYPES`?

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
+        query_metadata_table: QueryMetadataTable object containing all metadata collected during
+                              query processing, including location metadata (e.g. which locations
+                              are folded or optional).
+
+    Returns:
+        Dict[Tuple[str], LocationInfo], dictionary mapping query path to LocationInfo at that path.
+    """
+    query_path_to_location_info = {}
+    for location, location_info in query_metadata_table.registered_locations:
+        if location.query_path in query_path_to_location_info:
```
obi1kenobi: Not all `BaseLocation` objects (the type inside the iterable at `query_metadata_table.registered_locations`) have a `query_path` attribute. You need to filter out the ones that are instances of `Location` here.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
+
+        query_path_to_location_info[location.query_path] = location_info
+    return query_path_to_location_info
+
+
+def _location_infos_equal(left, right):
+    """Return True if LocationInfo objects are equivalent for the SQL backend, False otherwise."""
+    if not isinstance(left, LocationInfo) or not isinstance(right, LocationInfo):
+        raise AssertionError(
+            (u'Unsupported LocationInfo comparison between types {} and {} '
```
obi1kenobi: the `()` that start on this line are unnecessary

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
+
+    parent_query_paths_equal = (
+        (left.parent_location is None and right.parent_location is None) or
+        (left.parent_location.query_path == right.parent_location.query_path))
+
+    recursive_scopes_depths_equal = (left.recursive_scopes_depth == right.recursive_scopes_depth)
+
+    types_equal = left.type == right.type
+
+    all_equal = (optional_scopes_depth_equal and parent_query_paths_equal and
```
obi1kenobi: alternatively, you could use `all()` here

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
+            field_name=field_name,
+            output_name=output_name,
+            graphql_type=query_path_to_location_info[output_query_path].type)
+        output_field_mapping = query_path_to_output_fields.setdefault(output_query_path, [])
+        output_field_mapping.append(output_field_info)
+    return query_path_to_output_fields
+
+
+def _map_block_index_to_location(ir_blocks):
+    """Associate each IR block with it's corresponding location, by index."""
```
obi1kenobi: `it's` -> `its`

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -1,35 +1,148 @@
...
+    block_index_to_location = {}
+    current_block_ixs = []
+    for num, ir_block in enumerate(ir_blocks):
+        current_block_ixs.append(num)
+        if isinstance(ir_block, blocks.MarkLocation):
+            for ix in current_block_ixs:
+                block_index_to_location[ix] = ir_block.location
+            current_block_ixs = []
+    return block_index_to_location
```
obi1kenobi: Since `ir_blocks` does not end in a `MarkLocation` block, doesn't this mean that a number of blocks at the very end of the `ir_blocks` list will not be associated with a location?

Consider writing an assertion that `block_index_to_location` has exactly as many entries as `ir_blocks`, to be sure.

graphql_compiler/compiler/ir_lowering_sql/metadata.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class SqlMetadata(object):
+    """Metadata wrapper for use during compilation.
+
+    In order to transform GraphQL to SQL, there needs to be additional configuration specified
+    for mapping:
+        - GraphQL types -> SQL tables
+        - GraphQL fields -> SQL columns
+        - GraphQL Edges -> SQL JOINs
```
obi1kenobi: I'd probably lowercase `edges` here, for consistency with the previous two lines.

graphql_compiler/compiler/ir_lowering_sql/metadata.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    def __init__(self, dialect, sqlalchemy_metadata):
+        """Initialize a new sql metadata manager."""
```
obi1kenobi: nit: let's properly capitalize `SQL` in text

graphql_compiler/compiler/ir_lowering_sql/metadata.py
```diff
@@ -0,0 +1,41 @@
...
+        self.sqlalchemy_metadata = sqlalchemy_metadata
+        self._db_backend = SqlBackend(dialect)
+        self.table_name_to_table = {
+            name.lower(): table for name, table in self.sqlalchemy_metadata.tables.items()
```
obi1kenobi: `for` and onward on the next line please, this is pretty long. It might also be a good idea to use `six.iteritems()` instead of `.items()` here.

graphql_compiler/compiler/ir_lowering_sql/metadata.py
```diff
@@ -0,0 +1,41 @@
...
+        return self.table_name_to_table[table_name]
+
+    def has_table(self, schema_type):
+        """Retrieve a SQLAlchemy table based on the supplied GraphQL schema type name."""
+        table_name = schema_type.lower()
+        return table_name in self.table_name_to_table
+
+    @property
+    def db_backend(self):
+        """Retrieve this compilers DB backend."""
```
obi1kenobi: `compilers` -> `compiler's`

graphql_compiler/compiler/sql_context_helpers.py
```diff
@@ -0,0 +1,57 @@
...
+def get_column(column_name, node, context):
+    """Get a column by name from the selectable.
+
+    Args:
+        selectable: Selectable, selectable to attempt to retrieve the named column from.
+        column_name: str, name of the column to retrieve.
+
+    Returns:
+        column, the SQLAlchemy column if found. Raises an AssertionError otherwise.
+
```
obi1kenobi: remove blank lines from ends of docstrings

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1,7 +1,115 @@
 # Copyright 2018-present Kensho Technologies, LLC.
-"""Transform a tree representation of an SQL query into an executable SQLAlchemy query."""
+"""Transform a SqlNode tree into an executable SQLAlchemy query."""
+from collections import namedtuple
+
+from sqlalchemy import select
+
+from . import sql_context_helpers as context_helpers
```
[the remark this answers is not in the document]
jmeulemans: Sounds good, I don't have a strong argument for doing it.

graphql_compiler/compiler/emit_sql.py
jmeulemans: Yeah the base class is Column, but imported and used as column.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
jmeulemans: This assumption is handled in the validation pass now.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
jmeulemans: I didn't realize that, when does this occur?

graphql_compiler/compiler/ir_lowering_sql/__init__.py
jmeulemans: This is a good point. The assertion I arrived at was that any blocks after global operations start do not require location info for the SQL backend. To get there, I skip everything after global operations and make sure there was nothing left to be associated with a location before global operations. Does that seem reasonable?

jmeulemans:
I think I addressed most of your comments @obi1kenobi. Testing is now in place, and being run alongside match/gremlin.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
obi1kenobi: I wouldn't `pop()` here. The compiler generally tries to avoid mutating existing data structures (especially ones provided by the caller) since that tends to prevent a few different classes of bugs. Can you just select the block you need instead?

Also, I wouldn't put that assertion in the validation pass. When looking at this code, it's not obvious that such an assertion exists (so this code still raises red flags for me), and when looking at the validation pass, it's not clear what the assertion is meant to guard.

Consider making a helper function that in one go asserts what you need and then returns the data you need.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -32,4 +40,187 @@ def lower_ir(ir_blocks, query_metadata_table, type_equivalence_hints=None):
...
+def _validate_all_blocks_supported(ir_blocks, query_metadata_table):
+    """Validate that all IR blocks and ConstructResult fields passed to the are supported.
+
+    Args:
+        ir_blocks: List[BasicBlock], IR blocks to validate.
+        query_metadata_table: QueryMetadataTable, object containing all metadata collected during
+                              query processing, including location metadata (e.g. which locations
+                              are folded or optional).
+
+    Returns:
```
obi1kenobi: Omit the `Returns` section if the function doesn't return anything.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -32,4 +40,187 @@ def lower_ir(ir_blocks, query_metadata_table, type_equivalence_hints=None):
...
+
+    Raises:
+        NotImplementedError, if any block or ConstructResult field is unsupported.
+    """
+    if len(ir_blocks) < 3:
+        raise AssertionError(
+            u'Unexpectedly attempting to validate IR blocks with fewer than 3 blocks. A minimal '
+            u'query is expected to have at least a QueryRoot, GlobalOperationsStart, and '
+            u'ConstructResult block. The query metadata table is {}.'.format(query_metadata_table))
+    construct_result = ir_blocks[-1]
```
obi1kenobi: I wouldn't name this variable `construct_result` until you've checked its type. Otherwise, a future edit to this file may add some functionality between lines 95 and 105 that assumes its type is `ConstructResult`, which will break before the assertion catches it and therefore be difficult to debug.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
obi1kenobi: Locations inside a `@fold` are of type `FoldScopeLocation` which does not have a `query_path`. In principle, there may be more examples in the future.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -32,4 +40,187 @@ def lower_ir(ir_blocks, query_metadata_table, type_equivalence_hints=None):
...
+    # MarkLocation blocks occur after the blocks related to that location.
+    # The core approach here is to buffer blocks until their MarkLocation is encountered
+    # after which all buffered blocks can be associated with the encountered MarkLocation.location.
+    current_block_ixs = []
+    global_operations_started = False
+    for num, ir_block in enumerate(ir_blocks):
+        if global_operations_started:
+            continue
+        if isinstance(ir_block, blocks.GlobalOperationsStart):
+            global_operations_started = True
```
obi1kenobi: Why set the flag here, and then `continue` on the next iteration? This approach will add the `GlobalOperationsStart` block to `current_block_ixs` before breaking out of the loop, which doesn't seem like a good effect. Why not just a `break` at the bottom of the conditional that starts on line 214?

graphql_compiler/query_formatting/sql_formatting.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2018-present Kensho Technologies, LLC.
+######
+# Public API
+######
```
obi1kenobi: I'd put this below the imports.

graphql_compiler/tests/test_compiler.py
```diff
@@ -50,6 +24,61 @@ def setUp(self):
         self.maxDiff = None
         self.schema = get_schema()
 
+        _, sqlalchemy_metadata = get_animal_schema_sql_metadata()
+        self.sql_metadata = SqlMetadata(sqlite.dialect.name, sqlalchemy_metadata)
+
+    def check_test_data(self, test_case, test_data, expected_match, expected_gremlin, expected_sql):
```
obi1kenobi: This should either be a method on `CompilerTests`, or take `test_case` (i.e. the `self` at call sites) as the first argument, but not both.

This is also quite a large and unrelated change to this diff, which makes it harder to review this PR since the scope of the PR just broadened to include this, which means more comments and more round trips etc.

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -79,25 +80,37 @@ def tear_down_integration_test_backends(sql_test_backends):
...
-        Column('animal_id', Integer, primary_key=True),
-        Column('name', String(length=12), nullable=False),
-    )
+    table_name_to_table, metadata = get_animal_schema_sql_metadata()
     animal_rows = (
-        (1, 'Animal 1'),
-        (2, 'Animal 2'),
-        (3, 'Animal 3'),
-        (4, 'Animal 4'),
+        (1, 'Animal 1', 100, datetime.datetime(1900, 1, 1)),
```
obi1kenobi: add some times?

graphql_compiler/tests/test_compiler.py
jmeulemans: Reverted for now.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
jmeulemans: Good point, changed to break.

graphql_compiler/tests/test_data_tools/data_tool.py
jmeulemans: Reverted.

jmeulemans:
> Let's please not add more features or make other major changes to this diff until it can get merged.

Sorry for the confusion, I reverted the unit-test related stuff. That's an easy one for a separate PR.

PR #166, opened 2019-01-23
Adding support for Filter blocks to SQL backend

jmeulemans:


graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -20,13 +20,47 @@
...
+    u'||': Operator(u'or_', Cardinality.BINARY),
+    u'=': Operator(u'__eq__', Cardinality.UNARY),
+    u'<': Operator(u'__lt__', Cardinality.UNARY),
+    u'>': Operator(u'__gt__', Cardinality.UNARY),
+    u'<=': Operator(u'__le__', Cardinality.UNARY),
+    u'>=': Operator(u'__ge__', Cardinality.UNARY),
+    u'has_substring': Operator(u'contains', Cardinality.UNARY),
+}
+
+UNSUPPORTED_OPERATOR_NAMES = {
```
obi1kenobi: Any particular reason to specify this set in addition to `SUPPORTED_OPERATORS`? Seems kind of error-prone -- we'll always need to remember to add every new filtering operator to one of these constants here, or we'll get a random `KeyError` on line 205 of `emit_sql.py`. I'd suggest defining `SUPPORTED_OPERATORS` and then if the operator in the `BinaryComposition` isn't there then saying it's unsupported.

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -20,13 +20,47 @@
...
 
 
+class Operator(object):
+    def __init__(self, name, cardinality):
+        """Represent an operator and it's underlying method."""
+        self.name = name
+        self.cardinality = cardinality
+
+
+class Cardinality(object):
```
obi1kenobi: Any particular reason these should be attached to a class object, instead of being constants like `CARDINALITY_UNARY = 1`, `CARDINALITY_BINARY = 2` etc.?

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -20,13 +20,47 @@
...
 )
 
 SUPPORTED_OUTPUT_EXPRESSION_TYPES = (
     expressions.OutputContextField,
 )
 
 
+class Operator(object):
+    def __init__(self, name, cardinality):
+        """Represent an operator and it's underlying method."""
```
obi1kenobi: nit: it's -> its

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -20,13 +20,47 @@
...
     blocks.QueryRoot,
+    blocks.Filter,
 )
 
 SUPPORTED_OUTPUT_EXPRESSION_TYPES = (
     expressions.OutputContextField,
 )
 
 
+class Operator(object):
```
obi1kenobi: How do you feel about making this a `namedtuple`? That way, you get defined fields, immutability, and auto-defined operators like `==`.

graphql_compiler/compiler/ir_lowering_sql/constants.py
```diff
@@ -20,13 +20,47 @@
...
+class Cardinality(object):
+    """Cardinality for SQLAlchemy operators."""
+
+    UNARY = 1
+    BINARY = 2
+    LIST_VALUED = 3
+
+
+SUPPORTED_OPERATORS = {
+    u'contains': Operator(u'in_', Cardinality.LIST_VALUED),
```
obi1kenobi: Are the specific names (`Operator.name` values) here important? Are they used to look up things in SQLAlchemy? A comment might be helpful to explain why the existing compiler operator names are being remapped to new names.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -223,3 +231,20 @@ def _map_block_index_to_location(ir_blocks):
...
+
+
+def lower_unary_transformations(ir_blocks):
+    """Raise exception if any unary transformation block encountered."""
+    def visitor_fn(expression):
+        """Raise error if current expression is a UnaryTransformation."""
+        if not isinstance(expression, expressions.UnaryTransformation):
+            return expression
+        raise NotImplementedError(
+            u'"has_edge_degree" filter operation is unsupported by SQL backend.'
```
obi1kenobi: I find this a bit misleading. When this error gets hit, we've found a `UnaryTransformation` expression. While it might currently be the case that `has_edge_degree` is the only reason that such an expression might be generated at the moment, I don't think we should assume this will continue to always be the case. I'd make the error message complain about the specific expression that it encountered, and include it in the message.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -64,12 +69,15 @@ def lower_ir(ir_blocks, query_metadata_table, type_equivalence_hints=None):
                         block, tree_root, ir_blocks, query_metadata_table))
             tree_root = SqlNode(block=block, query_path=query_path)
             query_path_to_node[query_path] = tree_root
+        elif isinstance(block, blocks.Filter):
+            query_path_to_filters.setdefault(query_path, []).append((block, query_path))
```
obi1kenobi: I'm curious why append `(block, query_path)` when `query_path` is already the key that corresponds to the list of those values. That means that for a given `query_path` value of `q`, the value entry for `q` looks like `[(b1, q), (b2, q), (b3, q)]`. I don't see the value of including the exact same query path as the second value of every tuple for each key.

graphql_compiler/compiler/sql_context_helpers.py
```diff
@@ -20,6 +20,14 @@ def get_node_selectable(node, context):
     return selectable
 
 
+def get_node_at_path(query_path, context):
+    """Return the SqlNode associated with the query path."""
+    if query_path not in context.query_path_to_node:
+        raise AssertionError(u'Unable to find SqlNode for query path {}'.format(query_path))
```
obi1kenobi: Consider including the value of `context.query_path_to_node` and perhaps the entire `context`, for easier debugging.

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -85,12 +85,13 @@ def generate_sql_integration_data(sql_test_backends):
         metadata,
         Column('animal_id', Integer, primary_key=True),
         Column('name', String(length=12), nullable=False),
+        Column('net_worth', Float, nullable=False),
```
obi1kenobi: This doesn't look quite right. Isn't `net_worth` supposed to be a `Decimal`-typed field? I'm pretty sure most SQL databases support `Decimal` fields, though they might be fixed-precision in which case I'd say use 12 digits before the decimal point, and 2 decimal places.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -112,3 +125,154 @@ def _get_output_columns(node, context):
...
+
+    Args:
+        expression: expression, Variable compiler expression.
+        node: SqlNode, the SqlNode the expression applies to.
+        context: CompilationContext, global compilation state and metadata.
+
+    Returns:
+        Expression, SQLAlchemy expression.
+    """
+    variable_name = expression.variable_name
```
obi1kenobi: `expression` is meant to be of type `Variable`, no? If so, then line 260 should be an assertion -- variable names are *required* to start with `$`.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -112,3 +125,154 @@ def _get_output_columns(node, context):
...
+    """
+    if expression.operator in constants.UNSUPPORTED_OPERATOR_NAMES:
+        raise NotImplementedError(
+            u'Filter operation "{}" is not supported by the SQL backend.'.format(
+                expression.operator))
+    sql_operator = constants.SUPPORTED_OPERATORS[expression.operator]
+    left = _expression_to_sql(expression.left, node, context)
+    right = _expression_to_sql(expression.right, node, context)
+    if sql_operator.cardinality == constants.Cardinality.UNARY:
+        # ensure the operator is grabbed from the Column object
```
obi1kenobi: Seems like there should be an assertion here? The assumption appears to be that after this block, `left` is an instance of `Column`.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -112,3 +125,154 @@ def _get_output_columns(node, context):
...
+        return clause
+    elif sql_operator.cardinality == constants.Cardinality.BINARY:
+        clause = getattr(sql_expressions, sql_operator.name)(left, right)
+        return clause
+    elif sql_operator.cardinality == constants.Cardinality.LIST_VALUED:
+        if isinstance(left, BindParameter) and isinstance(right, Column):
+            left, right = right, left
+        if not isinstance(right, BindParameter):
+            raise AssertionError(
+                u'List valued operator expects column as left side of expression')
```
obi1kenobi: Always include the value that caused the error in the error message. Imagine debugging this problem -- the error message tells you that something went wrong, but you don't know precisely in what way it went wrong since the exact value of `right` isn't returned. So the first step afterward is to change the error message to give you more info so that you can actually start looking into it. Let's just skip that first step and make the error message include the data we'd need to start diagnosing the problem.

How about `List valued operator {} expects BindParameter as the right side of the expression, but instead got {}` or something to that effect?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -112,3 +125,154 @@ def _get_output_columns(node, context):
...
+        if not isinstance(left, Column):
+            raise AssertionError(
+                u'List valued operator expects bind parameter as right side of expression')
```
obi1kenobi: I think the error messages and checks are backwards relative to each other. Also, let's refer to types by their actual names: `BindParameter`, `Column`, etc.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -112,3 +125,154 @@ def _get_output_columns(node, context):
...
+        # ensure that SQLAlchemy treats the left bind parameter as list valued
```
obi1kenobi: The `right` is a BindParameter though, not the left.

graphql_compiler/compiler/sql_context_helpers.py
jmeulemans: Good suggestion, I added context dumping to all these errors, it is helpful.

graphql_compiler/tests/test_data_tools/data_tool.py
jmeulemans: TIL, there's a `Numeric` that takes care of what you're describing.

graphql_compiler/compiler/ir_lowering_sql/constants.py
jmeulemans: Nope, thanks fixed that.

graphql_compiler/compiler/ir_lowering_sql/constants.py
jmeulemans: No they don't have to be, I can use that style.

graphql_compiler/compiler/ir_lowering_sql/constants.py
jmeulemans: Done.

graphql_compiler/compiler/emit_sql.py
jmeulemans: Good to know, this is an assertion now.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
jmeulemans: This was important when it came time to join other tables. In that case what I had been doing was collapsing filters onto the root node, in which case each filter had to know the query path it applied to. That's a little convoluted so thinking about it now, when we do the joins, we'll keep track of the visited query paths, and then when it comes time to do filters we'll collect the filters at all visited query paths. That should be a little cleaner and easier to understand, and then things aren't moving around as much.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
jmeulemans: Fixed, the relevant expression is now output in the error

graphql_compiler/compiler/emit_sql.py
jmeulemans: This should be fixed now, the check is the same for UNARY and LIST_VALUED things, so they both call the same method with a more detailed error now.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -88,27 +100,187 @@ def _create_query(node, context):
...
+    elif sql_operator.cardinality == constants.CARDINALITY_BINARY:
+        clause = getattr(sql_expressions, sql_operator.name)(left, right)
+        return clause
+    elif sql_operator.cardinality == constants.CARDINALITY_LIST_VALUED:
+        left, right = _get_column_and_bindparam(left, right, sql_operator)
+        # ensure that SQLAlchemy treats the right bind parameter as list valued
+        right.expanding = True
+        clause = getattr(left, sql_operator.name)(right)
+        return clause
+    else:
```
obi1kenobi: nit: At one point there was a subtle bug in the compiler where an if-elif tree like this one, where each branch was meant to return, actually had a code path that didn't return. With the `raise AssertionError` in the `else` branch, that meant that no error was raised (since the `else` wasn't taken, one of other the branches was taken but didn't return), and instead the function simply returned `None` to the caller ... which promptly exploded elsewhere.

So if all branches are meant to `return`, I'd pull the `AssertionError` out of the else, and make it say `Unreachable state reached` (+ as much contextual info as possible) instead of necessarily blaming the cardinality.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -88,27 +100,187 @@ def _create_query(node, context):
...
+    if not isinstance(left, Column):
+        left, right = right, left
+    if not isinstance(left, Column):
+        raise AssertionError(
+            u'SQLAlchemy operator {} expects Column as left side the of expression, got {} '
+            u'of type {}'.format(operator, left, type(left)))
+    if not isinstance(right, BindParameter):
+        raise AssertionError(
+            u'SQLAlchemy operator {} expects BindParameter as the right side of the expression, got'
+            u' {} of type {}'.format(operator, right, type(right)))
```
obi1kenobi: nit: Can you push the whitespace to the previous line? It kind of looks awkward to start a string literal with a space.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -223,3 +231,20 @@ def _map_block_index_to_location(ir_blocks):
...
+
+
+def lower_unary_transformations(ir_blocks):
+    """Raise exception if any unary transformation block encountered."""
+    def visitor_fn(expression):
+        """Raise error if current expression is a UnaryTransformation."""
+        if not isinstance(expression, expressions.UnaryTransformation):
+            return expression
+        raise NotImplementedError(
+            u'UnaryTransformation expression "{}" is unsupported by SQL backend.'.format(expression)
```
obi1kenobi: nit: if you include the `ir_blocks` in the message here, it will be easier to figure out where the `UnaryTransformation` emerged from, and why

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -25,9 +27,24 @@ def sort_key(result):
     return sorted(results, key=sort_key)
 
 
+def try_convert_decimal_to_string(value):
+    """Return Decimals as string if value is a Decimal, return value otherwise."""
+    if isinstance(value, list):
+        return [try_convert_decimal_to_string(subvalue) for subvalue in value]
+    if not isinstance(value, Decimal):
+        return value
+    return str(value)
```
obi1kenobi: nit: I think this might be a bit easier to read if it were inverted:
```
if isinstance(value, Decimal):
    return str(value)

return value
```

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -25,9 +27,24 @@ def sort_key(result):
...
+
+
 def compile_and_run_match_query(schema, graphql_query, parameters, graph_client):
     """Compiles and runs a MATCH query against the supplied graph client."""
-    compilation_result = graphql_to_match(schema, graphql_query, parameters)
+    # OrientDB expects decimals to be passed as their string representation
```
obi1kenobi: nit: This is slightly misleading. The MATCH code the compiler emits expects Decimals to come in as strings, since they are wrapped in a statement that causes OrientDB to parse them into Decimal types before using them.

graphql_compiler/compiler/emit_sql.py
jmeulemans: That makes sense.  I remember that error, it is very hard to track down.

graphql_compiler/compiler/emit_sql.py
jmeulemans: Sure, yeah it does look a little weird.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
jmeulemans: Corrected, thanks for the explanation.

PR #176, opened 2019-01-31
Rename __count meta field to _x_count and release v1.10.0.

obi1kenobi:


nmutalik:
maybe worth linking to this? https://facebook.github.io/graphql/draft/#sec-Reserved-Names

obi1kenobi:
> maybe worth linking to this? https://facebook.github.io/graphql/draft/#sec-Reserved-Names

Added:
```markdown
We do not use the GraphQL standard double-underscore (`__`) prefix for meta fields,
since all names with that prefix are
[explicitly reserved and prohibited from being used](https://facebook.github.io/graphql/draft/#sec-Reserved-Names)
in directives, fields, or any other artifacts.
```

Issue #2, opened 2017-07-19
Using @tag and then @filter with that tag in the same scope might not work correctly


obi1kenobi:
Also unlikely to work, and should throw good errors: if @tag appears on the exact same property where the @filter is applied.

benjamincohen1:
+1 encountered this

PR #178, opened 2019-02-01
Fix bug with colocated @tag and @filter

bojanserafimov:
Fixes https://github.com/kensho-technologies/graphql-compiler/issues/2

### DIFF
--- a/graphql_compiler/compiler/filters.py
+++ b/graphql_compiler/compiler/filters.py
@@ -93,7 +93,7 @@ def _represent_argument(directive_location, context, argument, inferred_type):
     """Return a two-element tuple that represents the argument to the directive being processed.
 
     Args:
-        directive_location: Location where this directive is used.
+        directive_location: Location where the directive is used.
         context: dict, various per-compilation data (e.g. declared tags, whether the current block
                  is optional, etc.). May be mutated in-place in this function!
         argument: string, the name of the argument to the directive
@@ -142,12 +142,12 @@ def _represent_argument(directive_location, context, argument, inferred_type):
                                           u'not match the inferred required type for this filter: '
                                           u'{} vs {}'.format(tag_inferred_type, inferred_type))
 
-        # See if the argument is colocated with the directive
-        colocated = directive_location.at_vertex() == location.at_vertex()
+        # Check whether the argument is a field on the vertex on which the directive is applied.
+        field_is_local = directive_location.at_vertex() == location.at_vertex()
 
         non_existence_expression = None
         if optional:
-            if colocated:
+            if field_is_local:
                 non_existence_expression = expressions.FalseLiteral
             else:
                 non_existence_expression = expressions.BinaryComposition(
@@ -155,7 +155,7 @@ def _represent_argument(directive_location, context, argument, inferred_type):
                     expressions.ContextFieldExistence(location.at_vertex()),
                     expressions.FalseLiteral)
 
-        if colocated:
+        if field_is_local:
             representation = expressions.LocalField(argument_name)
         else:
             representation = expressions.ContextField(location)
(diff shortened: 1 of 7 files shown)


graphql_compiler/compiler/filters.py
```diff
@@ -650,4 +662,5 @@ def process_filter_directive(filter_operation_info, location, context):
         FilterInfo(fields=fields, op_name=op_name, args=tuple(operator_params))
     )
 
+    context['directive_location'] = location
```
bojanserafimov: Hacky. I'm not sure if this is the intended use of the context argument.

graphql_compiler/compiler/helpers.py
```diff
@@ -336,6 +336,10 @@ def __init__(self, query_path, field=None, visit_counter=1):
         # visit 'Y' in two different ways to generate colliding 'X__Y___1' identifiers.
         self.visit_counter = visit_counter
 
+    def remove_field(self):
```
obi1kenobi: See `at_vertex()`.

graphql_compiler/compiler/helpers.py
```diff
@@ -498,6 +502,10 @@ def get_first_folded_edge(self):
         first_folded_edge_direction, first_folded_edge_name = self.fold_path[0]
         return first_folded_edge_direction, first_folded_edge_name
 
+    def remove_field(self):
```
obi1kenobi: Make this `at_vertex()` similar to the one above.

graphql_compiler/compiler/filters.py
obi1kenobi: Why not just weave this into the filter operator functions, and then into the `_represent_argument`? This seems really hacky, and unnecessarily so.

graphql_compiler/compiler/helpers.py
```diff
@@ -498,6 +498,13 @@ def get_first_folded_edge(self):
         first_folded_edge_direction, first_folded_edge_name = self.fold_path[0]
         return first_folded_edge_direction, first_folded_edge_name
 
+    def at_vertex(self):
+        """Get the Location ignoring its field component."""
```
obi1kenobi: `Location != FoldScopeLocation` here. They are both `BaseLocation` objects, but this does not return a `Location`.

graphql_compiler/compiler/filters.py
```diff
@@ -89,10 +89,11 @@ def _is_tag_argument(argument_name):
     return argument_name.startswith('%')
 
 
-def _represent_argument(context, argument, inferred_type):
+def _represent_argument(directive_location, context, argument, inferred_type):
     """Return a two-element tuple that represents the argument to the directive being processed.
 
     Args:
+        directive_location: Location where this directive is used.
```
obi1kenobi: super nit: `this` in `this directive` throws me off a bit here. Contextually, the reader knows we're in the middle of processing a directive (in the caller function), but the directive is sort of irrelevant and not passed into this function, so referring to it as `this` seems a bit strange. Maybe use `the` instead?

graphql_compiler/compiler/filters.py
```diff
@@ -141,14 +142,24 @@ def _represent_argument(context, argument, inferred_type):
                                           u'not match the inferred required type for this filter: '
                                           u'{} vs {}'.format(tag_inferred_type, inferred_type))
 
+        # See if the argument is colocated with the directive
```
obi1kenobi: nit: Consider expanding on this comment a bit. What does `colocated` mean here, precisely? The reader of the code might be surprised to learn that `colocated` means `part of the same enclosing vertex field` (i.e. `directive_location.at_vertex() == location.at_vertex()`), rather than `on the same property field` (i.e. `directive_location == location`).

PR #179, opened 2019-02-04
Lint script

bojanserafimov:
scripts/lint.sh does all the standard linting, with an option --diff to only run on changed committed files.


scripts/lint.sh
```diff
@@ -0,0 +1,87 @@
...
+            diff_only=1
+            shift;;
+
+        *)
+            echo "Unknown option: $i";
+            exit 1;;
+    esac
+done
+
+cd "$(git rev-parse --show-toplevel)"
```
obi1kenobi: Do the `git -C` and `dirname` trick here that will allow running this script from any directory. See the other scripts in the `scripts` directory.

Issue #180, opened 2019-02-04
Allow "virtual" edges to be defined, and expanded using an AST-based macro system


obi1kenobi:
While normalized data representations are great for data quality and cleanliness, they often get in the way of ease of use, data discoverability, and navigation through the database.

Using the `Animals` schema in the compiler's tests as an example, it would be much easier to find a given animal's grandparents if an `out_Animal_Grandparent` edge existed. However, this edge is simply a two-fold traversal of the existing `in_Animal_ParentOf` edge; adding a `out_Animal_Grandparent` edge would denormalize the schema and would cause difficulties in maintaining the data.

Instead of adding such an edge to the database, we could define a macro that would be expanded by the GraphQL compiler before query compilation. That way, users can submit a query that relies on the `out_Animal_Grandparent` edge, and the compiler can use the macro system to rewrite that query into an equivalent query that relies only on existing schema elements.

obi1kenobi:
Closed by #356.

PR #181, opened 2019-02-04
Add a skeleton for the macro expansion system.

obi1kenobi:
No implementation yet, just a skeleton that defines some basic data structures and the overall package structure.


obi1kenobi:
Related to #180.

obi1kenobi:
@jmeulemans let me know if this makes sense. I'll start filling this in tomorrow morning, assuming no objections.

PR #183, opened 2019-02-05
Add a few helpers for macro validation.

obi1kenobi:
Minor refactor of AST manipulation functions. Add a new error type that is used when an error type is invalid.

### DIFF
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -1,9 +1,7 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from copy import copy
 
-from graphql.language.ast import (
-    Argument, Directive, Field, InlineFragment, Name, OperationDefinition, SelectionSet, StringValue
-)
+from graphql.language.ast import Field, InlineFragment, OperationDefinition, SelectionSet
 
 from ...ast_manipulation import get_human_friendly_ast_field_name
 from ...exceptions import GraphQLInvalidMacroError
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -1,5 +1,9 @@
 # Copyright 2019-present Kensho Technologies, LLC.
-from graphql.language.ast import Field, InlineFragment, OperationDefinition
+from copy import copy
+
+from graphql.language.ast import (
+    Argument, Directive, Field, InlineFragment, Name, OperationDefinition, SelectionSet, StringValue
+)
 
 from ...ast_manipulation import get_human_friendly_ast_field_name
 from ...exceptions import GraphQLInvalidMacroError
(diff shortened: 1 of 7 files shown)


graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -0,0 +1,122 @@
...
+
+def get_only_selection_from_ast(ast):
+    """Return the selected sub-ast, ensuring that there is precisely one."""
+    if ast.selection_set is None:
+        ast_name = get_human_friendly_ast_field_name(ast)
+        raise GraphQLInvalidMacroError(u'Expected an AST with exactly one selection, but got one '
+                                       u'with no selections. Error near AST node named: {}'
+                                       .format(ast_name))
+
+    selections = ast.selection_set.selections
```
bojanserafimov: `selections = [] if ast.selection_set is None else ast.selection_set.selections` looks more readable to me. The above lines will be unnecessary.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -0,0 +1,122 @@
...
+        ]
+        raise GraphQLInvalidMacroError(u'Expected an AST with exactly one selection, but found '
+                                       u'{} selections at AST node named {}: {}'
+                                       .format(len(selection_names), selection_names, ast_name))
+
+    return selections[0]
+
+
+def remove_directives_from_ast(ast, directive_names_to_omit):
+    """Return a copy of the AST but with instances of the named directives omitted.
```
bojanserafimov: This confusing, because it doesn't always return a copy. Is there a reason you make an effort to return the same object when no changes are necessary?

graphql_compiler/macros/macro_edge/helpers.py
obi1kenobi: In general, I'm not a huge fan of ternaries, but I can try to refactor and deduplicate it a bit. Just keep in mind that raising a user-friendly error here is the top priority, since the user doesn't know or care about the internals of the compiler.

graphql_compiler/macros/macro_edge/helpers.py
obi1kenobi: Runtime and memory concerns, primarily. We don't want to copy for no reason, especially because these AST objects tend to have a lot of children, whose children have a lot of children etc. We'd just be generating garbage for the GC to collect, even though we can easily avoid it.

Most ASTs don't need to be mutated, so we don't want to pay the copying cost if we don't have to. This approach buys us immutability while still being friendly to the GC.

bojanserafimov: I can see myself relying on this function to make a copy even if there's no change. How about `Remove instances of the named directives, making a copy if anything was removed`? If it doesn't fit in a line, maybe even say `Remove instances of the named directives`, and explain the copy behavior in detail later.

obi1kenobi: Why would you use a function like this to make a copy though? That would seem like "code smell" to me, since you could have just done a `copy.deepcopy(ast)` which would have been much more explicit.

As for explaining the copy behavior, is the part in the `Returns:` section not sufficient? Where do you think I should place it?

PR #185, opened 2019-02-07
Subclasses

bojanserafimov:
Implementation for subclasses, to be used in merging nested type coercions at the `@macro_edge_target`

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -91,6 +91,7 @@ from .helpers import (
     validate_safe_string
 )
 from .metadata import LocationInfo, QueryMetadataTable, RecurseInfo
+from .subclass import compute_subclass_sets
 
 
 # LocationStackEntry contains the following:
@@ -789,6 +790,8 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
         # 'type_equivalence_hints_inverse' is the inverse of type_equivalence_hints,
         # which is always invertible.
         'type_equivalence_hints_inverse': invert_dict(type_equivalence_hints),
+        # 'subclass_sets' is a dict mapping class names to the set of its subclass names
+        'subclass_sets': compute_subclass_sets(schema, type_equivalence_hints),
     }
 
     # Add the query root basic block to the output.
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -791,7 +791,8 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
         # which is always invertible.
         'type_equivalence_hints_inverse': invert_dict(type_equivalence_hints),
         # 'subclass_sets' is a dict mapping class names to the set of its subclass names
-        'subclass_sets': compute_subclass_sets(schema, type_equivalence_hints),
+        'subclass_sets': compute_subclass_sets(
+            schema, type_equivalence_hints=type_equivalence_hints),
     }
 
     # Add the query root basic block to the output.
--- a/graphql_compiler/compiler/subclass.py
+++ b/graphql_compiler/compiler/subclass.py
@@ -3,7 +3,7 @@ from graphql.type.definition import GraphQLObjectType, GraphQLUnionType
 import six
 
 
-def _transitive_closure(graph):
+def _add_transitive_closure(graph):
     """Compute the transitive closure of a graph represented as dict inplace."""
     # Floyd-Warshall O(N^3)
     for k, k_out in six.iteritems(graph):
@@ -11,10 +11,9 @@ def _transitive_closure(graph):
             for j, _ in six.iteritems(graph):
                 if k in i_out and j in k_out:
                     graph.setdefault(i, set()).add(j)
-    return graph
 
 
-def compute_subclass_sets(schema, type_equivalence_hints):
+def compute_subclass_sets(schema, type_equivalence_hints=None):
     """Return a dict mapping class names to the set of its subclass names.
 
     B is a subclass of A if any of the following conditions hold:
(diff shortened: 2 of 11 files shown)


graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2017-present Kensho Technologies, LLC.
+from graphql.type.definition import GraphQLObjectType, GraphQLUnionType
+import six
+
+
+def _transitive_closure(graph):
+    """Compute the transitive closure of a graph represented as dict inplace."""
```
obi1kenobi: For functions that mutate input, don't return anything. That will make it obvious to the user that the input has to have gotten mutated.

Also, this sentence structure feels a bit off, especially around the `a graph represented as dict inplace` part. Maybe move the `in-place` bit to earlier in the sentence.

graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,46 @@
...
+                if k in i_out and j in k_out:
+                    graph.setdefault(i, set()).add(j)
+    return graph
+
+
+def compute_subclass_sets(schema, type_equivalence_hints):
+    """Return a dict mapping class names to the set of its subclass names."""
+    subclass_set = dict()
+
+    for classname, cls in six.iteritems(schema.get_type_map()):
```
obi1kenobi: Don't use `cls`, its canonical use is for Python class objects and you can see that the syntax highlighting picks it up.

graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _transitive_closure(graph):
+    """Compute the transitive closure of a graph represented as dict inplace."""
+    # Floyd-Warshall O(N^3)
+    for k, k_out in six.iteritems(graph):
+        for i, i_out in six.iteritems(graph):
+            for j, j_out in six.iteritems(graph):
+                if k in i_out and j in k_out:
+                    graph.setdefault(i, set()).add(j)
```
obi1kenobi: Because of the "class is a subclass of itself" rule, I think the `setdefault` here is unnecessary.

graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,46 @@
...
+    return graph
+
+
+def compute_subclass_sets(schema, type_equivalence_hints):
+    """Return a dict mapping class names to the set of its subclass names."""
+    subclass_set = dict()
+
+    for classname, cls in six.iteritems(schema.get_type_map()):
+        # A class is a subclass of itself
+        subclass_set.setdefault(classname, set()).add(classname)
```
obi1kenobi: This might be premature optimization, but you could define the `subclass_set` using a dict comprehension as:
```
subclass_set = {
    classname: {classname}
    for classname in six.iterkeys(schema.get_type_map())
}
```
That would avoid all of the `setdefault` calls, since you are then guaranteed that every type in the type map has a key.

`setdefault` isn't that expensive on its own, but you are calling the `set()` constructor on every call, which is unnecessary work and causes memory pressure.

graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,46 @@
...
+    # Floyd-Warshall O(N^3)
+    for k, k_out in six.iteritems(graph):
+        for i, i_out in six.iteritems(graph):
+            for j, j_out in six.iteritems(graph):
+                if k in i_out and j in k_out:
+                    graph.setdefault(i, set()).add(j)
+    return graph
+
+
+def compute_subclass_sets(schema, type_equivalence_hints):
```
obi1kenobi: Should this handle the case where `type_equivalence_hints` are not provided? Presumably you can still compute the subclass sets of all non-unions?

graphql_compiler/tests/test_subclass.py
```diff
@@ -0,0 +1,39 @@
...
+            ('Food', 'FoodOrSpecies', True),
+            ('Species', 'FoodOrSpecies', True),
+            ('Animal', 'FoodOrSpecies', False),
+            ('Entity', 'FoodOrSpecies', False),
+            ('Event', 'EventOrBirthEvent', True),
+            ('BirthEvent', 'Event', True),  # Derived from the type_equivalence_hints
+        ]
+        for cls1, cls2, expected in cases:
+            is_subclass = cls1 in subclass_sets[cls2]
+            if expected != is_subclass:
```
obi1kenobi: why not directly do `self.assertEqual()` here, with a custom message?

bojanserafimov: That doesn't print out which case failed.

obi1kenobi: You can override the message to make it say whatever you want, including which test case failed.

graphql_compiler/compiler/subclass.py
bojanserafimov: But it's not specified in the spec of this function. I'd rather not complicate it.

obi1kenobi: I think it's better to specify it. Otherwise, this line is mutating the keys of a dict that is being iterated over, which is definitely bad practice. With that in the spec, it's only mutating the values, which is somewhat more palatable.

graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # The base of the union is a superclass of other members
+    for graphql_type, equivalent_type in six.iteritems(type_equivalence_hints):
+        if isinstance(equivalent_type, GraphQLUnionType):
+            for subclass in equivalent_type.types:
+                subclass_set[graphql_type.name].add(subclass.name)
+        else:
+            raise AssertionError(u'Unexpected type {}'.format(type(equivalent_type)))
+
+    # NOTE(bojanserafimov): Taking the transitive closure has no effect on the current schema.
```
obi1kenobi: You can probably remove the `NOTE(bojanserafimov)` here, I believe this is a true statement and will continue to hold :)

graphql_compiler/compiler/subclass.py
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    for classname, graphql_type in six.iteritems(schema.get_type_map()):
+        # A class is a subclass of interfaces it implements
+        if isinstance(graphql_type, GraphQLObjectType):
+            for interface in graphql_type.interfaces:
+                subclass_set[interface.name].add(classname)
+
+        # Members of a union subclass it
+        if isinstance(graphql_type, GraphQLUnionType):
```
obi1kenobi: Do we want union types in the subclass data at all? "Subclass of a union" seems a bit like an awkward concept. Would it be problematic to only handle unions through the type equivalence hint for that union (assuming it is provided), and otherwise not include them at all?

bojanserafimov: It makes sense to me. A subclass of X intuitively means something that has all the functionality of X, with the same meaning. With this definition we can say that only coercions to subclasses make sense. Removing the "subclass of a union" seems arbitrary. Talk in person if you disagree?

graphql_compiler/tests/test_subclass.py
```diff
@@ -0,0 +1,33 @@
...
+    def setUp(self):
+        """Initialize the test schema once for all tests."""
+        self.schema = get_schema()
+
+    def test_compute_subclass_sets(self):
+        type_equivalence_hints = {
+            self.schema.get_type('Event'): self.schema.get_type('EventOrBirthEvent'),
+        }
+
+        subclass_sets = compute_subclass_sets(self.schema, type_equivalence_hints)
```
obi1kenobi: pass `type_equivalence_hints` as a kwarg

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -789,6 +790,8 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
         # 'type_equivalence_hints_inverse' is the inverse of type_equivalence_hints,
         # which is always invertible.
         'type_equivalence_hints_inverse': invert_dict(type_equivalence_hints),
+        # 'subclass_sets' is a dict mapping class names to the set of its subclass names
+        'subclass_sets': compute_subclass_sets(schema, type_equivalence_hints),
```
obi1kenobi: `type_equivalence_hints` needs to be a kwarg here as well

PR #187, opened 2019-02-08
Add more validation, together with the new 2-directive macro format.

obi1kenobi:
The next PR will add the remainder of the validation, together with tests.

### DIFF
--- a/graphql_compiler/macros/macro_edge/validation.py
+++ b/graphql_compiler/macros/macro_edge/validation.py
@@ -1,9 +1,100 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from collections import namedtuple
+from copy import copy
+from itertools import chain
+
+from graphql.validation import validate
 
 from ...exceptions import GraphQLInvalidMacroError
-from .directives import MacroEdgeDefinitionDirective, MacroEdgeDirective, MacroEdgeTargetDirective
+from ...schema import VERTEX_FIELD_PREFIXES, is_vertex_field_name
+from ...ast_manipulation import get_ast_field_name, get_human_friendly_ast_field_name
+from .directives import MacroEdgeDefinitionDirective, MacroEdgeTargetDirective
 from .helpers import get_only_selection_from_ast
+
+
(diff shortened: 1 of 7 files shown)


graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -1,11 +1,102 @@
...
+    unexpected_directives = [
+        directive_name
+        for directive_name in directive_names
+        if directive_name != MacroEdgeDefinitionDirective.name
+    ]
+    if unexpected_directives:
+        raise GraphQLInvalidMacroError(
+            u'Found unexpected directives at the top level of the macro definition GraphQL: '
+            u'{}'.format(unexpected_directives))
+
```
bojanserafimov: Are you catching the case where the user puts two `@macro_edge_definition`s?

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -48,61 +139,26 @@ def get_and_validate_macro_edge_info(schema, ast, macro_directives, macro_edge_a
...
-            u'Unexpectedly found variable definitions at the top level of the GraphQL input. '
-            u'This is not supported. Variable definitions: {}'.format(ast.variable_definitions))
-
-    unique_and_parameterless_directives_to_check = (
-        MacroEdgeDirective,
-        MacroEdgeDefinitionDirective,
-        MacroEdgeTargetDirective,
-    )
-    macro_edge_sub_asts = {}
+    _validate_macro_ast_with_macro_directives(schema, ast, macro_directives)
```
bojanserafimov: FIx the docstring, there's no "three parts of a macro edge"

graphql_compiler/macros/macro_edge/validation.py
obi1kenobi: Yes, that should raise near `if len(macro_data) > 1:` since that's the number of occurrences for a given macro edge directive.

graphql_compiler/macros/macro_edge/validation.py
obi1kenobi: There are still three parts, they just aren't ASTs anymore :)

PR #188, opened 2019-02-11
Test invalid edge macros

bojanserafimov:
Add tests for invalid macros. The implementation does return the correct type of error, but the content of the error message is wrong in every case.

### DIFF
--- a/graphql_compiler/tests/test_macro_expansion_errors.py
+++ b/graphql_compiler/tests/test_macro_expansion_errors.py
@@ -1,4 +1,4 @@
-# Copyright 2017-present Kensho Technologies, LLC.
+# Copyright 2019-present Kensho Technologies, LLC.
 import unittest
 
 from ..exceptions import GraphQLInvalidMacroError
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/test_macro_expansion_errors.py
```diff
@@ -0,0 +1,175 @@
+# Copyright 2017-present Kensho Technologies, LLC.
```
obi1kenobi: 2019 please

obi1kenobi:
Why is the error message wrong in every case? Not sure what you mean here.

bojanserafimov:
> Why is the error message wrong in every case? Not sure what you mean here.

It gets into this branch every time:
`    if ast.directives is not None:
        directive_names = [directive.name.value for directive in ast.directives]
        raise GraphQLInvalidMacroError(
            u'Unexpectedly found directives at the top level of the GraphQL input. '
            u'This is not supported. Directives: {}'.format(directive_names))
`

Even if I bypass that, it doesn't recognize the macro directives.

PR #190, opened 2019-02-11
Test macro edge expansion

bojanserafimov:
All tests are skipped. They can be activated once the implementation is done.

### DIFF
--- a/graphql_compiler/macros/macro_edge/validation.py
+++ b/graphql_compiler/macros/macro_edge/validation.py
@@ -3,27 +3,19 @@ from collections import namedtuple
 from copy import copy
 from itertools import chain
 
-from graphql.language.ast import OperationDefinition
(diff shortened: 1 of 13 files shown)


graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -160,7 +160,7 @@ def get_and_validate_macro_edge_info(schema, ast, macro_edge_args,
 
     _validate_class_selection_ast(get_only_selection_from_ast(ast), macro_defn_ast)
     class_name = get_ast_field_name(macro_defn_ast)
-    macro_edge_name = macro_defn_directive.arguments['name'].value
+    macro_edge_name = macro_defn_directive.arguments[0].value.value
```
bojanserafimov: @obi1kenobi Please double check this makes sense

PR #197, opened 2019-02-14
Keep the @macro_edge_target in the macro descriptor

bojanserafimov:
If we remove the `@macro_edge_target` directive we permanently lose information on where the user code should be pasted inside the macro.


graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -177,6 +180,7 @@ def _make_macro_edge_descriptor(macro_definition_ast, macro_edge_args):
     directives_to_remove = {
         directive.name
         for directive in MACRO_EDGE_DIRECTIVES
+        if directive.name != MacroEdgeTargetDirective.name
```
obi1kenobi: @bojanserafimov Please update the docstring too.

PR #198, opened 2019-02-14
Expand specific macro

bojanserafimov:
Implement most of macro expansion merging, with some todos left, and unskip some tests.

### DIFF
--- a/graphql_compiler/macros/macro_expansion.py
+++ b/graphql_compiler/macros/macro_expansion.py
@@ -33,6 +33,8 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
     """Merge selection sets, merging directives on name conflict."""
     if selection_set_a is None:
         return selection_set_b
+    if selection_set_b is None:
+        return selection_set_a
 
     # Convert to dict
     selection_dict_a = get_uniquely_named_objects_by_name(selection_set_a.selections)
@@ -205,7 +207,7 @@ def _expand_macros_in_inner_ast(schema, macro_registry, current_schema_type,
                         macro_edge_descriptor.expansion_selection_set, selection_ast,
                         subclass_sets=subclass_sets)
                     extra_selection_set = _merge_selection_sets(
-                        extra_selection_set, SelectionSet(extra_selections))
+                        SelectionSet(extra_selections), extra_selection_set)
 
                     new_query_args = _merge_non_overlapping_dicts(
                         new_query_args, macro_edge_descriptor.macro_args)
--- a/graphql_compiler/macros/macro_expansion.py
+++ b/graphql_compiler/macros/macro_expansion.py
@@ -10,7 +10,7 @@ from ..ast_manipulation import (
     get_ast_field_name, get_only_query_definition, get_only_selection_from_ast
 )
 from ..compiler.helpers import get_vertex_field_type, get_uniquely_named_objects_by_name
-from ..exceptions import GraphQLInvalidMacroError
+from ..exceptions import GraphQLInvalidMacroError, GraphQLCompilationError
 from ..schema import is_vertex_field_name
 from .macro_edge.helpers import get_directives_for_ast
 from .macro_edge.directives import MacroEdgeTargetDirective
@@ -48,7 +48,7 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
         field_a = selection_dict_a[field_name]
         field_b = selection_dict_b[field_name]
         if field_a.selection_set is not None or field_b.selection_set is not None:
-            raise AssertionError('TODO')
+            raise GraphQLCompilationError('TODO')
 
         merged_field = deepcopy(field_a)
         merged_field.directives += field_b.directives
@@ -82,7 +82,7 @@ def _is_subclass(a, b):
     return True  # TODO implement
 
 
-def _expand_specific_macro_edge(macro_selection_set, selection_ast):
+def _expand_specific_macro_edge(schema, macro_selection_set, selection_ast):
     """Produce a tuple containing the new replacement selection AST, and a list of extra selections.
 
     Args:
@@ -104,15 +104,23 @@ def _expand_specific_macro_edge(macro_selection_set, selection_ast):
             target_ast, _ = directives[MacroEdgeTargetDirective.name][0]
             continuation_selection_set = selection_ast.selection_set
 
+            # Deal with coercions at the target
             first_selection = selection_ast.selection_set.selections[0]
-            if isinstance(first_selection, InlineFragment) and isinstance(target_ast, InlineFragment):
-                target_class = target_ast.type_condition.name.value
+            if isinstance(first_selection, InlineFragment):
                 coercion_class = first_selection.type_condition.name.value
-                if _is_subclass(coercion_class, target_class):
-                    continuation_selection_set = first_selection.selection_set
-                    target_ast.type_condition = first_selection.type_condition
+                if isinstance(target_ast, InlineFragment):
+                    target_class = target_ast.type_condition.name.value
+                    if _is_subclass(coercion_class, target_class):
+                        continuation_selection_set = first_selection.selection_set
+                        target_ast.type_condition = first_selection.type_condition
+                        # TODO(bojanserafimov): Merge directives on fragment
+                    else:
+                        raise AssertionError(u'Cannot coerce to non-subclass.')
                 else:
-                    raise AssertionError(u'Cannot coerce to non-subclass.')
+                    # TODO(bojanserafimov): When compiling the macro, compute the type at the
+                    #                       target and record that in the macro descriptor.
+                    raise NotImplementedError(u'Cannot coerce from macro that does not end'
+                                              u'in a coercion.')
 
             if replacement_selection_ast is not None:
                 raise AssertionError('TODO')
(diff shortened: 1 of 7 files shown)


graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+    # Compute intersection by name
+    common_selection_dict = dict()
+    common_fields = set(selection_dict_a.keys()) & set(selection_dict_b.keys())
+    for field_name in common_fields:
+        field_a = selection_dict_a[field_name]
+        field_b = selection_dict_b[field_name]
+        if field_a.selection_set is not None or field_b.selection_set is not None:
+            raise GraphQLCompilationError('TODO')
+
+        merged_field = deepcopy(field_a)
```
obi1kenobi: I don't think this needs a `deepcopy`, a shallow one should be fine if you explicitly assign a new `directives` list instead of mutating that one in place. A deep copy is likely to be quite expensive due to the number of nested AST objects that need to be constructed.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+    selection_dict_b = get_uniquely_named_objects_by_name(selection_set_b.selections)
+
+    # Compute intersection by name
+    common_selection_dict = dict()
+    common_fields = set(selection_dict_a.keys()) & set(selection_dict_b.keys())
+    for field_name in common_fields:
+        field_a = selection_dict_a[field_name]
+        field_b = selection_dict_b[field_name]
+        if field_a.selection_set is not None or field_b.selection_set is not None:
+            raise GraphQLCompilationError('TODO')
```
obi1kenobi: Please remember to add a message here, it's not obvious why this is an error nor what the error is.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+
+        merged_field = deepcopy(field_a)
+        merged_field.directives += field_b.directives
+        common_selection_dict[field_name] = merged_field
+
+    # Merge dicts
+    merged_selection_dict = copy(selection_dict_a)
+    merged_selection_dict.update(selection_dict_b)
+    merged_selection_dict.update(common_selection_dict)
```
obi1kenobi: This seems to be relying on overwriting common keys in a particular order, but this is not obvious. I would prefer to restructure the code to make it obvious. If that's too hard, I would at least explicitly mention it in a comment.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+
+    # Remove pro-forma fields if allowed
```
obi1kenobi: Please define `pro-forma fields` here.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+    if len(merged_selection_dict) > 1:
+        merged_selection_dict = {
+            name: ast
+            for name, ast in six.iteritems(merged_selection_dict)
+            if ast.selection_set is not None or len(ast.directives) > 0
```
obi1kenobi: It's not obvious to me why `ast.selection_set is not None` allows us to drop a field, nor why we only execute this logic if the length is greater than one.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+        }
+
+    # Get a deterministic ordering of the merged selections
+    selection_name_order = [
```
obi1kenobi: Why instantiate both lists, instead of chaining generators together and doing one `list` call?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+        for name in selection_name_order
+        if name in merged_selection_dict and merged_selection_dict[name].selection_set is not None
+    ])
+
+
+def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
+    """Add the selections, directives, and coercions from the selection_ast to the target_ast."""
+    # Deal with coercions at the target
+    continuation_ast = selection_ast
+    first_selection = selection_ast.selection_set.selections[0]
```
obi1kenobi: How do we know that `selection_set` is not `None` here? Also, the remainder of `selection_ast.selection_set.selections` is apparently never processed, is that okay? It seems suspicious.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+        coercion_class = first_selection.type_condition.name.value
+        if isinstance(target_ast, InlineFragment):
+            target_class = target_ast.type_condition.name.value
+            continuation_ast = first_selection
+            target_ast.type_condition = first_selection.type_condition
+
+            # coercion_class is required to subclass target_class so we can merge the type coercions
+            if coercion_class != target_class:
+                if subclass_sets is None:
+                    raise AssertionError(u'Cannot prove type coercion at macro target is valid. '
```
obi1kenobi: Is this something a user can trigger with an invalid use of a macro, or is it always a configuration error? `AssertionError` is fine for the latter, but not fine if a user can cause it with a bad GraphQL query on a correctly-configured system.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+                                                      u'is expected to subclass {}.'
+                                                      .format(coercion_class, target_class))
+        else:
+            # TODO(bojanserafimov): When compiling the macro, compute the type at the
+            #                       target and record that in the macro descriptor.
+            raise NotImplementedError(u'Cannot coerce from macro that does not end '
+                                      u'in a coercion.')
+
+    # Merge the continuation into the target
+    target_ast.directives += continuation_ast.directives
```
obi1kenobi: We are mutating a function parameter here. Let's not do this, unless there is very good reason to.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
     return result
 
 
-def _expand_specific_macro_edge(macro_definition_selection_set, selection_ast, subclass_sets=None):
+def _merge_selection_sets(selection_set_a, selection_set_b):
+    """Merge selection sets, merging directives on name conflict."""
```
obi1kenobi: Please write a full docstring for this function, it is far from simple.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+        if name in merged_selection_dict and merged_selection_dict[name].selection_set is None
+    ] + [
+        merged_selection_dict[name]
+        for name in selection_name_order
+        if name in merged_selection_dict and merged_selection_dict[name].selection_set is not None
+    ])
+
+
+def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
+    """Add the selections, directives, and coercions from the selection_ast to the target_ast."""
```
obi1kenobi: Please write a full docstring here as well.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
         The first value of the tuple is a replacement
     """
-    raise NotImplementedError()
+    replacement_selection_ast = None
+    extra_selections = []
+
+    # TODO(bojanserafimov): Rename macro tags if conflicting with user-defined tag names.
+    # TODO(bojanserafimov): Remove macro tags if the user has tagged the same field.
+
+    for macro_ast in deepcopy(macro_selection_set).selections:
```
obi1kenobi: Let's avoid this `deepcopy` if at all possible. Copying ASTs can get really expensive since they are highly nested and contain a very large number of objects.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+        directives = get_directives_for_ast(macro_ast)
+        if MacroEdgeTargetDirective.name in directives:
+            # This is not a copy. We intentionally mutate target_ast to make changes to macro_ast.
+            target_ast, _ = directives[MacroEdgeTargetDirective.name][0]
+            _merge_selection_into_target(target_ast, selection_ast, subclass_sets=subclass_sets)
+
+            if replacement_selection_ast is not None:
+                raise AssertionError(u'Only one selection should contain a target.')
```
obi1kenobi: Add more info here please. If we saw this error in production, we wouldn't know anything about what happened.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -28,21 +29,144 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+            replacement_selection_ast = macro_ast
+        else:
+            extra_selections.append(macro_ast)
+
+    if replacement_selection_ast is None:
+        raise AssertionError(u'At least one selection should contain a target.')
```
obi1kenobi: More info here as well please.

graphql_compiler/macros/macro_expansion.py
bojanserafimov: Good catch. I was making the assumption that if there's an inline fragment, it's the first selection and there's no more selections.

obi1kenobi:
That assumption is probably true, but should be backed up by an assertion
and ideally made more obvious in the code structure.

graphql_compiler/macros/macro_expansion.py
bojanserafimov: Removing this one is not simple at all. And even if we removed it, we would need to shallow copy all objects on the path to the `@macro_edge_target`. I'm not convinced that's any faster than a `deepcopy`. Also if we did that, changing the result of `expand_macros_in_query_ast` would change the macro definition in the registry, which is asking for bugs.

PR #199, opened 2019-02-15
Compile macro to ir

bojanserafimov:
Add a validation step that checks if the macro can successfully compile to IR.

### DIFF
--- a/graphql_compiler/macros/macro_edge/validation.py
+++ b/graphql_compiler/macros/macro_edge/validation.py
@@ -3,7 +3,7 @@ from collections import namedtuple
 from copy import copy
 from itertools import chain
 
-from graphql.language.ast import Argument, Document, Directive, Field, Name, StringValue
+from graphql.language.ast import Argument, Directive, Document, Field, Name, StringValue
 from graphql.validation import validate
 
 from ...ast_manipulation import (
@@ -135,6 +135,7 @@ def _get_minimal_query_ast_from_macro_ast(macro_ast):
 # ############
 # Public API #
 # ############
+
 
 MacroEdgeDescriptor = namedtuple(
     'MacroEdgeDescriptor', (
--- a/graphql_compiler/macros/macro_edge/validation.py
+++ b/graphql_compiler/macros/macro_edge/validation.py
@@ -114,6 +114,10 @@ def _get_minimal_query_ast_from_macro_ast(macro_ast):
         Argument(Name('out_name'), StringValue('dummy_output_name'))
     ])
 
+    # Prevent mutations to to macro_ast
+    query_ast = copy(query_ast)
+    query_ast.selection_set = copy(query_ast.selection_set)
+
     # Find or create a __typename field
     field = None
     top_level_selections = query_ast.selection_set.selections[0].selection_set.selections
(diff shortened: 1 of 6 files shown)


graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -99,6 +102,37 @@ def _validate_macro_edge_name_for_class_name(schema, class_name, macro_edge_name
...
+        #                       that we can output. The proper solution is to bypass the rule that
+        #                       a query has an output.
+        field = None
+        top_level_selections = query_ast.selection_set.selections[0].selection_set.selections
+        for selection in top_level_selections:
+            if isinstance(selection, Field):
+                if selection.name.value == 'uuid':
+                    field = selection
+        if field is None:
+            top_level_selections.insert(0, Field(Name('uuid'), directives=[]))
```
obi1kenobi: We seem to be setting `directives=[]` and then below mutating it with `field.directives.append` -- why not construct it fully the first time around?

bojanserafimov: It makes the logic simpler in the case when the field already exists. Will restructure to make it obvious.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -104,33 +104,29 @@ def _validate_macro_edge_name_for_class_name(schema, class_name, macro_edge_name
...
-            field = top_level_selections[0]
-        field.directives.append(output_directive)
+    # Add this output directive to make the ast a valid query
+    output_directive = Directive(Name('output'), arguments=[
+        Argument(Name('out_name'), StringValue('dummy_output_name'))
+    ])
+
+    # Find or create a __typename field
+    field = None
+    top_level_selections = query_ast.selection_set.selections[0].selection_set.selections
```
obi1kenobi: Let's not mutate the `query_ast` or the field objects within. While `remove_directives_from_ast` gives you a new AST object, it relies on structural sharing to minimize the amount of copying it does, so if you mutate underlying object, you can still cause serious issues for other ASTs that rely on the same information.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -99,6 +102,37 @@ def _validate_macro_edge_name_for_class_name(schema, class_name, macro_edge_name
...
+        directive.name
+        for directive in MACRO_EDGE_DIRECTIVES
+    })
+
+    # Add this output directive to make the ast a valid query
+    output_directive = Directive(Name('output'), arguments=[
+        Argument(Name('out_name'), StringValue('dummy_output_name'))
+    ])
+
+    # Prevent mutations to to macro_ast
```
obi1kenobi: nit: `to to`

graphql_compiler/macros/macro_edge/validation.py
obi1kenobi: I am fairly certain that the above copying is not enough to prevent mutating the `query_ast` object. The copies give you a new `query_ast` object, and a new `SelectionSet` object within, which holds a reference to the same `selections` list within, which has the same objects inside etc. So, this `top_level_selections` list that we're mutating here is within one of the entries of that `selections` list, i.e. it's shared with `query_ast`.

PR #200, opened 2019-02-21
Address comments on expand_specific_macro

bojanserafimov:
Link to comments addressed https://github.com/kensho-technologies/graphql-compiler/pull/199#discussion_r258673398

### DIFF
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -100,7 +100,21 @@ def remove_directives_from_ast(ast, directive_names_to_omit):
 
 
 def find_target_and_copy_path_to_it(ast):
-    """Return the same ast and the ast at the target, with the path to the target shallow copied."""
+    """Copy the AST objects on the path to the target, returning it and the target AST.
+
+    This function makes it easy to change the AST at the @target directive without mutating the
+    original object or doing a deepcopy.
+
+    Args:
+        ast: GraphQL library AST object
+
+    Returns:
+        pair containing:
+        - GraphQL library AST object equal to the input. Objects on the path to the @target
+          directive are shallow copied.
+        - GraphQL library AST object at the @target directive of the resulting AST, or None
+          if there was no @target directive in the AST.
+    """
     for directive in ast.directives:
         if directive.name.value == MacroEdgeTargetDirective.name:
             target_ast = copy(ast)
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -3,6 +3,8 @@ from copy import copy
 
 from graphql.language.ast import Field, InlineFragment, OperationDefinition, SelectionSet
 
+from ..macro_edge.directives import MacroEdgeTargetDirective
+
 
 def _yield_ast_nodes_with_directives(ast):
     """Get the AST objects where directives appear, anywhere in the given AST.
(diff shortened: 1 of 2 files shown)


graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -97,6 +99,33 @@ def remove_directives_from_ast(ast, directive_names_to_omit):
     return new_ast
 
 
+def find_target_and_copy_path_to_it(ast):
+    """Return the same ast and the ast at the target, with the path to the target shallow copied."""
```
obi1kenobi: Mind describing in a bit more detail what this function is intended to accomplish? I think I get what it is doing, but not why it's doing it. "Same AST" is also a bit misleading.

Also, a nit: I'd feel better about it with `ast` -> `AST` for consistency.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -30,7 +31,26 @@ def _merge_non_overlapping_dicts(merge_target, new_data):
...
+
+    Create a selection set that contains the selections of both inputs. If there is a name
+    collision on a property field, we take the directives from both inputs on that field and
+    merge them. We disallow name collision on a vertex field.
+
+    The value None represents an empty SelectionSet.
+
+    The order of selections in the resulting SelectionSet has the following properties:
+    - property fields are before vertex fields.
+    - fields present in selection_set_b are after all other fields.
```
obi1kenobi: presumably this only applies to vertex fields from B, right?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -47,75 +67,105 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
...
         if field_a.selection_set is not None or field_b.selection_set is not None:
-            raise GraphQLCompilationError('TODO')
+            raise GraphQLCompilationError(u'Macro expansion results in a query traversing the '
+                                          u'same edge {} twice, which is disallowed.'
+                                          .format(field_name))
 
-        merged_field = deepcopy(field_a)
-        merged_field.directives += field_b.directives
+        merged_field = copy(field_a)
+        merged_field.directives = field_a.directives + field_b.directives
```
obi1kenobi: My personal preference here would be to be explicit about the fact that we expect this to produce a new list copy using `list(chain(...))`, but I can live with it as-is if you prefer this.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -47,75 +67,105 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
...
         merged_selection_dict[name]
         for name in selection_name_order
         if name in merged_selection_dict and merged_selection_dict[name].selection_set is None
-    ] + [
+    ), (
         merged_selection_dict[name]
         for name in selection_name_order
         if name in merged_selection_dict and merged_selection_dict[name].selection_set is not None
-    ])
+    ))))
```
obi1kenobi: This is fairly repetitive. We already have a dependency on `funcy`, maybe `funcy.split` would help?

https://funcy.readthedocs.io/en/stable/seqs.html#lsplit

graphql_compiler/macros/macro_expansion.py
```diff
@@ -47,75 +67,105 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
...
             if coercion_class != target_class:
                 if subclass_sets is None:
-                    raise AssertionError(u'Cannot prove type coercion at macro target is valid. '
-                                         u'Please provide a proof that {} subclasses {} using the '
-                                         u'subclass_sets argument.'
-                                         .format(coercion_class, target_class))
+                    # TODO(bojanserafimov): Write test for this failure
+                    raise GraphQLCompilationError(
+                        u'Cannot prove type coercion at macro target is valid. Please provide a'
+                        u'proof that {} subclasses {} using the subclass_sets argument.'
```
obi1kenobi: Not sure that `proof` is the word we're looking for here... I think the user could conceivably "prove" something that is actually false.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -47,75 +67,105 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
...
+    if selection_ast.selection_set is None or not selection_ast.selection_set.selections:
+        raise AssertionError(u'Precondition violated. selection_ast is expected to be nonempty {}'
+                             .format(selection_ast))
+
+    # See if there's a type coercion in the selection_ast
+    coercion = None
+    for selection in selection_ast.selection_set.selections:
+        if isinstance(selection, InlineFragment):
+            if len(selection_ast.selection_set.selections) != 1:
+                raise GraphQLCompilationError(u'Found selections outside type coercion {}'
```
obi1kenobi: This is a user-facing error, so let's give the user some actionable advice, and let's avoid printing a giant AST selection set in the error message.

graphql_compiler/macros/macro_expansion.py
obi1kenobi: Also, I realize I suggested this approach on the previous iteration -- in retrospect that wasn't a good idea, sorry about that.

graphql_compiler/macros/macro_expansion.py
bojanserafimov: Applies to both vertex fields and property fields, but separately. Think of it as "It applies to everything, but it has lower priority than the first bullet point".

PR #202, opened 2019-02-26
Type coercion at @macro_edge_target

bojanserafimov:
Implement type coercion and unskip tests

### DIFF
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -4,6 +4,7 @@ from copy import copy
 from graphql.language.ast import Field, InlineFragment, OperationDefinition, SelectionSet
 
 from ..macro_edge.directives import MacroEdgeTargetDirective
+from ...compiler.helpers import get_vertex_field_type
 
 
 def _yield_ast_nodes_with_directives(ast):
@@ -115,11 +116,13 @@ def find_target_and_copy_path_to_it(ast):
         - GraphQL library AST object at the @target directive of the resulting AST, or None
           if there was no @target directive in the AST.
     """
+    # Base case
     for directive in ast.directives:
         if directive.name.value == MacroEdgeTargetDirective.name:
             target_ast = copy(ast)
             return target_ast, target_ast
 
+    # Recurse
     new_selections = []
     target_ast = None
     if isinstance(ast, (Field, InlineFragment, OperationDefinition)):
@@ -138,6 +141,34 @@ def find_target_and_copy_path_to_it(ast):
         new_ast = copy(ast)
         new_ast.selection_set = SelectionSet(new_selections)
         return new_ast, target_ast
+
+
+def get_type_at_target(schema, ast, current_type):
+    """Return type at target or none if there is no target."""
+    # Base case
+    for directive in ast.directives:
+        if directive.name.value == MacroEdgeTargetDirective.name:
+            return current_type
+
+    # Recurse
+    if isinstance(ast, (Field, InlineFragment, OperationDefinition)):
+        if ast.selection_set is not None:
+            for selection in ast.selection_set.selections:
+                if isinstance(selection, Field):
+                    type_in_selection = get_vertex_field_type(current_type, selection.name.value)
+                if isinstance(selection, InlineFragment):
+                    type_in_selection = schema.get_type(selection.type_condition.name.value)
+                else:
+                    raise AssertionError(u'Unexpected selection type received: {} {}'
+                                         .format(type(selection), selection))
+                type_at_target = get_type_at_target(schema, selection, type_in_selection)
+                if type_at_target is not None:
+                    return type_at_target
+
+    else:
+        raise AssertionError(u'Unexpected AST type received: {} {}'.format(type(ast), ast))
+
+    return None  # Didn't find target
 
 
 def omit_ast_from_ast_selections(ast, ast_to_omit):
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -178,7 +178,7 @@ def _get_type_at_macro_edge_target_using_current_type(schema, ast, current_type)
 
 
 def get_type_at_macro_edge_target(schema, ast):
-    """Return type at the @macro_edge_target or None if there is no target."""
+    """Return the GraphQL type at the @macro_edge_target or None if there is no target."""
     root_type = get_ast_field_name(ast)
     root_schema_type = get_field_type_from_schema(schema.get_query_type(), root_type)
     return _get_type_at_macro_edge_target_using_current_type(schema, ast, root_schema_type)
--- a/graphql_compiler/macros/macro_edge/helpers.py
+++ b/graphql_compiler/macros/macro_edge/helpers.py
@@ -3,9 +3,9 @@ from copy import copy
 
 from graphql.language.ast import Field, InlineFragment, OperationDefinition, SelectionSet
 
+from ...ast_manipulation import get_ast_field_name
+from ...compiler.helpers import get_field_type_from_schema, get_vertex_field_type
 from ..macro_edge.directives import MacroEdgeTargetDirective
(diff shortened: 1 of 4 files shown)


graphql_compiler/macros/macro_expansion.py
```diff
@@ -125,13 +125,21 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
...
     if selection_ast.selection_set is None or not selection_ast.selection_set.selections:
         raise AssertionError(u'Precondition violated. selection_ast is expected to be nonempty {}'
                              .format(selection_ast))
 
+    # Remove @macro_edge_target directive
+    target_ast.directives = [
+        directive
+        for directive in target_ast.directives
+        if directive.name.value != MacroEdgeTargetDirective.name
+    ]
```
bojanserafimov: moved this from the bottom of the function so I'd be free to change the pointer target_ast

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -140,6 +144,46 @@ def find_target_and_copy_path_to_it(ast):
         return new_ast, target_ast
 
 
+def _get_type_at_target(schema, ast, current_type):
+    """Return type at target or none if there is no target."""
```
obi1kenobi: `none` -> `None` and please clarify what you mean by `target` in the docstring and maybe also in the function name

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -140,6 +144,46 @@ def find_target_and_copy_path_to_it(ast):
...
+                    raise AssertionError(u'Unexpected selection type received: {} {}'
+                                         .format(type(selection), selection))
+
+                if type_in_selection is not None:
+                    type_at_target = _get_type_at_target(schema, selection, type_in_selection)
+                    if type_at_target is not None:
+                        return type_at_target
+
+    else:
+        raise AssertionError(u'Unexpected AST type received: {} {}'.format(type(ast), ast))
```
obi1kenobi: Maybe switch the order here and do a `if not isinstance(...): raise AssertionError()`, for the sake of readability? Then you don't need to indent the meat of the function.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -140,6 +144,46 @@ def find_target_and_copy_path_to_it(ast):
...
+
+    return None  # Didn't find target
+
+
+def get_type_at_target(schema, ast):
```
obi1kenobi: Let's not have two functions that are named exactly the same thing except for a leading underscore. Try to distinguish them somehow, for the sake of maintainability.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -146,48 +154,43 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
...
-                            coercion_class not in subclass_sets[target_class]):
-                        raise GraphQLCompilationError(u'Invalid type coercion at macro target. {} '
-                                                      u'is expected to subclass {}.'
-                                                      .format(coercion_class, target_class))
         else:
-            # TODO(bojanserafimov): Compute the type at the target and use it to validate.
-            raise NotImplementedError(u'Cannot coerce from macro that does not end '
-                                      u'in a coercion.')
+            target_ast.selection_set = SelectionSet([InlineFragment(
+                coercion.type_condition, target_ast.selection_set, directives=[])])
```
obi1kenobi: This is a bit difficult to read. Maybe construct the `InlineFragment` separately and then pass it to the `SelectionSet([...])`?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -146,48 +154,43 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
...
+        # macro selections inside the coercion, and merge the two coercions into one
+        if coercion_class != type_at_target:
+            if subclass_sets is None:
+                # TODO(bojanserafimov): Write test for this failure
+                raise GraphQLCompilationError(
+                    u'Cannot prove type coercion at macro target is valid. Please provide a '
+                    u'hint that {} subclasses {} using the subclass_sets argument.'
+                    .format(coercion_class, type_at_target))
+            else:
+                if (type_at_target not in subclass_sets or
```
obi1kenobi: Should this first condition be an `AssertionError`?

bojanserafimov: We never specified that property of the `subclass_sets`

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -140,6 +144,46 @@ def find_target_and_copy_path_to_it(ast):
...
+                type_at_target = _get_type_at_macro_edge_target_using_current_type(
+                    schema, selection, type_in_selection)
+                if type_at_target is not None:
+                    return type_at_target
+
+    return None  # Didn't find target
+
+
+def get_type_at_macro_edge_target(schema, ast):
+    """Return type at the @macro_edge_target or None if there is no target."""
```
obi1kenobi: grammar nit: `Return the type`? Maybe even `Return the GraphQL type`?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -125,13 +125,21 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
 
     Args:
         target_ast: AST at the @macro_edge_target directive
+        type_at_target: str name of the type at the target
```
obi1kenobi: I believe everywhere else we just pass around GraphQL type objects around, why not pass a type object here as well?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -146,48 +154,45 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
...
+        if coercion_class != type_at_target:
+            if subclass_sets is None:
+                # TODO(bojanserafimov): Write test for this failure
+                raise GraphQLCompilationError(
+                    u'Cannot prove type coercion at macro target is valid. Please provide a '
+                    u'hint that {} subclasses {} using the subclass_sets argument.'
+                    .format(coercion_class, type_at_target))
+            else:
+                if (type_at_target not in subclass_sets or
+                        coercion_class not in subclass_sets[type_at_target]):
```
obi1kenobi: nit: perhaps more concise: `coercion_class not in subclass_sets.get(type_at_target, set())` instead of the two-line, two clause approach?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -146,48 +154,45 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
...
+                    raise GraphQLCompilationError(u'Invalid type coercion at macro target. {} '
+                                                  u'is expected to subclass {}.'
```
obi1kenobi: nit: 'Foo is expected to subclass Bar' would be the message here, but this is confusing because it can be interpreted in two ways: either "Foo is a subclass of Bar, but your config is wrong -- make your friendly neighborhood backend engineer fix it", or "The macro wanted Foo to be a subclass of Bar, but it is not and therefore it is an invalid macro -- user error, fix your input." I believe the correct conclusion is the second one, but we should probably clarify the message so the user knows it for sure.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -146,48 +154,45 @@ def _merge_selection_into_target(target_ast, selection_ast, subclass_sets=None):
...
+                                          target_ast.selection_set, directives=[])
+            target_ast.selection_set = SelectionSet([new_coercion])
+            target_ast = new_coercion
+
+        # coercion_class is required to subclass type_at_target so we can merge the
+        # macro selections inside the coercion, and merge the two coercions into one
+        if coercion_class != type_at_target:
+            if subclass_sets is None:
+                # TODO(bojanserafimov): Write test for this failure
+                raise GraphQLCompilationError(
```
obi1kenobi: Should these errors be "invalid macro" errors instead?

PR #204, opened 2019-02-26
Add schema generation from orientdb records

pmantica1:


graphql_compiler/__init__.py
```diff
@@ -17,11 +17,20 @@
...
+from .schema_generation.schema_properties import ORIENTDB_BASE_VERTEX_CLASS_NAME
+from .schema_generation.graphql_schema import get_graphql_schema_from_schema_graph
+from .schema_generation.utils import toposort_classes
 
 
 __package_name__ = 'graphql-compiler'
 __version__ = '1.10.0'
 
+# Match query used to generate OrientDB records that are themselves used to generate GraphQL schema.
+ORIENTDB_SCHEMA_RECORDS_QUERY = 'SELECT FROM (SELECT expand(classes) FROM metadata:schema) '\
```
obi1kenobi: Use parens instead of `\`:

```
ORIENTDB_SCHEMA_RECORDS_QUERY = (
    'foo'
    'bar'
)
```

But also, this should probably not be in the root `__init__.py` file, since it's very much an implementation detail of the schema generation.

graphql_compiler/__init__.py
```diff
@@ -132,3 +141,36 @@ def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints
...
+            where it's first defined and all the the class's subclasses.
+        hidden_classes: optional set of Strings, classes to not include in the GraphQL schema.
+
+    Returns:
+        tuple (GraphQL schema object, GraphQL type equivalence hints dict), or (None, None)
+        if there is no schema data in the graph yet.
+        For example, the graph has no schema data if applying schema updates from the very
+        first update. We have to return None because the GraphQL library does not support
+        empty schema objects -- the root object must have some keys and values.
+
```
obi1kenobi: remove this blank line

graphql_compiler/__init__.py
```diff
@@ -132,3 +141,36 @@ def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints
...
+        class_to_field_type_overrides: optional dict, class name -> {field name -> field type},
+            (string -> {string -> GraphQLType}). Used to override the type of a field in the class
+            where it's first defined and all the the class's subclasses.
+        hidden_classes: optional set of Strings, classes to not include in the GraphQL schema.
+
+    Returns:
+        tuple (GraphQL schema object, GraphQL type equivalence hints dict), or (None, None)
+        if there is no schema data in the graph yet.
+        For example, the graph has no schema data if applying schema updates from the very
+        first update. We have to return None because the GraphQL library does not support
```
obi1kenobi: Let's just raise an error instead of `(None, None)`? There are no schema updates as far as the compiler is concerned, that's an implementation detail of the underlying abstractions.

graphql_compiler/__init__.py
```diff
@@ -132,3 +141,36 @@ def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints
...
+
+def get_graphql_schema_from_orientdb_records(schema_records, class_to_field_type_overrides=None,
+                                             hidden_classes=None):
+    """Construct a GraphQL schema from a list of OrientDB schema records.
+
+    Args:
+        schema_records: list of pyorient.otypes.OrientRecord describing the OrientDB schema.
+                        Generated by running ORIENTDB_SCHEMA_RECORDS_QUERY on OrientDB.
+        class_to_field_type_overrides: optional dict, class name -> {field name -> field type},
+            (string -> {string -> GraphQLType}). Used to override the type of a field in the class
```
obi1kenobi: Please make your docstrings consistent with those of the surrounding code. In this one docstring, the `Generated` is indented to under the `list` (i.e. visual indent after the `:`), whereas the second line of the `class_to_field_type_overrides` description is given only a single four-space indent. Especially for a file that will be commonly read by users such as the root `__init__.py` file, we have to make sure our style is consistent and reader-friendly.

graphql_compiler/__init__.py
```diff
@@ -132,3 +141,36 @@ def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints
...
+            where it's first defined and all the the class's subclasses.
+        hidden_classes: optional set of Strings, classes to not include in the GraphQL schema.
```
obi1kenobi: lowercase `string` please

graphql_compiler/__init__.py
```diff
@@ -132,3 +141,36 @@ def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints
...
     return compilation_result._replace(
         query=insert_arguments_into_query(compilation_result, parameters))
+
+
+def get_graphql_schema_from_orientdb_records(schema_records, class_to_field_type_overrides=None,
+                                             hidden_classes=None):
+    """Construct a GraphQL schema from a list of OrientDB schema records.
+
+    Args:
+        schema_records: list of pyorient.otypes.OrientRecord describing the OrientDB schema.
```
obi1kenobi: Let's not take in as input objects from a library that we don't depend on. Why not make the caller do the work of converting from whichever type their database client returns to something common like `dict`?

graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return class_list
+
+    # Map names to classes
+    class_map = {c['name']: c for c in classes}
+    seen_classes = set()
+
+    toposorted = []
+    for name in class_map.keys():
+        toposorted.extend(get_class_topolist(name, class_map, seen_classes, set([])))
```
obi1kenobi: nit: `set([])` is just `set()`

graphql_compiler/tests/test_helpers.py
```diff
@@ -18,6 +18,138 @@
 # flag to indicate a test component should be skipped
 SKIP_TEST = 'SKIP'
 
+# String of a GraphQL schema generated from OrientDB. Used for future reference testing.
+ORIENTDB_GENERATED_SCHEMA = '''
+       schema {
```
obi1kenobi: de-indent this, and let's make sure that the auto-generated schema and the hand-written one are equivalent

graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,70 @@
...
+        """Return a topologically sorted list of this class's dependencies and class itself
+
+        Args:
+            class_name: string, name of the class to process
+            name_to_class: dict, class_name -> descriptor
+            processed_classes: set of strings, a set of classes that have already been processed
+            current_trace: list of strings, list of classes traversed during the recursion.
+
+        Returns:
+            element of classes list sorted in topological order
```
MichaelaShtilmanMinkin: Please try to keep the first word of the `Returns` description as the type description (i.e. `list of .`). It makes it easier for the reader to infer the correct output type at a cursory glance. The same trend and reasoning also applies for `Args`.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -96,7 +93,7 @@ class SchemaElement(object):
     ELEMENT_KIND_EDGE = 'edge'
     ELEMENT_KIND_NON_GRAPH = 'non-graph'
 
-    def __init__(self, class_name, kind, abstract, properties, class_fields):
```
pmantica1: Deleted class fields since they were not used neither in Defacto nor in the compiler.

pmantica1: Since it relied on customFields from the OrientDB dict, deleting this made it easier to write the docs.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -184,61 +219,30 @@ def types_spec():
...
-        for superclass_name in schema_graph.get_inheritance_set(class_name):
-            if superclass_name != class_name:
-                superclass = schema_graph.get_element_by_class_name(superclass_name)
-                for field_name in field_type_overrides:
-                    if field_name in superclass.properties:
-                        raise AssertionError(
-                            u'Attempting to override field "{}" from class "{}", but the field is '
-                            u'defined in superclass "{}"'
-                            .format(field_name, class_name, superclass_name))
+    _validate_overriden_classes_have_no_superclasses(class_to_field_type_overrides, schema_graph)
```
pmantica1: Moved all of this into a function. The functions below are similar.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -20,6 +20,41 @@
 )
 
 
+def _get_inherited_field_types(class_to_field_type_overrides, schema_graph):
```
pmantica1: Move code below into these three next functions. Also instead of having hidden classes as a parameter, I just hide classes with no properties. Although it removes the sanity check that all classes have properties, it simplifies the user interface. Furthermore, I am going to add this sanity check back in defacto later by asserting that all classes are in the schema except for Vertex and Trait which have no fields.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -320,18 +324,21 @@ def get_graphql_schema_from_schema_graph(schema_graph, class_to_field_type_overr
     for non_graph_cls_name in sorted(schema_graph.non_graph_class_names):
         if non_graph_cls_name in hidden_classes:
             continue
+        if not schema_graph.get_element_by_class_name(non_graph_cls_name).abstract:
```
pmantica1: Merged this code from other branch. I added an implicit reference test. So the test that test that the string representation of the autogenerated schema is equal to SCHEMA_TEXT, implicitly tests that UniquelyIdentifiable is included in the schema.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -113,9 +110,11 @@ def __init__(self, class_name, kind, abstract, properties, class_fields):
             a SchemaElement with the given parameters
         """
         if kind == SchemaElement.ELEMENT_KIND_EDGE:
-            # Ensure that all edge classes have defined in/out properties.
-            _validate_edge_has_defined_endpoint_types(class_name, properties)
             _validate_edges_do_not_have_extra_links(class_name, properties)
+            if not abstract:
+                # Ensure that all non-abstract edge classes have defined in/out properties.
+                _validate_edge_has_defined_endpoint_types(class_name, properties)
```
pmantica1: This fix is one that we identified in defacto. Basically, edges only need to have defined edge points if they are not abstract.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -10,14 +10,11 @@
...
     validate_supported_property_type_id
 )
+from .utils import toposort_classes
 
 
 def _validate_edge_has_defined_endpoint_types(class_name, properties):
     """Validate that the edge properties dict has defined types for the in/out link properties."""
-    exception_classes = frozenset({ORIENTDB_BASE_EDGE_CLASS_NAME})
-    if class_name in exception_classes:
-        return
```
pmantica1: The base edge class is guaranteed to be abstract and so we'll never get here. Also see bottom comment. 
            if class_name in orientdb_base_classes:
                # Special-case the V and E base classes:
                # OrientDB won't let us make them abstract, but we don't want to create
                # any vertices or edges with those types either.
                # Pretend they are marked abstract in OrientDB's schema.
                abstract = True

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -203,18 +202,44 @@ class SchemaGraph(object):
     on the graph. It also holds a fully denormalized schema for the graph.
     """
 
-    def __init__(self, schema_query_result):
+    def __init__(self, schema_data):
         """Create a new SchemaGraph from the given schema and index query results.
```
bojanserafimov: what query results?

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -20,6 +20,41 @@
...
+    inherited_field_type_overrides = dict()
+    for superclass_name, field_type_overrides in class_to_field_type_overrides.items():
+        for subclass_name in schema_graph.get_subclass_set(superclass_name):
+            inherited_field_type_overrides.setdefault(subclass_name, dict())
+            inherited_field_type_overrides[subclass_name].update(field_type_overrides)
+    return inherited_field_type_overrides
+
+
+def _validate_overriden_classes_have_no_superclasses(class_to_field_type_overrides, schema_graph):
+    """Assert that the fields we want to override are not defined in a superclass."""
```
bojanserafimov: The name and the documentation talk of very different things. The documentation makes more sense.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -20,6 +20,41 @@
...
+                for field_name in field_type_overrides:
+                    if field_name in superclass.properties:
+                        raise AssertionError(
+                            u'Attempting to override field "{}" from class "{}", but the field is '
+                            u'defined in superclass "{}"'
+                            .format(field_name, class_name, superclass_name))
+
+
+def _get_classes_with_no_properties(schema_graph):
+    """Return the set of classes that have no properties. There are not representable in GraphQL."""
```
bojanserafimov: `There are` -> `They are`. Or maybe just remove the sentence. Decide whether this function gets classes with no properties, or classes that can't be represented (I vote for the former). Then make the title, documentation, and name of return value agree on this. Then when you use the function you can comment that currently the only classes we can't represent are those with no properties.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -20,6 +20,41 @@
...
+    hidden_classes = set()
+    # If user fails to input hidden types, assert that at all vertices have valid GraphQL reps.
```
bojanserafimov: How is the user inputting hidden types?

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -345,9 +352,8 @@ def get_graphql_schema_from_schema_graph(schema_graph, class_to_field_type_overr
                 graphql_types[non_graph_cls_name] = graphql_type
 
     if not graphql_types:
-        # After evaluating all subclasses of V, we weren't able to find any visible schema data
-        # to import into the GraphQL schema object.
-        return None, None
+        raise EmptySchemaError(u'After evaluating all subclasses of V, we were not able to find'
```
bojanserafimov: add space at the end of the first row.

graphql_compiler/schema_generation/schema_graph.py
bojanserafimov: Document in the name of the function the precondition that the class is abstract

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -203,18 +202,44 @@ class SchemaGraph(object):
...
 
         Args:
-            schema_query_result: list of dicts describing the OrientDB schema, in OrientDB's format:
-                                 e.g. "SELECT * FROM (SELECT expand(classes) FROM metadata:schema)"
-                                 Required to already be topologically sorted with respect to the
-                                 inheritance hierarchy.
+            schema_data: list of dicts describing the classes in the OrientDB schema.
```
bojanserafimov: Maybe mention as a hint that this is the way the data is structured in orientdb 2 and it's easy to get with a single query.

graphql_compiler/__init__.py
```diff
@@ -140,7 +140,11 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
     """Construct a GraphQL schema from an OrientDB schema.
 
     Args:
-        schema_data: list of dicts describing the classes in the OrientDB schema.
+        schema_data: list of dicts describing the classes in the OrientDB schema. The following
+                     format is the way the data is structured in OrientDB 2. See
+                     test_get_graphql_schema_from_orientdb_schema in
+                     graphql_compiler/tests/integration_tests/test_backends_integration.py
+                     for an example of how to query this data.
```
bojanserafimov: `example of how to get this data from orientdb`. Querying is easy, it's a dict :D

graphql_compiler/schema_generation/schema_graph.py
obi1kenobi: @pmantica1, I am a bit worried about this change and how the GraphQL schema generation code will handle it. The case I am worried about is this: the user has defined an abstract edge class in OrientDB, and has only defined the type of *one* of the edge's endpoints.

This is not a case that happens in our internal use case, but since this code is open-source, it may occur in someone else's use case. Would you mind adding a test (in a new PR) for the schema generation code that attempts to pin down what happens in such a situation? What should happen in that situation?

I don't think you need to actually have that schema exist in OrientDB -- you should be able to test for it by constructing a simple OrientDB schema list-of-dicts response by hand.

graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Match query used to generate OrientDB records that are themselves used to generate GraphQL schema.
+ORIENTDB_SCHEMA_RECORDS_QUERY = (
+    'SELECT FROM (SELECT expand(classes) FROM metadata:schema) '
+    'WHERE name NOT IN [\'ORole\', \'ORestricted\', \'OTriggered\', '
+    '\'ORIDs\', \'OUser\', \'OIdentity\', \'OSchedule\', \'OFunction\']'
+)
+
+
+def toposort_classes(classes):
```
obi1kenobi: @lodrion congratulations on your first contribution to the GraphQL compiler :)

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Predrag I think you might be reviewing an intermediate commit. I renamed this method to _validate_non_abstract_edge_has_defined_endpoint_types in the last commit of the pull request and I only run this validation if the edge is not abstract. So nothing should happen in the use case you mentioned?

PR #205, opened 2019-02-27
Macro edge validation for use of directives

bojanserafimov:
Check that macros are using non-macro directives correctly

### DIFF
--- a/graphql_compiler/macros/macro_edge/validation.py
+++ b/graphql_compiler/macros/macro_edge/validation.py
@@ -4,7 +4,8 @@ from copy import copy
 from itertools import chain
 
 from graphql.language.ast import (
-    Argument, Directive, Document, Field, Name, SelectionSet, StringValue
+    Argument, Directive, Document, Field, InlineFragment, Name, OperationDefinition, SelectionSet,
+    StringValue
 )
 from graphql.validation import validate
 
@@ -14,7 +15,10 @@ from ...ast_manipulation import (
 from ...compiler.compiler_frontend import ast_to_ir
 from ...exceptions import GraphQLInvalidMacroError
 from ...query_formatting.common import ensure_arguments_are_provided
-from ...schema import VERTEX_FIELD_PREFIXES, is_vertex_field_name
+from ...schema import (
+    VERTEX_FIELD_PREFIXES, FilterDirective, FoldDirective, OptionalDirective, OutputDirective,
+    OutputSourceDirective, RecurseDirective, TagDirective, is_vertex_field_name
+)
 from .directives import (
     MACRO_EDGE_DIRECTIVES, MacroEdgeDefinitionDirective, MacroEdgeTargetDirective
 )
@@ -59,6 +63,40 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
                 u'Required macro edge directive "@{}" was unexpectedly present more than once in '
                 u'the supplied macro edge definition GraphQL. It was found {} times.'
                 .format(directive_definition.name, len(macro_data)))
+
+
+# TODO(bojanserafimov): Write tests for each case
+def _validate_macro_ast_directives(ast, inside_fold_scope=False):
+    """Check that the macro is using non-macro direcives properly."""
+    subselection_inside_fold_scope = inside_fold_scope
+    for directive in ast.directives:
+        name = directive.name.value
+        if name == FilterDirective.name:
+            pass
+        elif name == TagDirective.name:
+            pass
+        elif name == OutputDirective.name:
+            raise GraphQLInvalidMacroError(u'Macros should not have outputs.')
+        elif name == OutputSourceDirective.name:
+            raise GraphQLInvalidMacroError(u'Macros should not have output source directives.')
+        elif name == OptionalDirective.name:
+            pass
+        elif name == RecurseDirective.name:
+            pass
+        elif name == FoldDirective.name:
+            subselection_inside_fold_scope = True
+        elif name == MacroEdgeTargetDirective.name:
+            if inside_fold_scope:
+                raise GraphQLInvalidMacroError(
+                    u'The @macro_edge_target cannot be inside a fold scope.')
+
+    if isinstance(ast, (Field, InlineFragment, OperationDefinition)):
+        if ast.selection_set is not None:
+            for selection in ast.selection_set.selections:
+                _validate_macro_ast_directives(
+                    selection, inside_fold_scope=subselection_inside_fold_scope)
+    else:
+        raise AssertionError(u'Unexpected AST type received: {} {}'.format(type(ast), ast))
 
 
 def _validate_class_selection_ast(ast, macro_defn_ast):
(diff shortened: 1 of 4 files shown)


graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,39 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
                 .format(directive_definition.name, len(macro_data)))
 
 
+def _validate_macro_ast_directives(ast, inside_fold_scope=False):
+    """Check that the macro is using non-macro direcives properly."""
+    subselection_inside_fold_scope = inside_fold_scope
+    for directive in ast.directives:
+        name = directive.name.value
+        if name == FilterDirective.name:
```
obi1kenobi: Unify all of these `pass` cases in a single case where you check `if name in <set>`.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,39 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+            pass
+        elif name == TagDirective.name:
+            pass
+        elif name == OutputDirective.name:
+            raise GraphQLInvalidMacroError(u'Macros should not have outputs.')
```
obi1kenobi: Let's give some more context in these error messages. Also, `should` -> `must` because it's a requirement.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,39 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+        elif name == OptionalDirective.name:
+            pass
+        elif name == RecurseDirective.name:
+            pass
+        elif name == FoldDirective.name:
+            subselection_inside_fold_scope = True
+        elif name == MacroEdgeTargetDirective.name:
+            if inside_fold_scope:
+                raise GraphQLInvalidMacroError(
+                    u'The @macro_edge_target cannot be inside a fold scope.')
```
obi1kenobi: add an `else` that raises `AssertionError` to make sure this matching is exhaustive

bojanserafimov: Good catch. I had forgotten about `@macro_edge_definition`

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,45 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
                 .format(directive_definition.name, len(macro_data)))
 
 
+def _validate_macro_ast_directives(ast, inside_fold_scope=False):
+    """Check that the macro is using non-macro direcives properly."""
```
jmeulemans: Consider a summary of the rules in the top-level docstring, it takes a little digging to figure out what's allowed for directives otherwise.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,45 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+        FilterDirective.name,
+        TagDirective.name,
+        OptionalDirective.name,
+        RecurseDirective.name,
+        MacroEdgeDefinitionDirective.name,
+    }
+    for directive in ast.directives:
+        name = directive.name.value
+        if name in directives_with_no_restrictions:
+            pass
```
jmeulemans: Are you sure you want `pass` here vs. an explicit `continue`?

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,45 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+        elif name == OutputDirective.name:
+            raise GraphQLInvalidMacroError(u'Macros must not have outputs. '
```
jmeulemans: These cases are very similar, what about generalizing and using the name of the directive in the error message? 
ie: `raise GraphQLInvalidMacroError(u'Macros are not allowed to use the "{}" directive. Found usage {}.'.format(name, directive)`

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,45 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+                                           u'Found output directive {}'.format(directive))
+        elif name == OutputSourceDirective.name:
+            raise GraphQLInvalidMacroError(u'Macros must not have output source directives. '
+                                           u'Found an outut_source directive {}'.format(directive))
```
jmeulemans: `outut_source` -> `output_source`

graphql_compiler/macros/macro_edge/validation.py
bojanserafimov: Yes. The `if/elif/else` block explicitly covers all cases. A `continue` is equivalent with the `pass` but makes the control flow more complicated. If there's any code added at the end of the for loop, there's no reason it shouldn't apply to the `directives_with_no_restrictions`.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,55 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
 
+def _validate_macro_ast_directives(ast, inside_fold_scope=False):
+    """Check that the macro is using non-macro direcives properly.
+
+    Restrictions on use of directives:
+    - @output and @output_source are disallowed
+    - @macro_edge_target is not allowed to be inside a @fold scope
+
+    Args:
+        ast: GraphQL AST describing a subtree of the macro
```
MichaelaShtilmanMinkin: What's the exact object name?

bojanserafimov: https://github.com/kensho-technologies/graphql-compiler/pull/210#discussion_r265748334

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,55 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+        inside_fold_scope: bool, whether the subtree is within a @fold scope
+    """
+    subselection_inside_fold_scope = inside_fold_scope
+    directives_with_no_restrictions = (
```
obi1kenobi: nit: We're using these with the `in` operator, any reason they shouldn't be frozensets instead?

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -61,6 +65,55 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
+                                           u'Found usage {}'.format(name, directive))
+        elif name == FoldDirective.name:
+            subselection_inside_fold_scope = True
+        elif name == MacroEdgeTargetDirective.name:
+            if inside_fold_scope:
+                raise GraphQLInvalidMacroError(
+                    u'The @macro_edge_target cannot be inside a fold scope.')
+        else:
+            raise AssertionError(u'Unexpected directive name found: {} {}'.
+                                 format(directive.name.value, directive))
```
obi1kenobi: nit: Very interesting that this works. I would have put the `.` together with `format` though, because you see that the syntax highlighting is a bit messed up at the moment (it thinks `format` is a global).

PR #206, opened 2019-02-27
Check macro argument types

bojanserafimov:


graphql_compiler/query_formatting/common.py
```diff
@@ -13,10 +19,64 @@
...
+    elif GraphQLDate.is_same_type(expected_type):
+        if not isinstance(value, datetime.date):
+            raise GraphQLInvalidArgumentError(u'Attempting to represent a non-date as a date: '
+                                              u'{}'.format(value))
+    elif GraphQLDateTime.is_same_type(expected_type):
+        if not isinstance(value, (datetime.date, arrow.Arrow)):
+            raise GraphQLInvalidArgumentError(u'Attempting to represent a non-date as a date: '
+                                              u'{}'.format(value))
+    elif isinstance(expected_type, GraphQLList):
+        if not isinstance(value, list):
```
obi1kenobi: Recursively check the type inside the list, too?

graphql_compiler/query_formatting/common.py
```diff
@@ -26,6 +86,9 @@ def ensure_arguments_are_provided(expected_types, arguments):
         raise GraphQLInvalidArgumentError(u'Missing or unexpected arguments found: '
                                           u'missing {}, unexpected '
                                           u'{}'.format(missing_args, unexpected_args))
+    if check_types:
```
obi1kenobi: Let's remove the option to skip type checking, and always type check. We can consider adding this short-circuiting option back if benchmarks show it to be necessary, but I suspect the query execution time will always dominate the time it takes to run `isinstance` on a few arguments.

graphql_compiler/query_formatting/common.py
```diff
@@ -13,10 +19,64 @@
...
+        if isinstance(value, bytes):  # should only happen in py3
+            value.decode('utf-8')  # decoding should not raise errors
+        else:
+            raise GraphQLInvalidArgumentError(u'Attempting to convert a non-string into a string: '
+                                              u'{}'.format(value))
+
+
+# TODO(bojanserafimov): test this function
+def _validate_argument_type(expected_type, value):
+    """Check if the value is appropriate for the type and usable in any of our backends."""
```
MichaelaShtilmanMinkin: Can you please be more specific about `our backends`?

graphql_compiler/tests/test_macro_validation.py
```diff
@@ -208,7 +208,29 @@ def test_macro_edge_extra_args(self):
...
+                color @filter(op_name: "=", value: ["$color"])
+                out_Animal_ParentOf {
+                    out_Animal_ParentOf @macro_edge_target {
+                        uuid
+                    }
+                }
+            }
+        }'''
+        args = {
+            'net_worth': 'five_cows',
```
MichaelaShtilmanMinkin: Suggestion: instead of `five_cows`, let's make this `incorrect_net_worth_type`? And maybe add a quick comment in the same line `# valid types are x and y`

graphql_compiler/query_formatting/common.py
```diff
@@ -13,10 +20,89 @@
 # Public API
 ######
 
+def _check_is_string_value(value):
```
MichaelaShtilmanMinkin: Should `_check_is_string_value` live within the public API and, if so, might we want to remove the prefixed `_`? This seems like something that should belong with validators.

graphql_compiler/query_formatting/common.py
```diff
@@ -13,10 +20,89 @@
...
+
+def validate_argument_type(expected_type, value):
+    """Check if the value is appropriate for the type and usable in any of our backends.
+
+    Backends are the database languages we have the ability to compile to, like OrientDB Match,
+    Gramlin, or SQLAlchemy. This function should be stricter than the validation done by any
+    specific backend. That way code that passes validation can be compiled to any backend.
+
+    Args:
+        expected_type: GraphQL type we expect
```
MichaelaShtilmanMinkin: Is the a case class for the GraphQL types we could use here to be more specific?

bojanserafimov: It's complicated https://github.com/graphql-python/graphql-core/blob/master/graphql/type/definition.py#L21

pmantica1: Consider at least using GraphQLType.

graphql_compiler/query_formatting/common.py
```diff
@@ -1,22 +1,109 @@
...
+        # Safeguard against this with an explicit check against bool type.
+        if isinstance(value, bool) or not isinstance(value, six.integer_types):
+            raise GraphQLInvalidArgumentError(u'Attempting to represent a non-int as an int: '
+                                              u'{} {}'.format(type(value), value))
+    elif GraphQLBoolean.is_same_type(expected_type):
+        if not isinstance(value, bool):
+            raise GraphQLInvalidArgumentError(u'Attempting to represent a non-bool as a bool: '
+                                              u'{} {}'.format(type(value), value))
+    elif GraphQLDecimal.is_same_type(expected_type):
+        if isinstance(value, bool):
```
obi1kenobi: `bool`? I am confused.

graphql_compiler/query_formatting/common.py
```diff
@@ -1,22 +1,109 @@
...
+from ..schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal
 from .gremlin_formatting import insert_arguments_into_gremlin_query
 from .match_formatting import insert_arguments_into_match_query
 from .sql_formatting import insert_arguments_into_sql_query
 
 
+def _check_is_string_value(value):
+    """Raise if the value is not a proper utf-8 string."""
+    if not isinstance(value, six.string_types):
+        if isinstance(value, bytes):  # should only happen in py3
```
obi1kenobi: Bytes are not strings, I think this should be an error. `GraphQLString` should only accept `six.string_types`.

graphql_compiler/query_formatting/common.py
```diff
@@ -1,22 +1,109 @@
...
+
+    Args:
+        expected_type: GraphQL type we expect
+        value: object that can be interpreted as being of that type
+    """
+    if GraphQLString.is_same_type(expected_type):
+        _check_is_string_value(value)
+    elif GraphQLID.is_same_type(expected_type):
+        # IDs can be strings or numbers, but the GraphQL library coerces them to strings.
+        # We will follow suit and treat them as strings.
```
obi1kenobi: It might be time to relax this and allow `int` as well. So the types accepted would be `six.string_types + (int,)`

graphql_compiler/query_formatting/common.py
obi1kenobi: I think my comments below will cause this function to be only used in one place and in simplified form, at which point it can just get inlined.

graphql_compiler/query_formatting/common.py
bojanserafimov: Let's do that when seperately, when we actually implement int parsing of ids?

graphql_compiler/query_formatting/common.py
```diff
@@ -63,6 +57,8 @@ def validate_argument_type(expected_type, value):
             raise GraphQLInvalidArgumentError(u'Attempting to represent a non-bool as a bool: '
                                               u'{} {}'.format(type(value), value))
     elif GraphQLDecimal.is_same_type(expected_type):
+        # Special case: in Python, isinstance(True, int) returns True.
+        # Safeguard against this with an explicit check against bool type.
```
obi1kenobi: Maybe preface this comment with the types we'd like to support: `int, float, Decimal`, but not `bool`. Then this explanation makes sense.

obi1kenobi: This is my only concern, otherwise this is good to go.

PR #208, opened 2019-02-27
Fixed bug that failed to include certain abstract classes in schema

pmantica1:


graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -320,18 +320,21 @@ def get_graphql_schema_from_schema_graph(schema_graph, class_to_field_type_overr
     for non_graph_cls_name in schema_graph.non_graph_class_names:
         if non_graph_cls_name in hidden_classes:
             continue
+        if not schema_graph.get_element_by_class_name(non_graph_cls_name).abstract:
```
obi1kenobi: Please add an explanation for why we're continuing here. The `hidden_classes` part is obvious, this is less so.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -320,18 +320,21 @@ def get_graphql_schema_from_schema_graph(schema_graph, class_to_field_type_overr
...
+            continue
 
         cls_subclasses = schema_graph.get_subclass_set(non_graph_cls_name)
         # No need to add the possible abstract class if it doesn't have subclasses besides itself.
         if len(cls_subclasses) > 1:
```
bojanserafimov: is this if statement necessary?

bojanserafimov: resolved offline

graphql_compiler/schema_generation/graphql_schema.py
pmantica1: There is a comment explaining that the class has to be  abstract at the top of the block.

pmantica1:
I'll add a reproducing test sometime tonight. Trying to change the schema for the autogenerated one right now.

PR #211, opened 2019-02-28
Change the schema from our manually written one to the autogenerated one

pmantica1:
This involved changing  some find and replace. 
Field type overriding when autogenerating schema (did this in a script not included in the pull request)
Slightly changing the SQL schema (in seperate pull request) 
Changing FedAt from pointing to Event to BirthEvent because the compiler would never assign an edge to be of Event type. Since it is a non-abstract class that is inherited by BirthEvent. In other words, the edge type would always be of type union of BirthEvent and Event. If I had replaced the type by an union type instead, then I would had to deal with the change of behavior in the tests, (what made it specifically problematic was the change in behavior due to the interaction between union's and `@optional`)


obi1kenobi:
I am afraid that changing the `Animal_FedAt` edge to point to `BirthEvent` is a no-go, since it is definitely very unexpected and unintuitive. I believe that having it as a union type is the correct thing to do -- how difficult would it be to alter the tests accordingly?

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -620,12 +620,6 @@ def test_filter_directive_wrong_location(self):
                     description @output(out_name: "description_text")
                 }
             }''',
-            '''{
-                Location @filter(op_name: "name_or_alias", value: ["$foo"]) {
-                    name @output(out_name: "name")
-                }
-            }''',
-
```
pmantica1: Changed the fields of Location slightly. Anyways, the few lines above cover the same case. So test was slightly redundant.

graphql_compiler/tests/snapshot_tests/snapshots/snap_test_orientdb_match_query.py
```diff
@@ -25,7 +25,7 @@
...
...0_5)_((18_4_5_7)_11_19_5)_(11_4_6_7)_7)_10_19_8)')}), 3), (frozenset({('animal_name', 'Hippogriff__((1_12_18_9)_11_13_16)')}), 3), (frozenset({('animal_name', 'Nazgul__(((18_19_2_6)_1_19_6)_(1_12_14_19)_19_5)')}), 2), (frozenset({('animal_name', 'Nazgul__(18_19_2_6)')}), 4), (frozenset({('animal_name', 'Nazgul__(((((18_19_2_6)_1_19_6)_15_16_2)_(((18_19_2_6)_1_19_6)_1_19_3)_(10_18_2_8)_2)_((((18_19_2_6)_1_19_6)_15_16_2)_11_14_6)_(1_12_14_19)_15)')}), 1), (frozenset({('animal_name', 'Hippogriff__((1_12_18_9)_17_3_4)')}), 3), (frozenset({('animal_name', 'Dragon__(((0_17_2_7)_12_16_18)_((1_16_2_6)_0_15_6)_11_4)')}), 2), (frozenset({('animal_name', 'Dragon__(1_16_2_6)')}), 4), (frozenset({('animal_name', 'Hippogriff__((0_12_15_17)_10_6_8)')}), 3), (frozenset({('animal_name', 'Nazgul__((((18_19_2_6)_1_19_6)_15_16_2)_11_14_6)')}), 3), (frozenset({('animal_name', 'Pteranodon__(((((0_10_12_4)_(10_11_13_2)_15_4)_0_10_4)_(0_10_12_4)_15_6)_((((0_10_12_4)_10_16_8)_13_2_3)_((0_10_12_4)_12_19_5)_10_14)_(0_10_12_8)_9)')}), 1), (frozenset({('animal_name', 'Pteranodon__(10_11_15_3)')}), 4), (frozenset({('animal_name', 'Pteranodon__((((0_10_12_4)_(10_11_13_2)_15_4)_0_10_4)_(0_10_12_4)_15_6)')}), 2)})
```
pmantica1: Change snapshot test because of change of using from Event to BirthEvent for the fedAt link

graphql_compiler/tests/snapshot_tests/snapshots/snap_test_orientdb_match_query.py
```diff
@@ -69,7 +69,7 @@
...
... ('sibling_and_self_names_list', ('Hippogriff__(((((0_12_15_17)_(10_19_2_7)_13_16)_18_2_7)_15_18_6)_((1_12_18_9)_11_13_16)_((1_12_18_9)_17_3_4)_10)', 'Hippogriff__((((0_12_15_17)_(10_19_2_7)_13_16)_18_2_7)_12_13_9)', 'Hippogriff__13', 'Hippogriff__5')), ('animal_name', 'Hippogriff__((((0_12_15_17)_(10_19_2_7)_13_16)_18_2_7)_12_13_9)')]), 1), (frozenset([('sibling_and_self_names_list', ()), ('animal_name', 'Hippogriff__(((0_12_15_17)_(10_19_2_7)_0_19)_1_17_18)'), ('sibling_and_self_uuids_list', ())]), 1), (frozenset([('sibling_and_self_names_list', ('Pteranodon__(10_11_15_3)', 'Pteranodon__1', 'Pteranodon__15', 'Pteranodon__17', 'Pteranodon__0', 'Pteranodon__14', 'Pteranodon__17', 'Pteranodon__19')), ('sibling_and_self_uuids_list', ('f5b7751f-497e-28aa-ce03-a34da022dfee', 'a867a0f0-a8e6-a772-b9cd-e60cab1e6794', '6f446806-c137-8e75-627b-9de79ae445e0', 'f313c0b0-76d1-42d0-5c66-a9bb62574c5a', 'c90ded29-5100-f186-d25a-b04930e9be17', 'efa2ce12-8c83-3848-11fb-a56e300a2187', 'f313c0b0-76d1-42d0-5c66-a9bb62574c5a', 'b47a5dce-cf70-9452-ce8c-037ee51c0cdb')), ('animal_name', 'Pteranodon__17')]), 1), (frozenset([('sibling_and_self_names_list', ()), ('animal_name', 'Dragon__((((13_14_8_9)_1_13_3
```
pmantica1: Same as above

bojanserafimov:
The problem with making `out_Animal_FedAt` a union type is that tests would now need coercions (it used to point to `Event` directly), and that will compromise the original idea of what the test was testing for (the combination of filters, coercions, optionals, etc). But I think we agreed to change the tests to use `out_Animal_BornAt`, as that edge points directly to a specific type, and not a union.

graphql_compiler/tests/test_helpers.py
```diff
@@ -94,137 +93,176 @@ def get_schema():
...
-        type RootSchemaQuery {
-            Animal: Animal
-            BirthEvent: BirthEvent
-            Entity: Entity
-            Event: Event
-            Food: Food
-            Species: Species
-            Location: Location
-        }
+    schema {
```
bojanserafimov: nit: indentation

PR #216, opened 2019-03-18
Added debug info to contributing.md

pmantica1:


CONTRIBUTING.md
```diff
@@ -27,7 +27,11 @@ pipenv shell
 
 py.test graphql_compiler/tests
 ```
-
+If you have any trouble starting the mysql database, make sure mysql service is not already running
+by typing:
+```
+service mysql stop
```
obi1kenobi: Is this specific to Linux, or will it work on e.g. OSX as well?

CONTRIBUTING.md
```diff
@@ -27,7 +27,11 @@ pipenv shell
 
 py.test graphql_compiler/tests
 ```
-
+If you have any trouble starting the mysql database, make sure mysql service is not already running
```
obi1kenobi: Might be worth making the distinction of "running outside of docker" here, so the user isn't confused about why they are stopping the service only to start it up again.

obi1kenobi: I'd leave blank lines before and after this "If you have any trouble" block.

CONTRIBUTING.md
pmantica1: Specific to linux. They are some stack overflow suggestions for OSX but I don't know which works because I work in Linux. Do you mind sharing the command that use for OSX to include it?

CONTRIBUTING.md
pmantica1: Makes sense

PR #219, opened 2019-03-19
Added two sanity tests to prevent someone from updating the current schema with an invalid one

pmantica1:


graphql_compiler/tests/test_helpers.py
```diff
@@ -267,6 +272,18 @@ def get_schema():
     return schema
 
 
+def generate_schema(graph_client):
+    """Generate schema and type equivalence dict from an OrientDB client"""
```
bojanserafimov: `pyorient client`

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -174,14 +172,18 @@ def test_has_substring_precedence(self, backend_name):
...
-            "Animal": {"birthday": GraphQLDate},
-            "Event": {"event_date": GraphQLDateTime}
-        }
-        schema, _ = get_graphql_schema_from_orientdb_schema_data(schema_data, type_overrides)
+    def test_snapshot_graphql_schema_from_orientdb_schema(self):
+        schema, _ = generate_schema(self.graph_client)
         compare_ignoring_whitespace(self, SCHEMA_TEXT, print_schema(schema), None)
+
+    @integration_fixtures
+    def test_override_field_types(self):
```
bojanserafimov: Add a comment explaining how it's being done. It's not obvious that this implementation does what the name of the test suggests.

graphql_compiler/tests/test_helpers.py
```diff
@@ -267,6 +272,18 @@ def get_schema():
     return schema
 
 
+def generate_schema(graph_client):
+    """Generate schema and type equivalence dict from a pyorient client"""
+    schema_records = graph_client.command(ORIENTDB_SCHEMA_RECORDS_QUERY)
+    schema_data = [x.oRecordData for x in schema_records]
+    type_overrides = {
+        "UniquelyIdentifiable": {"uuid": GraphQLID},
+        "Animal": {"birthday": GraphQLDate},
```
obi1kenobi: @pmantica1, do the `GraphQLDate` and `GraphQLDateTime` fields need type overrides? They should be stored as `Date` and `DateTime` objects in the database, so I thought they'd get auto-detected as such in our schema builder. Is that not the case?

PR #220, opened 2019-03-19
Fix and test schema generation interface

pmantica1:


bojanserafimov:
This api has no test coverage. We didn't know that `type_id` is not the correct name of the argument. Maybe add a test with some small made up schema? For your convenience, here's the input I used that exposed this bug:
```
    schema_graph = SchemaGraph([
        {
            'name': 'Entity',
            'abstract': True,
            'superClasses': [],
            'properties': [
                {
                    'name': 'name',
                    'type': schema_properties.PROPERTY_TYPE_STRING_ID,
                }
            ]
        },
        {
            'name': 'Animal',
            'abstract': False,
            'superClasses': ['Entity'],
            'properties': [
                {
                    'name': 'net_worth',
                    'type': schema_properties.PROPERTY_TYPE_DECIMAL_ID,
                }
            ]
        },
        {
            'name': 'Location',
            'abstract': False,
            'superClasses': ['Entity'],
            'properties': [
                {
                    'name': 'description',
                    'type': schema_properties.PROPERTY_TYPE_STRING_ID,
                }
            ]
        },
    ])

```

graphql_compiler/__init__.py
```diff
@@ -159,22 +157,31 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
...
-                                                                   name of an endpoint of the edge.
+                                                                    collection of class instances,
+                                                                    then it indicates the name of
+                                                                    the class. If class is an edge
+                                                                    class, and the field name is
+                                                                    either 'in' or 'out', then it
+                                                                    describes the name of an
+                                                                    endpoint of the edge.
+                                         - defaultValue: string, the textual representation of the
+                                                          default value for the property, as
```
obi1kenobi: Nit: the visual indentation is off by one space (compare `string` vs `default`)

graphql_compiler/tests/snapshot_tests/test_graphql_schema_from_schema_data.py
```diff
@@ -0,0 +1,88 @@
...
+                'properties': [
+                    {
+                        'name': 'description',
+                        'type': PROPERTY_TYPE_STRING_ID,
+                    }
+                ]
+            },
+        ]
+        schema, _ = get_graphql_schema_from_orientdb_schema_data(orientdb_schema_data)
+        self.assertMatchSnapshot(print_schema(schema))
```
obi1kenobi: Do we actually need this to be a snapshot test here? This feels like a unit test to me since we have well-defined input and output, and a single function call being tested. Invoking all the snapshot testing machinery feels like overkill for this use case.

graphql_compiler/tests/test_data_tools/schema.sql
```diff
@@ -64,6 +64,8 @@ CREATE INDEX Animal.net_worth NOTUNIQUE
 CREATE CLASS Animal_ParentOf EXTENDS E
 CREATE PROPERTY Animal_ParentOf.in LINK Animal
 CREATE PROPERTY Animal_ParentOf.out LINK Animal
+ALTER CLASS Animal_ParentOf CUSTOM human_name_in="Parent"
```
obi1kenobi: Add a test on the generated schema graph that these properties exist and have their expected values?

graphql_compiler/tests/snapshot_tests/test_graphql_schema_from_schema_data.py
pmantica1: I think it's a good idea because every time I've made a small change to the way we generate the schema, I've had to print it and then copy paste it into test_helpers. If I introduce this test as an unit test, every time we change the schema we now have to change two things.

obi1kenobi:
This also doesn't merge cleanly anymore, so I'll hold off on reviewing until you've had a chance to resolve the merge conflicts.

PR #221, opened 2019-03-19
Stop hidding classes

pmantica1:
Classes all have at least one filed, (__x__count). There is no need to hide them. I assume that whoever wrote the hidden classes stuff did it to deal with the fact that classes with no properties have no valid graphql rep.


obi1kenobi:
I don't think we should remove the capability to hide classes. Classes might be hidden because of a user-specified business need, not just because they are not representable in GraphQL. We should allow the user to specify that some classes should be hidden, even if they are representable in GraphQL.

PR #222, opened 2019-03-19
Added the capability to selectively hide classes

pmantica1:


graphql_compiler/tests/test_data_tools/schema.sql
```diff
@@ -27,7 +27,7 @@ CREATE INDEX Entity_Related ON Entity_Related (in, out) UNIQUE_HASH_INDEX
 ### Event ###
 CREATE CLASS Event EXTENDS Entity
 
-CREATE PROPERTY Event.event_date Date
+CREATE PROPERTY Event.event_date DateTime
```
obi1kenobi: I think you might have accidentally included this change from the other PR.

graphql_compiler/tests/test_helpers.py
```diff
@@ -193,6 +192,10 @@
       _x_count: Int
       uuid: ID
     }
+
+    interface V {
```
obi1kenobi: I still don't think we want to expose `V`. We were hiding it because it's meaningless (there's nothing you can do with it, since `_x_count` is a synthetic field that's only usable from a `@fold`, which requires an edge to it, of which there are none). We weren't hiding it just because it wasn't representable.

pmantica1: How about if it has no properties I hide it. (This would be in get_graphql_schema_from_schema graph). I would also add the comment that we hide it because it's meaningless.

obi1kenobi:
This doesn't merge cleanly anymore, so I'll hold off on reviews until you've had a chance to resolve the conflicts.

PR #223, opened 2019-03-20
Update SQL schema types and removed type override no-ops. 

pmantica1:


graphql_compiler/tests/test_helpers.py
```diff
@@ -278,7 +278,6 @@ def generate_schema(graph_client):
     schema_data = [x.oRecordData for x in schema_records]
     type_overrides = {
         "UniquelyIdentifiable": {"uuid": GraphQLID},
-        "Animal": {"birthday": GraphQLDate},
         "Event": {"event_date": GraphQLDateTime}
```
obi1kenobi: How about the `DateTime` one below? Is that one not inferred automatically?

pmantica1: It is inferred as Date.

pmantica1: We need it to be inferred as Datetime for other tests.

obi1kenobi: Let's change its definition in the schema from `CREATE PROPERTY Event.event_date Date` to `CREATE PROPERTY Event.event_date DateTime` then, otherwise our database and GraphQL schemas will disagree and we might get invalid compiled output.

graphql_compiler/tests/test_data_tools/schema.sql
```diff
@@ -27,7 +27,7 @@ CREATE INDEX Entity_Related ON Entity_Related (in, out) UNIQUE_HASH_INDEX
 ### Event ###
 CREATE CLASS Event EXTENDS Entity
 
-CREATE PROPERTY Event.event_date Date
+CREATE PROPERTY Event.event_date DateTime
```
obi1kenobi: Remember that we have data for snapshot tests that gets loaded into the database, and will need to be updated to load `DateTime` data instead of `Date` data. This data is generated by scripts, and those scripts might need updating as well.

pmantica1: The snapshot tests seem to already be using datetimes instead of date. 
('fed_at_datetimes_list', ('2008-09-23T00:00:00Z')),
('child_birthdays_list', ('2012-11-02'))

PR #224, opened 2019-03-20
Raise error for unused @tag directives

evantey14:
After the metadata table is created, check that all tags are used in some filter, and raise an error otherwise.


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -83,7 +83,7 @@
     validate_root_vertex_directives, validate_vertex_directives,
     validate_vertex_field_directive_in_context, validate_vertex_field_directive_interactions
 )
-from .filters import process_filter_directive
+from .filters import _is_tag_argument, process_filter_directive
```
evantey14: Should `_is_tag_argument` be moved / made public somewhere? what place would be good for this?

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -490,13 +490,20 @@ def test_tag_directive_constraints(self):
             }
         }''')
 
+        unused_tag = (GraphQLCompilationError, '''{
+            Animal {
+                name @tag(tag_name: "name")
+            }
+        }''')
+
         errors_and_inputs = (
```
evantey14: Should I add filters to the other tag constraint tests? This change makes it a bit hard to isolate problems (e.g. the filter might have to filter on a tag on a vertex field or with an illegally named argument)?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +718,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
     return basic_blocks
 
 
+def _validate_all_tags_are_used(context):
+    tag_names = set([tag_name for tag_name in context['tags']])
```
obi1kenobi: nit: why not just `set(context['tags'])`?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +718,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
...
+    filter_args = set([])
```
obi1kenobi: nit: `set()`?

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +718,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
...
+    for location, _ in context['metadata'].registered_locations:
```
obi1kenobi: I'd kind of rather do this in the opposite direction -- let's make sure the metadata table carries enough information to determine which tags are defined, and which are used. That will make the code a lot more stable since it won't be as deeply coupled to the various parts of the `context` in the frontend.

graphql_compiler/tests/test_ir_generation_errors.py
obi1kenobi: Yes, please do.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +718,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
...
+        filter_infos = context['metadata'].get_filter_infos(location)
+        for filter_info in filter_infos:
+            for arg in filter_info.args:
+                if _is_tag_argument(arg):
+                    filter_args.add(arg[1:])
```
pmantica1: Add a comment explaining the structure of args. Think of this sort of comments as guide comments for future readers.

obi1kenobi: We've added comments like that a lot, and it always felt a bit janky. Maybe it's time to have a helper method that gets the name of the argument without the prefix? Then we should be able to use it in `expressions.Variable` for validation as well, etc.

evantey14: temporarily added a comment. Is it better to create this method now and update other usages in another PR or leave this all for another PR?

graphql_compiler/compiler/compiler_frontend.py
evantey14: Added tag info into metadata in new commits. Do you have any thoughts on how this is implemented before I go and add more tests to `test_explain_info.py`?

graphql_compiler/compiler/compiler_frontend.py
obi1kenobi: Perhaps either `metadata.py` or `helpers.py`, but move it only together with its dual, `is_variable_argument()`.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +719,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
...
 
+def _validate_all_tags_are_used(metadata):
+    """Ensure all tags are used in some filter."""
+    tag_names = set(metadata.tags)
+    filter_args = set()
+    for location, _ in metadata.registered_locations:
+        for filter_info in metadata.get_filter_infos(location):
+            for arg in filter_info.args:
+                if _is_tag_argument(arg):
+                    filter_args.add(arg[1:])  # Don't include % in tag name
```
obi1kenobi: Let's maybe also add a helper for this, in the same place where `_is_tag_argument()` is getting moved to.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +719,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
...
+
+    unused_tags = tag_names - filter_args
+    if len(unused_tags) > 0:
```
obi1kenobi: nit: you could leave this as just `if unused_tags`, it will do the same thing.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -718,6 +719,21 @@ def _compile_ast_node_to_ir(schema, current_schema_type, ast, location, context)
...
+        raise GraphQLCompilationError(u'Found at least one unused @tag: {}.'.format(unused_tags))
```
obi1kenobi: Good error messages are fundamental to a good user experience, and ones that are due to user error should always be actionable for the user -- they must know how to resolve their error. How about something like:
`This GraphQL query contains @tag directives whose values are not used: {}. This is not allowed. Please either use them in a filter, or remove them entirely.`

graphql_compiler/compiler/metadata.py
```diff
@@ -154,11 +161,20 @@ def get_location_info(self, location):
                                  u'{}'.format(location))
         return location_info
 
+    @property
+    def tags(self):
+        """Return a dict of tag name to TagInfo for the query."""
+        return self._tags
```
obi1kenobi: Dicts are mutable, let's not return references to mutable objects which the caller may abuse. Look at how the `registered_locations` property gets around this problem.

graphql_compiler/compiler/metadata.py
```diff
@@ -154,11 +161,20 @@ def get_location_info(self, location):
...
+
+    def record_tag_info(self, tag_name, tag_info):
+        """Record information about the tag."""
+        self._tags[tag_name] = tag_info
```
obi1kenobi: Is it ever okay to overwrite an existing tag's info? If not, let's assert that the tag being recorded hasn't been recorded previously.

graphql_compiler/compiler/metadata.py
```diff
@@ -154,11 +161,20 @@ def get_location_info(self, location):
...
+
+    def get_tag_info(self, tag_name):
+        return self._tags.get(tag_name, None)
```
obi1kenobi: Please add a docstring here.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -462,13 +465,19 @@ def test_tag_directive_constraints(self):
...
+                }
             }
         }''')
 
         tag_with_illegal_name = (GraphQLCompilationError, '''{
             Animal @filter(op_name: "name_or_alias", value: ["$animal_name"]) {
                 name @tag(tag_name: "name'\\\\\\"")
                 uuid @output(out_name: "uuid")
+                in_Animal_ParentOf {
+                    name @filter(op_name: "!=", value: ["%name'\\\\\\""])
```
obi1kenobi: This doesn't look quite right.

graphql_compiler/tests/test_ir_generation_errors.py
```diff
@@ -487,6 +499,15 @@ def test_tag_directive_constraints(self):
...
                 }
+                in_Animal_ParentOf {
+                   _x_count @filter(op_name: "=", value: ["%count"])
+                }
+            }
+        }''')
+
+        unused_tag = (GraphQLCompilationError, '''{
+            Animal {
+                name @tag(tag_name: "name")
```
obi1kenobi: This query might also fail because nothing is output, let's make sure that it has at least one output so that we have a better sense of where the compilation error is coming from.

graphql_compiler/compiler/compiler_frontend.py
pmantica1: Also can you explain what args mean? I think Predrag already mentioned a way to address this issue before.

PR #225, opened 2019-03-20
Add separate lockfiles for py2/py3, and new flake8 plugins for better linting

obi1kenobi:
Add:
- `flake8-print`, a linter plugin for catching leftover `print` statements in code.
- `flake8-quotes`, a linter plugin for catching use of inappropriate quotes in code (wrong quotes for docstrings etc.)
- Separate lockfiles for Python 2 and Python 3, since the dev dependencies for the two Python versions are mutually incompatible.
- A script for auto-generating both sets of lockfiles, since it's a multi-step process.
- A script for auto-installing a Python 2 virtualenv with `pipenv`, since by default a Python 3 virtualenv will be created.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,10 @@
 Thank you for taking the time to contribute to this project!
 
 To get started, make sure that you have `pipenv`, `docker` and `docker-compose` installed
-on your computer.
+on your computer. Additionally, since this is a package that supports both Python 2 and Python 3,
+please make sure you have Python 2.7.15+ and Python 3.6+ installed locally. This project assumes
+that they are available on the system as `python2` and `python3`, respectively. If you do not
+already have them installed, consider doing so using [pyenv](https://github.com/pyenv/pyenv).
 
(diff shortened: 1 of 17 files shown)


CONTRIBUTING.md
```diff
@@ -58,10 +61,34 @@ of both the contributors and the project.
...
+If you need to set up a Python 2 virtualenv locally, simply run the following script:
+```
+./scripts/make_py2_venv.sh
+```
+
+If you change the Pipfile or the package requirements, please make sure to regenerate the
+lockfiles for both Python versions. The easiest way to do so is with the following script:
+```
+./scripts/make_pipenv_lockfiles.sh
+```
```
bojanserafimov: What kind of error will developers get when they forget to do this?

obi1kenobi: We install dependencies for tests using `pipenv install --dev --deploy --system (see the `.travis.yml` file), which promises to assert that the lockfile and Pipfile match up:

```
--system                 System pip management.  [env var: PIPENV_SYSTEM]
--deploy                 Abort if the Pipfile.lock is out-of-date, or Python
                           version is wrong.
```

PR #228, opened 2019-03-21
Added myself as an author

pmantica1:


obi1kenobi:
@bojanserafimov are you going to add yourself as an author, or do you prefer to stay under the radar? :)

PR #229, opened 2019-03-22
Include location type in ContextField and related expressions.

obi1kenobi:
This should make emitting correct SQL a bit easier, as it brings `ContextField` expressions a bit closer to `OutputContextField` expressions in terms of the information they carry. In a subsequent PR, we can add location types for `LocalField` expressions as well.


graphql_compiler/tests/test_emit_output.py
```diff
@@ -94,74 +92,6 @@ def test_simple_traverse_filter_output(self):
         received_match = emit_match.emit_code_from_ir(compound_match_query, None)
         compare_match(self, expected_match, received_match)
 
-    def test_output_inside_optional_traversal(self):
```
obi1kenobi: I removed this test because of a couple of reasons:
- The test tests functionality that is already covered in `test_compiler.py`.
- Unfortunately, even though it aims to test a narrow bit of functionality (emitting MATCH code), it depends on a vast portion of the compiler: the post-lowering IR, conversion to `MatchQuery` and then `CompoundMatchQuery` object, and, had I kept it, would have also depended on the `QueryMetadataTable`.
- Also unfortunately, the test was not written using the same schema as the other tests, which meant that adding the `QueryMetadataTable` required creating real GraphQL types for this new schema, etc.

All of this put together meant that this was not an effective test anymore, and was not worth the amount of effort it would have taken to turn it into one.

graphql_compiler/tests/test_helpers.py
```diff
@@ -26,7 +26,7 @@
 # without requiring a massive number of types and interfaces.
 SCHEMA_TEXT = '''
     schema {
-      query: RootSchemaQuery
+        query: RootSchemaQuery
```
obi1kenobi: The schema text was 2-space indented, which caused editors such as Sublime to try to make the entire file 2-space indented. This is a simple quality-of-life fix, with no functional changes.

graphql_compiler/compiler/compiler_entities.py
```diff
@@ -43,9 +44,23 @@ def __repr__(self):
     # pylint: disable=protected-access
     def __eq__(self, other):
         """Return True if the CompilerEntity objects are equal, and False otherwise."""
-        return (type(self) == type(other) and
-                self._print_args == other._print_args and
-                self._print_kwargs == other._print_kwargs)
+        if not type(self) == type(other):
```
jmeulemans: Consider a `!=` here, the `not ... ==` looks a little strange to my eye.

graphql_compiler/compiler/ir_lowering_common.py
```diff
@@ -53,30 +53,34 @@ def to_match(self):
...
-            ContextField(expression.location),
+            ContextField(expression.location, location_type),
             NullLiteral)
 
     def construct_result_visitor_fn(expression):
         """Expression visitor function that rewrites ContextFieldExistence expressions."""
         if not isinstance(expression, ContextFieldExistence):
             return expression
 
+        location_type = query_metadata_table.get_location_info(expression.location).type
```
jmeulemans: Would it be worth making this type grab a method proper on the `query_metadata_table`? It's used at least 3 times that I see.

obi1kenobi: After thinking about it for a bit, I think saving a `.type` at the end of the existing function call is not worth defining a new function call on `query_metadata_table` -- it would expand the contract of the metadata table and mean we have to test a larger surface area, for comparatively little benefit. Let me know if you think I'm missing something or if you feel strongly about it.

graphql_compiler/compiler/compiler_entities.py
jmeulemans: Where did you land on this?

obi1kenobi: I think I just failed to push my local changes up, sorry about that.

obi1kenobi: Pushed now.

obi1kenobi:
I'd like to get this merged before too long, since it's on the long-ish side as far as PRs go.

PR #231, opened 2019-03-25
Added type equivalence dict test

pmantica1:


graphql_compiler/tests/test_schema_generation.py
```diff
@@ -188,3 +200,29 @@ def test_parsed_class_fields(self):
         person_lives_in_edge = schema_graph.get_element_by_class_name('Person_LivesIn')
         self.assertEqual(PERSON_LIVES_IN_EDGE_SCHEMA_DATA['customFields'],
                          person_lives_in_edge.class_fields)
+
+    def test_type_equivalence_dicts(self):
+        schema_data = [BASE_VERTEX_SCHEMA_DATA, ENTITY_SCHEMA_DATA,
+                       BASE_EDGE_SCHEMA_DATA, PERSON_SCHEMA_DATA, BABY_SCHEMA_DATA,
+                       BASE_EDGE_SCHEMA_DATA, LOCATION_SCHEMA_DATA,
```
obi1kenobi: `BASE_EDGE_SCHEMA_DATA` seems to appear twice here, that seems problematic.

graphql_compiler/tests/test_schema_generation.py
```diff
@@ -181,10 +221,46 @@ def test_link_property(self):
...
+            BASE_EDGE_SCHEMA_DATA,
+            BASE_VERTEX_SCHEMA_DATA,
+            BABY_SCHEMA_DATA,
+            ENTITY_SCHEMA_DATA,
+            LOCATION_SCHEMA_DATA,
+            PERSON_LIVES_IN_EDGE_SCHEMA_DATA,
+            PERSON_SCHEMA_DATA,
+        ]
+        schema, type_equivalence_dicts = get_graphql_schema_from_orientdb_schema_data(schema_data)
+        person, person_baby_union = next(six.iteritems(type_equivalence_dicts))
```
obi1kenobi: Doesn't this implicitly depend on ordering of a dict, which is undefined except for one-element dicts? Since we know the key we're looking for, why not just look it up and assert that it exists?

Also, since we're trying to test `type_equivalence_hints`, shouldn't we ensure that the detected hints are exactly as we expect? Right now, we're checking a single key, but we don't e.g. make sure that's the only key in there (assuming that's what we expected).

graphql_compiler/tests/test_schema_generation.py
```diff
@@ -181,10 +221,46 @@ def test_link_property(self):
...
+        baby = schema.get_type('Baby')
+        location = schema.get_type('Location')
+
+        # Assert that the Person class is part of the schema.
+        self.assertEqual(person, schema.get_type('Person'))
+
+        # Assert that the union consists of the Baby and Person classes
+        self.assertEqual(person_baby_union.types, [baby, person])
```
obi1kenobi: Are we sure that the order of types in the union is always going to be this, and not the other way around? Would it be better to check this with set equality instead?

PR #232, opened 2019-03-26
Implement schema with macros

bojanserafimov:
1. make `validate_schema_and_ast` public so we can validate before macro expansion
2. implement and better specify `get_schema_with_macros`
3. use schema validation before macro expansion
4. add tests

### DIFF
--- a/graphql_compiler/macros/__init__.py
+++ b/graphql_compiler/macros/__init__.py
@@ -10,48 +10,61 @@ from .macro_expansion import expand_macros_in_query_ast
 
 MacroRegistry = namedtuple(
     'MacroRegistry', (
-        'macro_edges',  # Dict[str, Dict[str, MacroEdgeDescriptor]] mapping:
-                        # class name -> (macro edge name -> MacroEdgeDescriptor)
+        # GraphQLSchema, created using the GraphQL library
+        'schema_without_macros',
+
+        # Optional dict of GraphQL interface or type -> GraphQL union.
+        # Used as a workaround for GraphQL's lack of support for
+        # inheritance across "types" (i.e. non-interfaces), as well as a
+        # workaround for Gremlin's total lack of inheritance-awareness.
+        # The key-value pairs in the dict specify that the "key" type
+        # is equivalent to the "value" type, i.e. that the GraphQL type or
+        # interface in the key is the most-derived common supertype
+        # of every GraphQL type in the "value" GraphQL union.
+        # Recursive expansion of type equivalence hints is not performed,
+        # and only type-level correctness of this argument is enforced.
+        # See README.md for more details on everything this parameter does.
+        # *****
+        # Be very careful with this option, as bad input here will
+        # lead to incorrect output queries being generated.
+        # *****
+        'type_equivalence_hints',
+
+        # Optional dict mapping class names to the set of its subclass names.
+        # A class in this context means a GraphQL object type or interface.
+        'subclass_sets',
+
+        # Dict[str, Dict[str, MacroEdgeDescriptor]] mapping:
+        # class name -> (macro edge name -> MacroEdgeDescriptor)
+        'macro_edges',
+
         # Any other macro types we may add in the future go here.
     )
 )
 
 
-def create_macro_registry():
+def create_macro_registry(schema, type_equivalence_hints=None, subclass_sets=None):
     """Create and return a new empty macro registry."""
-    return MacroRegistry(macro_edges=dict())
+    return MacroRegistry(
+        schema_without_macros=schema,
+        type_equivalence_hints=type_equivalence_hints,
+        subclass_sets=subclass_sets,
+        macro_edges=dict())
 
 
-def register_macro_edge(macro_registry, schema, macro_edge_graphql, macro_edge_args,
-                        type_equivalence_hints=None):
+def register_macro_edge(macro_registry, macro_edge_graphql, macro_edge_args):
     """Add the new macro edge descriptor to the provided MacroRegistry object, mutating it.
 
     Args:
         macro_registry: MacroRegistry object containing macro descriptors, where the new
                         macro edge descriptor should be added.
-        schema: GraphQL schema object, created using the GraphQL library
         macro_edge_graphql: string, GraphQL defining how the new macro edge should be expanded
         macro_edge_args: dict mapping strings to any type, containing any arguments the macro edge
                          requires in order to function.
-        type_equivalence_hints: optional dict of GraphQL interface or type -> GraphQL union.
-                                Used as a workaround for GraphQL's lack of support for
-                                inheritance across "types" (i.e. non-interfaces), as well as a
-                                workaround for Gremlin's total lack of inheritance-awareness.
-                                The key-value pairs in the dict specify that the "key" type
-                                is equivalent to the "value" type, i.e. that the GraphQL type or
-                                interface in the key is the most-derived common supertype
-                                of every GraphQL type in the "value" GraphQL union.
-                                Recursive expansion of type equivalence hints is not performed,
-                                and only type-level correctness of this argument is enforced.
-                                See README.md for more details on everything this parameter does.
-                                *****
-                                Be very careful with this option, as bad input here will
-                                lead to incorrect output queries being generated.
-                                *****
     """
     class_name, macro_edge_name, macro_descriptor = make_macro_edge_descriptor(
-        schema, macro_edge_graphql, macro_edge_args,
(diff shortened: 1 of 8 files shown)


graphql_compiler/macros/__init__.py
```diff
@@ -88,14 +97,23 @@ def register_macro_edge(macro_registry, macro_edge_graphql, macro_edge_args):
 def get_schema_with_macros(macro_registry):
     """Get a new GraphQLSchema with fields where macro edges can be used.
 
-    Specification:
+    Preconditions:
+    1. No macro in the registry has the same name as a field on the vertex where it applies,
+       or any interfaces it implements.
```
pmantica1: This is redundant info. The fields of a vertex is a superset of the fields of the interface it implements.

graphql_compiler/macros/__init__.py
```diff
@@ -88,14 +97,23 @@ def register_macro_edge(macro_registry, macro_edge_graphql, macro_edge_args):
...
+    2. Members of a union type do not have outgoing macros with the same name.
```
pmantica1: nit: an union

graphql_compiler/macros/__init__.py
```diff
@@ -104,7 +122,25 @@ def get_schema_with_macros(macro_registry):
...
+                        macro_edge_descriptor.expansion_ast)
+                    arguments = []
+                    definition.fields.append(FieldDefinition(
+                        Name(macro_edge_name), arguments,
+                        NamedType(Name(type_at_target.name)), directives=[]))
+    return build_ast_schema(schema_ast)
+
+
+def get_schema_for_macro_definition(schema):
+    """Returns a schema with macro directives."""
```
pmantica1: Can you elaborate more on the spec of this function? I don't see the link between the docstring and the function name.

graphql_compiler/macros/__init__.py
```diff
@@ -121,6 +157,10 @@ def perform_macro_expansion(macro_registry, graphql_with_macro, graphql_args):
         the returned values are guaranteed to be identical to the input query and args.
     """
     query_ast = safe_parse_graphql(graphql_with_macro)
+    schema_with_macros = get_schema_with_macros(macro_registry)
+    validation_errors = validate_schema_and_ast(schema_with_macros, query_ast)
+    if validation_errors:
+        raise GraphQLValidationError(u'String does not validate: {}'.format(validation_errors))
```
pmantica1: This pattern is used multiple time in the code, (if validation_errors ...). I think it might be best to put this in validate_schema_and_ast. Up to you though.

graphql_compiler/tests/test_macro_expansion_errors.py
```diff
@@ -106,3 +104,16 @@ def test_macro_edge_target_on_union_type(self):
...
+            Animal {
+                out_Animal_GrandparentOf {
+                    names @output(out_name: "grandkid")
+                }
+            }
+        }'''
+        args = {}
+
+        with self.assertRaises(GraphQLValidationError):
+            perform_macro_expansion(self.macro_registry, query, args)
```
pmantica1: Please comment why the schema usage is incorrect.

graphql_compiler/tests/test_macro_schema.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def test_get_schema_with_macros_original_schema_unchanged(self):
+        empty_macro_registry = get_empty_test_macro_registry()
+        original_printed_schema = print_schema(self.macro_registry.schema_without_macros)
+        printed_schema_with_0_macros = print_schema(get_schema_with_macros(empty_macro_registry))
+        printed_schema_afterwards = print_schema(self.macro_registry.schema_without_macros)
+        self.assertEqual(original_printed_schema, printed_schema_afterwards)
+        self.assertEqual(original_printed_schema, printed_schema_with_0_macros)
+
+    def test_get_schema_with_macros_basic(self):
```
pmantica1: Can you a slightly more complex test? For this test the object, parent and grandparent are all of the same type so the test will pass if the type is of any of the three.

graphql_compiler/macros/__init__.py
obi1kenobi: Agreed that it's redundant, but that's not necessarily a bad thing. We're going for making the code super clear and easy to understand, not for making the info fully normalized.

graphql_compiler/macros/__init__.py
obi1kenobi: I believe that `a union` might actually be correct, though it's definitely tricky and I had to look it up to be sure: https://english.stackexchange.com/questions/266309/why-is-union-an-exception-to-the-a-an-rule

graphql_compiler/macros/__init__.py
```diff
@@ -88,14 +97,23 @@ def register_macro_edge(macro_registry, macro_edge_graphql, macro_edge_args):
...
+    Postconditions:
     1. Every GraphQLQuery that uses macros from this registry appropriately should
        successfully type-check against the schema generated from this function.
     2. A GraphQLQuery that uses macros not present in the registry, or uses valid
        macros but on types they are not defined at should fail schema validation with
        the schema generated from this function.
-    3. This function is total -- A macro registry with all macros inserted by using
-       register_macro_edge should not fail to create a GraphQL schema with macros.
+    3. This function is total -- A valid macro registry should not fail to create a
+       graphql schema with macros.
```
obi1kenobi: `graphql` -> `GraphQL`

graphql_compiler/macros/__init__.py
```diff
@@ -104,7 +122,25 @@ def get_schema_with_macros(macro_registry):
     Returns:
         GraphQLSchema with additional fields where macroe edges can be used.
     """
-    raise NotImplementedError()  # TODO(bojanserafimov) implement
+    schema_ast = parse(print_schema(macro_registry.schema_without_macros))
```
obi1kenobi: I'd add a comment explaining what this line is doing. If we hadn't discussed it in person yesterday, I would have been very confused right now.

graphql_compiler/macros/__init__.py
```diff
@@ -104,7 +122,25 @@ def get_schema_with_macros(macro_registry):
...
+            if definition.name.value in macro_registry.macro_edges:
+                macro_edges = macro_registry.macro_edges[definition.name.value]
+                for macro_edge_name, macro_edge_descriptor in six.iteritems(macro_edges):
+                    type_at_target = get_type_at_macro_edge_target(
+                        macro_registry.schema_without_macros,
+                        macro_edge_descriptor.expansion_ast)
+                    arguments = []
+                    definition.fields.append(FieldDefinition(
+                        Name(macro_edge_name), arguments,
+                        NamedType(Name(type_at_target.name)), directives=[]))
```
obi1kenobi: I think for this PR this is fine, but before we merge this into `master` I think we should add a `@macro_edge` directive here, to signal that the edge is from a macro. This is important since macro edges have limitations on e.g. recursion and optionals, and the user should have a way to know that ahead of time, instead of only finding it out once they write a query and it fails to compile.

graphql_compiler/macros/__init__.py
bojanserafimov: why?

graphql_compiler/macros/__init__.py
obi1kenobi: The tricky thing is that if `validate_schema_and_ast` raises the error, it will be unable to include the GraphQL  string that was passed in -- since it's analyzing an AST rather than the string itself. I think there are few enough places where this pattern is used that maybe it's okay that we have to raise the error outside the function that detects it.

I do have one nitpick observation to make, though. Imagine if this error bubbled up from some UI: `String does not validate`. Would you think it was a bug in the UI, or in the thing you wrote? I'd say it's probably a 50-50. How about we make it `The provided GraphQL input is not valid:` instead?

bojanserafimov: It's only used twice. And we might want to do something else with the errors, other than raising. For example, you can write tests that not only assert that something is invalid, but that it's invalid for the right reason. Or maybe return all errors to the user, instead of just the first one.

graphql_compiler/macros/__init__.py
bojanserafimov: Will do in the next PR

graphql_compiler/tests/test_macro_schema.py
obi1kenobi: @pmantica1 thoughts on addressing this in a subsequent PR, so we can merge this, upload a new pre-release, and unblock work in other projects while we improve the test suite here?

PR #234, opened 2019-03-29
Fix missing union types bug

pmantica1:


obi1kenobi:
Add a test that the autogenerated schema contains all the expected union types?

graphql_compiler/tests/test_helpers.py
```diff
@@ -167,6 +167,8 @@
       FoodOrSpecies: FoodOrSpecies
       Location: Location
       Species: Species
+      Union__BirthEvent__Event__FeedingEvent: Union__BirthEvent__Event__FeedingEvent
```
obi1kenobi: I don't think unions should be part of the `RootSchemaQuery`. I don't think starting a query at a union is something that should be possible -- users should be able to simply start querying at the appropriate base type of the union.

PR #235, opened 2019-03-29
Fix macro on superclass

bojanserafimov:


graphql_compiler/macros/__init__.py
```diff
@@ -127,20 +128,22 @@ def get_schema_with_macros(macro_registry):
...
+        for macro_edge_name, macro_edge_descriptor in six.iteritems(macros_for_base_class):
+            type_at_target = get_type_at_macro_edge_target(
+                macro_registry.schema_without_macros,
+                macro_edge_descriptor.expansion_ast)
+            list_type_at_target = ListType(NamedType(Name(type_at_target.name)))
+            arguments = []
+            directives = [Directive(Name(MacroEdgeDirective.name))]
+            for subclass in macro_registry.subclass_sets[macro_base_class_name]:
+                definitions_by_name[subclass].fields.append(FieldDefinition(
+                    Name(macro_edge_name), arguments, list_type_at_target, directives=directives))
```
obi1kenobi: suggestion: I feel like a blank line before the `return` would help readability here

graphql_compiler/macros/macro_expansion.py
```diff
@@ -246,7 +250,7 @@ def _expand_macros_in_inner_ast(macro_registry, current_schema_type, ast, query_
         # No macro expansion happens at this level if there are no selections.
         return ast, query_args
 
-    macro_edges_at_this_type = macro_registry.macro_edges.get(current_schema_type.name, dict())
+    macro_edges_at_this_type = inherited_macro_edges.get(current_schema_type.name, dict())
```
obi1kenobi: This looks like just a denormalization of the `macro_registry.macro_edges`. Any reason not to store it in the `macro_registry` itself? That way we know what was created from where, and we don't have to pass extra args.

Also, that would mean that we pay the denormalization cost once per macro edge definition at initialization, and not once per query on each run.

bojanserafimov: I tried to be consistent with how we treat `schema_with_macros`. If we want to denormalize, I would denormalize both, in a separate PR.

obi1kenobi: Sure, let's do that. Separate PR is fine.

PR #237, opened 2019-03-29
Remove type equivalences with no outgoing or incoming edges 

pmantica1:


graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -21,6 +21,19 @@
 )
 
 
+def _get_referenced_type_equivances(graphql_types, type_equivalence_hints):
```
obi1kenobi: `equivances` seems like a typo

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -21,6 +21,19 @@
...
+    """Filter union types with no edges from the type equivalence hints dict."""
+    referenced_types = set()
+    for graphql_type in graphql_types.values():
+        if isinstance(graphql_type, (GraphQLObjectType, GraphQLInterfaceType)):
+            for _, field in graphql_type.fields.items():
+                if isinstance(field.type, GraphQLList):
+                    referenced_types.add(field.type.of_type.name)
+    return {original: union
```
obi1kenobi: Style consistency suggestion: we generally prefer this style for multi-line comprehensions:
```
foo = {
    key: value
    for key, value in stuff
    if key > 0
}
```

PR #239, opened 2019-03-31
Add Quick Overview to README.md

pmantica1:


README.md
```diff
@@ -14,15 +14,21 @@ Turn complex GraphQL queries into optimized database queries.
...
+
+Furthermore, the GraphQL compiler validates queries through the use of a GraphQLSchema object 
+that specifies the underlying schema of the database. We can currently autogenerate the 
+GraphQLSchema by querying an OrientDB database, (see  [Querying OrientDB with GraphQL](#querying-orientdb-with-graphql)). In the nearby future, we plan to add 
+schema autogeneration from SQLAlchemy metadata. 
+
 For a more detailed overview and getting started guide, please see
 [our blog post](https://blog.kensho.com/compiled-graphql-as-a-database-query-language-72e106844282).
 
-To pretty-print GraphQL queries, use the included pretty-printer:
```
obi1kenobi: Why remove this? I believe this is the only mention of this capability of the tool, so if we remove this mention without adding it elsewhere, users will be entirely unable to find it.

README.md
```diff
@@ -14,15 +14,21 @@ Turn complex GraphQL queries into optimized database queries.
 pip install graphql-compiler
 ```
 
+## Quick Overview 
+
+The GraphQL compiler turns read-only queries written in GraphQL syntax to different
```
obi1kenobi: Nice summary! What do you think about flipping the order of these two sentences? That way, we start with the answer to "why should I care?" rather than "how does it work?", which might be a bit more user-friendly.

README.md
```diff
@@ -14,15 +14,21 @@ Turn complex GraphQL queries into optimized database queries.
...
 ```
 
+## Quick Overview 
+
+The GraphQL compiler turns read-only queries written in GraphQL syntax to different 
+query languages. Through the GraphQL compiler, users can write powerful queries that uncover 
+deep relationships in the data while not having to worry about the underlying database query 
+language.
+
+Furthermore, the GraphQL compiler validates queries through the use of a GraphQLSchema object
```
obi1kenobi: `GraphQLSchema object` uses the internal name of the type i.e. an implementation detail. Let's refer to it everywhere as `GraphQL schema` instead.

README.md
```diff
@@ -14,15 +14,21 @@ Turn complex GraphQL queries into optimized database queries.
...
+## Quick Overview 
+
+The GraphQL compiler turns read-only queries written in GraphQL syntax to different 
+query languages. Through the GraphQL compiler, users can write powerful queries that uncover 
+deep relationships in the data while not having to worry about the underlying database query 
+language.
+
+Furthermore, the GraphQL compiler validates queries through the use of a GraphQLSchema object 
+that specifies the underlying schema of the database. We can currently autogenerate the 
+GraphQLSchema by querying an OrientDB database, (see  [Querying OrientDB with GraphQL](#querying-orientdb-with-graphql)). In the nearby future, we plan to add
```
obi1kenobi: Suggestion: `introspecting` instead of `querying`. Also:
- There are two spaces before the `[` of the link.
- How about putting the next sentence on its own line, to keep the file's line length low, and replacing `nearby` with `near`?

README.md
```diff
@@ -14,15 +14,21 @@ Turn complex GraphQL queries into optimized database queries.
...
+
+The GraphQL compiler turns read-only queries written in GraphQL syntax to different 
+query languages. Through the GraphQL compiler, users can write powerful queries that uncover 
+deep relationships in the data while not having to worry about the underlying database query 
+language.
+
+Furthermore, the GraphQL compiler validates queries through the use of a GraphQLSchema object 
+that specifies the underlying schema of the database. We can currently autogenerate the 
+GraphQLSchema by querying an OrientDB database, (see  [Querying OrientDB with GraphQL](#querying-orientdb-with-graphql)). In the nearby future, we plan to add 
+schema autogeneration from SQLAlchemy metadata.
```
obi1kenobi: Suggestion: add `as well` to the end of the sentence?

README.md
pmantica1: I don't think that the feature is important enough to have at the top of the Github repo. Also, it's a bit vague also what it means to pretty-print a GraphQL query. I am also not sure why this one of the first things I would like to learn about the compiler.

README.md
```diff
@@ -55,6 +62,7 @@ It's modeled after Python's `json.tool`, reading from stdin and writing to stdou
      * [End-To-End SQL Example](#end-to-end-sql-example)
      * [Configuring the SQL Database to Match the GraphQL Schema](#configuring-the-sql-database-to-match-the-graphql-schema)
   * [Miscellaneous](#miscellaneous)
+     * [Pretty Printing GraphQL Queries](#pretty-printing-graphql-queries)
```
obi1kenobi: Let's hyphenate `Pretty-Printing` since it should be one word (see line 1316 below, for example).

obi1kenobi: I used the online editor to address this myself.

PR #241, opened 2019-04-02
Remove links as properties

pmantica1:


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -488,19 +447,19 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
-                property_name = property_definition['name']
+            for orientdb_property_definition in chain.from_iterable(all_orientdb_property_lists):
+                property_name = orientdb_property_definition['name']
 
                 # The only properties we allow to be redefined are the in/out properties
                 # of edge classes. All other properties may only be defined once
                 # in the entire inheritance hierarchy of any schema class, of any kind.
                 duplication_allowed = all((
-                    property_name in allowed_duplicated_edge_property_names,
+                    property_name in set(links.keys()),
```
obi1kenobi: What is the type of `links` here? If it's a dict, why not just `property_name in links`?

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -124,14 +123,12 @@ def _get_fields_for_class(schema_graph, graphql_types, field_type_overrides, hid
     schema_element = schema_graph.get_element_by_class_name(cls_name)
     outbound_edges = (
         ('out_{}'.format(out_edge_name),
-         schema_graph.get_element_by_class_name(out_edge_name).properties[
-             EDGE_DESTINATION_PROPERTY_NAME].qualifier)
+         next(iter(schema_graph.get_element_by_class_name(out_edge_name).out_connections)))
```
obi1kenobi: This seems like it could benefit from an assertion that the element we're fetching is indeed unique. Try using the `get_only_element_from_collection` function, and consider moving it out of the `compiler` submodule and into some more generic location.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -181,6 +131,18 @@ def freeze(self):
         """Make the SchemaElement's connections immutable."""
         self.in_connections = frozenset(self.in_connections)
         self.out_connections = frozenset(self.out_connections)
+        self._validate_endpoints()
+
+    def _validate_endpoints(self):
+        if self._kind == SchemaElement.ELEMENT_KIND_EDGE:
+            if not self.abstract:
```
obi1kenobi: How about switching the order of the `if` and `else` here, and avoiding the `not` on this line? Too many "not"s make the code harder to read, and removing this one seems to be straightforward.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -618,29 +552,55 @@ def _create_descriptor_from_property_definition(self, class_name, property_defin
...
-            from_class_name = edge_element.properties[EDGE_SOURCE_PROPERTY_NAME].qualifier
-            to_class_name = edge_element.properties[EDGE_DESTINATION_PROPERTY_NAME].qualifier
+            if edge.abstract:
+                continue
+            elif not(len(edge.in_connections) == 1 and len(edge.out_connections) == 1):
+                raise AssertionError(u'Found a non-abstract edge class with an incorrect number '
+                                     u'of endpoints: {}'.format(edge))
 
-            edge_schema_element = self._elements[edge_class_name]
+            from_class_name = next(iter(edge.in_connections))
```
obi1kenobi: Let's still use the `get_only_element_from_collection` helper here, even though we did assert it above.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -618,29 +552,55 @@ def _create_descriptor_from_property_definition(self, class_name, property_defin
...
 
             # Link edge_class_name with to_class_name
             for to_class in self._subclass_sets[to_class_name]:
                 to_schema_element = self._elements[to_class]
-                edge_schema_element.out_connections.add(to_class)
                 to_schema_element.in_connections.add(edge_class_name)
+
+    def _validate_link(self, orientdb_property_definition, class_name,
+                       class_name_to_definition):
+        """Validates that the class is a vertex and that the link is correctly defined.."""
```
obi1kenobi: nit: two periods at the end of the docstring

PR #242, opened 2019-04-02
Remove links as properties

pmantica1:


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -509,84 +470,59 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
                                          u'more than once, this is not allowed!'
                                          .format(property_name, class_name))
 
-                property_descriptor = self._create_descriptor_from_property_definition(
-                    class_name, property_definition, class_name_to_definition)
-
-                if property_name in allowed_duplicated_edge_property_names:
-                    links[property_name].append(property_descriptor)
+                if property_name in set(links.keys()):
```
bojanserafimov: nit: no need to make a set, just do `property_name in links`

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -509,84 +470,59 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
                     property_name_to_descriptor[property_name] = property_descriptor
 
-            for property_name in allowed_duplicated_edge_property_names:
-                elements = {
-                    property_descriptor.qualifier
-                    for property_descriptor in links[property_name]
-                }
+            self._elements[class_name] = SchemaElement(class_name, kind, abstract,
+                                                       property_name_to_descriptor, class_fields)
+            for link_direction in links.keys():
```
bojanserafimov: nit: just iterate the keys and values of the dict, you're using the value anyway. This way I won't have to do mental gymnastics to figure out if `link_direction` is in `links` when you access it.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -509,84 +470,59 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
-                                                 .format(current_descriptor,
-                                                         property_descriptor, class_name))
-                        property_name_to_descriptor[property_name] = property_descriptor
-
-                if (property_name not in property_name_to_descriptor and not abstract and
-                        kind == SchemaElement.ELEMENT_KIND_EDGE):
+                for link in links[link_direction]:
+                    subclass_set = self._subclass_sets[link]
+                    if len(set(links[link_direction]).intersection(subclass_set)) == 1:
+                        if leaf_link:
```
bojanserafimov: nit: be explicit `leaf_link is None` so I don't need to check if a non-`None` value could be falsy.

graphql_compiler/schema_generation/schema_graph.py
bojanserafimov: Can you explain what this whole code block does? Where is `leaf_link` used?

graphql_compiler/compiler/emit_match.py
```diff
@@ -4,9 +4,11 @@
 
 import six
 
+from graphql_compiler.helpers import get_only_element_from_collection
```
obi1kenobi: When writing a python package, please always use relative imports (`from .helpers`) instead of absolute ones (`from graphql_compiler.helpers`) to reference the package itself. You can use `..` to go up a directory.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -509,84 +471,56 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
-                                         u'no such subclass-of-all-elements exists.'
-                                         .format(property_name, class_name))
-
             self._elements[class_name] = SchemaElement(class_name, kind, abstract,
                                                        property_name_to_descriptor, class_fields)
 
-    def _create_descriptor_from_property_definition(self, class_name, property_definition,
-                                                    class_name_to_definition):
+            if kind == SchemaElement.ELEMENT_KIND_EDGE:
+                in_leaf_endpoint = self._try_get_base_endpoint(class_name,
```
obi1kenobi: Consider starting the arguments to the function on a new line with one indent. The visual indent at the `(` like this produces lots of whitespace on the left, and these columns of code on the right that at least I find difficult to read.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -182,6 +133,17 @@ def freeze(self):
         self.in_connections = frozenset(self.in_connections)
         self.out_connections = frozenset(self.out_connections)
 
+    def validate_endpoints(self):
```
obi1kenobi: All the other validation functions are not defined on instances of objects, and are just defined at module level. Let's stick to that paradigm to avoid confusing the reader unless there is very strong reason not to do so.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -182,6 +133,17 @@ def freeze(self):
...
+        """Validate that the class has valid endpoints."""
+        if self._kind == SchemaElement.ELEMENT_KIND_EDGE:
```
obi1kenobi: Should it be an error to call this function on a non-edge element?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -509,84 +471,56 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
+                                                                link_direction_to_endpoint_classes,
+                                                                abstract)
+                edge = self._elements[class_name]
+                # Either of the endpoints may not be defined if the edge is abstract.
+                if in_leaf_endpoint is not None:
+                    edge.in_connections.add(in_leaf_endpoint)
+                if out_leaf_endpoint is not None:
+                    edge.out_connections.add(out_leaf_endpoint)
+                edge.validate_endpoints()
+                edge.freeze()
```
obi1kenobi: Honestly, I'm pretty lost in this function, since it's ~80 lines of fairly complex and intricate code. Try breaking it up into a few smaller ones for the sake of maintainability.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -509,84 +471,56 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
+                                                               EDGE_SOURCE_PROPERTY_NAME,
+                                                               link_direction_to_endpoint_classes,
+                                                               abstract)
+                out_leaf_endpoint = self._try_get_base_endpoint(class_name,
+                                                                EDGE_DESTINATION_PROPERTY_NAME,
+                                                                link_direction_to_endpoint_classes,
+                                                                abstract)
+                edge = self._elements[class_name]
+                # Either of the endpoints may not be defined if the edge is abstract.
+                if in_leaf_endpoint is not None:
```
obi1kenobi: We have `leaf_endpoint` here and `base_endpoint` on the function that produced this value. Let's pick one term and stick with it everywhere, otherwise everything will get very confusing very quickly.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,187 +430,231 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
+               classes that can appear on that end of the edge class. The link
+               direction can be one of 'in' or 'out'.
+        abstract: bool, describes whether the class is abstract.
+        subclass_sets: dict, string -> [string], mapping a class to its subclasses.
+
+    Returns:
+        optional string, describing the leaf link. We define the links of an edge's end to be the
+        set of classes that could appear in that end of the edge. We define the leaf link as the
+        link that is not inherited by any of the other links. Each edge's end can have at
+        most 1 leaf link. The ends of non-abstract edges must each have a leaf link.
```
bojanserafimov: I still think this is not clear.
1. The name leaf is inappropriate because subclass relationships don't form a tree.
2. I can't see how "the link that is not inherited by any of the other links" is useful in constructing the schema.
3. Can you justify some facts? Why are we guaranteed to find at most one leaf link? Why do the ends of non-abstract edges have to have a leaf link?

The way I understanding this by reading the code is we want to derive the end type of an edge based on the end types of its subclasses. Can we have the documentation (and function name) reflect that, and then as an implementation detail, mention how it's constructed?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,187 +430,231 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
+
+            # Set edge connections. (Abstract classes may have missing leaf links).
+            in_leaf_link = _try_get_leaf_link(
+                class_name, EDGE_SOURCE_PROPERTY_NAME, links, abstract, self._subclass_sets)
+            out_leaf_link = _try_get_leaf_link(
+                class_name, EDGE_DESTINATION_PROPERTY_NAME, links, abstract, self._subclass_sets)
+            if in_leaf_link is not None:
+                elem.in_connections.add(in_leaf_link)
+            if out_leaf_link is not None:
+                elem.out_connections.add(out_leaf_link)
```
bojanserafimov: Do we only need these leaf links when the edge is not abstract? Can we not call `_try_get_leaf_link` on abstract edges at all? That will make the function spec and the usage much simpler.

pmantica1: Nope we need for abstract classes too.

pmantica1: Actually we need it almost solely for abstract edges. The leaf links of a non-abstract edge are just the links that were defined in it.

graphql_compiler/helpers.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+import funcy
```
obi1kenobi: Try naming this file something different than `helpers` so we don't have ambiguous imports of `helpers` depending on how many levels up in the hierarchy one goes. How about `global_utils`?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,187 +429,224 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
-                raise AssertionError(u'Property "{}" is an embedded collection but has '
-                                     u'neither a linked class nor a linked type: '
-                                     u'{}'.format(name, property_definition))
+    def _set_up_non_graph_elements(self, class_name_to_definition):
+        """Set up non graph elements."""
+        kind = SchemaElement.ELEMENT_KIND_NON_GRAPH
+        for class_name in self.non_graph_class_names:
+            # Create non-graph SchemaElement.
+            inheritance_set = self._inheritance_sets[class_name]
+            elem, links = _get_schema_element_and_links(class_name, class_name_to_definition,
```
obi1kenobi: Is there a good reason for abbreviating `element` to `elem`? If not, let's not abbreviate.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,187 +429,224 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
+def _get_schema_element_and_links(class_name, class_name_to_definition, kind, inheritance_set,
+                                  non_graph_class_names):
+    """Return the schema element and its links. Only Edges have links."""
+    class_definition = class_name_to_definition[class_name]
+    class_fields = class_definition.get('customFields')
+    if class_fields is None:
+        # OrientDB likes to make empty collections be None instead.
+        # We convert this field back to an empty dict, for our general sanity.
+        class_fields = dict()
+    property_name_to_descriptor, links = (
```
obi1kenobi: nit: why the `(` on its own line, instead of putting `_get_class_properties_and_links(` on that line?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,187 +429,224 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
+            self._elements[class_name].freeze()
+
+
+def _get_schema_element_and_links(class_name, class_name_to_definition, kind, inheritance_set,
+                                  non_graph_class_names):
+    """Return the schema element and its links. Only Edges have links."""
+    class_definition = class_name_to_definition[class_name]
+    class_fields = class_definition.get('customFields')
+    if class_fields is None:
+        # OrientDB likes to make empty collections be None instead.
```
obi1kenobi: Shouldn't this file be oblivious of specific backends' implementations and idiosyncrasies?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,187 +429,224 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
+        _get_class_properties_and_links(class_name, class_name_to_definition, inheritance_set,
+                                        non_graph_class_names))
+    elem = SchemaElement(class_name, kind, _is_abstract(class_definition),
+                         property_name_to_descriptor, class_fields)
+    return elem, links
+
+
+def _get_class_properties_and_links(class_name, class_name_to_definition, inheritance_set,
+                                    non_graph_class_names):
+    """Return the properties descriptors and links of the class. Only Edges have links."""
```
obi1kenobi: typo: `properties descriptors`. Also, `Edges` being capitalized makes me feel like it's a type, when I don't believe it is.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -124,14 +124,14 @@ def _get_fields_for_class(schema_graph, graphql_types, field_type_overrides, hid
     schema_element = schema_graph.get_element_by_class_name(cls_name)
     outbound_edges = (
         ('out_{}'.format(out_edge_name),
-         schema_graph.get_element_by_class_name(out_edge_name).properties[
-             EDGE_DESTINATION_PROPERTY_NAME].qualifier)
+         get_only_element_from_collection(schema_graph.get_element_by_class_name(
```
obi1kenobi: I disagree with this change. I think `in_connections` and `out_connections` should contain _all_ types that a given type could be connected to, including the subclasses of the type specified by the schema.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Yes. I'll make that refactor in a later branch when I separate the OrientDB constructor into its own file.

PR #249, opened 2019-05-01
Create SchemaElement hierarchy

pmantica1:


graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -122,45 +122,47 @@ def _get_fields_for_class(schema_graph, graphql_types, field_type_overrides, hid
...
-        else:
-            if to_type_name not in hidden_classes:
-                edge_endpoint_type_name = to_type_name
-
-        if edge_endpoint_type_name is not None:
-            # If we decided to not hide this edge due to its endpoint type being non-representable,
-            # represent the edge field as the GraphQL type List(edge_endpoint_type_name).
-            result[field_name] = GraphQLList(graphql_types[edge_endpoint_type_name])
+    if not schema_element.is_non_graph:
+        outbound_edges = (
```
pmantica1: The change below is essentially just shifting everything 4 spaces to the right and dealing with the 100 character limit.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -89,18 +90,16 @@ def get_superclasses_from_class_definition(class_definition):
 
 
 class SchemaElement(object):
-    ELEMENT_KIND_VERTEX = 'vertex'
-    ELEMENT_KIND_EDGE = 'edge'
-    ELEMENT_KIND_NON_GRAPH = 'non-graph'
+    # pylint: disable=metaclass-assignment
+    __metaclass__ = ABCMeta
```
pmantica1: I assigned the metaclass to ABCMeta in order to make the class an abstract class in a python2/python3 compatible way.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -88,19 +89,15 @@ def get_superclasses_from_class_definition(class_definition):
     return []
 
 
+@six.add_metaclass(ABCMeta)
```
pmantica1: Tested that this works as expected for python2/python3. In other words, tested that we get an error if we try to instantiate a class with this decorator, but we don't if we instantiate a subclass that overrides its abstract methods.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -88,19 +89,15 @@ def get_superclasses_from_class_definition(class_definition):
...
 class SchemaElement(object):
-    ELEMENT_KIND_VERTEX = 'vertex'
-    ELEMENT_KIND_EDGE = 'edge'
-    ELEMENT_KIND_NON_GRAPH = 'non-graph'
 
-    def __init__(self, class_name, kind, abstract, properties, class_fields):
+    @abstractmethod
```
obi1kenobi: `__init__` is not abstract, it's implemented. This should get removed.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -122,45 +122,47 @@ def _get_fields_for_class(schema_graph, graphql_types, field_type_overrides, hid
...
+            ('out_{}'.format(out_edge_name),
+             schema_graph.get_element_by_class_name(out_edge_name).properties[
+                 EDGE_DESTINATION_PROPERTY_NAME].qualifier)
+            for out_edge_name in schema_element.out_connections
+        )
+        inbound_edges = (
+            ('in_{}'.format(in_edge_name),
+             schema_graph.get_element_by_class_name(in_edge_name).properties[
+                 EDGE_SOURCE_PROPERTY_NAME].qualifier)
```
obi1kenobi: Do we still want `qualifier` to exist as a field? How does this fit in with using the GraphQL type system for types? Assuming this is staying for now and then going away in a subsequent PR?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -165,34 +136,105 @@ def class_fields(self):
...
+        return isinstance(self, Edge)
 
     @property
     def is_non_graph(self):
         """Return True if the schema element represents a non-graph type, and False otherwise."""
-        return self._kind == SchemaElement.ELEMENT_KIND_NON_GRAPH
+        return isinstance(self, NonGraphElement)
+
+    @abstractmethod
+    def __str__(self):
```
obi1kenobi: Not sure why the linter didn't complain here, but these functions need docstrings. Also, look at the implementation of `CompilerEntity` for an example of how you can implement `__str__` and `__repr__` generically. You'll have to do a bit of magic with the constructor, and to get the names of args defined by subtypes' constructors you'll have to pass them in as kwargs to this constructor, but it can work and simplify your code overall.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -165,34 +136,105 @@ def class_fields(self):
...
+        self.in_connections = set()
+        self.out_connections = set()
 
     def freeze(self):
         """Make the SchemaElement's connections immutable."""
         self.in_connections = frozenset(self.in_connections)
         self.out_connections = frozenset(self.out_connections)
 
+
+class Vertex(GraphElement):
```
obi1kenobi: I'm a bit worried that `Vertex` and `Edge` might get confused with the actual data vertices and edges. Maybe `VertexType/EdgeType` or something like that?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -88,19 +89,15 @@ def get_superclasses_from_class_definition(class_definition):
     return []
 
 
+@six.add_metaclass(ABCMeta)
 class SchemaElement(object):
```
obi1kenobi: We probably want `@six.python_2_unicode_compatible` here as well, check out the docs for it.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -165,34 +136,105 @@ def class_fields(self):
...
+        #
+        # For edge classes:
+        #   in  = the tail side of the edge
+        #   out = the head / arrow side of the edge
+        #
+        # For non-graph classes, these properties are always empty sets.
+        self.in_connections = set()
+        self.out_connections = set()
 
     def freeze(self):
```
obi1kenobi: Let's define `freeze()` on `SchemaElement` as well, and make sure to always call the parent class' `freeze()` function.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -165,34 +136,105 @@ def class_fields(self):
...
+
+    __repr__ = __str__
+
+
+class NonGraphElement(SchemaElement):
+    def __init__(self, class_name, abstract, properties, class_fields):
+        super(NonGraphElement, self).__init__(class_name, abstract, properties, class_fields)
+
+        # Non-edges must not have properties like "in" or "out" defined, and
+        # must not have properties of type "Link".
```
obi1kenobi: Presuming the "type Link" stuff is going away in the next PR?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -537,13 +580,13 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
...
                 if (property_name not in property_name_to_descriptor and not abstract and
-                        kind == SchemaElement.ELEMENT_KIND_EDGE):
+                        kind == Edge):
                     raise AssertionError(u'For property "{}" of non-abstract edge class "{}", '
                                          u'no such subclass-of-all-elements exists.'
                                          .format(property_name, class_name))
 
-            self._elements[class_name] = SchemaElement(class_name, kind, abstract,
-                                                       property_name_to_descriptor, class_fields)
+            self._elements[class_name] = kind(class_name, abstract, property_name_to_descriptor,
```
obi1kenobi: If you are going to call it as a constructor, call it `element_cls` or something like that, to make it obvious that it's a type and it can be called as a constructor.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -537,13 +580,13 @@ def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_
                         property_name_to_descriptor[property_name] = property_descriptor
 
                 if (property_name not in property_name_to_descriptor and not abstract and
-                        kind == SchemaElement.ELEMENT_KIND_EDGE):
+                        kind == Edge):
```
obi1kenobi: It's almost never a good idea to compare types for equality. Use `issubclass` instead: https://docs.python.org/3/library/functions.html#issubclass

graphql_compiler/schema_generation/graphql_schema.py
pmantica1: Yes this is going away in a future PR. In order to remove the qualifier field, I first need to remove links as properties and make the SchemaGraph use GraphQL's type system.

pmantica1:
In this PR I intend to make the minimal changes necessary to separate each SchemaElement kind into its own class and to build the SchemaElement hierarchy.  In future PRs, I intend to remove links as properties, make the SchemaGraph use GraphQL's type system,  and remove the PropertyDescriptor's qualifier field.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -460,7 +496,7 @@ def _split_classes_by_kind(self, class_name_to_definition):
         self._edge_class_names = frozenset(self._edge_class_names)
         self._non_graph_class_names = frozenset(self._non_graph_class_names)
 
-    def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind, class_names):
+    def _set_up_schema_elements_of_kind(self, class_name_to_definition, kind_cls, class_names):
```
pmantica1: I am not satisfied with kind_cls, but I will heavily refactor  _set_up_schema_elements_of_kind entirely in a future commit anyways. It is too large of a function and a bit hard to understand. I might also split it up into _set_up_vertex_elements, _set_up_edge_elements and _set_up_graph_elements.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: https://stackoverflow.com/questions/44800659/python-abstract-class-init

graphql_compiler/schema_generation/schema_graph.py
bojanserafimov: I think that linter was in arc lint, not sure if we picked it up.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -88,19 +89,19 @@ def get_superclasses_from_class_definition(class_definition):
...
+@six.add_metaclass(ABCMeta)
 class SchemaElement(object):
-    ELEMENT_KIND_VERTEX = 'vertex'
-    ELEMENT_KIND_EDGE = 'edge'
-    ELEMENT_KIND_NON_GRAPH = 'non-graph'
 
-    def __init__(self, class_name, kind, abstract, properties, class_fields):
+    # Since an abstract class without any abstract methods can be instantiated, we make the
+    # init method abstract in order to make it impossible to instantiate this class.
+    # Inspiration came from: https://stackoverflow.com/questions/44800659/python-abstract-class.
```
obi1kenobi: The link is broken: `https://stackoverflow.com/questions/44800659/python-abstract-class` is wrong -- compare it with the PR comment below. It currently happens to redirect to the right place, but let's not depend on that and fix it instead. Also, `.` is a valid URL character, so avoid putting `.` at the end of links.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -165,32 +143,96 @@ def class_fields(self):
...
+        return isinstance(self, EdgeType)
 
     @property
     def is_non_graph(self):
         """Return True if the schema element represents a non-graph type, and False otherwise."""
-        return self._kind == SchemaElement.ELEMENT_KIND_NON_GRAPH
+        return isinstance(self, NonGraphElement)
+
+    def freeze(self):
+        """Do nothing."""
```
obi1kenobi: Explain what `freeze` is supposed to do. The `pass` makes it obvious that it doesn't happen here, but you should explain to implementers of subclasses what this method is intended to do, so they know when to override it with a custom implementation.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -165,32 +143,96 @@ def class_fields(self):
...
+    def __repr__(self):
+        """Return a human-readable str representation of the SchemaElement object."""
+        return self.__str__()
+
+
+class GraphElement(SchemaElement):
+
+    # Since an abstract class without any abstract methods can be instantiated, we make the
+    # init method abstract in order to make it impossible to instantiate this class.
+    # Inspiration came from: https://stackoverflow.com/questions/44800659/python-abstract-class.
```
obi1kenobi: Same broken link here too.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Fixed in "nits" pr.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Fixed in "nits" pr.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Fixed in "nits" pr.

PR #250, opened 2019-05-01
Decompose _setup_schema_elements_of_kind

pmantica1:
This PR intends to decompose _setup_schema_elements_of_kind into smaller functions. In a future PR I will decompose _get_element_properties into two functions: one addressing non-link properties and another addressing link-properties.

### DIFF
--- a/graphql_compiler/schema_generation/schema_graph.py
+++ b/graphql_compiler/schema_generation/schema_graph.py
@@ -509,13 +509,13 @@ class SchemaGraph(object):
             class_fields = _get_class_fields(class_definition)
             abstract = _is_abstract(class_definition)
             property_name_to_descriptor = self._get_element_properties(
-                abstract, class_name,class_name_to_definition, kind_cls)
+                abstract, class_name, class_name_to_definition, kind_cls)
 
             self._elements[class_name] = kind_cls(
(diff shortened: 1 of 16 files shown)


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -504,88 +504,69 @@ def _split_classes_by_kind(self, class_name_to_definition):
...
+                for property_descriptor in links[property_name]
+            }
+            # If there are multiple in/out properties, we choose to include the one that
+            # is a subclass of all the elements present in the in/out properties.
+            for property_descriptor in links[property_name]:
+                subclass_set = self._subclass_sets[property_descriptor.qualifier]
+                if len(elements.intersection(subclass_set)) == 1:
+                    current_descriptor = property_name_to_descriptor.get(property_name, None)
+                    if current_descriptor and current_descriptor != property_descriptor:
+                        raise AssertionError(u'There already exists property "{}" in addition '
```
obi1kenobi: The sentence in the error is somewhat confusing. I know it was there already, but we should probably try to clarify it anyway. Good error messages are one of the hallmarks of a mature software project.

pmantica1: I agree. I plan on moving all of this into its own function.  That will come in a future PR.

PR #251, opened 2019-05-01
Decomposed _get_element_properties into two functions

pmantica1:
This PR intends to do the minimum change required to decompose _get_element_properties into two functions: one addressing link properties and the other addressing non-link properties. In a future PR, I will break down _get_link_properties into smaller functions since its logic is a bit complex. My next PR will focus on changing _set_up_schema_elements_of_kind into a different function for each kind. Once that PR is approved, I will remove links as properties and store them as a field of edges instead.


graphql_compiler/global_utils.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+def partition(iterable, pred):
```
obi1kenobi: I believe we already have `funcy` as a dependency, let's use it. Be careful to use it in a python-version-agnostic way, it does some strange things if you don't specify py2 vs py3 semantics.

https://funcy.readthedocs.io/en/stable/seqs.html#split

PR #253, opened 2019-05-02
Fix _x_count and optional filter creating duplicate GlobalOperationsStart

bojanserafimov:


graphql_compiler/tests/test_compiler.py
```diff
@@ -660,6 +660,45 @@ def test_filter_in_optional_block(self):
...
+                    as: Species___1
+                }}.in('Animal_OfSpecies') {{
+                   where: ((name = {animal_name})),
+                   optional: true,
+                   as: Species__in_Animal_OfSpecies___1
+                }}
+                RETURN $matches
+            )
+            LET $Species___1___in_Species_Eats = Species___1.in("Species_Eats").asList()
+                WHERE (
```
obi1kenobi: De-indent the `WHERE` and the expression inside it, it is at the same level as the `LET` and `FROM` clauses.

graphql_compiler/tests/test_input_data.py
```diff
@@ -314,6 +314,34 @@ def filter_in_optional_block():
...
+    graphql_input = '''{
+        Species {
+            name @output(out_name: "species_name")
+
+            in_Animal_OfSpecies @optional {
+                 name @filter(op_name: "=", value: ["$animal_name"])
+            }
+
+            in_Species_Eats @fold {
+                 _x_count @filter(op_name: ">=", value: ["$predators"])
```
obi1kenobi: This appears to be a 5-space indent, instead of a 4-space indent. Same above.

PR #259, opened 2019-05-06
Refactor _try_get_base_connection

pmantica1:


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -569,16 +569,15 @@ def _try_get_base_connections(self, class_name, class_name_to_definition,
...
-                subclass_set = self._subclass_sets[linked_class]
-                if len(linked_classes.intersection(subclass_set)) == 1:
-                    base_direction_connection = base_connections.get(end_direction, None)
-                    if base_direction_connection and base_direction_connection != linked_class:
-                        raise AssertionError(u'There already exists class "{}" in addition '
-                                             u'to class "{}" which is a subclass of all '
-                                             u'{} properties for class "{}".'
-                                             .format(base_direction_connection,
-                                                     linked_class, end_direction, class_name))
+                inheritance_set = self._subclass_sets[linked_class]
```
bojanserafimov: why change the name to something more ambiguous?

pmantica1: This seems to be an uncaught bug. Let me add a test for it.

PR #263, opened 2019-05-07
Compute only direct subclasses

pmantica1:


pmantica1:
Yes we are breaking functionality. This PR was intended as a quick fix to deal with the O(n^3) runtime. The other two options that we have is to improve its performance by using 2D arrays instead of sets or integrate the SchemaGraph, which seems to be a bit more complex of a change. Since this fix seems unacceptable, I am going to attempt to integrate the SchemaGraph instead.

obi1kenobi:
Sounds good to me. Do you have a rough ETA on how long that will take?

PR #264, opened 2019-05-08
Replace compute_subclass_sets functionality through SchemaGraph 

pmantica1:


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -787,9 +786,6 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
         # 'type_equivalence_hints_inverse' is the inverse of type_equivalence_hints,
         # which is always invertible.
         'type_equivalence_hints_inverse': invert_dict(type_equivalence_hints),
-        # 'subclass_sets' is a dict mapping class names to the set of its subclass names
-        'subclass_sets': compute_subclass_sets(
```
pmantica1: There is no usage of 'subclass_sets' field in the entire codebase so I felt free to delete it. @bojanserafimov is there an use case I am missing?

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -171,8 +173,21 @@ def test_has_substring_precedence(self, backend_name):
...
 
+    @integration_fixtures
+    def test_snapshot_schema_data_from_orientdb_schema(self):
+        schema_data = ordered(generate_schema_data(self.graph_client))
+
+        # To update run:
+        # import json
+        # from ..test_helpers import SCHEMA_DATA_JSON_FILE
+        # with open(SCHEMA_DATA_JSON_FILE, 'w') as outfile:
+        #     json.dump(generate_schema_data(self.graph_client), outfile)
```
pmantica1: This seemed like the most convenient place to put this code.

graphql_compiler/tests/test_data_tools/orientdb_schema_data.json
```diff
@@ -0,0 +1 @@
...s": "Event", "superClasses": ["Event"], "customFields": null}, {"name": "Animal_OfSpecies", "shortName": null, "description": null, "defaultClusterId": 201, "clusterIds": [201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212], "clusterSelection": "round-robin", "overSize": 0.0, "strictMode": false, "abstract": false, "properties": [{"name": "in", "type": 13, "globalId": 26, "mandatory": false, "readonly": false, "notNull": false, "defaultValue": null, "min": null, "max": null, "linkedClass": "Species", "customFields": null, "collate": "default", "description": null}, {"name": "out", "type": 13, "globalId": 27, "mandatory": false, "readonly": false, "notNull": false, "defaultValue": null, "min": null, "max": null, "linkedClass": "Animal", "customFields": null, "collate": "default", "description": null}], "superClass": "E", "superClasses": ["E"], "customFields": null}, {"name": "V", "shortName": null, "description": null, "defaultClusterId": 9, "clusterIds": [9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20], "clusterSelection": "round-robin", "overSize": 0.0, "strictMode": false, "abstract": false, "properties": [], "superClass": null, "superClasses": null, "customFields": null}]
```
obi1kenobi: Pretty-print this, please. Otherwise it's unreadable and unusable.

graphql_compiler/tests/test_helpers.py
```diff
@@ -281,14 +284,35 @@ def get_schema():
...
 
 
+def get_schema_graph():
+    """Get a SchemaGraph object for testing."""
+    return SchemaGraph(get_schema_data())
+
+
+def get_schema_data():
+    """Return the OrientDB schema data for testing."""
+    with open(SCHEMA_DATA_JSON_FILE) as infile:
```
obi1kenobi: This opens a file relative to the current working directory, which may change and may break the tests as a result. Load the file relative to the current file's location instead, that way the test doesn't care about the current working directory.

graphql_compiler/tests/test_helpers.py
```diff
@@ -281,14 +284,35 @@ def get_schema():
     return schema
 
 
+def get_schema_graph():
```
obi1kenobi: Is this a specific OrientDB-based schema graph, or should it be interchangeable with the rest?

graphql_compiler/tests/test_helpers.py
```diff
@@ -281,14 +284,35 @@ def get_schema():
...
-    schema_data = [x.oRecordData for x in schema_records]
+    schema_graph = generate_schema_graph(graph_client)
     type_overrides = {
         'UniquelyIdentifiable': {'uuid': GraphQLID},
     }
-    return get_graphql_schema_from_orientdb_schema_data(schema_data, type_overrides)
+    return get_graphql_schema_from_schema_graph(schema_graph, type_overrides)
+
+
+def generate_schema_graph(graph_client):
```
obi1kenobi: Make the difference between this function and `get_schema_graph()` painfully clear, in both the name and the docstring. Otherwise, users won't know which one to use / which ones they _should_ use.

graphql_compiler/tests/test_helpers.py
```diff
@@ -407,3 +435,13 @@ def get_test_macro_registry():
     for graphql, args in valid_macros:
         register_macro_edge(macro_registry, graphql, args)
     return macro_registry
+
+
+def ordered(obj):
+    """Return an ordered representation of a json-serializable object"""
+    if isinstance(obj, dict):
```
obi1kenobi: Why do we need dicts to be ordered? Dicts (not `OrderedDict`, just vanilla ones) should compare equal if they contain the same keys and values, regardless of order. I think this part is probably not necessary.

graphql_compiler/tests/test_helpers.py
```diff
@@ -407,3 +435,13 @@ def get_test_macro_registry():
...
+        return OrderedDict(sorted([(k, ordered(v)) for k, v in obj.items()], key=str))
+    if isinstance(obj, list):
+        return sorted([ordered(x) for x in obj], key=str)
```
obi1kenobi: This is probably fine but somewhat risky. `str` _should_ be deterministic, but it is definitely getting called multiple times and so if it ever isn't deterministic, we're going to have a really rough time figuring out what went wrong.

graphql_compiler/compiler/compiler_frontend.py
obi1kenobi: I think the use might not have gotten added yet, since this is an unmerged branch :)

graphql_compiler/tests/integration_tests/test_backends_integration.py
obi1kenobi: This should be added as code and tested as well, otherwise it might drift. Bonus points if you add the pretty-printing step here, so the JSON file is always pretty-printed.

pmantica1:
Closing pull request until refactor-schema-graph branch is ready.

PR #265, opened 2019-05-08
Ignore unsupported properties in SchemaGraph.

pmantica1:
Instead of requiring all properties to have supported types, I ignore those that don't and issue a warning. Next PR will focus on removing GraphQLAny type.

### DIFF
--- a/graphql_compiler/schema_generation/schema_graph.py
+++ b/graphql_compiler/schema_generation/schema_graph.py
@@ -616,7 +616,7 @@ class SchemaGraph(object):
                                      .format(property_name, class_name))
 
             maybe_graphql_type = self._try_get_graphql_type(class_name, property_definition)
-            if maybe_graphql_type is None:
+            if maybe_graphql_type is not None:
                 default_value = _get_default_value(class_name, property_definition)
                 property_descriptor = PropertyDescriptor(maybe_graphql_type, default_value)
                 property_name_to_descriptor[property_name] = property_descriptor
@@ -641,7 +641,7 @@ class SchemaGraph(object):
             elif linked_type is not None and linked_class is None:
                 # No linked class, must be a linked native OrientDB type.
                 maybe_inner_type = try_get_graphql_scalar_type(name + ' inner type', linked_type)
-                if maybe_inner_type is None:
+                if maybe_inner_type is not None:
                     maybe_graphql_type = GraphQLList(maybe_inner_type)
             elif linked_class is not None and linked_type is None:
                 # No linked type, must be a linked non-graph user-defined type.
--- a/graphql_compiler/schema_generation/schema_graph.py
+++ b/graphql_compiler/schema_generation/schema_graph.py
@@ -616,14 +616,14 @@ class SchemaGraph(object):
                                      .format(property_name, class_name))
 
             maybe_graphql_type = self._try_get_graphql_type(class_name, property_definition)
-            if maybe_graphql_type:
+            if maybe_graphql_type is None:
                 default_value = _get_default_value(class_name, property_definition)
                 property_descriptor = PropertyDescriptor(maybe_graphql_type, default_value)
                 property_name_to_descriptor[property_name] = property_descriptor
         return property_name_to_descriptor
 
     def _try_get_graphql_type(self, class_name, property_definition):
-        """Return the GraphQLType corresponding to the non-link property definition or None."""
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -615,20 +615,21 @@ def _get_element_properties(self, class_name, non_link_property_definitions):
...
+            if maybe_graphql_type:
+                default_value = _get_default_value(class_name, property_definition)
+                property_descriptor = PropertyDescriptor(maybe_graphql_type, default_value)
+                property_name_to_descriptor[property_name] = property_descriptor
         return property_name_to_descriptor
 
-    def _get_graphql_type(self, class_name, property_definition):
-        """Return the GraphQLType corresponding to the non-link property definition."""
+    def _try_get_graphql_type(self, class_name, property_definition):
+        """Return the GraphQLType corresponding to the non-link property definition or None."""
```
obi1kenobi: Explain in the docstring when `None` is going to be returned.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -615,20 +615,21 @@ def _get_element_properties(self, class_name, non_link_property_definitions):
                                      u'more than once, this is not allowed!'
                                      .format(property_name, class_name))
 
-            graphql_type = self._get_graphql_type(class_name, property_definition)
-            default_value = _get_default_value(class_name, property_definition)
-            property_descriptor = PropertyDescriptor(graphql_type, default_value)
-            property_name_to_descriptor[property_name] = property_descriptor
+            maybe_graphql_type = self._try_get_graphql_type(class_name, property_definition)
+            if maybe_graphql_type:
```
obi1kenobi: I think you specifically want to check against `is None` here, as opposed to checking for truthiness in general.

graphql_compiler/schema_generation/schema_properties.py
```diff
@@ -111,12 +112,12 @@
...
-    """Return the matching GraphQLScalarType for the property type id, asserting it exists."""
+def try_get_graphql_scalar_type(property_name, property_type_id):
+    """Return the matching GraphQLScalarType for the property type id or None."""
     if property_type_id not in ORIENTDB_TO_GRAPHQL_SCALARS:
-        raise AssertionError(u'Property "{}" has unsupported property type id: '
-                             u'{}'.format(property_name, property_type_id))
-    return ORIENTDB_TO_GRAPHQL_SCALARS[property_type_id]
+        warnings.warn(u'Ignoring property "{}" with unsupported property type id: 'u'{}'
+                      .format(property_name, property_type_id))
+    return ORIENTDB_TO_GRAPHQL_SCALARS.get(property_type_id, None)
```
obi1kenobi: Let's not fetch data twice from the `ORIENTDB_TO_GRAPHQL_SCALARS` dict, that just opens us up to bugs. Instead, do `ORIENTDB_TO_GRAPHQL_SCALARS.get(property_type_id, None)` first, assign it to a variable, and check it for being equal to `None`, then warn.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -654,15 +656,15 @@ def _get_graphql_type(self, class_name, property_definition):
                                          .format(class_name, property_definition))
                 # Don't include the fields and implemented interfaces, this information is already
```
obi1kenobi: I still strongly disagree with this.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -639,8 +640,9 @@ def _get_graphql_type(self, class_name, property_definition):
                                      u'a linked type: {}'.format(name, property_definition))
             elif linked_type is not None and linked_class is None:
                 # No linked class, must be a linked native OrientDB type.
-                inner_type = get_graphql_scalar_type_or_raise(name + ' inner type', linked_type)
-                graphql_type = GraphQLList(inner_type)
+                maybe_inner_type = try_get_graphql_scalar_type(name + ' inner type', linked_type)
+                if maybe_inner_type:
```
obi1kenobi: Here too, you want to compare against `None`.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: I agree with your argument why we shouldn't do this, but this is not the scope of this PR.

PR #266, opened 2019-05-08
Add warning for filtered class collection properties

pmantica1:


graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -77,9 +78,16 @@ def _get_fields_for_class(schema_graph, graphql_types, field_type_overrides, hid
         property_name: graphql_type
         for property_name, graphql_type in six.iteritems(all_properties)
         if not (isinstance(graphql_type, GraphQLList) and
-                isinstance(graphql_type.of_type, GraphQLObjectType))
+                not isinstance(graphql_type.of_type, GraphQLScalarType))
```
obi1kenobi: You need `strip_non_null_from_type()` here. The type `GraphQLList(GraphQLNonNull(GraphQLInt))` for example is representable and I believe should be supported. Same goes for `GraphQLNonNull(GraphQLList(GraphQLInt))`.

I think the first clause should be `isinstance(strip_non_null_from_type(graphql_type), GraphQLList)`, and the second clause should be `not isinstance(strip_non_null_from_type(graphql_type.of_type), GraphQLScalarType)`.

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -77,9 +78,16 @@ def _get_fields_for_class(schema_graph, graphql_types, field_type_overrides, hid
...
     }
 
+    class_collection_properties = set(all_properties.keys()).difference(result.keys())
+    if len(class_collection_properties) > 0:
+        warnings.warn('The fields {} of class {} were ignored since they are GraphQLLists of '
```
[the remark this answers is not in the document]
pmantica1: Sorry, I don't see how the warning is inaccurate. It seems to me that that is exactly what the code above does.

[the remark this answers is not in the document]
obi1kenobi: Ah yes I think you are right, my bad. I was confused because we first capture all properties, then select a subset of them with an `if not`, and then we do set difference. This is a bit convoluted and we may be able to simplify it. How about refactoring so that we explicitly select the properties we want to exclude (i.e. directly compute `class_collection_properties`) and then compute the result by excluding the unsupported properties from the all-properties dict?

evantey14: In the test, can we assert that this warning gets raised?

pmantica1: Seems reasonable. I will add this in a test in the next pr update.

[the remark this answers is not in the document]
pmantica1: Will add this change in next pr update.

pmantica1: python2's unittest.Test is missing an self.assertWarns method. So adding this test is a bit problematic. The effort to figure out how to add this test does not seem to match the reward so I've decided not to do it.

graphql_compiler/schema_generation/graphql_schema.py
pmantica1: Fixed.

PR #268, opened 2019-05-08
Refactor schema graph class collections

pmantica1:


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -652,9 +653,20 @@ def _get_graphql_type(self, class_name, property_definition):
                                          'collection property {}. Only graph classes are allowed '
                                          'to have collections as properties.'
                                          .format(class_name, property_definition))
-                # Don't include the fields and implemented interfaces, this information is already
-                # stored in the SchemaGraph.
-                graphql_type = GraphQLList(GraphQLObjectType(linked_class, {}, []))
+                if len(self._inheritance_sets[linked_class]) > 1:
+                    raise AssertionError('Class {} contains an invalid collection of class {} '
+                                         'elements. Inner classes are not allowed to have '
+                                         'multiple superclasses.'.format(class_name, linked_class))
```
evantey14: I think this message is a bit confusing? Why is it specifically `>1` (rather than `0`)?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -652,9 +653,20 @@ def _get_graphql_type(self, class_name, property_definition):
...
+
+                if linked_class not in self._non_graph_class_to_graphql_rep:
+                    fields = {
+                        name: property_obj.type
+                        for property_obj in self._elements[linked_class].properties.items()
+                    }
+                    self._non_graph_class_to_graphql_rep[linked_class] = (
```
evantey14: If we are only adding things to this mapping here, this would mean the mapping isn't necessarily complete (other functions could try to get from the dict and not find anything). Can we do some setup somewhere? or at least leave a note?

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Ok so each class is a superclass of itself. I will change the if condition to be more explicit about this in the next update.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: The field self._non_graph_class_to_graphql_rep will for sure get deleted anyways in my next PR where I refactor the SchemaGraph constructor so I will leave this unaddressed.

PR #270, opened 2019-05-09
Refactor schema graph constructor

pmantica1:
Since we want to be able to construct SchemaGraphs multiple different types of databases, we must change the OrientDB specific constructor. To that end, I created the get_orientdb_schema_graph function and refactored the SchemaGraph constructor. 

This pull request is only to show a high level overview and get some preliminary feedback. Most of this PR is just copy paste. I've tried to include comments for when I made a particular design decision.

In the upcoming PRs I will add validation for SchemaGraph constructor input.


graphql_compiler/schema_generation/exceptions.py
```diff
@@ -36,3 +36,21 @@ class IllegalSchemaStateError(SchemaError):
 
 class EmptySchemaError(SchemaError):
     """Raised when there are no visible vertex types to import into the GraphQL schema object."""
+
+
+ILLEGAL_PROPERTY_NAME_PREFIXES = (
```
pmantica1: I thought it made most sense to move this to the exceptions file.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,462 @@
...
+from .schema_properties import (
+    COLLECTION_PROPERTY_TYPES, EDGE_DESTINATION_PROPERTY_NAME, EDGE_END_NAMES,
+    EDGE_SOURCE_PROPERTY_NAME, ORIENTDB_BASE_EDGE_CLASS_NAME, ORIENTDB_BASE_VERTEX_CLASS_NAME,
+    PROPERTY_TYPE_LINK_ID, PropertyDescriptor, get_graphql_scalar_type_or_raise,
+    parse_default_property_value
+)
+from .utils import toposort_classes
+
+
+class Kind(Enum):
```
pmantica1: I don't love this Kind Enum, but its simple and doesn't affect code quality that much. Another alternative would be to use the EdgeType, VertexType and NonGraphElement classes directly as the kind.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,464 @@
...
+    Vertex = 1
+    Edge = 2
+    NonGraph = 3
+
+
+def get_orientdb_schema_graph(schema_data):
+    """Create a new SchemaGraph from the OrientDB schema data.
```
pmantica1: Docstring is copy-paste.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,464 @@
...
+
+        non_graph_elements[class_name] = NonGraphElement(
+            class_name, abstract, property_name_to_descriptor, class_fields)
+    return non_graph_elements
+
+
+def _get_edge_elements(class_name_to_definition, inheritance_sets):
+    """Return a dict mapping class name to EdgeType."""
+    edge_elements = dict()
+    subclass_sets = get_subclass_sets_from_inheritance_sets(inheritance_sets)
```
pmantica1: I choose to recompute the subclass_sets multiple times throughout the code instead of passing it as an argument to reduce the number of arguments.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,464 @@
...
+
+    edge_class_names = _get_class_names_of_kind(inheritance_sets, Kind.Edge)
+    vertex_class_names = _get_class_names_of_kind(inheritance_sets, Kind.Vertex)
+    non_graph_class_names = _get_class_names_of_kind(inheritance_sets, Kind.NonGraph)
```
pmantica1: Same here.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,464 @@
...
+                all_subclasses_are_vertices = False
+                break
+        if not (is_linked_class_abstract and all_subclasses_are_vertices):
+            raise AssertionError(u'Property "{}" is declared as a Link to class {}, but '
+                                 u'that class is neither a vertex nor is it an '
+                                 u'abstract class whose subclasses are all vertices!'
+                                 .format(name, linked_class))
+
+
+def _get_end_direction_to_superclasses(link_property_definitions):
```
pmantica1: New helper function for _try_get_base_directions. Essentially copy pasted code.

graphql_compiler/schema_generation/orientdb/schema_properties.py
```diff
@@ -16,23 +16,6 @@
 ORIENTDB_BASE_VERTEX_CLASS_NAME = 'V'
 ORIENTDB_BASE_EDGE_CLASS_NAME = 'E'
 
-ILLEGAL_PROPERTY_NAME_PREFIXES = (
```
pmantica1: Moved this to the exceptions file.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -36,23 +25,6 @@ def _validate_property_names(class_name, properties):
                                           u'{}'.format(class_name, property_name))
 
 
-def get_superclasses_from_class_definition(class_definition):
```
pmantica1: Moved this to the get_orientdb_schema_graph file.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,488 @@
...
+            for superclass_name in immediate_superclass_names
+        ))
+
+        # Freeze the inheritance set so it can't ever be modified again.
+        inheritance_sets[class_name] = frozenset(inheritance_set)
+    return inheritance_sets
+
+
+def get_superclasses_from_class_definition(class_definition):
+    """Extract a list of all superclass names from a class definition dict."""
```
bojanserafimov: Can we decide on one of `inheritance_set` and `superclass_set`? I prefer superclass, because inheritance goes both ways.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,488 @@
...
+
+        property_name_to_descriptor = _get_element_properties(
+            class_name, non_link_property_definitions, non_graph_class_names, inner_collection_objs)
+
+        vertex_elements[class_name] = VertexType(
+            class_name, abstract, property_name_to_descriptor, class_fields)
+    return vertex_elements
+
+
+def _get_class_names_of_kind(inheritance_sets, kind):
```
bojanserafimov: It's weird that `_get_class_names_of_kind` would need `inheritance_sets`. If you renamed the function to `get_subclasses(inheritance_sets, class_name)`, it will be clear with no extra documentation why the `inheritance_sets` are needed. This will also get rid of all use cases of the `Kind` enum, which sounds like it has to do with metaprogramming.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
pmantica1: This seems good to me. However, I'll do this in a following PR. I want to keep this diff as much as copy paste as possible.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
pmantica1: This is not quite true.  Finding the non-graph class names would involve finding the set difference between all the class names and union of the vertex and edge class names. I am going to reuse what we had before: `_split_elements_of_kind`. However, I am going to modify it to use the subclass_sets instead. I think it makes more sense.

pmantica1: I removed _get_class_names_of_kind. At the end, I decided against using subclass_sets instead of inheritance_sets. I would had to compute the subclass_sets in several parts of the code and that would make this pr even larger. I kept this code as much as a copy paste as possible by just modifying `_split_classes_of_kind` into the function `_get_vertex_edge_and_non_graph_class_names`.

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,477 @@
...
+        property_name_to_descriptor = _get_element_properties(
+            class_name, non_link_property_definitions, non_graph_class_names, inner_collection_objs)
+
+        vertex_elements[class_name] = VertexType(
+            class_name, abstract, property_name_to_descriptor, class_fields)
+    return vertex_elements
+
+
+def _get_vertex_edge_and_non_graph_class_names(inheritance_sets):
+    """Return the vertex, edge and non-graph class names."""
```
pmantica1: New function. Refactored from `_split_classes_by_kind`

graphql_compiler/schema_generation/orientdb/get_orientdb_schema_graph.py
```diff
@@ -0,0 +1,477 @@
...
+            if maybe_inner_type is not None:
+                maybe_graphql_type = GraphQLList(maybe_inner_type)
+        elif linked_class is not None and linked_type is None:
+            # No linked type, must be a linked non-graph user-defined type.
+            if class_name in non_graph_class_names:
+                raise AssertionError('Class {} is a non-graph class that contains a '
+                                     'collection property {}. Only graph classes are allowed '
+                                     'to have collections as properties.'
+                                     .format(class_name, property_definition))
+            if linked_class not in inner_collection_objs:
```
pmantica1: Changed this assertion slightly.

PR #275, opened 2019-05-13
Add test and fix broken tests in macro expansion

bojanserafimov:


graphql_compiler/tests/test_macro_expansion.py
```diff
@@ -426,10 +426,11 @@ def test_macro_edge_tag_collision(self):
             Animal {
                 net_worth @tag(tag_name: "parent_net_worth")
                 in_Animal_ParentOf {
-                    net_worth @tag(tag_name: "parent_net_worth_1")
-                    out_Animal_ParentOf @macro_edge_target {
-                        net_worth @filter(op_name: ">", value: ["%parent_net_worth_1"])
+                    net_worth @tag(tag_name: "parent_net_worth_copy_0")
```
obi1kenobi: Why rename the tag?

bojanserafimov: When implementing the name conflict resolution, I thought it's better to be more explicit and closer to user expectations (copying a file in any filesystem will generate a name like this).

bojanserafimov: Now that I think about it, the word `copy` makes it sound like the tag is colocated, which is not. I'll revert to not using `copy`

obi1kenobi: How about `_macro_edge_0` or something like that then? That seems even more explicit about where the tag came from.

PR #276, opened 2019-05-13
Fix subclass sets slow computation

bojanserafimov:
It takes 400 ms, so I removed it from the front-end. We weren't using it there anyway, and we won't use it in the future because we have the `schema_graph`.

I could completely remove the `subclass` module but that requires plumbing the `schema_graph` into the macro system, and there's no need to do it now.


obi1kenobi:
Do you remember what this functionality was meant to be for? I seem to recall it was a way to fix a bug of some sort, I just don't think we ever finished it.

bojanserafimov:
> Do you remember what this functionality was meant to be for? I seem to recall it was a way to fix a bug of some sort, I just don't think we ever finished it.

I don't remember either.

PR #277, opened 2019-05-13
Validate macro edge does not exist on sub/superclass

bojanserafimov:
Address TODO (see red code).

### DIFF
--- a/graphql_compiler/macros/__init__.py
+++ b/graphql_compiler/macros/__init__.py
@@ -75,40 +75,41 @@ def register_macro_edge(macro_registry, macro_edge_graphql, macro_edge_args):
         macro_edge_args: dict mapping strings to any type, containing any arguments the macro edge
                          requires in order to function.
     """
-    class_name, macro_edge_name, macro_descriptor = make_macro_edge_descriptor(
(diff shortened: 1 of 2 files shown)


graphql_compiler/macros/__init__.py
```diff
@@ -88,11 +89,24 @@ def register_macro_edge(macro_registry, macro_edge_graphql, macro_edge_args):
...
-    # TODO(predrag): Write a more stringent check that makes sure that two types A and B,
-    #                where A is a superclass of B, cannot define the same macro edge.
-    #                Right now, both A and B can independently define a macro edge out_Foo,
-    #                which would result in an illegal schema as B would be required to have
-    #                two different descriptors for the same out_Foo edge.
+    # Ensure there's no conflict with macro edges defined on subclasses and superclasses.
+    class_sets_to_check = (
+        ('subclass', macro_registry.subclass_sets[class_name]),
+        ('superclass', {
+            cls
```
obi1kenobi: Are these just class names, or actual class objects of some kind? If they are only names, let's go with `cls_name` instead, just to be extra explicit. Same below.

bojanserafimov: Renamed `class_name` to `new_macro_edge_class_name`, which frees up the name for use in loops like this.

PR #279, opened 2019-05-13
Add macro edge target validation

bojanserafimov:


graphql_compiler/macros/macro_edge/.#validation.py
```diff
@@ -0,0 +1 @@
+bojan@kadmin-XPS-15-9570.19104:1557756771
```
pmantica1: Are you sure you wanted to add this?

PR #281, opened 2019-05-13
Macro tag names

bojanserafimov:
When a macro uses the same tag name as the user, or another macro, assign new names to the macro tags.

If a macro tags a field that the user or another macro also tags, we can remove that tag. This is more difficult and is left for a separate PR.


graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
...
+    Returns:
+        dict mapping the new names to other unique names not present in existing_names
+    """
+    name_mapping = dict()
+    for name in new_names:
+        # We try adding different suffixes to disambiguate from the existing names.
+        disambiguation = name
+        index = 0
+        while disambiguation in existing_names or disambiguation in name_mapping:
+            disambiguation = disambiguation + '_macro_edge_' + str(index)
```
pmantica1: This name resolution seems word. If you fail two times it generates, name+'_macro_edge_1_macro_edge_2'. Are you sure you didn't want simply name+'_macro_edge_2'

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
     return result
 
 
+def get_all_tag_names(ast):
+    """Return a set of strings containing tag names that appear in the query.
+    Args:
```
obi1kenobi: newline before `Args`, and please add a `Returns:` section

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
...
+                    new_value_list.append(StringValue('%' + new_name))
+                elif is_variable_argument(value.value):
+                    new_value_list.append(value)
+                else:
+                    raise AssertionError(u'Value was neither tag nor filter: {} {}'
+                                         .format(filter_directive, value.value))
+            filter_with_renamed_args.arguments.append(
+                Argument(Name('value'), value=ListValue(new_value_list)))
+        else:
+            raise AssertionError(u'Unknown argument name {} in filter'
```
obi1kenobi: Add info on the filter, otherwise this is going to be pretty hard to track down.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
...
+            made_changes = made_changes or made_changes_to_tag
+            new_directives.append(renamed_tag_directive)
+        elif directive.name.value == FilterDirective.name:
+            filter_with_renamed_args, made_changes_to_filter = (
+                _replace_tag_names_in_filter_directive(name_change_map, directive))
+            made_changes = made_changes or made_changes_to_filter
+            new_directives.append(filter_with_renamed_args)
+        else:
+            new_directives.append(directive)
+    return made_changes, new_directives
```
obi1kenobi: Let's keep the API consistent across functions. This function returns `(made_changes, stuff)` but others (e.g. on line 113) return `(stuff, made_changes)`.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
...
+            filter_with_renamed_args.arguments.append(
+                Argument(Name('value'), value=ListValue(new_value_list)))
+        else:
+            raise AssertionError(u'Unknown argument name {} in filter'
+                                 .format(argument.name.value))
+    return filter_with_renamed_args, made_changes
+
+
+def _replace_tag_names_at_current_node(name_change_map, ast):
+    """Replace tag names that are already in use at the root of the AST."""
```
obi1kenobi: The behavior and return value of this function aren't obvious. It kind of sounds like the AST object might be mutated in-place, which does not appear to be the case after inspecting the implementation.

Please write a full docstring that describes what is done and what is returned.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
...
+
+    new_ast = copy(ast)
+    new_ast.selection_set = new_selection_set
+    new_ast.directives = new_directives
+    return new_ast
+
+
+def generate_disambiguations(existing_names, new_names):
+    """Return a dict mapping the new names to similar names not conflicting with existing names.
+    Args:
```
obi1kenobi: Newlines before the starts of new docstring sections.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,137 @@ def get_directives_for_ast(ast):
...
+    Returns:
+        dict mapping the new names to other unique names not present in existing_names
+    """
+    name_mapping = dict()
+    for name in new_names:
+        # We try adding different suffixes to disambiguate from the existing names.
+        disambiguation = name
+        index = 0
+        while disambiguation in existing_names or disambiguation in name_mapping:
+            disambiguation = name + '_macro_edge_' + str(index)
```
obi1kenobi: `'{}_macro_edge_{}'.format(name, index)` might be a bit more convenient here.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -292,9 +296,15 @@ class name -> (macro edge name -> MacroEdgeDescriptor)
                     _check_that_expansion_directives_are_supported(selection_ast)
                     macro_edge_descriptor = macro_edges_at_this_type[field_name]
 
+                    # Sanitize the macro, making sure it doesn't use any taken tag names.
+                    macro_tag_names = get_all_tag_names(macro_edge_descriptor.expansion_ast)
+                    name_change_map = generate_disambiguations(tag_names, macro_tag_names)
+                    tag_names |= set(name_change_map.values())
```
obi1kenobi: Nit: how about `tag_names.update(name_change_map.values())`? It saves a set construction by just walking the iterable directly.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,143 @@ def get_directives_for_ast(ast):
...
+            filter_with_renamed_args.arguments.append(
+                Argument(Name('value'), value=ListValue(new_value_list)))
+        else:
+            raise AssertionError(u'Unknown argument name {} in filter: {}'
+                                 .format(argument.name.value, ast))
+    return filter_with_renamed_args, made_changes
+
+
+def _replace_tag_names_at_current_node(name_change_map, ast):
+    """Return a new ast with tag names replaced according to the name_change_map."""
```
obi1kenobi: This docstring is very misleading. Having read this, I would expect the function's signature to be:
```
T = TypeVar('T', Field, OperationDefinition, InlineFragment)

def _replace_tag_names_at_current_node(name_change_map: Dict[str, str], ast: T) -> T:
    ...
```

However, the actual signature's return type is `-> Tuple[List[Directive], bool]`.

Please write out a full docstring, including `Args/Returns`.

obi1kenobi: While we're at it, the function's name is also misleading in the same way.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -88,7 +99,19 @@ def _replace_tag_names_in_tag_directive(name_change_map, tag_directive):
 
 
 def _replace_tag_names_in_filter_directive(name_change_map, filter_directive, ast):
-    """Return the new directive, and whether it is different from the old."""
+    """Return the new directive, and whether it is different from the old.
+
+    Args:
+        name_change_map: Dict[str, str] mapping tag names to new names
+        tag_directive: GraphQL library filter directive that potentially uses tagged parameters.
+                       all such tagged parameters should be in the name_change_map.
```
obi1kenobi: Copy pasta. `tag_directive` is not an argument, and `filter_directive` and `ast` are not described.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -78,7 +78,18 @@ def get_all_tag_names(ast):
...
 def _replace_tag_names_in_tag_directive(name_change_map, tag_directive):
-    """Return the new directive, and whether it is different from the old."""
+    """Return the new directive, and whether it is different from the old.
+
+    Args:
+        name_change_map: Dict[str, str] mapping tag names to new names
+        tag_directive: GraphQL library directive whose name is in the name_change_map
+
+    Returns:
+        pair containing:
```
obi1kenobi: Tuple, not pair.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -118,7 +141,18 @@ def _replace_tag_names_in_filter_directive(name_change_map, filter_directive, as
...
-    """Return a new ast with tag names replaced according to the name_change_map."""
+    """Return a new ast with tag names replaced according to the name_change_map.
+
+    Args:
+        name_change_map: Dict[str, str] mapping all tag names in the ast to new names
+        ast: GraphQL library AST object, such as a Field, InlineFragment, or OperationDefinition
+
+    Returns:
+        pair containing:
+        - GraphQL library AST object, equivalent to the input one, with all tag names at its root
```
obi1kenobi: I don't think we return an AST object. `new_directives` is a list.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -139,7 +173,16 @@ def _replace_tag_names_at_current_node(name_change_map, ast):
 
 
 def replace_tag_names(name_change_map, ast):
-    """Replace tag names that are already in use in the whole AST."""
+    """Return a new ast with tag names replaced according to the name_change_map.
```
obi1kenobi: Everywhere where you depend on the exact same object being returned in the "no changes" scenario, please make sure to call that out in the docstring. Otherwise, it's not a postcondition and the code that depends on that behavior is technically wrong.

bojanserafimov: It is not a postcondition. I don't rely on it.

bojanserafimov: Is it misleading to say `Return a new ast...`, but in the case where nothing needs to be done, return the same one? Is `Return an ast ...` clearer?

obi1kenobi: I'd just append a new sentence, something like `If no tag names were replaced, the same AST object that was provided as input is returned.` What do you think?

graphql_compiler/macros/macro_edge/helpers.py
obi1kenobi: Thoughts on this, here and elsewhere in this file?

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,191 @@ def get_directives_for_ast(ast):
...
+
+
+def _replace_tag_names_in_filter_directive(name_change_map, filter_directive, ast):
+    """Return the new directive, and whether it is different from the old.
+
+    Args:
+        name_change_map: Dict[str, str] mapping tag names to new names
+        filter_directive: GraphQL library filter directive that potentially uses tagged parameters.
+                          all such tagged parameters should be in the name_change_map. This
+                          directive is not mutated.
```
obi1kenobi: missing `ast` argument description

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,191 @@ def get_directives_for_ast(ast):
...
+    """
+    made_changes = False
+    filter_with_renamed_args = copy(filter_directive)
+    filter_with_renamed_args.arguments = []
+    for argument in filter_directive.arguments:
+        if argument.name.value == 'op_name':
+            filter_with_renamed_args.arguments.append(argument)
+        elif argument.name.value == 'value':
+            new_value_list = []
+            for value in argument.value.values:
```
obi1kenobi: Why not pull `value.value` into a named local, so it's more obvious what is going on?

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -53,6 +60,191 @@ def get_directives_for_ast(ast):
...
+                if is_tagged_parameter(value.value):
+                    current_name = get_parameter_name(value.value)
+                    new_name = name_change_map[current_name]
+                    if new_name != current_name:
+                        made_changes = True
+                    new_value_list.append(StringValue('%' + new_name))
+                elif is_runtime_parameter(value.value):
+                    new_value_list.append(value)
+                else:
+                    raise AssertionError(u'Value was neither tag nor filter: {} {}'
```
obi1kenobi: This error message is confusing. `Value` is vague, and `tag nor filter` makes it sound like the problem is with the directive. Let's stick to the naming conventions we've adopted in the README: `runtime parameter` and `tagged parameter`.

graphql_compiler/macros/macro_edge/helpers.py
```diff
@@ -123,22 +124,25 @@ def _replace_tag_names_in_filter_directive(name_change_map, filter_directive, as
...
                     new_value_list.append(StringValue('%' + new_name))
-                elif is_runtime_parameter(value.value):
+                elif is_runtime_parameter(parameter):
                     new_value_list.append(value)
                 else:
-                    raise AssertionError(u'Value was neither tag nor filter: {} {}'
-                                         .format(filter_directive, value.value))
+                    raise AssertionError(u'Parameter {} used in directive {} is expected to '
+                                         u'be a tagged parameter (starting with %) or a runtime '
+                                         u'parameter (starting with $), but it was neither. {}',
```
obi1kenobi: You accidentally deleted the `.format()` here, and are instead passing 4 arguments to the `AssertionError` constructor.

PR #282, opened 2019-05-13
Only allow supported directives at macro expansion

bojanserafimov:


graphql_compiler/tests/test_macro_expansion_errors.py
```diff
@@ -118,3 +118,29 @@ def test_incorrect_schema_usage(self):
...
+            Animal {
+                out_Animal_GrandparentOf @optional {
+                    name @output(out_name: "grandkid")
+                }
+            }
+        }'''
+        args = {}
+
+        with self.assertRaises(GraphQLCompilationError):
+            perform_macro_expansion(self.macro_registry, query, args)
```
pmantica1: Aren't you missing output source test?

PR #284, opened 2019-05-15
Allow list types in RootSchemaQuery

bojanserafimov:
I verified manually that this works. I won't change the actual schema we use in tests now, because I will also need to change the OrientDB schema generation to match that, which will lead to merge conflicts with Pedro's feature branch.


bojanserafimov:
If we're doing this, we might also want to allow non-list types for vertex fields, which can be used for one-to-many relationships.

pmantica1:
resolves #284

obi1kenobi:
> If we're doing this, we might also want to allow non-list types for vertex fields, which can be used for many-to-one relationships.

Agreed, though I'd probably leave that for a separate diff. This has been something I've wanted to do ever since I spoke to @hartig a few months ago, when he showed me his ideas on expressing relationship metadata like this. Here's a link to his blog post for inspiration: http://blog.liu.se/olafhartig/documents/graphql-schemas-for-property-graphs/

We are spinning a few too many plates at the same time: SQL backend, macro edges, schema graph, GraphQL over schema, etc. Let's get them merged into `master` and released, and then let's do a pass over how we communicate these schema constraints. @hartig and his colleagues have worked very hard on these ideas, and we should avoid reinventing the wheel whenever possible.

Also, @hartig, if you and your colleagues are able and willing to work with us on this push, your help would be very much welcome and appreciated!

hartig:
Hi, thanks for the mention!

In addition to the blog post mentioned by @obi1kenobi we have now published the following research paper about the approach:

Olaf Hartig and Jan Hidders: Defining Schemas for Property Graphs by using the GraphQL Schema Definition Language. In Proceedings of the 2nd Joint International Workshop on Graph Data Management Experiences & Systems (GRADES) and Network Data Analytics (NDA) (GRADES-NDA'19), 2019. http://olafhartig.de/files/HartigHidders_GRADES2019_Preprint.pdf 

Moreover, as kinda the next step, we are working on an approach to extend such a database schema S (defined using the GraphQL SDL as per our approach) into an actual GraphQL schema for a GraphQL API over a database that conforms to the schema S. Find an informal overview of this approach at https://github.com/LiUGraphQL/pgschema-to-apischema/wiki/Overview-of-the-Approach

bojanserafimov:
> Could you add a test? If you want to wait for my branch to be merged that's ok.

Now I'm running all tests with list types in the `RootSchemaQuery` by default. Do you think there's a need for a test the other way (that it works with bare types in `RootSchemaQuery`)? Or should we do this when we address the problem globally?

pmantica1:
Now that I look at it,  it doesn't seem the code would act any differently if we used bare types instead so I personally don't think we need more tests for now.

PR #287, opened 2019-05-16
Allow connections for NonGraphElements

pmantica1:
I will add SchemaGraph input validation and validation tests in future diffs. Right now we don't even have a notion of what SQL object would NonGraphElement could be so there is no point adding general SchemaGraph validation for NonGraphElements.

### DIFF
--- a/graphql_compiler/schema_generation/schema_graph.py
+++ b/graphql_compiler/schema_generation/schema_graph.py
@@ -205,6 +205,22 @@ class SchemaElement(object):
         self._properties = properties
         self._class_fields = class_fields
 
+        # In the schema graph, both vertices and edges are represented with vertices.
+        # These dicts have the name of the adjacent schema vertex in the appropriate direction.
+        #
+        # For vertex classes:
+        #   in  = the edge is attached with its head / arrow side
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -205,6 +205,22 @@ def __init__(self, class_name, abstract, properties, class_fields, *args, **kwar
...
+        # For vertex classes:
+        #   in  = the edge is attached with its head / arrow side
+        #   out = the edge is attached with its tail side
+        #
+        # For edge classes and non-graph classes:
+        #   in  = the tail side of the edge
+        #   out = the head / arrow side of the edge
+        #
+        # A non-graph class may only have connections if it's abstract and all of its non-abstract
+        # subclasses are vertices.
```
bojanserafimov: Is there any validation for this?

pmantica1: https://github.com/kensho-technologies/graphql-compiler/blob/allow-connections-for-non-graph-elements/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py#L387

PR #296, opened 2019-05-17
Remove automatic OrientDBBaseVertex schema hiding

pmantica1:
We should have no reference to the OrientDB base vertex class name in get_graphql_schema_from_schema_graph.


graphql_compiler/__init__.py
```diff
@@ -184,16 +188,16 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
                                        (string -> {string -> GraphQLType}). Used to override the
                                        type of a field in the class where it's first defined and all
                                        the class's subclasses.
-        hidden_classes: optional set of strings, classes to not include in the GraphQL schema.
+        hidden_classes: set of strings, classes to not include in the GraphQL schema.
+                        By default we hide the base OrientDB vertex class to make the schema
+                        more syntactically sweet.
```
bojanserafimov: It's still `optional set of strings`

pmantica1:
No. If you put it None, you'll get a runtime error.

graphql_compiler/__init__.py
```diff
@@ -137,8 +138,11 @@ def graphql_to_gremlin(schema, graphql_query, parameters, type_equivalence_hints
         query=insert_arguments_into_query(compilation_result, parameters))
 
 
-def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_type_overrides=None,
-                                                 hidden_classes=None):
+def get_graphql_schema_from_orientdb_schema_data(
+        schema_data,
```
obi1kenobi: Why the double indent?

pmantica1: Will fix.

graphql_compiler/__init__.py
pmantica1: I think that the terminology that we've used for maybe None is optional.

pmantica1: I mean to say that this field cannot be None.

obi1kenobi: This is probably safe, but it's still better practice in Python to set it to `None` and then internally to the function check for `None` and set it to whatever the default value is. To see why, try making a function with a mutable default argument, then mutate it, and see what the default argument is on the next call.

bojanserafimov: haha we were just talking about this

pmantica1: I think having a frozenset in this specific case is better. 

Using the 'if is None" approach does not communicate as well to the reader that the default value of hidden_classes is {OrientDBBaseVertexClassName}. In this case, it's also unclear what happens when you modify the `hidden_classes` parameter. Do we still hide the OrientDBBaseVertex class name in addition to the hidden classes you specified? You have to look at the code to see.

obi1kenobi: This is precisely what the docstring is for, however. All of these things should be made clear in the docstring (and probably tested, as well). Relying on the code to communicate this is bad practice -- what happens when the code is refactored?

Using a `frozenset` means that the code is currently correct. However, it also means that we have to continue to be mindful to not accidentally change the type back to `set`, because of the problem we discussed above. Going from `frozenset` to `set` is a much smaller change than going from `None` to `set`, and I am not confident I'll remember to always watch out for it in code reviews on this function in the future. Are you?

pmantica1: Ok fair. Will fix.

graphql_compiler/tests/test_helpers.py
```diff
@@ -279,6 +279,10 @@ def generate_schema_graph(graph_client):
 
 def generate_schema(graph_client, class_to_field_type_overrides=None, hidden_classes=None):
     """Generate schema and type equivalence dict from a pyorient client"""
+    if class_to_field_type_overrides is None:
+        class_to_field_type_overrides = dict()
+    if hidden_classes is None:
+        hidden_classes = set()
```
bojanserafimov: This is getting complicated. This function has a default `set()`. The function it calls has a default `{'V'}`. The function calling this one gives `{'V'}` as an argument. Why not make the default `set()` everywhere, and make the user explicitly ignore it?

Another weird side effect of `{'V'}` being default is that if you want to actually hide another class, you can accidentally unhide `{'V'}`.

pmantica1: Ok makes sense. I will do this.

PR #298, opened 2019-05-17
Move schema element linking to SchemaGraph

pmantica1:
This method is not OrientDB specific and I plan to use reuse it for the SQLAlchemy SchemaGraph generation.

### DIFF
--- a/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
+++ b/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
@@ -434,7 +434,8 @@ def _try_get_base_connections(class_name, superclass_sets, links, abstract):
         base_connections.get(EDGE_DESTINATION_PROPERTY_NAME, None),
     )
 
-def _get_graphql_rep_of_non_graph_elements(non_graph_elements, inheritance_sets):
+
+def _get_graphql_rep_of_non_graph_elements(non_graph_elements, superclass_sets):
     """Return a dict mapping name to GraphQL Object for non graph elements without superclasses."""
     graphql_reps = {}
     for element_name, element in six.iteritems(non_graph_elements):
--- a/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
+++ b/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
@@ -8,7 +8,7 @@ import six
 from ..exceptions import IllegalSchemaStateError
 from ..schema_graph import (
     EdgeType, NonGraphElement, PropertyDescriptor, SchemaGraph, VertexType,
-    get_subclass_sets_from_inheritance_sets
+    get_subclass_sets_from_inheritance_sets, link_schema_elements
 )
 from .schema_properties import (
     COLLECTION_PROPERTY_TYPES, EDGE_DESTINATION_PROPERTY_NAME, EDGE_END_NAMES,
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -356,6 +348,15 @@ def _validate_property_names(class_name, properties):
                                           u'{}'.format(class_name, property_name))
 
 
+def _get_element_names_of_class(elements, cls):
+    """Return a frozenset of the names of the elements are instances of the class."""
```
bojanserafimov: `that are`

graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
```diff
@@ -434,37 +434,7 @@ def _try_get_base_connections(class_name, superclass_sets, links, abstract):
...
-
-        # Link edge_class_name with to_class_name
-        for to_class in subclass_sets[to_class_name]:
-            to_schema_element = elements[to_class]
-            edge_schema_element.out_connections.add(to_class)
-            to_schema_element.in_connections.add(edge_class_name)
-
-
-def _get_graphql_rep_of_non_graph_elements(non_graph_elements, superclass_sets):
+def _get_graphql_rep_of_non_graph_elements(non_graph_elements, inheritance_sets):
```
obi1kenobi: What is `rep`? Let's not use abbreviations, because economizing on typing out names is always a losing proposition.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -372,6 +373,34 @@ def get_subclass_sets_from_superclass_sets(superclass_sets):
     return subclass_sets
 
 
+def link_schema_elements(elements, inheritance_sets):
+    """For each edge, link the schema elements it connects to each other."""
+    subclass_sets = get_subclass_sets_from_inheritance_sets(inheritance_sets)
```
obi1kenobi: Doesn't this mean that we'll end up computing subclass sets twice? Once in this function, and once when actually building the SchemaGraph?

This is a bit of a code smell. ( https://en.wikipedia.org/wiki/Code_smell in case you haven't heard the term before)

pmantica1: Yes. It does.

pmantica1: So another option for this could be passing in the subclass_sets when we construct the SchemaGraph but then we'd have to perform additional validation to check that the subclass_sets and the superclass_sets are consistent.

obi1kenobi: @bojanserafimov thoughts? On one hand, constructing subclass sets twice is somewhat expensive and gives us more places to write buggy code. On the other, having to pass both subclass and superclass sets into the constructor and then needing to validate them against each other makes no sense either. Any ideas on how we can design our way out of this situation?

pmantica1: What if we created an inheritance sets named tuple that created the subclass sets from the superclass sets? (It would also be entirely immutable).

obi1kenobi: Yeah, something like that could work. Maybe if the namedtuple contained both the sub- and the superclass sets at once, and was constructed immediately from the toposorted schema information. So instead of calling `_get_superclass_sets_from_schema_data()`, call the function that constructs this namedtuple that contains both sub- and superclass information. Then if we always pass the namedtuple itself instead of its components, we know it was well-formed and don't need to validate.

bojanserafimov: that makes sense

pmantica1: Ok, I'll make a separate PR for this.

PR #299, opened 2019-05-17
Add basic sqlalchemy vertex generation

pmantica1:


graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,81 @@
...
+    for i, class_ in enumerate(sql_classes):
+        for other_class_ in sql_classes[i + 1:]:
+            if issubclass(other_class_, class_):
+                raise AssertionError("SQL_CLASS_TO_GRAPHQL_TYPE is not toposorted by class "
+                                     "inheritance. Class {} is a subclass of {} but is appears  "
+                                     "later in the OrderedDict.")
+
+
+def _try_get_graphql_scalar_type(column_name, column_type):
+    """Return the most precise GraphQLScalarType for the SQL datatype class.
```
pmantica1: This function has the advantage that we can also map custom types to GraphQLTypes provided the custom types inherit from the supported SQL datatypes.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,81 @@
...
+
+    For instance, if class of the column_type is the Numeric class we return GraphQLDecimal.
+    If the class of the column_type is Float, a subclass of Numeric, we return GraphQLFloat.
+    """
+    maybe_graphql_type = None
+    for sql_class, graphql_type in SQL_CLASS_TO_GRAPHQL_TYPE.items():
+        if isinstance(column_type, sql_class):
+            maybe_graphql_type = graphql_type
+            break
+    if not maybe_graphql_type:
```
bojanserafimov: can you make a list of the unsupported ones, and check if it's there? Then if it's not either supported or unsupported, you can raise an `AssertionError`. This way we will know exactly what's unsupported.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,81 @@
...
+    (sqltypes.DateTime, GraphQLDateTime),
+    (sqltypes.Date, GraphQLDate),
+    (sqltypes.Boolean, GraphQLBoolean),
+])
+
+
+# TODO(pmantica1): Map foreign keys to edges.
+# TODO(pmantica1): Represent table inheritance in SchemaGraph.
+# TODO(pmantica1): Add option to map tables to EdgeTypes instead of VertexTypes.
+# TODO(pmantica1): Map arrays to GraphQLLists.
```
bojanserafimov: Do the IR, and the SQL backend support lists?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: I think we should warn over raising an AssertionError like we currently do. I have that list in the tests.

pmantica1: The list is in the test. I don't think we should raise an AssertionError because we won't be able to generate a schema then unless we support all the types. I think it's best to just ignore some columns.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: @obi1kenobi ?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+
+from ...schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal
+from ..schema_graph import PropertyDescriptor, SchemaGraph, VertexType
+
+
+# TODO(pmantica1): Add scalar mapping for the following classes: Interval, and Time.
+# We do not currently plan to add a mapping for JSON and Binary objects.
+SQL_CLASS_TO_GRAPHQL_TYPE = OrderedDict([
+    (sqltypes.String, GraphQLString),
+    (sqltypes.Integer, GraphQLInt),
```
bojanserafimov: Are these actually the same? What's the largest `sqltypes.Integer`?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
bojanserafimov: The `AssertionError` would be raised if we get a type that's neither supported nor unsupported. This can happen if we fail to account for all types that exist. For example, it would be really sad if SQL had 2 different types of ints, and we only supported one, without knowing that we just dropped all ints of the second type.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Big int is signed 8 bytes. GraphQLInt is signed 4 bytes.

bojanserafimov: More generally, we should look at all minor differences among the Orientdb, python, and sqlalchemy versions of the same type, and make sure we have a plan on how to cross boundaries.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+
+def _try_get_graphql_scalar_type(column_name, column_type):
+    """Return the most precise GraphQLScalarType for the SQL datatype class.
+
+    For instance, if class of the column_type is the Numeric class we return GraphQLDecimal.
+    If the class of the column_type is Float, a subclass of Numeric, we return GraphQLFloat.
+    """
+    maybe_graphql_type = None
+    for sql_class, graphql_type in SQL_CLASS_TO_GRAPHQL_TYPE.items():
+        if isinstance(column_type, sql_class):
```
bojanserafimov: Sqlalchemy primitive types can inherit from multiple primitive types. So the order of elements in the `SQL_CLASS_TO_GRAPHQL_TYPE` matters. I would completely ignore the class hierarchy of primitive types and map each one separately to a GraphQL type.

bojanserafimov:
I have some nits too, but I'll leave them for later.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
obi1kenobi: The IR has no problem with list-typed values. I think not all SQL databases support array-typed columns, though, so there's a bit more work to be done before we can compile those successfully in the backend.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
obi1kenobi: We'll probably need to (eventually) add a new custom scalar type, maybe call it `Int64` to signify it's 64-bit. For now, we can do the easy but wrong thing and use the `Int` type, so long as we open an issue so we remember to fix it later.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+    for table_name, table in six.iteritems(sqlalchemy_metadata.tables):
+        elements[table_name] = _get_vertex_type_from_sqlalchemy_table(table)
+    inheritance_sets = {element_name: {element_name} for element_name in elements}
+    return SchemaGraph(elements, inheritance_sets)
+
+
+def _validate_sql_to_graphql_is_toposorted_by_class_inheritance():
+    """Validate that SQL_SCALAR_CLASS_TO_GRAPHQL_TYPE dict is toposorted by class inheritance."""
+    sql_classes = list(SQL_CLASS_TO_GRAPHQL_TYPE.keys())
+    for i, class_ in enumerate(sql_classes):
```
obi1kenobi: Please avoid suffixing variable names with underscores. Is there a better name for this?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+])
+
+
+# TODO(pmantica1): Map foreign keys to edges.
+# TODO(pmantica1): Represent table inheritance in SchemaGraph.
+# TODO(pmantica1): Add option to map tables to EdgeTypes instead of VertexTypes.
+# TODO(pmantica1): Map arrays to GraphQLLists.
+def get_schema_graph_from_sql_alchemy_metadata(sqlalchemy_metadata):
+    """Return the matching SchemaGraph for the SQLAlchemy Metadata object"""
+    _validate_sql_to_graphql_is_toposorted_by_class_inheritance()
```
obi1kenobi: Since this is verifying that a global constant is valid, why not make it a test case instead of validating it on every call of this function?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+    return SchemaGraph(elements, inheritance_sets)
+
+
+def _validate_sql_to_graphql_is_toposorted_by_class_inheritance():
+    """Validate that SQL_SCALAR_CLASS_TO_GRAPHQL_TYPE dict is toposorted by class inheritance."""
+    sql_classes = list(SQL_CLASS_TO_GRAPHQL_TYPE.keys())
+    for i, class_ in enumerate(sql_classes):
+        for other_class in sql_classes[i + 1:]:
+            if issubclass(other_class, class_):
+                raise AssertionError('SQL_CLASS_TO_GRAPHQL_TYPE is not toposorted by class '
```
obi1kenobi: Remember the `u` prefix to make these unicode literals, we still support py2.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
obi1kenobi: Great -- document that in the docstring, please. Also, whenever your docstring is more than just the one-liner, please remember to include `Args:` and `Returns:` sections.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
obi1kenobi: @bojanserafimov would it help if our tests are run in such a way that warnings are treated as exceptions, so that they are guaranteed to fail if this warning is hit in the test environment?

What I am a bit worried about is this scenario:
- SQLAlchemy updates the SQL types it exposes, adding a new type that is not in our list.
- The compiler breaks with that version of SQLAlchemy, and we have to release a new patch on an emergency basis that updates that list to make it explicitly unsupported.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    maybe_graphql_type = None
+    for sql_class, graphql_type in SQL_CLASS_TO_GRAPHQL_TYPE.items():
+        if isinstance(column_type, sql_class):
+            maybe_graphql_type = graphql_type
+            break
+    if not maybe_graphql_type:
+        # Try get the string representation of SQLAlchemy JSON and ARRAY types
+        # will lead to an error. We therefore use repr instead.
+        warnings.warn(u'Ignoring column "{}" with unsupported SQL datatype class: 'u'{}'
```
obi1kenobi: This seems like a formatting problem: `datatype class: 'u'{}'`

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,82 @@
...
+# TODO(pmantica1): Address default values of columns.
+# TODO(pmantica1): Map Enum to the GraphQL Enum type.
+def _get_vertex_type_from_sqlalchemy_table(table):
+    """Return the VertexType corresponding to the SQLALchemyTable object."""
+    properties = dict()
+    for column in table.get_children():
+        name = column.key
+        default = None
+        maybe_property_type = _try_get_graphql_scalar_type(name, column.type)
+        if maybe_property_type:
```
obi1kenobi: Do you mean `if maybe_property_type is not None:` here?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: We have this same issue in OrientDB.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Good idea.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Yep mb.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Ok makes sense. Seems simpler. I'll do that. Can you show me an example though of a class inheriting from multiple primitive types? This seems to be a very important detail.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: I am going to follow Bojan's suggestion for this code to simplify it.  I am just going to add a direct map from primitive to GraphQLType. One advantage that this approach seemed to have is that it dealt with custom types that inherited from primitives, but the complexity doesn't seem worth it.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Thanks will fix.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: There is no longer any need for the dict to be toposorted.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,97 @@
...
+import sqlalchemy.sql.sqltypes as sqltypes
+
+from ...schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal, GraphQLInt
+from ..schema_graph import PropertyDescriptor, SchemaGraph, VertexType
+
+
+# TODO(pmantica1): Add scalar mapping for the following classes: Interval.
+# We do not currently plan to add a mapping for JSON and Binary objects.
+SQL_CLASS_TO_GRAPHQL_TYPE = {
+    sqltypes.BIGINT: GraphQLInt,
```
pmantica1: An explanation that I found for classes written in all-caps :

> This category of types refers to types that are either part of the SQL standard, or are potentially found within a subset of database backends. Unlike the “generic” types, the SQL standard/multi-vendor types have no guarantee of working on all backends, and will only work on those backends that explicitly support them by name. That is, the type will always emit its exact name in DDL with CREATE TABLE is issued.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Updated TODO statement to account for SQL array compiler targeting not being ready.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: @bojanserafimov ?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: This makes sense to me. @bojanserafimov what do you think?

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
bojanserafimov: enum is a string and also a schema type

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
bojanserafimov: I see, that would be bad. Agree.

A slight problem with this is that tests will never hit that warning because they only test with a hard-coded list of types. If we can assert that the types are what we think we are, that would be great. But I don't see a way to do that.

pmantica1: Ok, I added a warning test.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,97 @@
...
+
+# TODO(pmantica1): Map foreign keys to edges.
+# TODO(pmantica1): Represent table inheritance in SchemaGraph.
+# TODO(pmantica1): Add option to map tables to EdgeTypes instead of VertexTypes.
+def get_schema_graph_from_sql_alchemy_metadata(sqlalchemy_metadata):
+    """Return the matching SchemaGraph for the SQLAlchemy Metadata object"""
+    elements = dict()
+    for table_name, table in six.iteritems(sqlalchemy_metadata.tables):
+        elements[table_name] = _get_vertex_type_from_sqlalchemy_table(table)
+    inheritance_sets = {element_name: {element_name} for element_name in elements}
```
bojanserafimov: `superclass_sets`

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
obi1kenobi: Can you add this excerpt together with the documentation link where you found it? Otherwise, we'll say this looks good now, and then 6 months from now we'll have forgotten about it and be confused again.

In general, if you feel the need to explain your code in a review comment, strongly consider adding a comment in the code itself, for the above reason.

PR #300, opened 2019-05-20
Explain unintuitive naming convention

pmantica1:


graphql_compiler/schema_generation/orientdb/schema_properties.py
```diff
@@ -8,9 +8,13 @@
 
 from ...schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal
 
-
+# The following variable assignment stems from confusing OrientDB conventions.
+# In the following url, the 'out' link of Owns edge is the source class Person and the 'in'
```
obi1kenobi: I think we're better off making our own example with our own schema to explain this concept. Forcing the reader to read another article on another website (which might move / get edited / etc.) and understand another schema before understanding the code is not scalable.

pmantica1: I just wanted to use the reference as "proof" of the naming convention. I don't think we really need an example for this. How about:
# The naming stems from OrientDB naming conventions for builtin link properties.

obi1kenobi: Sure, that seems fine. I'd append `for edges` to the end of that sentence, just to extra clear. My concern with the reference is that it's not to the official documentation, but to something that looks like a ported-over blog post from OrientDB's old site, and the unfortunate formatting in the link doesn't really inspire confidence.

obi1kenobi:
The PR is failing lint, otherwise it looks good.

PR #304, opened 2019-05-20
Document common Docker Compose issue

ReaLNeroM:


TROUBLESHOOTING.md
```diff
@@ -48,3 +48,15 @@ Then, make sure that `clang` is able to find it by adding the following line to
 ```
 export LIBRARY_PATH=$LIBRARY_PATH:/usr/local/opt/openssl/lib/
 ```
+
+## Issues starting PostgreSQL: Port already in use
```
pmantica1: Since this section is very similar to the MySQL section in the top do you mind combining these two into a single section?

PR #305, opened 2019-05-20
Add SQLAlchemy default values

pmantica1:


graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -97,7 +96,7 @@ def _get_vertex_type_from_sqlalchemy_table(table):
     properties = dict()
     for column in table.get_children():
         name = column.key
-        default = None
+        default = column.default.arg if column.default is not None else None
```
bojanserafimov: Can we type check the default values, or write some tests? For example, I'm not sure if time and timestamp would actually return something we can use as a datetime directly without any processing.

obi1kenobi: Good idea -- does SQLAlchemy do any checking here?

bojanserafimov: Even if they do, their idea of time might be "seconds since midnight".

PR #308, opened 2019-05-21
Move element linking to SchemaGraph

pmantica1:
This method can be used by also be used by other schema builders, (specifically the SQL schema builder), and we should, therefore, put it in the schema_graph file.

### DIFF
--- a/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
+++ b/graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
@@ -8,7 +8,7 @@ import six
 from ..exceptions import IllegalSchemaStateError
 from ..schema_graph import (
     EdgeType, InheritanceStructure, NonGraphElement, PropertyDescriptor, SchemaGraph, VertexType,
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
```diff
@@ -437,35 +437,7 @@ def _try_get_base_connections(class_name, inheritance_structure, links, abstract
...
-
-        # Link edge_class_name with to_class_name
-        for to_class in inheritance_structure.subclass_sets[to_class_name]:
-            to_schema_element = elements[to_class]
-            edge_schema_element.out_connections.add(to_class)
-            to_schema_element.in_connections.add(edge_class_name)
-
-
-def _get_graphql_rep_of_non_graph_elements(non_graph_elements, inheritance_structure):
+def _get_graphql_representation_of_non_graph_elements(non_graph_elements, inheritance_structure):
```
pmantica1: This nit is out of the scope of the pull request, but it doesn't hurt to add it.

PR #310, opened 2019-05-21
Fix sql alchemy schema inheritance

pmantica1:


pmantica1:
I also renamed the OrientDB schema generation tests in this branch.

PR #311, opened 2019-05-22
Add output info to query metadata

evantey14:
Add basic information about outputs to the query metadata table.


graphql_compiler/compiler/metadata.py
```diff
@@ -24,6 +24,13 @@
 )
 
 
+OutputInfo = namedtuple(
+    'OutputInfo',
+    (
+        'location',
```
obi1kenobi: Let's include the type and whether the output is optional as well?

The QueryMetadataTable object should arguably be a superset of all other metadata, so in principle if we do `record_output_info()` on line 321 in the file above, we probably shouldn't need to also record this info in `context['outputs'][output_name]` as well -- instead, that should be autogenerated from the `OutputInfo` object from the QueryMetadataTable.

evantey14: I think the idea that it should be a superset makes sense. 

What do you mean by autogenerated? (that `context['outputs']` should be generated from `context['metadata']`?) Either way, why have `context['outputs']` if all the information is in `context['metadata']`? 

Generally, is the idea of context to store local information that we need for making decisions? If so, is there any problem with storing the info twice?

bojanserafimov: Yes, I think we want to remove `context['outputs']` if that information is in the `QueryMetadataTable`. We don't need to do it in this PR, but it'd be nice to work in that direction by adding the output and optional information in the table.

obi1kenobi: Agreed with @bojanserafimov. The `context` predates the `QueryMetadataTable` by a long time, and is how we stored query-level metadata before we had this newer, more structured approach. Over time, I expect we'll remove many if not most parts of the `context` dict.

evantey14:
Added tests and removed `context['outputs']`

graphql_compiler/compiler/metadata.py
```diff
@@ -161,6 +172,24 @@ def get_location_info(self, location):
                                  u'{}'.format(location))
         return location_info
 
+    @property
+    def outputs(self):
+        for output_name, output_info in six.iteritems(self._outputs):
+            yield output_name, output_info
```
bojanserafimov: why not just `return six.iteritems(self._outputs)`?

graphql_compiler/compiler/metadata.py
```diff
@@ -161,6 +172,24 @@ def get_location_info(self, location):
                                  u'{}'.format(location))
         return location_info
 
+    @property
+    def outputs(self):
```
bojanserafimov: can you rename it to `get_all_outputs` for consistency with other getters here, and add a docstring explaining that it returns an iterable of tuples?

bojanserafimov:
I got two nits, looks good otherwise

graphql_compiler/compiler/metadata.py
evantey14: This is meant to be a *property* consistent with `tags` and `registered_locations`. Not sure if we have a preference between properties and get_all. Maybe in a future PR?

graphql_compiler/compiler/metadata.py
evantey14: Hmm I'm not sure if we have a huge reason for one or the other? This is consistent with the other properties.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -524,7 +525,7 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
         if edge_traversal_is_folded:
             has_count_filter = has_fold_count_filter(context)
             _validate_fold_has_outputs_or_count_filter(
-                get_context_fold_info(context), has_count_filter, context['outputs'])
+                get_context_fold_info(context), has_count_filter, context['metadata'])
```
obi1kenobi: `context['metadata']` is already available as `query_metadata_table` on line 400. This is absolutely not obvious at all since this function is quite large at this point, and I'll probably get around to refactoring it before too long...

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -560,7 +561,9 @@ def _compile_vertex_ast(schema, current_schema_type, ast,
     return basic_blocks
 
 
-def _validate_fold_has_outputs_or_count_filter(fold_scope_location, fold_has_count_filter, outputs):
+def _validate_fold_has_outputs_or_count_filter(
+    fold_scope_location, fold_has_count_filter, metadata
```
obi1kenobi: Let's always try to refer to the metadata table object as `query_metadata_table` , for consistency.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -570,8 +573,8 @@ def _validate_fold_has_outputs_or_count_filter(fold_scope_location, fold_has_cou
 
     # At least one output in the outputs list must point to the fold_scope_location,
     # or the scope corresponding to fold_scope_location had no @outputs and is illegal.
-    for output in six.itervalues(outputs):
-        if output['fold'] == fold_scope_location:
+    for _, output_info in metadata.outputs:
+        if output_info.fold == fold_scope_location:
```
obi1kenobi: Let's try to get rid of the `.fold` attribute, since it was a hack meant to work around the lack of a metadata table. I suspect you'll be able to reconstruct its value from the `.location` attribute that is available on the output info object.

graphql_compiler/compiler/metadata.py
obi1kenobi: I don't think we have a great reason. It's probably fine either way, we can do some cleanup in a future PR.

PR #312, opened 2019-05-22
Add indexes

pmantica1:


graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
```diff
@@ -448,3 +468,67 @@ def _get_graphql_representation_of_non_graph_elements(non_graph_elements, inheri
...
+    for index in index_data:
+        index_name = index['name']
+        index_type = index['type']
+        index_unique = index_type in UNIQUE_INDEX_TYPES
+        index_ordered = index_type in ORDERED_INDEX_TYPES
+        index_definition = index['indexDefinition']
+        index_base_classname = index_definition['className']
+        index_ignore_nulls = index_definition['nullValuesIgnored']
+
+        # Exclude indexes on non-graph classes (e.g. OUser).
```
bojanserafimov: I'd call this orientdb metadata rather than non-graph classes, which mean something else in this repo

graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
```diff
@@ -448,3 +468,67 @@ def _get_graphql_representation_of_non_graph_elements(non_graph_elements, inheri
...
+        if index_base_classname not in all_class_names:
+            continue
+
+        is_edge_index = isinstance(elements[index_base_classname], EdgeType)
+        if is_edge_index:
+            # OrientDB sometimes decides to make edge indexes ignore nulls.
+            # This is not something we want to allow, hard-code it to False.
+            # While this makes us diverge from OrientDB's schema,
+            # it's safe to be *more strict* than OrientDB.
+            index_ignore_nulls = False
```
bojanserafimov: I'm not sure if we should do this. Looks like it's Kensho business logic, and can be fixed in the `index_data` input dict. But also, it might be a sane thing to do for any user of this library, given that we document it.

graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
```diff
@@ -7,18 +7,19 @@
...
-    PROPERTY_TYPE_LINK_ID, parse_default_property_value, try_get_graphql_scalar_type
+    EDGE_SOURCE_PROPERTY_NAME, ORDERED_INDEX_TYPES, ORIENTDB_BASE_EDGE_CLASS_NAME,
+    ORIENTDB_BASE_VERTEX_CLASS_NAME, PROPERTY_TYPE_LINK_ID, UNIQUE_INDEX_TYPES,
+    parse_default_property_value, try_get_graphql_scalar_type
 )
 from .utils import toposort_classes
 
 
-def get_orientdb_schema_graph(schema_data):
+def get_orientdb_schema_graph(schema_data, index_data=None):
```
obi1kenobi: Why make this optional? Seems like we are adding a lot more surface area we have to test, and I don't see what this is buying us.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -166,6 +213,36 @@ def non_graph_class_names(self):
         """Return the set of non-graph class names in the SchemaGraph."""
         return self._non_graph_class_names
 
+    @property
+    def all_indexes(self):
+        """Return the set of all indexes in the graph."""
```
obi1kenobi: probably `in the schema` instead of `in the graph`?

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -166,6 +213,36 @@ def non_graph_class_names(self):
...
+        return self._all_indexes
+
+    @property
+    def unique_indexes(self):
+        """Return the set of all unique indexes in the graph."""
```
obi1kenobi: same here

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -403,10 +480,23 @@ def link_schema_elements(elements, inheritance_structure):
 #              explicit value for this property. Set to None if no default is given in the schema.
 PropertyDescriptor = namedtuple('PropertyDescriptor', ('type', 'default'))
 
-# A way to describe a schema's normalized inheritance structure.
+
+# A way to describe a schema's normalized inheritance structure:
```
obi1kenobi: Please don't modify code or comments unrelated to the current PR, especially for things like whitespace and punctuation. The benefit that introduces is not worth it relative to the cost of making the git history of the file more convoluted.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -110,6 +113,50 @@ def get_edge_schema_element_or_raise(self, edge_classname):
...
+        """Return a frozenset of IndexDefinitions of unique indexes that apply to this class."""
+        return frozenset({
+            index_definition
+            for index_definition in self.get_all_indexes_for_class(cls)
+            if index_definition.unique
+        })
+
+    def get_properties_covered_by_index(self, index_definition, classname, props):
+        """Return the dict of values captured by the index, or None if the index does not apply.
+
```
bojanserafimov: I would add the sentence `An index capturing properties means that it prevents these property values from appearing in two different entries`.

graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -403,10 +480,23 @@ def link_schema_elements(elements, inheritance_structure):
...
 InheritanceStructure = namedtuple('InheritanceStructure', ('superclass_sets', 'subclass_sets'))
+
+
+# A way to describe an index:
+#   - name: string, the name of the index.
+#   - base_classname: string, the name of the class on which the index is defined.
+#   - fields: set of strings, the set of fields which the index encompasses.
+#   - unique: bool, indicating whether this index is unique.
+#   - ignore_nulls: bool, indicating if the index ignores null values.
+#   - ordered: bool, indicating whether this index is ordered.
```
obi1kenobi: I'd keep `unique` and `ordered` adjacent to each other, and move `ignore_nulls` either above or below them, since that one is the unusual property here.

graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
obi1kenobi: Yeah it's a good question. I think I originally wrote this code when I realized that OrientDB across different versions in the 2.x series had inconsistent defaults on whether nulls are allowed in unique indexes that apply to edges. In general, though, `null` values on edge endpoints don't make sense -- they would imply that you could have edges with only one endpoint, or worse, no endpoints.

So overall, I am leaning toward leaving it in, but I am open to being convinced otherwise.

graphql_compiler/schema_generation/schema_graph.py
pmantica1: Is this always true? What if the unique field is false?

obi1kenobi: If we change `captured` to `covered`, we'd be using the standard nomenclature for this and we probably wouldn't need additional explanation.

graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
bojanserafimov: but this only relates to edge endpoints. It still makes sense to ignore nulls on other edge properties, no? Since we already assert you can only ignore nulls if the index is defined on a single column, we should at least check if that column is an endpoint, and only apply the rule in that case.

obi1kenobi: There are some subtleties of OrientDB's index description format that we are glossing over, but I don't think it's worth going into them. Let's just remove this entire block, since it's the definition of legacy code :)

graphql_compiler/schema_generation/orientdb/schema_graph_builder.py
```diff
@@ -448,3 +468,59 @@ def _get_graphql_representation_of_non_graph_elements(non_graph_elements, inheri
...
+            name=index_name,
+            base_classname=index_base_classname,
+            fields=index_fields,
+            unique=index_unique,
+            ignore_nulls=index_ignore_nulls,
+            ordered=index_ordered,
+        )
+        all_indexes.add(definition)
+
+    frozenset(all_indexes)
```
obi1kenobi: This line doesn't do anything. Did you mean to return this instead?

graphql_compiler/tests/test_helpers.py
```diff
@@ -274,7 +274,7 @@ def generate_schema_graph(graph_client):
     """Generate SchemaGraph from a pyorient client"""
     schema_records = graph_client.command(ORIENTDB_SCHEMA_RECORDS_QUERY)
     schema_data = [x.oRecordData for x in schema_records]
-    return get_orientdb_schema_graph(schema_data)
+    return get_orientdb_schema_graph(schema_data, set())
```
obi1kenobi: I assume you're going to make a separate PR to update all of these locations to actually fetch index data?

pmantica1: Yes, I will.

PR #313, opened 2019-05-22
Schema for macro definitions

ReaLNeroM:


graphql_compiler/tests/test_macro_schema.py
```diff
@@ -32,3 +34,27 @@ def test_get_schema_with_macros_basic(self):
...
+
+    def test_get_schema_for_macro_definition_unchanged(self):
+        original_schema = self.macro_registry.schema_without_macros
+        original_printed_schema = print_schema(original_schema)
+        macro_definition_schema = get_schema_for_macro_definition(original_schema)
+        macro_definition_printed_schema = print_schema(macro_definition_schema)
+        required_directives = get_required_macro_edge_directives()
+        # pylint: disable=protected-access
+        for directive in required_directives:
+            self.assertTrue(directive in macro_definition_schema._directives)
```
pmantica1: Please use get_directives and remove the don't disable pylint.

graphql_compiler/tests/test_macro_schema.py
```diff
@@ -32,3 +34,27 @@ def test_get_schema_with_macros_basic(self):
...
+
+    def test_get_schema_for_macro_definition_basic(self):
+        schema_with_macros = get_schema_with_macros(self.macro_registry)
+        printed_schema_with_macros = print_schema(schema_with_macros)
+        macro_definition_schema = get_schema_for_macro_definition(schema_with_macros)
+        macro_definition_printed_schema = print_schema(macro_definition_schema)
+        required_directives = get_required_macro_edge_directives()
+        # pylint: disable=protected-access
+        for directive in required_directives:
+            self.assertTrue(directive in macro_definition_schema._directives)
```
pmantica1: Same as above.

graphql_compiler/macros/__init__.py
```diff
@@ -176,7 +181,16 @@ def get_schema_for_macro_definition(schema):
     Returns:
         GraphQLSchema usable for writing macros
     """
-    raise NotImplementedError()  # TODO(bojanserafimov): Implement
+
+    required_macro_directives = (MacroEdgeDefinitionDirective, MacroEdgeTargetDirective)
+
+    # pylint: disable=protected-access
+    schema_for_macro_definition = copy(schema)
+    schema_for_macro_definition._directives = list(chain(
```
pmantica1: What if the macros are already defined in the schema?

graphql_compiler/macros/__init__.py
```diff
@@ -176,7 +181,16 @@ def get_schema_for_macro_definition(schema):
     Returns:
         GraphQLSchema usable for writing macros
     """
-    raise NotImplementedError()  # TODO(bojanserafimov): Implement
+
+    required_macro_directives = (MacroEdgeDefinitionDirective, MacroEdgeTargetDirective)
```
pmantica1: @bojanserafimov Why don't we just add macro directives in all schemas?

bojanserafimov: We don't want the graphql editor to tell the user to maybe define a macro in the middle of their query, because that doesn't work.

graphql_compiler/macros/__init__.py
```diff
@@ -176,7 +181,16 @@ def get_schema_for_macro_definition(schema):
...
+
+    # pylint: disable=protected-access
+    schema_for_macro_definition = copy(schema)
+    schema_for_macro_definition._directives = list(chain(
+        schema_for_macro_definition._directives, required_macro_directives))
```
bojanserafimov: Can we also not include the output and output_source directives?

graphql_compiler/tests/test_macro_schema.py
```diff
@@ -32,3 +34,27 @@ def test_get_schema_with_macros_basic(self):
...
+        # pylint: enable=protected-access
+        self.assertNotEqual(printed_schema_with_macros, macro_definition_printed_schema)
```
bojanserafimov: A good test to have is to see that a macro passes validation with this schema.

graphql_compiler/tests/test_helpers.py
```diff
@@ -407,3 +408,8 @@ def get_test_macro_registry():
     for graphql, args in valid_macros:
         register_macro_edge(macro_registry, graphql, args)
     return macro_registry
+
+
+def get_required_macro_edge_directives():
+    """Return a tuple containing the directives necessary for defining a macro edge"""
+    return (MacroEdgeDefinitionDirective, MacroEdgeTargetDirective)
```
bojanserafimov: Great idea materializing this tuple. But it might be better to move it `macros/directives.py`, as a constant. This is mostly for consistency with the codebase (that's where people would expect to find it), but also because this information is used outside of tests too.

graphql_compiler/tests/test_macro_schema.py
ReaLNeroM: Good point! Fixed in commit "Address comments."

graphql_compiler/tests/test_macro_schema.py
ReaLNeroM: Fixed in commit "Address Comments." x2

graphql_compiler/macros/__init__.py
ReaLNeroM: In the next commit, the list is first converted to a set, to eliminate duplicates.

graphql_compiler/tests/test_helpers.py
ReaLNeroM: Fixed in commit "Address comments". I added sets in macro_edges/directives.py for allowed & disallowed directives in a macro, as well as directives necessary for defining a macro.

graphql_compiler/macros/__init__.py
ReaLNeroM: Good point. Added in commit "Finish adding sets ..."

graphql_compiler/macros/__init__.py
```diff
@@ -167,16 +171,25 @@ def get_schema_for_macro_definition(schema):
     """Returns a schema with macro directives.
 
     This returned schema can be used to validate macro definitions, and support GraphQL
-    macro editors, enabling them to autocomplete one the @macro_edge_definition and
-    @macro_edge_target directives.
+    macro editors, enabling them to autocomplete on the @macro_edge_definition and
+    @macro_edge_target directives.*
```
bojanserafimov: Oopsie. Doing `git diff --staged` before you commit will tell you what's going in, and you can spot these kinds of things.

graphql_compiler/macros/__init__.py
```diff
@@ -167,16 +171,25 @@ def get_schema_for_macro_definition(schema):
...
     Args:
         schema: GraphQLSchema over which we want to write macros
 
     Returns:
         GraphQLSchema usable for writing macros
     """
-    raise NotImplementedError()  # TODO(bojanserafimov): Implement
+
+    macro_definition_schema = copy(schema)
+    macro_definition_schema_directives_set = set(chain(macro_definition_schema.get_directives(),
```
bojanserafimov: I would replace all directives with the usual ones, instead of keeping the ones defined in this schema. The macro system works only with our usual directives, and makes the extra effort to make sure that all the folds, tags and recurses work as expected. If this schema has anything other than the usual directives, then we can't promise the macro system will work at all.

If you do this, also make sure to either document that all directives are replaced, or add a precondition and assertion that this schema uses the usual directives.

graphql_compiler/macros/macro_edge/directives.py
```diff
@@ -36,9 +41,26 @@
...
+
+MACRO_EDGE_DISALLOWED_SUBDIRECTIVES = frozenset({
+    OutputDirective.name,
+    OutputSourceDirective.name,
+})
+
+MACRO_EDGE_DEFINITION_DIRECTIVES = frozenset({
+    MacroEdgeDefinitionDirective,
+    MacroEdgeTargetDirective
+})
```
bojanserafimov: It would be nice to document in one line what these mean. In what context are they allowed/disallowed? Someone might read this as "directives you are allowed to use on usage of macro edges in queries". See `_check_that_expansion_directives_are_supported` in `macro_expansion.py`.

graphql_compiler/macros/__init__.py
ReaLNeroM: That was a typo I fixed. I didn't mention that in my commit message -- I'll do that from now on!

graphql_compiler/macros/__init__.py
ReaLNeroM: I'm thinking of asserting that all directives are the default rather than adding a precondition. Is this a performance concern?

bojanserafimov: nope, go for it. It's only a concern if this function is called often, but it shouldn't.

PR #315, opened 2019-05-23
Add indexes to test SchemaGraph

pmantica1:


graphql_compiler/tests/test_helpers.py
```diff
@@ -274,7 +276,9 @@ def generate_schema_graph(graph_client):
     """Generate SchemaGraph from a pyorient client"""
     schema_records = graph_client.command(ORIENTDB_SCHEMA_RECORDS_QUERY)
     schema_data = [x.oRecordData for x in schema_records]
-    return get_orientdb_schema_graph(schema_data, set())
+    index_records = graph_client.command(ORIENTDB_INDEX_RECORDS_QUERY)
+    index_query_data = [x.oRecordData for x in index_records]
```
evantey14: small thing: looks like this is an empty list if there are no `index_records` but before you were passing in an empty set? also in `get_orientdb_schema_graph` looks like you check if the index data is `None`. Seems like it'll work either way, but (unless I'm missing something?) might be good to neaten up the code now / in the future.

PR #316, opened 2019-05-23
Add lists for required and allowed directives within a macro edge definition

ReaLNeroM:


pmantica1:
Why did you decide to open a new branch that is essentially a copy of this old one: https://github.com/kensho-technologies/graphql-compiler/pull/313 ?

In general, please try to avoid doing this. It is quite helpful to see the comments that have been previously written for this branch and how you've addressed them. Also, it seems that you are missing the tests that you had earlier?

graphql_compiler/macros/macro_edge/directives.py
```diff
@@ -36,9 +38,23 @@
...
 )
+
+# Directives required present in a macro edge definition
+DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION = frozenset({
+    MacroEdgeDefinitionDirective,
+    MacroEdgeTargetDirective
+})
+
+# Directives tolerated without restrictions within a macro edge definition
+DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION = frozenset({
```
pmantica1: This code is inconsistent. The set above uses actual directives while in this set you include strings yet both names start with "DIRECTIVE". Please either change the naming to DIRECTIVE_NAMES_ALLOWED_IN_MACRO_EDGE_DEFINTION or change the types of the set elements.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -78,24 +74,12 @@ def _validate_macro_ast_directives(ast, inside_fold_scope=False):
...
-    disallowed_directives = frozenset({
-        OutputDirective.name,
-        OutputSourceDirective.name,
-    })
     for directive in ast.directives:
         name = directive.name.value
-        if name in directives_with_no_restrictions:
+        if name in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION:
+            pass
+        elif name == MacroEdgeDefinitionDirective.name:
```
pmantica1: Shouldn't the MacroEdgeDefinition directive be in the DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINTION set?

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -78,24 +74,12 @@ def _validate_macro_ast_directives(ast, inside_fold_scope=False):
...
             pass
-        elif name in disallowed_directives:
-            raise GraphQLInvalidMacroError(u'Macros are not allowed to use the {} directive. '
-                                           u'Found usage {}'.format(name, directive))
         elif name == FoldDirective.name:
```
pmantica1: Same as above,

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -78,24 +74,12 @@ def _validate_macro_ast_directives(ast, inside_fold_scope=False):
...
             subselection_inside_fold_scope = True
         elif name == MacroEdgeTargetDirective.name:
```
pmantica1: Same as above.

ReaLNeroM:
> Why did you decide to open a new branch that is essentially a copy of this old one: #313 ?
> 
> In general, please try to avoid doing this. It is quite helpful to see the comments that have been previously written for this branch and how you've addressed them. Also, it seems that you are missing the tests that you had earlier?

The old PR addressed generating schemas for macro definitions, while this new PR addresses the `magic directive lists` throughout validation.py, solidifying them in directives.py. These two aren't really related -- my old PR was just bloated. I'll first wait for this to get merged, then I'll work on the other PR.

graphql_compiler/macros/macro_edge/directives.py
ReaLNeroM: Good point. This is fixed in the next commit.

graphql_compiler/macros/macro_edge/validation.py
ReaLNeroM: Renamed .._ALLOWED_... to .._OPTIONAL_.. to better indicate the directives listed are not related to macros.

graphql_compiler/macros/macro_edge/validation.py
ReaLNeroM: Now that the directive list has been renamed to OPTIONAL_DIRECTIVES_UNRESTRICTED_..., FoldDirective clearly doesn't belong in the set, as right here we check whether certain usage conditions for FoldDirective are met.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -78,24 +75,16 @@ def _validate_macro_ast_directives(ast, inside_fold_scope=False):
...
-        TagDirective.name,
-        OptionalDirective.name,
-        RecurseDirective.name,
-        MacroEdgeDefinitionDirective.name,
-    })
-    disallowed_directives = frozenset({
-        OutputDirective.name,
-        OutputSourceDirective.name,
-    })
+    optional_directive_names = [directive.name
```
obi1kenobi: Two things here:
- When a comprehension of any kind (list, set, dict, ...) requires multiple lines, we always prefer the following formatting:
```
foo = [
    bar
    for bar in baz
    if quux
]
```
- On line 84 below, you use the `in` operator against this variable. Why not make it a set instead of using a list? Let's not pay `O(n)` runtime costs if we don't have to.

obi1kenobi: One more -- reconsider the name `optional_directive_names`. Ignoring the code that computes its value, which of these do you think the reader will think it refers to:
- names of the `@optional` directive
- names of directives which are somehow optional
- names, if they happen to exist, of directives

graphql_compiler/macros/macro_edge/directives.py
```diff
@@ -36,9 +38,24 @@
...
+
+# Directives required present in a macro edge definition
+DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION = frozenset({
+    MacroEdgeDefinitionDirective,
+    MacroEdgeTargetDirective
+})
+
+# Directives (excluding MACRO_EDGE_DIRECTIVES) tolerated without restrictions within a macro
+# edge definition
+OPTIONAL_DIRECTIVES_UNRESTRICTED_IN_MACRO_EDGE_DEFINITION = frozenset({
```
obi1kenobi: I don't think the `OPTIONAL` is helpful here. How about `UNRESTRICTED_DIRECTIVES_IN_MACRO_EDGE_DEFINITION` plus the comment you wrote right above it?

Naming is hard. But it's also very important.

ReaLNeroM: - If we're working with macro edges and directives, we'll need three sets: all allowed, required for macro-edge, and non-macro-edge (optional) specific. Since we're working with sets, given two of these, we can find the third using either a union or difference. So we should implement two. Just by going over the functions we have now, the first two seem to be most useful. So I agree, the optional directives set isn't needed.
- I don't think an UNRESTRICTED directives set should stay here. How can we communicate FoldDirective is allowed within macro edges? I think the better solution is to create an allowed directives set, whether they have restrictions or not, and then manually check each directive for its necessary conditions (FoldDirectives, MacroEdgeDefinition, MacroEdgeTarget).

graphql_compiler/macros/macro_edge/validation.py
ReaLNeroM: - Since there's now an allowed directive list, I've renamed this variable to names_of_allowed_directives.

PR #319, opened 2019-05-27
Schema for macro definitions

ReaLNeroM:
In this PR, get_schema_for_macro_definitions is implemented. 
Here's my methodology: 

- We check for non-default directives, and report an error.
- We add the directives required for a macro edge definition(@macro_edge_definition and @macro_edge-target). 
- Then any directives not in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION are removed from the set.
- We create a shallow copy of the schema, where the directives list is replaced by a list containing the elements of the set. @reviewers: Could the shallow copy cause issues, if for example, the macro editor modifies the schema? Is this a vulnerability? EDIT: Updated the function docstring to reflect modifying the schema is **undefined behavior**

### DIFF
--- a/graphql_compiler/macros/__init__.py
+++ b/graphql_compiler/macros/__init__.py
@@ -175,6 +175,7 @@ def get_schema_for_macro_definition(schema):
     macro editors, enabling them to autocomplete on the @macro_edge_definition and
     @macro_edge_target directives. Some directives that are disallowed in macro edge definitions,
     like @output and @output_source, will be removed from the directives list.
+    Raises an error if any non-default directives are present in the given schema.
 
     Args:
         schema: GraphQLSchema over which we want to write macros
--- a/graphql_compiler/macros/__init__.py
+++ b/graphql_compiler/macros/__init__.py
@@ -10,7 +10,6 @@ from graphql import parse
 from graphql.language.printer import print_ast
 from graphql.utils.build_ast_schema import build_ast_schema
 from graphql.utils.schema_printer import print_schema
-from graphql_compiler.schema import DIRECTIVES
 
 from ..compiler.compiler_frontend import validate_schema_and_ast
 from ..ast_manipulation import safe_parse_graphql
@@ -20,7 +19,7 @@ from .macro_edge.directives import (MacroEdgeDirective,
                                     DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION,
                                     DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION)
 from .macro_expansion import expand_macros_in_query_ast
-from ..exceptions import GraphQLInvalidMacroError, GraphQLValidationError, GraphQLCompilationError
+from ..exceptions import GraphQLInvalidMacroError, GraphQLValidationError
 
 
 MacroRegistry = namedtuple(
@@ -170,9 +169,8 @@ def get_schema_with_macros(macro_registry):
 def get_schema_for_macro_definition(schema):
     """Returns a schema with macro directives.
 
-    Precondition:
+    Preconditions:
     1. No user-created directives are in schema.
-
     This returned schema can be used to validate macro definitions, and support GraphQL
     macro editors, enabling them to autocomplete on the @macro_edge_definition and
     @macro_edge_target directives. Some directives that are disallowed in macro edge definitions,
@@ -187,12 +185,11 @@ def get_schema_for_macro_definition(schema):
 
     macro_definition_schema = copy(schema)
     macro_definition_schema_directives_set = set(macro_definition_schema.get_directives())
+
     # Include required directives
-    macro_definition_schema_directives_set += DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION
+    macro_definition_schema_directives_set |= set(DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION)
     # Exclude disallowed directives
-    for directive in macro_definition_schema_directives_set:
-        if directive.name not in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION:
-            macro_definition_schema_directives_set.remove(directive)
+    macro_definition_schema_directives_set &= set(DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION)
 
     # pylint: disable=protected-access
     macro_definition_schema._directives = list(macro_definition_schema_directives_set)
--- a/graphql_compiler/macros/__init__.py
+++ b/graphql_compiler/macros/__init__.py
@@ -17,7 +17,7 @@ from .macro_edge import make_macro_edge_descriptor
 from .macro_edge.helpers import get_type_at_macro_edge_target
 from .macro_edge.directives import (MacroEdgeDirective, DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION,
                                     DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION)
-from ..schema import _check_for_nondefault_directives
+from ..schema import _check_for_nondefault_directive_names
 from .macro_expansion import expand_macros_in_query_ast
 from ..exceptions import GraphQLInvalidMacroError, GraphQLValidationError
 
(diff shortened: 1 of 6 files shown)


graphql_compiler/macros/__init__.py
```diff
@@ -167,16 +170,26 @@ def get_schema_for_macro_definition(schema):
...
+    like @output and @output_source, will be removed from the directives list.
 
     Args:
         schema: GraphQLSchema over which we want to write macros
 
     Returns:
         GraphQLSchema usable for writing macros
     """
-    raise NotImplementedError()  # TODO(bojanserafimov): Implement
+    macro_definition_schema = copy(schema)
```
pmantica1: The copying seems fine to me. I don't understand what you mean by vulnerabilities.

graphql_compiler/macros/__init__.py
```diff
@@ -167,16 +170,26 @@ def get_schema_for_macro_definition(schema):
...
+    macro_definition_schema_directives_set = set(macro_definition_schema.get_directives()).union(
+        DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION)
+    # Exclude disallowed directives
+    macro_definition_schema_directives_set &= DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION
```
pmantica1: I think it's better to validate that all of the current schema directives are allowed and raise a validation error if one of them isn't. (Look for the exception that fits best). Then I would add a test that validation works and update the docstring with a "Raises" block.

pmantica1: If you would continue with your current approach I would separate this into functions: one to remove the unsupported directives and another to add the required directives.

bojanserafimov: Agree with Pedro on raising an exception. The macro system is not a macro system for any graphql schema, but specifically a macro system for this graphql compiler. If it's being used with anything else, it should not promise to work and pretend that everything is ok.

graphql_compiler/macros/__init__.py
ReaLNeroM: The returned schema's fields are the same copy as the argument schema's fields. So if a macro editor edits the schema returned from this function, we might run into issues. How can we make the fields immutable?

graphql_compiler/macros/__init__.py
ReaLNeroM: Added schema.py/_check_for_nondefault_macro_directives() to detect this issue.

graphql_compiler/macros/__init__.py
```diff
@@ -167,16 +172,30 @@ def get_schema_for_macro_definition(schema):
...
         GraphQLSchema usable for writing macros
     """
-    raise NotImplementedError()  # TODO(bojanserafimov): Implement
+
+    macro_definition_schema = copy(schema)
+    macro_definition_schema_directives = schema.get_directives()
+    _check_for_nondefault_macro_directives(macro_definition_schema_directives)
+    macro_definition_schema_directives = include_required_macro_directives(
+        macro_definition_schema_directives)
+    macro_definition_schema_directives = exclude_disallowed_directives_in_macros(
```
pmantica1: Why are we excluding disallowed directives? Didn't Bojan and I agree on just asserting that we simply raise an error if any of schema directives is not allowed?

ReaLNeroM: Some default directives, like @output and @output_source are disallowed. However, if we raised errors, we wouldn't be able to use @output and macro edges in the same schema.

graphql_compiler/schema.py
```diff
@@ -349,3 +349,15 @@ def insert_meta_fields_into_existing_schema(graphql_schema):
                                      .format(meta_field_name))
 
             type_obj.fields[meta_field_name] = meta_field
+
+
+def _check_for_nondefault_macro_directives(directives):
```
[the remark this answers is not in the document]
ReaLNeroM: The function can either leave a directive as-is, remove it from the schema(@output), or raise an error(any user-created directive). Since we don't want a DEFAULT_DIRECTIVES_DISALLOWED_IN_MACRO_EDGE_DEFINITION per [this comment](https://github.com/kensho-technologies/graphql-compiler/pull/313#commitcomment-33655951), we need whether all directives are default.

graphql_compiler/schema.py
```diff
@@ -349,3 +349,17 @@ def insert_meta_fields_into_existing_schema(graphql_schema):
                                      .format(meta_field_name))
 
             type_obj.fields[meta_field_name] = meta_field
+
+
+def _check_for_nondefault_directives(directives):
+    """Check if any user-created directives are present"""
+    deprecated_directive_names = {'deprecated', 'skip', 'include'}
```
bojanserafimov: These directives are not deprecated. They're the default.

graphql_compiler/schema.py
```diff
@@ -349,3 +349,17 @@ def insert_meta_fields_into_existing_schema(graphql_schema):
...
+    default_directive_names = {directive.name for directive in DIRECTIVES}
```
bojanserafimov: Might be a good idea to check these for equality, instead of just comparing names. It's possible that someone defines a different `@optional` directive for example. Not likely to happen, but if it does it will be difficult to find because we think we are addressing the problem.

graphql_compiler/tests/test_macro_schema.py
```diff
@@ -32,3 +37,115 @@ def test_get_schema_with_macros_basic(self):
...
+            }''',
+            '''{
+                Animal @macro_edge_definition(name: "out_Animal_RelatedEntity") {
+                    in_Entity_Related {
+                        ... on Entity @macro_edge_target {
+                            uuid
+                        }
+                    }
+                }
+            }''']
```
bojanserafimov: You can import these from `test_helpers.py`

graphql_compiler/tests/test_macro_schema.py
```diff
@@ -32,3 +37,115 @@ def test_get_schema_with_macros_basic(self):
...
+        for directive in original_schema.get_directives():
+            if directive in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION:
+                self.assertTrue(directive in macro_definition_schema.get_directives())
+
+    def test_get_schema_for_macro_definition_removal(self):
+        schema_with_macros = get_schema_with_macros(self.macro_registry)
+        macro_definition_schema = get_schema_for_macro_definition(schema_with_macros)
+        for directive in macro_definition_schema.get_directives():
+            self.assertTrue(directive.name != '@output')
+            self.assertTrue(directive.name != '@output_source')
```
bojanserafimov: `directive.name` is a `Name` object. You can get it's string contents by calling `directive.name.value`. If you cargo-culted this code from somewhere else, please check that we are not making the same mistake there too :) It's possible.

bojanserafimov: The `@` symbol will not be in the directive name. You can use `OutputSourceDirective.name` to avoid magic strings, and having to remember this.

graphql_compiler/schema.py
ReaLNeroM: Directive definitions include a description, but the argument directives do not. If we check for equality, the descriptions get compared as well. 
What if we just create a copy of DIRECTIVES, with descriptions nulled out? What about deprecated, skip, and include then?
It's way too adversarial to assume users will create their own directives, with the same name as the ones we've reserved.

graphql_compiler/tests/test_macro_schema.py
ReaLNeroM: directive.name is a str. Check again.

graphql_compiler/schema.py
ReaLNeroM: Changed the function name to reflect it only compares names.
Get_schema_for_macro_definitons now has a precondition assuming all directives we support like @fold, @output, etc. have their default behavior.

graphql_compiler/schema.py
```diff
@@ -291,7 +292,6 @@ def _parse_datetime_value(value):
     FoldDirective,
 )
 
-
```
pmantica1: nit: Please don't change line formatting that is unrelated to this pr.

graphql_compiler/tests/test_macro_schema.py
bojanserafimov: Ah, correct. It's only a `Name` in directive instances.

PR #322, opened 2019-05-28
Make sure the tag is not reordered afer the filter

bojanserafimov:
Just a failing test case. Fix coming soon, in this PR.


obi1kenobi:
Very interesting, nice catch! Were we not guaranteeing the order of clauses within the macro expansion? That sounds like something we should do (just like how we guarantee the order of data output clauses), if nothing else then just for the sake of having deterministic outputs and non-flaky tests.

bojanserafimov:
> Very interesting, nice catch! Were we not guaranteeing the order of clauses within the macro expansion? That sounds like something we should do (just like how we guarantee the order of data output clauses), if nothing else then just for the sake of having deterministic outputs and non-flaky tests.

Yes. Currently when you call a macro, the macro gets expanded at that specific location, with any "extra selections" prepended. I'm guessing this was to preserve the original look of the query. The right thing to do is to preserve the original relative ordering of selections in the macro. I'll get rid of the concept of extra selections in the process of fixing it.

bojanserafimov:
If the user and the macro tag and filter on the same two fields, but in different order, then there's no way to expand the macro. We can either document this as a limitation or fix it by allowing tag-filter cycles in the queries. @obi1kenobi thoughts?

obi1kenobi:
I'd say document it as a limitation for now. I don't think it will come into play much, and it's non-trivial to fix: we'd need to do a topological sort between the tags and filters and account for handling of things like folds and optional traversals. I don't think the engineering effort is best spent solving that problem at the moment, so leave it as a limitation for now.

graphql_compiler/tests/test_macro_expansion.py
```diff
@@ -118,9 +118,9 @@ def test_macro_edge_source_merging(self):
 
         expected_query = '''{
             Animal {
-                net_worth @tag(tag_name: "net_worth")
-                          @filter(op_name: "<", value: ["$net_worth_upper_bound"])
+                net_worth @filter(op_name: "<", value: ["$net_worth_upper_bound"])
                           @output(out_name: "net_worth")
+                          @tag(tag_name: "net_worth")
```
bojanserafimov: Macro directives now come after non-macro directives as a result.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -323,18 +330,15 @@ class name -> (macro edge name -> MacroEdgeDescriptor)
                     macro_registry, inherited_macro_edges, vertex_field_type,
                     new_selection_ast, new_query_args, tag_names)
 
-        if new_selection_ast is not selection_ast:
+        if new_selection_ast is not selection_ast or selections_before or selections_after:
```
obi1kenobi: Could we have a situation where `new_selection_ast is not selection_ast` is False but the rest of this is `True`? Am I missing something, or should this maybe be an `AssertionError`? If there is such a case, it's probably a good idea to both describe it explicitly in a comment here, and also include a test case for it, since it's far from obvious.

bojanserafimov:
Since now the order of tags and filters within the same vertex does not matter, all we have to do is make sure property fields come before vertex fields, and that the order of vertex fields in the query and in the macro is preserved after expansion.

graphql_compiler/tests/test_helpers.py
```diff
@@ -389,6 +389,23 @@ def get_test_macro_registry():
                 }
             }
         }''', {}),
+        # The same as out_AnimalRichYoungerSiblings, but with a filter after the target.
+        ('''{
+            Animal @macro_edge_definition(name: "out_Animal_RichYoungerSiblings_2") {
```
obi1kenobi: Oh dang, that's a wild one... I never even thought of this.

@lodrion @eyurtsev check out this macro edge definition.

graphql_compiler/macros/macro_expansion.py
```diff
@@ -331,7 +331,12 @@ class name -> (macro edge name -> MacroEdgeDescriptor)
                     macro_registry, inherited_macro_edges, vertex_field_type,
                     new_selection_ast, new_query_args, tag_names)
 
-        if new_selection_ast is not selection_ast or selections_before or selections_after:
+        if new_selection_ast is selection_ast and (selections_before or selections_after):
+            raise AssertionError(u'No macro expansion happened but there are selections '
```
obi1kenobi: Do you think we could somehow work in the macro we thought we were expanding here, and what field we were in / which field in there we were expanding? Otherwise this would be pretty hard to debug, if it ever happened.

PR #323, opened 2019-05-28
Improve getting started documentation

pmantica1:


graphql_compiler/tests/conftest.py
```diff
@@ -24,7 +24,6 @@ def init_snapshot_graph_client():
     return _init_graph_client(load_schema, generate_orient_snapshot_data)
 
 
-@pytest.fixture(scope='session')
```
pmantica1: I had to remove this as a fixture to fix the test.

README.md
```diff
@@ -109,24 +122,16 @@ parameters = {
     'net_worth': '100',
 }
 
-# Step 4: Use autogenerated GraphQL schema to compile query into Match, an OrientDB query language
+# Step 3: Use autogenerated GraphQL schema to compile query into the target database language.
 compilation_result = graphql_to_match(schema, graphql_query, parameters, type_equivalence_hints)
 
-# Step 5: Run query in OrientDB
+# Step 5: Run compiled compiled query.
```
obi1kenobi: `compiled compiled`

graphql_compiler/tests/conftest.py
obi1kenobi: What's the reasoning for removing this? Doesn't this mean that the client will get initialized for every single test class annotated with the other fixture?

Also, I'm not a fan of changing test setup together with changing documentation.

pmantica1: So this is a bug fix. The README.md example uses this as a function but it's a fixture. 

How about I also create an Init_integration_graph_client_fixture function to avoid initializing the client every single test case.

README.md
obi1kenobi: This is presumably `Step 4` now?

README.md
```diff
@@ -76,27 +71,72 @@ For a more detailed overview and getting started guide, please see
    these databases. See the [SQL](#sql) section for more details.
 * **GraphQL Language Features:**  We prioritized and implemented a subset of all functionality supported by the GraphQL language. We hope to add more functionality over time.
 
-## Querying OrientDB with GraphQL
+## Getting Started
+The first thing you'll need is to build a GraphQL schema. The GraphQL schema is a GraphQL
```
obi1kenobi: I think `to build` can go away and the rest of the sentence remains true.

README.md
```diff
@@ -76,27 +71,72 @@ For a more detailed overview and getting started guide, please see
    these databases. See the [SQL](#sql) section for more details.
 * **GraphQL Language Features:**  We prioritized and implemented a subset of all functionality supported by the GraphQL language. We hope to add more functionality over time.
 
-## Querying OrientDB with GraphQL
+## Getting Started
+The first thing you'll need is to a GraphQL schema. The GraphQL schema is a GraphQL
```
obi1kenobi: `to a`

README.md
ReaLNeroM: Nit: "First, you'll need to ..."

README.md
```diff
@@ -97,45 +97,17 @@ from graphql_compiler import (
     get_graphql_schema_from_orientdb_schema_data, graphql_to_match
 )
 from graphql_compiler.schema_generation.orientdb.utils import ORIENTDB_SCHEMA_RECORDS_QUERY
-from pyorient import OrientDB
-from pyorient.constants import DB_TYPE_GRAPH
-from pyorient.ogm import Config, Graph
 
+# This code is a mock example. It is not executable. It assumes that:
+# - The client variable refers to a pyorient client.
```
obi1kenobi: Why not express this through code?

```
client = your_function_that_returns_an_orientdb_client()

< do stuff with the client >

schema, type_equivalence_hints = get_graphql_schema_from_orientdb_schema_data(schema_data)

print(print_schema(schema))
# Assume that the above has output a type definition that looks like this:
# type Animal extends Entity 
# ... < copy the type definition from the tests here >

# Step 2: Write GraphQL query that does your desired operation, in this case getting the names of all animals.
...
```

README.md
```diff
@@ -453,7 +422,7 @@ The database captures the value of the parent animal's name as the `parent_name`
 value is then used as the `%parent_name` tagged parameter in the child animal's `@filter`.
 
 We considered and **rejected** the idea of allowing literal values (e.g. `123`)
-as `@filter` parameters, for several reasons:
+as `@filter` parameters, for several r thanreasons:
```
obi1kenobi: this seems like an unintentional typo

README.md
pmantica1: Ok sounds like a good idea.  Will do.

README.md
```diff
@@ -76,27 +71,73 @@ For a more detailed overview and getting started guide, please see
    these databases. See the [SQL](#sql) section for more details.
 * **GraphQL Language Features:**  We prioritized and implemented a subset of all functionality supported by the GraphQL language. We hope to add more functionality over time.
 
-## Querying OrientDB with GraphQL
+## Getting Started
```
MichaelaShtilmanMinkin: This may be personal preference, but if there are x steps I need to take towards getting started, I would prefer they would either in an ordered list and/or in distinct sections - kind of like treating it as a checklist. The former (enumerated or bulleted list) is probably the better option here since distinct sections should probably only get used when you have more complicated and/or detailed steps.

README.md
```diff
@@ -76,27 +71,73 @@ For a more detailed overview and getting started guide, please see
...
-# unless you are in the development environment outlined by CONTRIBUTING.md.
-
-# Step 1: Initialize dummy OrientDB database and get pyorient OrientDB client
-client = init_integration_graph_client()
 
-# Step 2: Generate GraphQL schema from queried OrientDB schema records
+# Step 1: Get schema metadata from hypothetical Animals database.
+client = your_function_that_returns_an_orientdb_client()
 schema_records = client.command(ORIENTDB_SCHEMA_RECORDS_QUERY)
 schema_data = [x.oRecordData for x in schema_records]
```
MichaelaShtilmanMinkin: Since this is written to be read I would replace `x` with `record`.

README.md
pmantica1: I am not sure how to add the ordered list without being too verbose. If added the ordered list, I would still have to have comments in the code that separated it into different steps.

MichaelaShtilmanMinkin: I think my main reservation with this section is that it doesn't actually help with getting started - there's no install, config, etc process. This section tells me that I need a GraphQL schema and that I will need a query. This is more of a this-is-what-you'll-need section.

pmantica1: I agree with your comment. I am going to refactor this section.

pmantica1: I removed the "Getting Started" section for now and reduced the scope of this PR.

PR #324, opened 2019-05-28
Refactor and clean up IR lowering code. Add common IR lowering module.

obi1kenobi:
This PR refactors the IR lowering code and extracts backend-agnostic lowering code into its own module, fixing multiple small issues along the way:
- Refactor location rewriting logic so that it is backend agnostic, and move it to this new module.
- Remove the last vestiges of the pre-`QueryMetadataTable` location type passing code in the MATCH backend.
- Ensure Gremlin correctly rewrites location objects, no matter what kind of Expression they are part of.
- Fix some trivial bugs and documentation mismatches related to IR Expression objects.

This the set of issues I came across while prototyping Cypher support, and I extracted these from that branch so that these could be merged earlier.

### DIFF
--- a/graphql_compiler/compiler/ir_lowering_gremlin/ir_lowering.py
+++ b/graphql_compiler/compiler/ir_lowering_gremlin/ir_lowering.py
@@ -1,27 +1,32 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Perform optimizations and lowering of the IR that allows the compiler to emit Gremlin queries.
 
-The compiler IR allows blocks and expressions that cannot be directly compiled to Gremlin or MATCH.
-For example, ContextFieldExistence is an Expression that returns True iff its given vertex exists,
-but the produced Gremlin and MATCH outputs for this purpose are entirely different and not easy
-to generate directly from this Expression object. An output-language-aware IR lowering step allows
-us to convert this Expression into other Expressions, using data already present in the IR,
-to simplify the final code generation step.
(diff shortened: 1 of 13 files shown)


graphql_compiler/compiler/ir_lowering_gremlin/ir_lowering.py
```diff
@@ -148,16 +153,19 @@ def validate(self):
                     u'Allowed types are {}.'
                     .format(type(block), self.folded_ir_blocks, allowed_block_types))
 
-        if not isinstance(self.field_type, GraphQLList):
-            raise ValueError(u'Invalid value of "field_type", expected a list type but got: '
+        if isinstance(self.field_type, GraphQLList):
```
pmantica1: Can the field_type be null?

obi1kenobi: Not in this location, which is why a `None` value will fall through to the `else` clause and raise an error. There are some situations where the type of the field doesn't really fit in with our type definitions (e.g. `@this` or `@rid` in OrientDB), but this is not one of them.

pmantica1: Can it be a GraphQLNonNull type?

obi1kenobi: We do not yet support GraphQLNonNull types in the compiler, so it shouldn't be. I added a call to `strip_non_null_from_type()` just in case, though.

PR #325, opened 2019-05-29
Add an OpenCypher backend, as implemented by Neo4j and RedisGraph

obi1kenobi:
Implemented using the same IR as all other backends, plus a minimal number of custom IR lowering steps and a `CypherQuery` object similar to the `MatchQuery` one used in the MATCH backend.

Burndown list:
- [x] Projecting out outputs
- [x] Non-optional edge traversals
- [x] Filtering on non-custom scalars and lists of them `*1`
- [x] Support for `@recurse`
- [x] Outputting data from optional edge traversals `*2`
- [x] Filtering inside optional edge traversals `*2`
- [ ] Support for date/datetime/decimal scalars and lists of them `*1`
- [ ] Support for type coercions
- [ ] Support for `@fold` scope `*3`
- [ ] Support for `_x_count` within a `@fold` scope `*3`
- [ ] Support for filtering with the `has_edge_degree` operator
- [ ] Hook up test infrastructure to run generated queries against Neo4j and RedisGraph

Entries marked with an `*` (followed by a number) are partially or fully disallowed by current (as of 2019-05-28) RedisGraph limitations, though I'll continue keeping an eye on it:
https://oss.redislabs.com/redisgraph/cypher_support/
https://oss.redislabs.com/redisgraph/known_limitations/

Specifically:
- `*1` entries require support for list-valued properties, which is currently lacking in RedisGraph: https://oss.redislabs.com/redisgraph/cypher_support/#composite-types
- `*2` entries require support for `OPTIONAL MATCH`, which is currently lacking in RedisGraph: https://oss.redislabs.com/redisgraph/cypher_support/#clauses
- `*3` entries require support for `collect()`, which is currently lacking in RedisGraph: https://oss.redislabs.com/redisgraph/cypher_support/#aggregating-functions

The following features are still up in the air since I am not yet sure whether they are expressible in _any_ OpenCypher implementation:
- [ ] Support for outputting the `__typename` field
- [ ] Support for filtering on the `__typename` field


obi1kenobi:
I'd like to knock out a few more of the entries on the burndown list before merging and releasing this functionality, but I thought I'd put it up there in draft form just in case anyone is curious. Feel free to ignore it for now, if you'd like.

obi1kenobi:
Also, I'm happy to break this up into smaller PRs for easier reviewing. I wanted to open one PR where Travis can ensure I don't break anything in the process of adding functionality, but that one doesn't have to be the one that gets merged. For example, here's one possible split (one PR per entry):
- user-visible components (e.g. stuff in `compiler/__init__.py` and `compiler/common.py`)
- modifications to `compiler/compiler_entities.py` and `compiler/expressions.py`
- CypherQuery object and associated logic
- IR lowering code for Cypher
- emitting Cypher query strings
- unit tests that check generated vs expected Cypher
- docker-compose setup for Neo4j and RedisGraph
- integration tests against these real databases

pmantica1:
Right now Travis only generates builds on PRs. (Not draft PRs). I can turn it back on again easily if you'd like.

obi1kenobi:
> Right now Travis only generates builds on PRs. (Not draft PRs). I can have it build on push also if you'd like.

I take it there's no option to generate builds on both draft and non-draft PRs?

pmantica1:
No, there is not. Travis does not support draft prs. 
https://docs.travis-ci.com/user/pull-requests/

obi1kenobi:
Added support for filtering within optional traversals, subject to #337 being merged first.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+##################################
+# Optimization / lowering passes #
+##################################
+
+def insert_explicit_type_bounds(ir_blocks, query_metadata_table, type_equivalence_hints=None):
+    """Add a CoerceType block after every Traverse/Fold/Recurse, to hint to the Cypher scheduler."""
+    # Cypher might not be aware of the fact that all our edges' endpoints are strictly typed,
+    # so we expose the implicit types of edges' endpoints explicitly, by adding CoerceType blocks.
```
bojanserafimov: Is the cypher-generated graphql schema going to know about endpoint types? How?

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -0,0 +1,185 @@
...
+    make_location_rewriter_visitor_fn, make_revisit_location_translations
+)
+from ..helpers import FoldScopeLocation
+
+
+##################################
+# Optimization / lowering passes #
+##################################
+
+def insert_explicit_type_bounds(ir_blocks, query_metadata_table, type_equivalence_hints=None):
```
bojanserafimov: In retrospect, doesn't it make more sense to have the type bounds for all vertices in the IR, instead of using `QueryRoot` and `CoerceType`? It seems like all backends use this information differently. For example, the Match backend derives `type_bounds` information from `QueryRoot` and `CoerceType` blocks, and replaces them with `CoerceType` and `Filter` blocks to enforce execution order.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
bojanserafimov: The same question will be relevant for the SPARQL backend.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -0,0 +1,185 @@
...
+    """Add a CoerceType block after every Traverse/Fold/Recurse, to hint to the Cypher scheduler."""
```
bojanserafimov: How about adding a `QueryRoot` if the root type can be inferred from an outgoing edge?

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
obi1kenobi: I think the schema generation process for these types of databases will need to be somewhat different compared to the existing ones. For example, it could be human-written, or it could be generated from a SHACL specification: https://www.w3.org/TR/shacl/

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
obi1kenobi: If we had a chance to redesign the IR from scratch right now, that would probably have been a good idea. However, I don't think the cost-benefit tradeoff is worth it in this case, since the type of each location is captured in the query metadata already.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
obi1kenobi: I think that might end up being quite confusing, since that's not really what `QueryRoot` blocks are meant for.

obi1kenobi:
Merged `master` into `cypher_support` and added a bit more test coverage. @LWprogramming I'll let you handle adding the expected Cypher for the `not_in_collection` and `not_contains` filter operators.

This PR now only contains the "emit Cypher" stage, the additions to the root `__init__.py` that expose the new compilation target, and the compiler tests. Everything else was split off and merged separately.

obi1kenobi:
Moving the checklist to the Cypher issue, and closing this PR since we are chopping up the work into smaller pieces.

PR #330, opened 2019-05-30
Bug in schema generation or integration tests

bojanserafimov:
Adding data makes schema generation fail. I don't know what's the issue, might be something very stupid.


pmantica1:
Solved.

bojanserafimov:
If you use an edge that doesn't exist, orientdb is so nice it creates the schema for you.

PR #332, opened 2019-05-30
Find bug in match backend

bojanserafimov:
Not sure what's the cause. Looking.


graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
```diff
@@ -3,3 +3,6 @@ create vertex Animal set name = 'Animal 1', net_worth = Decimal('100'), uuid = '
 create vertex Animal set name = 'Animal 2', net_worth = Decimal('200'), uuid = 'cfc6e625-8594-0927-468f-f53d864a7a52'
 create vertex Animal set name = 'Animal 3', net_worth = Decimal('300'), uuid = 'cfc6e625-8594-0927-468f-f53d864a7a53'
 create vertex Animal set name = 'Animal 4', net_worth = Decimal('400'), uuid = 'cfc6e625-8594-0927-468f-f53d864a7a54'
+create vertex Species set name = 'Species 1', uuid = 'ce4f0889-ecdb-4e27-8ffa-3140eb507549', alias = ['Species 1']
+create vertex Species set name = 'Species 2', uuid = '660b4d07-37fe-4eeb-bb9f-fbc3845e35a9', alias = ['Species 2']
+create edge Entity_Related from (select from Species where name = 'Species 1') to (select from Species where name = 'Species 2')
```
pmantica1: This doesn't seem relevant?

bojanserafimov: The query asks for animals, and gets back species. This is just the data needed to expose the bug.

bojanserafimov:
One of the match passes replaces `class: Animal` with `where: ((@this INSTANCEOF 'Animal'))` (see `orientdb_query_execution.py`. In this case, the class was removed, but the instanceof was not added. Not sure for what reason. Looking..

graphql_compiler/compiler/workarounds/orientdb_query_execution.py
```diff
@@ -273,6 +273,8 @@ def _expose_only_preferred_locations(match_query, location_types, coerced_locati
                         # or has already been applied at a previous occurrence of this location.
                         # We can simply delete the QueryRoot / CoerceType blocks that impart it.
                         if isinstance(match_step.root_block, QueryRoot):
+                            # Question: What does it mean for the bound to be implied by the
+                            # query structure? I think we need an INSTANCEOF filter in this case.
```
bojanserafimov: @obi1kenobi

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
obi1kenobi: Maybe make the aliases be something that is not equal to the `name` field? Maybe add an underscore instead of a space, or something like that.

graphql_compiler/compiler/workarounds/orientdb_query_execution.py
obi1kenobi: Oh hmm, I think in my haste to develop a fix for the query scheduling bug we observed in OrientDB that this code attempts to work around, I might have missed an edge case here. Good catch!

If you have an edge from types `A -> B`, and you have already enforced the type bound `A` before traversing the edge, then after traversing the edge (per the schema and query structure), you already know you are at a `B` and don't need an explicit `INSTANCEOF` check.

What your query points out, though, is an interesting edge case:
- The edge is defined as from types `A -> B`.
- `B` is a preferred location (due to its filter), so it runs first. The other side of the edge is an eligible location but not preferred.
- However, in the query, the class used on the "out" side of the edge is not `A` but a class `X` that is a subclass of `A`. This `X` is not implied by the schema and query structure, and requires an `INSTANCEOF` check. This check is not applied, producing incorrect results.

I would propose two actions here:
- The presence of a `Filter` that does not contain a `Variable` expression (i.e. filters that use tagged parameters instead of runtime parameters) should not make a location preferred, only eligible. This is because OrientDB (to my knowledge) is not able to take advantage of indexes in situations involving tagged parameters (i.e. `ContextField` expressions). This action does not remove the bug, but does remove an unintended consequence of my rushed implementation here, and I suspect will improve scheduling outcomes for a subset of queries.
- Improve the logic for eliminating `INSTANCEOF` checks on the basis of schema and query structure information, such that the edge case above is eliminated.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -124,6 +124,23 @@ def test_simple_filter(self, backend_name):
 
         self.assertResultsEqual(graphql_query, parameters, backend_name, expected_results)
 
+    @integration_fixtures
+    def test_colocated_filter_and_tag(self):
+        graphql_query = '''{
+            Animal {
+                name @output(out_name: "animal_name")
```
obi1kenobi: There's already a `colocated_filter_and_tag` test input in the test inputs file. Mind syncing up this query and that one so they are exactly identical, and updating any tests that are affected?

obi1kenobi: Also, the generated MATCH for `test_colocated_filter_and_tag()` is incorrect -- it is missing the `Animal` type bound. I'm guessing that the author of the test didn't write it by hand, but just ran the code and copy-pasted the output into the "expected output" field. Let's try to not do that, because this is how we get bitten when we do.

graphql_compiler/compiler/workarounds/orientdb_query_execution.py
bojanserafimov: My approach would be to rewrite this function. It's too hairy. The `previous_type_bound` and the concept of `implied by query structure` are just too vague and complicated. Too much is done at the same time, it's likely I will have an error in my fix.

obi1kenobi: That's probably fine, have a go at it and let's see what it looks like.

graphql_compiler/tests/test_data_tools/integration_data/create_orientdb_integration.sql
bojanserafimov: Changed the test to use a runtime parameter instead of a colocated tagged parameter. It is simpler, removes the need for aliases, and does what I need.

PR #334, opened 2019-05-30
Add EdgeType generation from SQLAlchemy foreign keys

pmantica1:


graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -102,3 +114,17 @@ def _get_vertex_type_from_sqlalchemy_table(table):
...
             properties[name] = PropertyDescriptor(maybe_property_type, default)
     return VertexType(table.name, False, properties, {})
+
+
+def _get_edge_types(tables):
+    """Return the EdgeType objects corresponding to SQLAlchemy ForeignKey objects."""
+    edge_types = []
+    for table in six.itervalues(tables):
+        for column in table.get_children():
+            for foreign_key in column.foreign_keys:
```
bojanserafimov: Is this multiple foreign keys on the same column?

bojanserafimov:
Remember to be careful with answering edge degree queries, since we make edges for all foreign keys even if they don't have a foreign key constraint.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: Yes. I just realized that because of this the edge key names are not unique even if we prepend the table name to the column name. So I guess we have two options. One is to prepend the table name and append the outgoing table name to the EdgeTyoe. The other is to not allow columns to have multiple foreign keys.

pmantica1:
ForeignKey objects are always associated with a ForeignKeyConstraint in SQLAlchemy.

```
> :
>         Construct a column-level FOREIGN KEY.
> 
>         The :class:`.ForeignKey` object when constructed generates a
>         :class:`.ForeignKeyConstraint` which is associated with the parent
>         :class:`.Table` object's collection of constraints.
```

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
bojanserafimov: (had the conversation offline) Makes more sense to not allow multiple foreign keys on the same column.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -126,10 +126,21 @@ def _get_edge_types(tables):
     edge_types = []
     for table in six.itervalues(tables):
         for column in table.get_children():
-            for foreign_key in column.foreign_keys:
+            if len(column.foreign_keys) > 1:
+                raise IllegalSchemaInputError(u'Multiple foreign keys are defined on column {} of '
+                                              u'table {}.'.format(column.key, table.name))
+            if len(column.foreign_keys) > 0:
```
ReaLNeroM: Nit: check len(column.foreign_keys) == 1, not inequality. This `if` basically has a dependency on the above condition.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -126,10 +126,21 @@ def _get_edge_types(tables):
...
+
+                outgoing_table = foreign_key.column.table
+                if not foreign_key.column.primary_key:
+                    raise AssertionError(u'Foreign key defined in column {} in table {} references '
+                                         u'column {} in table {}, which is not primary key.'
+                                         .format(table.name, column.key, foreign_key.column.name,
+                                                 outgoing_table.name))
+
                 # We prepend the table name to ensure that the name is unique.
                 edge_name = '{}_{}'.format(table.name, column.name)
```
ReaLNeroM: Could this result in collisions, thinking adversarially?

pmantica1: Yes it could. I guess that we must both preprend and append the column name to ensure that the schema element is unique.

obi1kenobi: Also, how do we handle edges that result from join tables? This is a very complex design question, please provide an overview of your proposed design in the PR summary.

pmantica1: Yea not that I think about it prohibiting tables with  '_' is a terrible idea.

bojanserafimov: A strictly better idea is detecting and prohibiting precisely schemas that collide in this way. I don't expect it to happen at all IRL. But I'm also not saying this is the best solution.

pmantica1:
The way we implement edges for SQL schemas has changed since this pr was made. I think we are likely going to want to generate edges differently now.

PR #335, opened 2019-05-31
Add graphql_schema_from_sqlachemy_metadata

pmantica1:


graphql_compiler/tests/test_sqlalchemy_schema_generation.py
```diff
@@ -55,3 +61,37 @@ def test_sql_self_inheritance(self):
         # Tests that a class is in its own subclass/superclass set.
         self.assertEqual(self.schema_graph.get_subclass_set('A'), {'A'})
         self.assertEqual(self.schema_graph.get_superclass_set('A'), {'A'})
+
+    def test_sqlalchemy_graphql_schema_generation(self):
```
bojanserafimov: Can we add the ordering of directives, types, etc to the spec of the function?

graphql_compiler/__init__.py
```diff
@@ -190,11 +193,27 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
...
+                                       type of a field in the class where it's first defined and all
+                                       the class's subclasses.
+        hidden_classes: optional set of strings, classes to not include in the GraphQL schema.
+
+    Returns:
+        tuple of (GraphQL schema object, GraphQL type equivalence hints dict).
+        The tuple is of type (GraphQLSchema, {GraphQLObjectType -> GraphQLUnionType}).
+    """
+    schema_graph = get_sqlalchemy_schema_graph(metadata)
+    return get_graphql_schema_from_schema_graph(schema_graph, class_to_field_type_overrides,
```
obi1kenobi: `class_to_field_type_overrides` and `hidden_classes` are becoming kwargs, per your changes in the file below, and yet here you are passing them positionally instead of as kwargs. Consider this: what happens if someone were to add another positional argument in the signature of `get_graphql_schema_from_schema_graph`?

pmantica1: I see your point. Will fix.

graphql_compiler/tests/test_sqlalchemy_schema_generation.py
pmantica1: I'll do this but in a different branch and then I'll merge it onto this one.

pmantica1: Right now the generated string is always deterministically the same given two equivalent SchemaGraph objects. However, the ordering specification is quite complicated right now. In another PR I will simplify the ordering and write the specification.

pmantica1: If you look at it right now, it seems that the directives are listed at the top and the rest is alphabetical. This is not always the case. The perfectly alphabetical ordering is a bit of a coincidence.

PR #337, opened 2019-05-31
Record the GraphQL type of LocalField expressions.

obi1kenobi:
Required to make progress on the Cypher backend (#325), since a lowering step there converts LocalField expressions into ContextField expressions. Prior to this PR, ContextField expressions include the GraphQL type of the field, whereas LocalField expressions do not, making that lowering step impossible to implement.

Note that in a few cases involving OrientDB backends and edge-related or special local fields like `@this`, including the proper type is the `LocalField` object would have been tricky. For example, there exist special fields (`@rid` is one example), whose types are not really representable in the GraphQL type system at the moment. This is a problem that is not worth solving at the moment; instead, I have left `HACK` comments explaining my reasoning and opened/linked to issues here on the issue tracker.


graphql_compiler/compiler/expressions.py
```diff
@@ -233,12 +233,21 @@ def __ne__(self, other):
...
 
-    def __init__(self, field_name):
-        """Construct a new LocalField object that references a field at the current position."""
-        super(LocalField, self).__init__(field_name)
+    def __init__(self, field_name, field_type):
+        """Construct a new LocalField object that references a field at the current position.
+
+        Args:
+            field_name: string, the name of the local field being referenced
+            field_type: GraphQL type object describing the type of the referenced field. For some
```
pmantica1: Two nits. GraphQLType not GraphQL Type.  If the field can be None, please start with the word optional.

graphql_compiler/compiler/expressions.py
```diff
@@ -248,6 +257,8 @@ def get_local_object_gremlin_name(self):
     def validate(self):
         """Validate that the LocalField is correctly representable."""
         validate_safe_string(self.field_name)
+        if self.field_type is not None and not is_graphql_type(self.field_type):
+            raise ValueError(u'Invalid value of "field_type": {}'.format(self.field_type))
```
pmantica1: nit: Add the field_name to this assertion error.

graphql_compiler/compiler/filters.py
```diff
@@ -321,19 +341,22 @@ def _process_name_or_alias_filter_directive(filter_operation_info, location, con
...
+                                              alias_field_name, name_field_type,
+                                              alias_field_inner_type))
 
     argument_inferred_type = name_field_type
     argument_expression, non_existence_expression = _represent_argument(
         location, context, parameters[0], argument_inferred_type)
 
     check_against_name = expressions.BinaryComposition(
-        u'=', expressions.LocalField('name'), argument_expression)
+        u'=', expressions.LocalField(name_field_name, name_field.type), argument_expression)
```
pmantica1: name_field_name seems to be a constant. Why not uppercase it to make that clear?

graphql_compiler/compiler/filters.py
```diff
@@ -321,19 +341,22 @@ def _process_name_or_alias_filter_directive(filter_operation_info, location, con
...
     check_against_alias = expressions.BinaryComposition(
-        u'contains', expressions.LocalField('alias'), argument_expression)
+        u'contains',
+        expressions.LocalField(alias_field_name, alias_field.type),
```
pmantica1: Same as above.

graphql_compiler/compiler/expressions.py
obi1kenobi: I would disagree on the "optional" part, and this is something that will get better once we drop support for Python 2.x and enforce type hints everywhere. The type of the value is a "maybe" type (`Optional[GraphQLType]`), but the argument is not optional -- it is positional and required, and it's allowed in some circumstances to supply the value `None`.

I would prefer to avoid describing it as "optional" since it's not an optional argument, but a required argument with an optional type.

graphql_compiler/compiler/expressions.py
obi1kenobi: I'll force it to write out the entire `LocalField` so we are guaranteed to have all the arguments, even if we add more in the future.

graphql_compiler/compiler/filters.py
obi1kenobi: It's only a function-level constant, not a module-level one. The Google Python Style Guide says that only module-level constants are capitalized, and all function-level values (constant or variable) are lowercase: http://google.github.io/styleguide/pyguide.html#316-naming

PR #345, opened 2019-06-04
Query cardinality estimation

ReaLNeroM:
Here are the assumptions we've made so far in the query cost estimator. Later on, we'll see which of these are most important, and which aren't.

- every “leaf” outputs something
- edges are uniformly distributed among “parents”
- “sibling” estimates are independent
- subclasses act like the rest of the superclass
- recurse depths > 1 are treated like depth = 1.
- ‘=’-filters on uniquely indexed properties always yield 1 result
- 'in_collection'-filters on uniquely indexed properties always yield absolute selectivities of the size of the collection.
- for all other filters, assume all results will be returned.


graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -0,0 +1,847 @@
...
+
+def create_lookup_counts(count_data):
+    """Create lookup_counts function for use in estimating query cost."""
+    def lookup_counts(name):
+        """Lookup the total number of instances and subinstances of a given class name."""
+        return count_data[name]
+    return lookup_counts
+
+
+# The following TestCase class uses the 'graph_client' fixture
```
obi1kenobi: I think you might have already renamed this fixture in `master` so this merge might fail to apply correctly.

obi1kenobi:
Looks good to me. I'll let @bojanserafimov and @pmantica1 take a look as well before we merge, just in case I missed something.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -0,0 +1,272 @@
...
+    else:
+        raise AssertionError(u'Unexpected location encountered: {}'.format(location))
+    return edge_direction, edge_name
+
+
+def _get_parent_and_child_name_from_edge(schema_graph, child_location):
+    """Get the base classname of a location and its parent from the last edge information."""
+    edge_direction, edge_name = _get_last_edge_direction_and_name_to_location(child_location)
+    edge_element = schema_graph.get_edge_schema_element_or_raise(edge_name)
+    if edge_direction == EDGE_DESTINATION_PROPERTY_NAME:
```
pmantica1: One nit: EDGE_DESTINATION_PROPERTY_NAME is an OrientDB variable that is very counterintuitive. (It is equal to 'out'). Could you instead use OUTBOUND_EDGE_DIRECTION and INBOUND_EDGE_DIRECTION? They are defined on graphql_compiler/compiler/helpers.py

graphql_compiler/cost_estimation/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Type casting and directives can affect these calculations in many different ways. We naively
+    handle type casting, as well as optional, fold, recurse, and some filter directives. Additional
+    statistics can be recorded to improve the coverage and accuracy of these adjustments.
+
+TODOs
+=====
+    - Estimate execution cost by augmenting the cardinality calculation.
+    - Add recurse handling.
+    - Add basic filter handling (for unique indexes).
```
bojanserafimov: This one is stale.

PR #347, opened 2019-06-04
Delete references to non-existent non-graph interfaces

pmantica1:
This PR removes references to filtered NonGraphElements. A minor change in this PR is that I removed the "SCHEMA_DATA" suffix of the test dictionaries in the test_schema_generation file since it was unnecessary and verbose.

### DIFF
--- a/graphql_compiler/schema_generation/graphql_schema.py
+++ b/graphql_compiler/schema_generation/graphql_schema.py
@@ -21,9 +21,9 @@ def _has_only_vertex_non_abstract_subclasses(cls_name, schema_graph):
 
     all_non_abstract_subclasses_are_vertices = True
     for subclass_name in cls_subclasses:
-        subclass = schema_graph.get_element_by_class_name(subclass_name)
+        subclass_element = schema_graph.get_element_by_class_name(subclass_name)
         if subclass_name != cls_name:
-            if not subclass.abstract and not subclass.is_vertex:
+            if not subclass_element.abstract and not subclass_element.is_vertex:
                 all_non_abstract_subclasses_are_vertices = False
                 break
 
--- a/graphql_compiler/schema_generation/graphql_schema.py
+++ b/graphql_compiler/schema_generation/graphql_schema.py
@@ -15,6 +15,21 @@ from ..schema import DIRECTIVES, EXTENDED_META_FIELD_DEFINITIONS
 from .exceptions import EmptySchemaError
 
 
+def _has_only_vertex_non_abstract_subclasses(cls_name, schema_graph):
+    """Return True if all the classes non-abstract subclasses are vertices. Else, return False."""
+    cls_subclasses = schema_graph.get_subclass_set(cls_name)
+
+    all_non_abstract_subclasses_are_vertices = True
+    for subclass_name in cls_subclasses:
+        subclass = schema_graph.get_element_by_class_name(subclass_name)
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -15,6 +15,21 @@
 from .exceptions import EmptySchemaError
 
 
+def _has_only_vertex_non_abstract_subclasses(cls_name, schema_graph):
+    """Return True if all the classes non-abstract subclasses are vertices. Else, return False."""
+    cls_subclasses = schema_graph.get_subclass_set(cls_name)
+
+    all_non_abstract_subclasses_are_vertices = True
+    for subclass_name in cls_subclasses:
+        subclass = schema_graph.get_element_by_class_name(subclass_name)
```
obi1kenobi: `subclass` might get confused for being a class object, rather than a `SchemaElement`. Consider `subclass_element` as an alternative name?

graphql_compiler/tests/test_schema_generation.py
```diff
@@ -146,75 +146,135 @@
     ],
 })
 
+ABSTRACT_NON_GRAPH_CLASS_WITH_NON_VERTEX_CONCRETE_SUBCLASS = frozendict({
+    'name': 'AbstractNonGraphClassWithNonVertexConcreteSubclass',
+    'abstract': True,
+    'superClasses': [],
+    'properties': [],
```
obi1kenobi: If `properties` is empty, then these are not representable in GraphQL regardless of everything else. Try to make the test data as "real" as possible, otherwise we can very easily end up with tests that pass by accident, rather than passing because the code is actually correct.

graphql_compiler/tests/test_schema_generation.py
```diff
@@ -336,9 +430,42 @@ def test_edge_inheritance(self):
...
+        ]
+
+        vertex_clases = [
+            ARBITRARY_CONCRETE_VERTEX_CLASS,
+            BASE_VERTEX,
+        ]
+
+        schema_data = non_graph_classes_to_include + non_graph_classes_to_ignore + vertex_clases
+
+        names_of_non_graph_classes_to_ignore = set(
```
obi1kenobi: Why not use a set comprehension here instead of `set(<generator expression>)`?

graphql_compiler/tests/test_schema_generation.py
```diff
@@ -336,9 +430,42 @@ def test_edge_inheritance(self):
...
+            non_graph_class['name']
+            for non_graph_class in non_graph_classes_to_ignore
+        )
+
+        graphql_schema, _ = get_graphql_schema_from_orientdb_schema_data(schema_data)
+        for name, _ in six.iteritems(graphql_schema.get_type_map()):
```
obi1kenobi: `six.iterkeys()` and you don't have to discard the second value

graphql_compiler/schema_generation/graphql_schema.py
```diff
@@ -15,6 +15,21 @@
 from .exceptions import EmptySchemaError
 
 
+def _has_only_vertex_non_abstract_subclasses(cls_name, schema_graph):
+    """Return True if all the classes non-abstract subclasses are vertices. Else, return False."""
```
ReaLNeroM: Typo: "all the classes' non-abstract"

PR #349, opened 2019-06-07
Create not-contains filter

LWprogramming:
Implement a not-contains filter (negation of CONTAINS) along with unit and snapshot tests.

### DIFF
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -811,6 +811,7 @@ class BinaryComposition(Expression):
         regular_operator_format = '(%(left)s %(operator)s %(right)s)'
         inverted_operator_format = '(%(right)s %(operator)s %(left)s)'  # noqa
         intersects_operator_format = '(%(operator)s(%(left)s, %(right)s).asList().size() > 0)'
+        negated_regular_operator_format = 'NOT (%(left)s %(operator)s %(right)s)'
         # pylint: enable=unused-variable
 
         # Null literals use the OrientDB 'IS/IS NOT' (in)equality operators,
@@ -832,7 +833,7 @@ class BinaryComposition(Expression):
                 u'||': (u'OR', regular_operator_format),
                 u'&&': (u'AND', regular_operator_format),
                 u'contains': (u'CONTAINS', regular_operator_format),
-                u'not_contains': (u'NOT CONTAINS', regular_operator_format),
+                u'not_contains': (u'CONTAINS', negated_regular_operator_format),
                 u'intersects': (u'intersect', intersects_operator_format),
                 u'has_substring': (None, None),  # must be lowered into compatible form using LIKE
 
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -811,7 +811,7 @@ class BinaryComposition(Expression):
         regular_operator_format = '(%(left)s %(operator)s %(right)s)'
         inverted_operator_format = '(%(right)s %(operator)s %(left)s)'  # noqa
         intersects_operator_format = '(%(operator)s(%(left)s, %(right)s).asList().size() > 0)'
-        negated_regular_operator_format = 'NOT (%(left)s %(operator)s %(right)s)'
+        negated_regular_operator_format = '(NOT (%(left)s %(operator)s %(right)s))'
         # pylint: enable=unused-variable
 
         # Null literals use the OrientDB 'IS/IS NOT' (in)equality operators,
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -758,7 +758,7 @@ class BinaryComposition(Expression):
 
     SUPPORTED_OPERATORS = frozenset({
         u'=', u'!=', u'>=', u'<=', u'>', u'<', u'+', u'||', u'&&',
-        u'contains', u'intersects', u'has_substring', u'LIKE', u'INSTANCEOF',
+        u'contains', u'not_contains', u'intersects', u'has_substring', u'LIKE', u'INSTANCEOF',
     })
 
     __slots__ = ('operator', 'left', 'right')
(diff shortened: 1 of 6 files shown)


graphql_compiler/tests/test_compiler.py
```diff
@@ -2145,6 +2145,96 @@ def test_contains_op_filter_with_optional_tag(self):
...
+            FROM (
+                MATCH {{
+                    class: Animal,
+                    where: ((alias NOT CONTAINS {wanted})),
+                    as: Animal___1
+                }}
+                RETURN $matches
+            )
+        '''
+        # "NOT" is not supported? https://orientdb.com/docs/2.0/orientdb.wiki/SQL-Where.html
```
obi1kenobi: I don't think it's supported like this -- this would make `NOT CONTAINS` an operator. `NOT` is supported as a general unary operator, so in this case you'd use it as `NOT (alias CONTAINS {wanted})`.

graphql_compiler/tests/test_compiler.py
```diff
@@ -2145,6 +2145,96 @@ def test_contains_op_filter_with_optional_tag(self):
...
+    def test_not_contains_op_filter_with_variable(self):
+        test_data = test_input_data.contains_op_filter_with_variable()
+
+        expected_match = '''
+            SELECT
+                Animal___1.name AS `animal_name`
+            FROM (
+                MATCH {{
+                    class: Animal,
+                    where: ((alias NOT CONTAINS {wanted})),
```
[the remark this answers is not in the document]
LWprogramming: oops-- you're totally right, when I was looking for tests for `in_collection` I grepped for "contains" which led me to do that. I'll implement `not_contains` first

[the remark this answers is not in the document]
LWprogramming: argh it seems like I can't rename the branch without needing to redo this PR and losing all the comments, so it might be a little bit confusing

graphql_compiler/tests/test_compiler.py
LWprogramming: aha, I found it!

LWprogramming:
This might sound really silly but it turns out that just a bang `!` does the trick instead of some arrangement of `not` and `contains`. I checked that the bang does give the negation.

LWprogramming:
Ok, ready for review!

graphql_compiler/compiler/expressions.py
```diff
@@ -811,6 +811,7 @@ def to_match(self):
         regular_operator_format = '(%(left)s %(operator)s %(right)s)'
         inverted_operator_format = '(%(right)s %(operator)s %(left)s)'  # noqa
         intersects_operator_format = '(%(operator)s(%(left)s, %(right)s).asList().size() > 0)'
+        negated_regular_operator_format = 'NOT (%(left)s %(operator)s %(right)s)'
```
obi1kenobi: Note that all the other format strings wrap everything in parens. This is to avoid any chance of confusion about order of operations in the generated query. Let's make sure to wrap the entire expression here in parens as well, to continue that approach.

graphql_compiler/compiler/expressions.py
```diff
@@ -856,6 +858,7 @@ def to_gremlin(self):
         immediate_operator_format = u'({left} {operator} {right})'
         dotted_operator_format = u'{left}.{operator}({right})'
         intersects_operator_format = u'(!{left}.{operator}({right}).empty)'
+        negated_dotted_operator_format = u'!{left}.{operator}({right})'
```
obi1kenobi: The `intersects_operator_format` and the `negated_dotted_operator_format` seem fairly similar. Do you think it makes sense to deduplicate their content and express one in terms of the other?

graphql_compiler/compiler/expressions.py
LWprogramming: Sounds good, will do

graphql_compiler/compiler/expressions.py
LWprogramming: I'm not exactly sure intersects is like negation-- we might've also implemented this with no bang by checking if the size is nonzero (assuming there's a way to get size of the results), so I suspect it might just be a coincidence.

I think if we run into more cases of where we have the bang in front (e.g. it turns out to be useful to have a filter for non-empty lists or something, in which case you check if it's not empty), then we could look into deduplicating it.

graphql_compiler/compiler/filters.py
```diff
@@ -534,6 +534,47 @@ def _process_contains_filter_directive(filter_operation_info, location, context,
...
+    """Return a Filter basic block that checks if the directive arg is not contained in the field.
+
+    Args:
+        filter_operation_info: FilterOperationInfo object, containing the directive and field info
+                               of the field where the filter is to be applied.
+        location: Location where this filter is used.
+        context: dict, various per-compilation data (e.g. declared tags, whether the current block
+                 is optional, etc.). May be mutated in-place in this function!
+        parameters: list of 1 element, specifying the collection in which the value must exist;
+                    if the collection is optional and missing, the check will return True
```
pmantica1: nit: at a dot at the end of this line.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -796,5 +823,4 @@ def test_filter_within_fold_scope(self):
 
         self.assertMatchSnapshot(rows)
 
-
```
pmantica1: nit: don't format lines unrelated to diff.

pmantica1:
Looks good. Just two minor nits since you have to fix failing CI anyways.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
LWprogramming: done!

graphql_compiler/compiler/filters.py
LWprogramming: done

obi1kenobi:
Looks good to me, but I can't merge until you've resolved the merge conflict. As soon as you're done with that, I'm happy to push the button.

LWprogramming:
ok, everything passed this time!

PR #350, opened 2019-06-11
Not in collection filter

LWprogramming:
This is the actual not-in-collection feature (the previous PR with not-contains had an incorrectly named branch).

### DIFF
--- a/graphql_compiler/compiler/filters.py
+++ b/graphql_compiler/compiler/filters.py
@@ -452,6 +452,43 @@ def _process_in_collection_filter_directive(filter_operation_info, location, con
     return blocks.Filter(filter_predicate)
 
 
+@scalar_leaf_only(u'not_in_collection')
+@takes_parameters(1)
+def _process_not_in_collection_filter_directive(filter_operation_info, location, context, parameters):
+    """Return a Filter basic block that checks for a value's non-existence in a collection.
+
+    Args:
+        filter_operation_info: FilterOperationInfo object, containing the directive and field info
+                               of the field where the filter is to be applied.
+        location: Location where this filter is used.
+        context: dict, various per-compilation data (e.g. declared tags, whether the current block
+                 is optional, etc.). May be mutated in-place in this function!
+        parameters: list of 1 element, specifying the collection in which the value must exist;
+                    if the collection is optional and missing, the check will return True.
+
+    Returns:
+        a Filter basic block that performs the collection existence check
+    """
+    filtered_field_type = filter_operation_info.field_type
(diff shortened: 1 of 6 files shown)


LWprogramming:
Minor nit question: linter says line 457 of filters.py is too long (by just 2 characters). Is there a style guide for where to break the line, or better yet an auto-formatter somewhere? Maybe in Pycharm settings?

graphql_compiler/compiler/filters.py
```diff
@@ -452,6 +452,43 @@ def _process_in_collection_filter_directive(filter_operation_info, location, con
     return blocks.Filter(filter_predicate)
 
 
+@scalar_leaf_only(u'not_in_collection')
+@takes_parameters(1)
+def _process_not_in_collection_filter_directive(filter_operation_info, location, context, parameters):
```
obi1kenobi: You could break the line right at the comma and put `parameters` on the next line, aligned under `filter_operation_info`.

obi1kenobi:
> Minor nit question: linter says line 457 of filters.py is too long (by just 2 characters). Is there a style guide for where to break the line, or better yet an auto-formatter somewhere? Maybe in Pycharm settings?

We don't yet have an auto-formatter. After January 1, 2020, we will drop Python 2 support together with the rest of the Python ecosystem, and with that we'll adopt `black` as the default auto-formatter. That's a few months away, though.

graphql_compiler/compiler/filters.py
```diff
@@ -640,6 +678,7 @@ def _get_filter_op_name_and_values(directive):
 COMPARISON_OPERATORS = frozenset({u'=', u'!=', u'>', u'<', u'>=', u'<='})
 PROPERTY_FIELD_OPERATORS = COMPARISON_OPERATORS | frozenset({
     u'between',
+    u'not_in_collection',
```
obi1kenobi: Nit: Let's maintain a consistent relative order between operators here. `not_contains` comes after `contains`, so `not_in_collection` should come after `in_collection`.

LWprogramming: done!

LWprogramming: Did this get merged before I changed that?

obi1kenobi: I merged `master` in because this branch was behind and our current setup is such that only branches that are up-to-date can be merged into `master`, to avoid breaking the build. You'll have to pull the change into your local copy of the branch, make the change, commit it, and then push again.

LWprogramming: just pushed the nit! ready for merge into master.

PR #351, opened 2019-06-11
Relax tag filter constraint

bojanserafimov:
Tags are now allowed to appear after filters, if they are on a different property of the same vertex.

Changes made:
- Removed uses of undocumented `context['tags']` and replaced them with `TagInfo` from the metadata table.
- Changed the directive processing order, so that tags at all property fields on a vertex are processed before we compile any of those property fields.

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -28,19 +28,21 @@ To get from GraphQL AST to IR, we follow the following pattern:
     The root AST node is always a vertex AST node.
 
     *** P-steps ***
-    step P-2. Process property-only directives, like @tag and @output.
+    step P-2. Process @output directives.
     ***************
 
     *** V-steps ***
-    step V-2. Recurse into any property field children of the current AST node
+    step V-2. Process @tag directives at all property field children of the current AST node.
+
+    step V-3. Recurse into any property field children of the current AST node
               (property fields cannot have property fields of their own, see _compile_vertex_ast()).
 
-    step V-3. Property field processing complete:  (see _compile_vertex_ast())
+    step V-4. Property field processing complete:  (see _compile_vertex_ast())
         - mark the current location in the query, since all @filter directives that apply to the
           current field have already been processed;
         - process the output_source directive, if it exists
 
-    step V-4. Recurse into any vertex field children of the current AST node:
+    step V-5. Recurse into any vertex field children of the current AST node:
               (see _compile_vertex_ast())
         - before recursing into each vertex:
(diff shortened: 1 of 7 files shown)


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -244,6 +246,38 @@ def _process_output_source_directive(schema, current_schema_type, ast,
...
+
+    Args:
+        context: dict, various per-compilation data (e.g. declared tags, whether the current block
+                 is optional, etc.). May be mutated in-place in this function!
+        current_schema_type: GraphQLType, the schema type at the current location
+        location: Location object representing the current location in the query
+        tag_directive: GraphQL Directive that we want to process
+    """
+    if is_in_fold_scope(context):
+        raise GraphQLCompilationError(u'Tagging values within a @fold vertex field is '
```
obi1kenobi: Let's make sure we have a test for this. I think the module docstring's summary, if followed in the implementation, will ensure that this is handled correctly, but best to prove it with a passing test case.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -270,30 +304,6 @@ def _compile_property_ast(schema, current_schema_type, ast, location,
                                           .format(COUNT_META_FIELD_NAME, location))
 
     # step P-2: process property-only directives
```
obi1kenobi: Update this comment to the new P-2 description in the module docstring

graphql_compiler/compiler/filters.py
```diff
@@ -113,13 +113,13 @@ def _represent_argument(directive_location, context, argument, inferred_type):
 
         return (expressions.Variable(argument, inferred_type), None)
     elif is_tagged_parameter(argument):
-        argument_context = context['tags'].get(argument_name, None)
```
obi1kenobi: Thanks! Is anything else using `context['tags']`? If not, it's probably safe to delete.

bojanserafimov: Nothing else. I did delete it.

graphql_compiler/compiler/compiler_frontend.py
bojanserafimov: We do.

graphql_compiler/compiler/filters.py
obi1kenobi: Let's make sure `context['tags']` is never set up in the first place, and confirm with a grep for `context['tags']`:
https://github.com/kensho-technologies/graphql-compiler/pull/351/files#diff-fffee83da4144bc70ab3578390d57761R827

PR #353, opened 2019-06-12
Add to_cypher() methods to Expression objects.

obi1kenobi:
Small, reviewable part of #325 that adds the `to_cypher()` method to the compiler primitives.

Whichever one of this PR and #350 merges first will cause the other to have to incorporate the necessary changes in `BinaryComposition.to_cypher()` to express the new operators correctly. It's not a difficult change, so I'm not worried about it.

### DIFF
--- a/graphql_compiler/compiler/compiler_entities.py
+++ b/graphql_compiler/compiler/compiler_entities.py
@@ -142,3 +142,10 @@ class MarkerBlock(BasicBlock):
         The effect of MarkerBlocks is applied during optimization and code generation steps.
         """
         return u''
+
+    def to_cypher(self):
+        """Return the Cypher representation of the block, which should almost always be empty.
+
+        The effect of MarkerBlocks is applied during optimization and code generation steps.
+        """
+        return u''
--- a/graphql_compiler/compiler/compiler_entities.py
+++ b/graphql_compiler/compiler/compiler_entities.py
@@ -72,6 +72,11 @@ class CompilerEntity(object):
         """Return the Gremlin unicode string representation of this object."""
         raise NotImplementedError()
 
+    @abstractmethod
+    def to_cypher(self):
+        """Return the Cypher unicode string representation of this object."""
+        raise NotImplementedError()
+
 
 @six.add_metaclass(ABCMeta)
 class Expression(CompilerEntity):
--- a/graphql_compiler/compiler/compiler_entities.py
+++ b/graphql_compiler/compiler/compiler_entities.py
@@ -72,11 +72,6 @@ class CompilerEntity(object):
         """Return the Gremlin unicode string representation of this object."""
         raise NotImplementedError()
 
-    @abstractmethod
-    def to_cypher(self):
-        """Return the Cypher unicode string representation of this object."""
-        raise NotImplementedError()
-
 
 @six.add_metaclass(ABCMeta)
 class Expression(CompilerEntity):
@@ -143,13 +138,6 @@ class MarkerBlock(BasicBlock):
 
     def to_gremlin(self):
         """Return the Gremlin representation of the block, which should almost always be empty.
-
-        The effect of MarkerBlocks is applied during optimization and code generation steps.
-        """
-        return u''
-
-    def to_cypher(self):
-        """Return the Cypher representation of the block, which should almost always be empty.
 
         The effect of MarkerBlocks is applied during optimization and code generation steps.
         """
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -968,7 +968,7 @@ class BinaryComposition(Expression):
         else:
             translation_table = {
                 u'=': (u'=', regular_operator_format),
-                u'!=': (u'!=', regular_operator_format),
+                u'!=': (u'<>', regular_operator_format),
                 u'>=': (u'>=', regular_operator_format),
                 u'<=': (u'<=', regular_operator_format),
                 u'>': (u'>', regular_operator_format),
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -95,8 +95,9 @@ class Literal(Expression):
         raise GraphQLCompilationError(u'Cannot represent literal: {}'.format(self.value))
 
     def _to_output_code(self):
-        """Return a unicode object with the Gremlin/MATCH representation of this Literal."""
-        # All supported Literal objects serialize to identical strings both in Gremlin and MATCH.
+        """Return a unicode object with the Gremlin/MATCH/Cypher representation of this Literal."""
+        # All supported Literal objects serialize to identical strings
+        # in all of Gremlin, Cypher, and MATCH.
         self.validate()
         if self.value is None:
             return u'null'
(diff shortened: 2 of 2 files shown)


graphql_compiler/compiler/compiler_entities.py
```diff
@@ -142,3 +142,10 @@ def to_gremlin(self):
         The effect of MarkerBlocks is applied during optimization and code generation steps.
         """
         return u''
+
+    def to_cypher(self):
```
ReaLNeroM: If we're adding to_cypher() methods, shouldn't we also add an abstract method to_cypher() for CompilerEntity after line 74?

graphql_compiler/compiler/expressions.py
```diff
@@ -120,6 +120,7 @@ def _to_output_code(self):
 
     to_gremlin = _to_output_code
     to_match = _to_output_code
+    to_cypher = _to_output_code
```
ReaLNeroM: I couldn't select _to_output_code() line 98, this is the closest I could select.
Update the documentation of _to_output_code to reflect our support of Cypher.

graphql_compiler/compiler/expressions.py
```diff
@@ -392,6 +411,21 @@ def to_gremlin(self):
 
         return template.format(mark_name=mark_name, field_name=field_name)
 
+    def to_cypher(self):
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        mark_name, field_name = self.location.get_location_name()
+        if field_name is not None:
```
ReaLNeroM: Nit: Add a newline between 418 and 419 for consistency between the to_match and to_gremlin methods.

graphql_compiler/compiler/expressions.py
```diff
@@ -480,6 +514,19 @@ def to_gremlin(self):
         return template.format(mark_name=mark_name, field_name=field_name,
                                format=format_value)
 
+    def to_cypher(self):
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        mark_name, field_name = self.location.get_location_name()
+
```
ReaLNeroM: Nit: remove this newline for consistency with to_match and to_gremlin.

graphql_compiler/compiler/compiler_entities.py
obi1kenobi: Good catch, thank you.

obi1kenobi: Actually, as @pmantica1 points out, we don't output anything other than `Expression` objects this way, for any output target other than Gremlin. So we don't need `to_cypher()` methods on `CompilerEntity` or `MarkerBlock`.

graphql_compiler/compiler/expressions.py
```diff
@@ -885,6 +949,47 @@ def to_gremlin(self):
...
+            translation_table = {
+                u'=': (u'=', regular_operator_format),
+                u'!=': (u'!=', regular_operator_format),
+                u'>=': (u'>=', regular_operator_format),
+                u'<=': (u'<=', regular_operator_format),
+                u'>': (u'>', regular_operator_format),
+                u'<': (u'<', regular_operator_format),
+                u'||': (u'OR', regular_operator_format),
+                u'&&': (u'AND', regular_operator_format),
+                u'contains': (u'IN', inverted_operator_format),
```
LWprogramming: Seems like contains and not_contains (as in does the list L contain item I) aren't [valid operators](https://neo4j.com/docs/cypher-manual/current/syntax/operators/#query-operators-comparison) in Cypher, but it's valid for searching in strings.

If we remove both of them, we can also take out the negated_inverted_operator format.

graphql_compiler/compiler/expressions.py
```diff
@@ -885,6 +949,47 @@ def to_gremlin(self):
...
+        # Null literals use 'is/is not' as (in)equality operators, while other values use '=/<>'.
+        if self.left == NullLiteral or self.right == NullLiteral:
+            translation_table = {
+                u'=': (u'IS', regular_operator_format),
+                u'!=': (u'IS NOT', regular_operator_format),
+            }
+        else:
+            translation_table = {
+                u'=': (u'=', regular_operator_format),
+                u'!=': (u'!=', regular_operator_format),
```
LWprogramming: cypher inequality check is <>

graphql_compiler/compiler/expressions.py
```diff
@@ -576,6 +624,10 @@ def to_gremlin(self):
         raise AssertionError(u'FoldedContextField are not used during the query emission process '
                              u'in Gremlin, so this is a bug. This function should not be called.')
 
+    def to_cypher(self):
+        """Not implemented yet."""
+        raise NotImplementedError()
```
LWprogramming: Do we want to implement this (FoldedContextField) in this PR, since we're implementing to_cypher methods for all the Expression objects here? Same question for the to_cypher methods in FoldCountContextField and UnaryTransformation. 

If it makes sense to do them in this PR rather than doing it in a separate PR, I can look into implementing those methods right now.

graphql_compiler/compiler/expressions.py
obi1kenobi: `contains` and `not_contains` are what we call the operators, not what Cypher calls them. If you trace through the code, you'll see that we actually use Cypher's `IN` operator to emit the expressions that use `contains` and `not_contains`.

graphql_compiler/compiler/expressions.py
obi1kenobi: Doing this well would probably require implementing `@fold` in general, otherwise there's no possible way to test that we've implemented it correctly. I think that would be too big of a lift for this PR.

graphql_compiler/compiler/expressions.py
obi1kenobi: Thanks!

graphql_compiler/compiler/expressions.py
LWprogramming: Ah, gotcha-- my initial reading of the docs suggested that `IN` is just for searching for a substring in a string but I just tried it running a query to search in a list and it works, so my initial reading was mistaken.

graphql_compiler/compiler/expressions.py
```diff
@@ -885,6 +949,47 @@ def to_gremlin(self):
...
 
+    def to_cypher(self):
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        # The Cypher versions of some operators require an inverted order of arguments.
+        regular_operator_format = u'({left} {operator} {right})'
+        inverted_operator_format = u'({right} {operator} {left})'
+        negated_inverted_operator_format = u'(NOT ({right} {operator} {left}))'
+        intersects_operator_format = u'any(_ {operator} {left} WHERE _ {operator} {right})'
```
LWprogramming: Nit: capitalize ANY (I've tested that this is case-insensitive and the other keywords are all capitalized).

LWprogramming: Looks good to me once this is changed!

LWprogramming: Ah, good point-- I'm too used to [SQL formatting](https://www.w3schools.com/sql/sql_ref_sqlserver.asp). In that case, everything looks good to me!

obi1kenobi: Thanks for looking it over. If you'd like to express your acceptance by formally submitting an accepting review, I'll be able to merge this PR :)

LWprogramming: done!

PR #354, opened 2019-06-12
IR lowering code for Cypher backend.

obi1kenobi:
The IR lowering component of #325.

Going to add the emit-Cypher part in the next PR, and then all the tests #325 has so far in the PR after that.

### DIFF
--- a/graphql_compiler/compiler/cypher_query.py
+++ b/graphql_compiler/compiler/cypher_query.py
@@ -82,6 +82,9 @@ def _make_query_root_cypher_step(query_metadata_table, linked_location, current_
     """Return a CypherStep for a list of IR blocks that start with a QueryRoot block."""
     current_step_block_types = tuple(type(block) for block in current_step_blocks)
 
+    # There are only two "shapes" of blocks that can form the first CypherStep.
+    # The first CypherStep always has to be made of a QueryRoot block to start, and a MarkLocation
+    # at the end, and might also have a Filter in between.
     if current_step_block_types == (QueryRoot, Filter, MarkLocation):
         step_block, where_block, as_block = current_step_blocks
     elif current_step_block_types == (QueryRoot, MarkLocation):
@@ -115,7 +118,18 @@ def convert_to_cypher_query(ir_blocks, query_metadata_table, type_equivalence_hi
     global_operations_index = None
 
     for current_block_index, block in enumerate(remaining_ir_blocks):
-        if isinstance(block, QueryRoot):
+        if isinstance(block, GlobalOperationsStart):
+            # We hit the end of the IR blocks that describe the query structure,
+            # and we are about to start the global operations section. Finish adding
+            # the last CypherStep, then break out of the loop.
+            cypher_step = _make_cypher_step(
+                query_metadata_table, linked_location, current_step_blocks)
+            steps.append(cypher_step)
+            current_step_blocks = None
+
+            global_operations_index = current_block_index
+            break
+        elif isinstance(block, QueryRoot):
             if current_step_blocks is not None:
                 raise AssertionError(u'Unexpectedly encountered a QueryRoot block that was not '
                                      u'the first block in the IR: {} {}'
@@ -131,22 +145,18 @@ def convert_to_cypher_query(ir_blocks, query_metadata_table, type_equivalence_hi
             next_linked_location = None
 
             current_step_blocks = [block]
-        elif isinstance(block, GlobalOperationsStart):
-            global_operations_index = current_block_index
-            break
         elif isinstance(block, linked_location_source_types):
             next_linked_location = block.location
             if isinstance(block, MarkLocation):
                 current_step_blocks.append(block)
         elif isinstance(block, discarded_block_types):
+            # These blocks do not require any changes to the generated Cypher query. Skip them.
             pass
         elif isinstance(block, (Filter, CoerceType)):
             current_step_blocks.append(block)
         else:
             raise AssertionError(u'Unexpected block encountered: {} {}'
                                  .format(block, ir_blocks))
-
-    steps.append(_make_cypher_step(query_metadata_table, linked_location, current_step_blocks))
 
     if global_operations_index is None:
         raise AssertionError(u'Unexpectedly, no GlobalOperationsStart block was found in '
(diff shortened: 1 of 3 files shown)


graphql_compiler/compiler/cypher_query.py
```diff
@@ -0,0 +1,168 @@
...
+                current_step_blocks.append(block)
+        elif isinstance(block, discarded_block_types):
+            pass
+        elif isinstance(block, (Filter, CoerceType)):
+            current_step_blocks.append(block)
+        else:
+            raise AssertionError(u'Unexpected block encountered: {} {}'
+                                 .format(block, ir_blocks))
+
+    steps.append(_make_cypher_step(query_metadata_table, linked_location, current_step_blocks))
```
LWprogramming: It seems like every query has a GlobalOperationsStart block, which means that the only way to break out of the for-loop above is by coming across that block. 

Would it make more sense to put this line of code in the GlobalOperationsStart case above instead of as a standalone line of code? This makes it clearer that we're adding one last cypher step.

obi1kenobi: Good idea!

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -0,0 +1,184 @@
...
+
+    return new_ir_blocks
+
+
+def move_filters_in_optional_locations_to_global_operations(cypher_query, query_metadata_table):
+    """Move Filter blocks found within @optional traversals to the global operations section.
+
+    This transformation is necessary to uphold the compiler's chosen semantics around filters
+    within optional traversals: if the edge exists but the filter fails to match, we can't pretend
+    the edge didn't exist. The Cypher specification chooses the opposite approach. The
```
pmantica1: Can you elaborate a bit more what you mean by "if the edge exists but the filter fails to match, we can't pretend
the edge didn't exist"?

LWprogramming: Suppose we have two Person nodes with names Albert and Betty such that there's an edge Albert-knows->Betty. 

Then we run a query

```
MATCH (p:Person {name:"Albert"})
OPTIONAL MATCH (p)-[r:KNOWS]->(p2:Person)
where p2.name="Charles"
return p,r,p2
```

Neo4j returns `Albert` for `p`, `null` for `r`, and null for `p2`.

In this case, we do have an "knows" edge from Albert to some other person but the filter fails because the other person isn't named Charles. However, if Albert has no outgoing "knows" edges, then the filter wouldn't apply at all, so effectively Neo4j pretends that the outgoing edge doesn't exist.

pmantica1: Why can't we pretend that the edge does not exist? Isn't that exactly what we are supposed to do?

obi1kenobi: Per our spec: https://github.com/kensho-technologies/graphql-compiler#constraints-and-rules

```When filtering (via @filter) or type coercion (via e.g. ... on Animal) are applied at or within a vertex field marked @optional, the @optional is given precedence:

- If a given result set cannot produce a value for the optional vertex field, it is preserved: the @optional directive is applied first, and no filtering or type coercion can happen.
- If a given result set is able to produce a value for the optional vertex field, the @optional does not apply, and that value is then checked against the filtering or type coercion. These subsequent operations may then cause the result set to be discarded if it does not match.```

So we may not pretend that the edge does not exist.

File: graphql_compiler/compiler/cypher_query.py [hunk 3]
```diff
@@ -0,0 +1,168 @@
...
+#                  or None if no filtering is required
+#   - as_block: MarkLocation block describing the location of the query step
+CypherStep = namedtuple(
+    'CypherStep',
+    ('linked_location', 'step_block', 'step_types', 'where_block', 'as_block'))
+
+
+step_block_types = (Traverse, Recurse)
+linked_location_source_types = (MarkLocation, Backtrack)
+discarded_block_types = (Unfold, OutputSource, EndOptional)
```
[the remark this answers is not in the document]
obi1kenobi: I added the comment in the place where this value is used. That way we minimize the risk that the use of the value is changed, and the comment is not updated.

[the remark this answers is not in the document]
LWprogramming: I don't see the comment on line 141?

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
obi1kenobi: Btw, @LWprogramming the `Albert-knows->Betty` example is really good. Consider adding (maybe a query-language-agnostic variant) to the spec section that I quoted? I think it might really help give people some intuition about what is going on there.

graphql_compiler/compiler/cypher_query.py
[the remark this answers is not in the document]
obi1kenobi: I haven't pushed yet, give me a second :)

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
LWprogramming: Sure! I'll make a PR.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -0,0 +1,184 @@
...
+            else:
+                raise AssertionError(u'Illegal IR blocks found. Block {} at index {} does not have '
+                                     u'a MarkLocation or CoerceType block after it: {}'
+                                     .format(block, current_index, ir_blocks))
+
+    return new_ir_blocks
+
+
+def remove_mark_location_after_optional_backtrack(ir_blocks, query_metadata_table):
+    """Remove location revisits, since they are not required in Cypher."""
```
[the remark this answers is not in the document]
obi1kenobi: Let me know if this is better.

graphql_compiler/compiler/cypher_query.py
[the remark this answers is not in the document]
obi1kenobi: Alright @LWprogramming try now.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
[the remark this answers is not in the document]
pmantica1: Yea makes sense.

PR #357, opened 2019-06-14
Cost statistics

ReaLNeroM:


graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+
+##
+# A GraphQLStatistics object bundles statistics function regarding GraphQL objects and vertex field
```
obi1kenobi: We might be better off making an abstract base class that describes the interface we want (`get_class_count` etc.), and then having several different implementations of the interface (e.g. `LocalStatistics` if everything is precomputed and stored in-process, `RedisStatistics` if cached in Redis, etc.).

Otherwise we're sort-of faking the ABC+multiple implementations by using this namedtuple and the helper methods below.

ReaLNeroM: Very elegant solution, thanks!

ReaLNeroM:
I'm going to close this branch, and reopen it in another PR, to remove all the commit artifacts from master.

PR #359, opened 2019-06-17
Cost statistics v2

ReaLNeroM:
Created an abstract statistics class to bundle all our statistics together, and a LocalStatistics class so that we can provide statistics at initialization time.
We have three statistics `edge_count_between_vertex_pair`, `get_domain_count` and `histogram` that we do not use currently. I'll try to address better filter support and type coercion support in a later PR.

### DIFF
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -79,6 +79,8 @@ def _get_filter_selectivity(
         Selectivity object, the selectivity of a specific filter at a given location.
     """
     result_selectivity = Selectivity(kind=FRACTIONAL_SELECTIVITY, value=1.0)
+    #TODO(vlad): Add support for inequality operations '>', '<', 'between' using histogram statistic
+    #TODO(vlad): Add support for other filters like '!='
 
     if filter_info.op_name == '=':
         result_selectivity = _estimate_filter_selectivity_of_equality(
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -3,6 +3,9 @@ from collections import namedtuple
 import sys
 
 from ..compiler.helpers import get_parameter_name
+from .statistics import (
+    _get_class_count, _get_vertex_pair_links_count, _get_domain_count, _get_histograms
+)
 
 
 # The Selectivity represents the selectivity of a filter or a set of filters
@@ -33,6 +36,85 @@ def _has_any_absolute(selectivities):
             return True
     return False
 
+def _assert_interval_type_is_number(interval):
+    for element in interval:
+        element_type = type(element)
+        if element_type != IntType and element_type != FloatType:
+            raise AssertionError('Provided non-number type as argument {}'.format(element_type))
+
+
+def _get_intersection_length(interval_a, interval_b):
+    """Find the length of intersection between two intervals given as a tuple of ints or floats."""
+    intersection_of_bucket_and_interval = min(interval_a[1], interval_b[1]) - 
+                                          max(interval_a[0], interval_b[0])
+    return intersection_of_bucket_and_interval
+
+def _estimate_count_in_interval_using_bucket(bucket, interval):
+    """Estimate number of elements in bucket (start, end, bucketCount) contained in interval"""
+
+
+def _attempt_estimate_count_in_interval_from_histogram(vertex_name, field_name, lower_bound,
+                                                    upper_bound):
+    """Using histograms, attempts to estimate """
+    histogram = self._histograms(vertex_name, field_name)
+    if histogram is None:
+        return None
+
+    result_count = 0
+    for histogram_bucket in histogram:
+        length_of_intervals_intersection = _get_intersection_length((lower_bound, upper_bound),
+                                                                 (histogram_bucket[0], 
+                                                                  histogram_bucket[1]))
+        bucket_length = histogram_bucket[1] - histogram_bucket[0]
+        intersection_as_fraction_of_bucket = length_of_intervals_intersection / bucket_length
+        elements_in_bucket = histogram_bucket[2]
+
+        # We assume elements are uniformly distributed within the bucket's interval and include the
+        # fraction of the bucket's entries that are within the interval we're interested in
+        result_count += elements_in_bucket * intersection_as_fraction_of_bucket
+    return result_count
+
+
+def _attempt_estimate_count_in_interval_from_boundary_values(vertex_name, field_name, lower_bound,
+                                                          upper_bound):
+    boundary_values = self._boundary_values(vertex_name, field_name)
+    if boundary_values is None:
+        return None
+
+    # Assume field_name values are uniformly distributed among the minimum and maximum values
+    # TODO(vlad): if the min and max values are different by several orders of magnitude, assume
+    #             elements are distributed logarithmically instead of uniformly distributed
+    vertex_count = self._class_counts(vertex_name)
+    length_of_intervals_intersection = _get_intersection_length((lower_bound, upper_bound),
+                                                             boundary_values)
+    boundary_interval_length = boundary_values[1] - boundary_values[0]
+    intersection_as_fraction_of_boundary_interval = length_of_intervals_intersection /
+                                                     boundary_interval_length
+    result_count = vertex_count * intersection_as_fraction_of_bound_interval
+    return result_count
+
+
+def estimate_count_in_interval(vertex_name, field_name, query_interval):
+    """Estimates how many vertices have fields inside the interval of (lower_bound, upper_bound)"""
+    _assert_interval_type_is_number(query_interval)
+    # Without this check, we may return negative results for estimates.
+    if interval[1] > interval[0]: 
+        return 0
+
+    histogramEstimate = _attempt_estimate_count_in_interval_from_histogram(
+        vertex_name, field_name, lower_bound, upper_bound
+    )
+    # Histograms give more detail than domain counts, so they are preferred
+    if histogramEstimate is not None:
+        return histogramEstimate
+
+    domainEstimate = _attempt_estimate_count_in_interval_from_boundary_values(
+        vertex_name, field_name, lower_bound, upper_bound
+    )
+    if domainEstimate is not None:
+        return domainEstimate
+
+    return None
 
 def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
     """Returns True if the field(s) being filtered are uniquely indexed."""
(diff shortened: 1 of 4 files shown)


graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,207 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from abc import ABCMeta, abstractmethod
+
+import six
+
+
+@six.python_2_unicode_compatible
+@six.add_metaclass(ABCMeta)
+class Statistics(object):
+    """Abstract class for statistics regarding GraphQL objects and vertex field values.
```
pmantica1: nit: GraphQL objects and their fields.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Args:
+            class_name: str, either a vertex class name or an edge class name defined in the
+                        GraphQL schema.
+
+        Returns:
+            - Number of objects with, or inheriting the given class name
+
+        Raises:
+            AssertionError, if statistic for class_name was not provided.
```
pmantica1: nit: do you mean "does not exist" instead of "was not provided"

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,207 @@
...
+    @abstractmethod
+    def get_class_count(self, class_name):
+        """Return how many objects have, or inherit, the given class name.
+
+        Args:
+            class_name: str, either a vertex class name or an edge class name defined in the
+                        GraphQL schema.
+
+        Returns:
+            - Number of objects with, or inheriting the given class name
```
pmantica1: nit: Please start with the type of the return value first in this docstring and in the ones below.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,207 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_edge_count_between_vertex_pair(
+        self, vertex_out_class_name, vertex_in_class_name, edge_class_name
+    ):
+        """Return the edge count between two vertices of a given type using an edge of a given type.
+
+        Edge traversal cost between two vertices using an edge that connects the vertices' using
+        their abstract inherited types is difficult to estimate, so we provide an interface for
```
pmantica1: What interface are you referring to?

pmantica1: It's not clear what you mean by this block of the docstring.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,207 @@
...
+        1. field_name must be a valid vertex field of vertex_name
+
+        Args:
+            vertex_name: str, name of a vertex.
+            field_name: str, name of a vertex field.
+
+        Returns:
+            - A histogram as a list of tuples with 3 elements each if the statistic exists.
+              Each tuple represents a bin as (start, end, elementCount), where start and end are
+              the interval boundaries (inclusively), and elementCount is the bin element count.
```
pmantica1: Including the elements at both boundaries in each bin might lead to overcounting. Why not have the count be inclusive of the elements at the start boundary and not the end boundary?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,207 @@
...
+        return None
+
+
+class LocalStatistics(Statistics):
+    """Provides statistics using ones given at initialization."""
+    def __init__(self, class_count, edge_count_between_vertex_pair, domain_count, histogram):
+        """Initializes statistics with the given data.
+
+        Args:
+            class_count: dict, vertex/edge class name -> class count.
```
pmantica1: For these arguments how about the following format: 
dict, keyType -> valueType, mapping what the keys abstractly represent to what the values abstractly represent.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Edges are GraphQL objects, but don't have any fields (as of now). I'll instead modify it to "GraphQL objects and vertex fields.`

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,37 +44,71 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
-    return Selectivity(kind=FRACTIONAL_SELECTIVITY, value=1.0)
+        # Assumption: The selectivity is proportional to the number of entries in the collection
+        # This will not hold in case of duplicates.
+        if _is_absolute(selectivity_per_entry_in_collection):
+            result_selectivity = Selectivity(
+                kind=ABSOLUTE_SELECTIVITY,
+                value=float(collection_size) * selectivity_per_entry_in_collection.value
+            )
+        elif _is_fractional(selectivity_per_entry_in_collection):
+            if selectivity_per_entry_in_collection != Selectivity(kind=FRACTIONAL_SELECTIVITY,
```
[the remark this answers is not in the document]
ReaLNeroM: Nice catch, thanks!

graphql_compiler/cost_estimation/statistics.py
obi1kenobi: I think Pedro's point is that the type of a field can itself be a GraphQL object. So the statistics are covering objects and their fields, regardless of whether the objects in question represent vertices or edges or something else entirely.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Reworded the docstring.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Do you mean VertexClassName -> ClassCount for example? This would lead to confusion about the object type.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Oh now I see! But if a type of a field is a GraphQL object, that's either a scalar or an edge. We only have statistics if it's a scalar, so it's a lie to say we provide statistics about vertex fields referring to edges.

graphql_compiler/cost_estimation/statistics.py
pmantica1: For instance here I mean:
dict, string -> int, mapping vertex/edge class name to class count.

ReaLNeroM: Good idea!

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+            field_name: str, name of a vertex field.
+
+        Returns:
+            - list[tuple(int, int, int)], histogram as a list of tuples with 3 elements each if the
+              statistic exists. Each tuple represents a bin as (start, end, elementCount), where
+              [start, end) is the range of values in the bin, and elementCount is the bin element
+              count.
+            - None otherwise
+        """
+        if vertex_name not in self._histogram or field_name not in self._histogram[vertex_name]:
```
obi1kenobi: This could just as easily be `self._histogram.get(vertex_name, dict()).get(field_name, None)`

ReaLNeroM: To be honest, that reads like something I'd see in a code-golf. I much prefer the version we have now. 
If I'm missing something or this is something much better, please tell me.

obi1kenobi: Let's ignore for a second the performance cost of those key lookups (there are 5 in yours and 2 in mine).

The remaining difference is this: in my one-liner, we only specify *once* what key goes first and what key goes second. In yours, count in how many places it is either implicitly or explicitly defined that `vertex_name` is the first key, and `field_name` the second? Each one of those places (after the first) is an opportunity to make a mistake and introduce a bug.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+            return None
+        return self._histogram[vertex_name][field_name]
```
ReaLNeroM: I just realized this style of dict is different from the others that receive a tuple of the arguments. I'll look into refactoring this.

pmantica1: I also just noticed that the `self._histogram` variable is inconsistent with the init docstring.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+            edge_class_name: str, edge class name.
+
+        Returns:
+            - int, the count of edges if the statistic exists
+            - None otherwise
+        """
+        statistic_id = (vertex_in_class_name, vertex_out_class_name, edge_class_name)
+        if statistic_id not in self._edge_count_between_vertex_pair:
+            return None
+        return self._edge_count_between_vertex_pair[statistic_id]
```
pmantica1: nit: simplify to: `return self._edge_count_between_vertex_pair.get(statistic_id)`

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+            field_name: str, name of a vertex field.
+
+        Returns:
+            - int, domain size of the vertex field's values if the statistic exists
+            - None otherwise
+        """
+        statistic_id = (vertex_name, field_name)
+        if statistic_id not in self._domain_count:
+            return None
+        return self._domain_count[statistic_id]
```
pmantica1: same as above

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+                count of edge type between vertex types.
+            domain_count: dict, (str, str) -> int, mapping vertex class and field on that
+                vertex class to domain size.
+            histogram: dict, (str, str) -> list[tuple(int, int, int)], mapping vertex class and
+                field on that vertex class to histogram as list[tuple(int, int, int)].
+        """
+        self._class_count = class_count
+        self._edge_count_between_vertex_pair = edge_count_between_vertex_pair
+        self._domain_count = domain_count
+        self._histogram = histogram
```
pmantica1: Can you make all these variables plural? For instance, `self._histogram` to `self._histograms`.

graphql_compiler/cost_estimation/statistics.py
pmantica1: I think that the best solution is the following:
```
histogram_key = (vertex_name, field_name)
`return self._histograms.get(histogram_key)
```
It is consistent with the style of the other methods and avoids the complexity of having a nested dict.(Note that I renamed `self._histogram` to `self._histograms`. Please see comment above.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+    """Provides statistics using ones given at initialization."""
+    def __init__(self, class_count, edge_count_between_vertex_pair, domain_count, histogram):
+        """Initializes statistics with the given data.
+
+        Args:
+            class_count: dict, str -> int, mapping vertex/edge class name to class count.
+            edge_count_between_vertex_pair: dict, (str, str, str) -> int,
+                mapping tuple of (vertex out class name, vertex in class name, edge class name) to
+                count of edge type between vertex types.
+            domain_count: dict, (str, str) -> int, mapping vertex class and field on that
```
pmantica1: Can you clarify what domain size is?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+class LocalStatistics(Statistics):
+    """Provides statistics using ones given at initialization."""
+    def __init__(self, class_count, edge_count_between_vertex_pair, domain_count, histogram):
+        """Initializes statistics with the given data.
+
+        Args:
+            class_count: dict, str -> int, mapping vertex/edge class name to class count.
+            edge_count_between_vertex_pair: dict, (str, str, str) -> int,
+                mapping tuple of (vertex out class name, vertex in class name, edge class name) to
+                count of edge type between vertex types.
```
pmantica1: nit: count of edges between the two vertices. You are not really referring to types here but concrete values.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Preconditions:
+        1. field_name must be a valid vertex field of vertex_name
+
+        Args:
+            vertex_name: str, name of a vertex.
+            field_name: str, name of a vertex field.
+
+        Returns:
+            - list[tuple(int, int, int)], histogram as a list of tuples with 3 elements each if the
```
pmantica1: Are you sure you want the inner list type not to be `tuple(float, float, int)` instead? Please see the following link for relevant info:
https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Args:
+            class_count: dict, str -> int, mapping vertex/edge class name to class count.
+            edge_count_between_vertex_pair: dict, (str, str, str) -> int,
+                mapping tuple of (vertex out class name, vertex in class name, edge class name) to
+                count of edge type between vertex types.
+            domain_count: dict, (str, str) -> int, mapping vertex class and field on that
+                vertex class to domain size.
+            histogram: dict, (str, str) -> list[tuple(int, int, int)], mapping vertex class and
+                field on that vertex class to histogram as list[tuple(int, int, int)].
```
pmantica1: nit: "to a histogram, providing statistics about range values". You do not need to include `list[tuple(int, int, int)]` again in the docstring.

graphql_compiler/cost_estimation/statistics.py
pmantica1: If you decide to change the docstring here please remember to change the other docstrings where this type description is mentioned.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+                                 u'{}'.format(class_name))
+        return self._class_count[class_name]
+
+    def get_edge_count_between_vertex_pair(
+        self, vertex_out_class_name, vertex_in_class_name, edge_class_name
+    ):
+        """Return number of edges of given class connecting two vertex types.
+
+        For cost estimation, sometimes we need to predict traversal costs between two vertex types
+        using an edge type that connects the vertices' abstract types. If this statistic isn't
```
pmantica1: nit: Instead of writing edge type and vertex type please write edge class and vertex class. Even though type and class are interchangeable, I think it's best if we stick with only one of the two and I think we use class more consistently. If you decide to change to using "class" instead please remember to make this change in the other docstrings too.

pmantica1: I think you mean to say: an edge between the vertices' superclasses.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,211 @@
...
+        provided, it can be roughly estimated using multiple calls of get_class_count().
```
pmantica1: nit: instead of "provided" how about "directly available"

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I'll push this in another PR shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I'll push this in another PR shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I'll push this in another PR shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I don't see how a newcomer reading the docstring could misunderstand this.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I'll push this is another PR shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Removed this in a PR that I'll push shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Will fix this is a PR shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Fixed in a PR I'll push shortly.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: The types are not necessarily concrete, for example, we might have this statistic between abstract to abstract vertices. Even though we won't use the abstract to abstract statistics in the short term, I don't see why we should artificially limit ourselves.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I'll push this change in another PR.

PR #361, opened 2019-06-17
Broaden query formatting APIs, and ensure they handle non-null types

obi1kenobi:
PR 2/n for merging the `macro_system` branch into `master`. See #356 for the full context of where we're going, and how we're getting there.

Relative to #356, I added handling for non-null types and a test case to make sure that code is correct.

### DIFF
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -1,18 +1,106 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Safely insert runtime arguments into compiled GraphQL queries."""
+import datetime
+import decimal
+
+import arrow
+from graphql import GraphQLBoolean, GraphQLFloat, GraphQLID, GraphQLInt, GraphQLList, GraphQLString
 import six
 
 from ..compiler import GREMLIN_LANGUAGE, MATCH_LANGUAGE, SQL_LANGUAGE
+from ..compiler.helpers import strip_non_null_from_type
 from ..exceptions import GraphQLInvalidArgumentError
+from ..schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal
 from .gremlin_formatting import insert_arguments_into_gremlin_query
 from .match_formatting import insert_arguments_into_match_query
 from .sql_formatting import insert_arguments_into_sql_query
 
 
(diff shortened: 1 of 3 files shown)


graphql_compiler/query_formatting/common.py
```diff
@@ -1,18 +1,106 @@
...
+                .format(type(value), value))
+        try:
+            stripped_type.serialize(value)
+        except ValueError as e:
+            raise GraphQLInvalidArgumentError(e)
+    elif isinstance(stripped_type, GraphQLList):
+        if not isinstance(value, list):
+            raise GraphQLInvalidArgumentError(u'Attempting to represent a non-list as a list: '
+                                              u'{} {}'.format(type(value), value))
+        inner_type = strip_non_null_from_type(stripped_type.of_type)
```
pmantica1: not: Delete this line.

pmantica1: Can you delete the `strip_non_null_from_type` here? There is a `strip_non_null_from_type` at the beginning of the function anyways so this doesn't do anything.

obi1kenobi: It definitely does do something :)

`GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLString)))` is a valid type. The first `strip_non_null_from_type` will cause `stripped_type = GraphQLList(GraphQLNonNull(GraphQLString))`. This second call takes `stripped_type.of_type` and strips that, to cause `inner_type = GraphQLString`. If I remove this call, we'll have `inner_type = GraphQLNonNull(GraphQLString)` which will then fail the new test I added.

pmantica1: No it will not. We will pass `GraphQLNonNull(GraphQLString)` into `validate_argument_type` which will then create the variable `stripped_type` with value `GraphQLString` and validate the argument based on the `stripped_type`.

obi1kenobi: I agree, it won't fail, so I take that part back. But why unwrap the type on every single iteration, when we can unwrap it once and for all? That loop is `O(len(value))` and `value` is user-supplied so it may be quite large, and it doesn't cost us much to avoid a bunch of inner loop work.

PR #362, opened 2019-06-18
Add filter-on-optional vertex example

LWprogramming:
References pull request #354

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ with runtime parameter
   "name": "Charles"
 }
 ```
-because the `Person_Knows` edge from `Albert` to `Betty` satisfies the `@optional` directive, but `Betty` doesn't match the filter checking for a node with name `Charles`.
+would output an empty list because the `Person_Knows` edge from `Albert` to `Betty` satisfies the `@optional` directive, but `Betty` doesn't match the filter checking for a node with name `Charles`.
 
 However, if no such `Person_Knows` edge existed from `Albert`, then the output would be
 ```python
--- a/README.md
+++ b/README.md
@@ -230,33 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
 coercion. These subsequent operations may then cause the result set to be discarded if it does
 not match.
 
-For example, suppose we have two Person nodes with names `Albert` and `Betty` such that there's an edge 
-`Albert-knows->Betty`.
+For example, suppose we have two `Person` vertices with names `Albert` and `Betty` such that there is a `Person_Knows` edge from `Albert` to `Betty`.
 
-Then for the following query:
+Then the following query:
 ```graphql
 {
   Person {
     out_Person_Knows @optional {
-      name @filter(op_name: "=", value: "Charles")
+      name @filter(op_name: "=", value: ["$name"])
     }
     name @output(out_name: "person_name")
   }
 }
 ```
-then the output would be 
-```graphql
-{}
+with runtime parameter
+```python
+{
+  "name": "Charles"
+}
 ```
-because the `Albert-knows->Betty` edge satisfies the `@optional` directive, but this edge is then discarded because it doesn't match the filter checking if Betty's name is Charles. 
+because the `Person_Knows` edge from `Albert` to `Betty` satisfies the `@optional` directive, but `Betty` doesn't match the filter checking for a node with name `Charles`.
 
-However, if no such outgoing `knows` edge existed from Albert, then the output would be
-```graphql
+However, if no such `Person_Knows` edge existed from `Albert`, then the output would be
+```python
 {
   name: 'Albert'
 }
 ```
-because no such edge can satisfy the @optional directive, and no filtering happens.
+because no such edge can satisfy the `@optional` directive, and no filtering happens.
 
 ### @output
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
 coercion. These subsequent operations may then cause the result set to be discarded if it does
 not match.
 
+For example, suppose we have two Person nodes with names `Albert` and `Betty` such that there's an edge 
+`Albert-knows->Betty`.
```
pmantica1: I think it'd be best if we used proper GraphQL terminology here. How about:

For example, suppose we have two `Person` vertices with names `Albert` and `Betty` such that there is a `Person_Knows`  edge from `Albert` to `Betty`.

README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
...
+  Person {
+    out_Person_Knows @optional {
+      name @filter(op_name: "=", value: "Charles")
+    }
+    name @output(out_name: "person_name")
+  }
+}
+```
+then the output would be 
+```graphql
```
pmantica1: nit: This should be python syntax highlighting.

README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
...
+{}
+```
+because the `Albert-knows->Betty` edge satisfies the `@optional` directive, but this edge is then discarded because it doesn't match the filter checking if Betty's name is Charles. 
+
+However, if no such outgoing `knows` edge existed from Albert, then the output would be
```
pmantica1: nit: if no such outgoing `Person_Knows` edge

README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
 coercion. These subsequent operations may then cause the result set to be discarded if it does
 not match.
 
+For example, suppose we have two Person nodes with names `Albert` and `Betty` such that there's an edge 
+`Albert-knows->Betty`.
+
+Then for the following query:
```
pmantica1: nit: Then the following query:

README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
...
+    }
+    name @output(out_name: "person_name")
+  }
+}
+```
+then the output would be 
+```graphql
+{}
+```
+because the `Albert-knows->Betty` edge satisfies the `@optional` directive, but this edge is then discarded because it doesn't match the filter checking if Betty's name is Charles.
```
pmantica1: This part of the sentence is not technically correct. For instance, the result set is discarded not the edge. I am having trouble figuring out how to rephrase this. @obi1kenobi what do you think?

README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
...
+```
+because the `Albert-knows->Betty` edge satisfies the `@optional` directive, but this edge is then discarded because it doesn't match the filter checking if Betty's name is Charles. 
+
+However, if no such outgoing `knows` edge existed from Albert, then the output would be
+```graphql
+{
+  name: 'Albert'
+}
+```
+because no such edge can satisfy the @optional directive, and no filtering happens.
```
pmantica1: same as above.

README.md
```diff
@@ -230,6 +230,34 @@ the `@optional` does not apply, and that value is then checked against the filte
...
 
+For example, suppose we have two Person nodes with names `Albert` and `Betty` such that there's an edge 
+`Albert-knows->Betty`.
+
+Then for the following query:
+```graphql
+{
+  Person {
+    out_Person_Knows @optional {
+      name @filter(op_name: "=", value: "Charles")
```
obi1kenobi: This is not a legal GraphQL query, this has to be a runtime parameter since literals are not allowed.

README.md
obi1kenobi: Per the GraphQL query below, the edge name is `Person_Knows` so let's stay as consistent as possible.

README.md
obi1kenobi: We can just say "would return an empty list".

README.md
obi1kenobi: Remember to put backticks around `@optional`.

README.md
pmantica1: @obi1kenobi ?

LWprogramming: How's this: 

because the `Person_Knows` edge from `Albert` to `Betty` satisfies the `@optional` directive, but `Betty` doesn't match the filter checking for a node with name `Charles`.

README.md
LWprogramming: Done-- will be reflected in the next commit (same goes for all the conversations I resolve here before committing, just convenient to keep track of things).

README.md
pmantica1: Ok sounds good.

README.md
```diff
@@ -230,6 +230,35 @@ the `@optional` does not apply, and that value is then checked against the filte
...
+  }
+}
+```
+with runtime parameter
+```python
+{
+  "name": "Charles"
+}
+```
+because the `Person_Knows` edge from `Albert` to `Betty` satisfies the `@optional` directive, but `Betty` doesn't match the filter checking for a node with name `Charles`.
```
pmantica1: I think you are missing "would output an empty list"

LWprogramming: D'oh, surprised that got deleted. Fixed now.

PR #366, opened 2019-06-24
Create Cypher integration tests

LWprogramming:
- Created a client for Neo4j backend, including docker-compose file.

- Several TODOs remain: One in conftest.py, two in cypher_formatting.py. I'll continue working on the one for dates, datetimes, etc. next-- in the meantime, feel free to read over the other TODOs.

- Small issue: although the language spec says that we can specify the Cypher version in queries, when I tested this (both via the integration tests as well as manually in my browser, the queries failed).

```cypher
CYPHER 9
MATCH (n)
RETURN n
```

failed, but 

```cypher
CYPHER
MATCH (n)
RETURN n
```

was fine. Changing this is simple (just find-and-replace for all instances of `CYPHER 9`), but let's confirm that this is a legitimate issue and not just some setup issue on my computer first.


LWprogramming:
will be breaking this up into one PR for the compiler changes and another for the integration tests themselves, so I'll be closing this PR.

PR #368, opened 2019-06-24
Cypher support compiler changes

LWprogramming:
Part of PR #366. This PR is just for the compiler changes themselves.

Comments from the original PR:
- Two TODOs in cypher_formatting.py. I'll continue working on the one for dates, datetimes, etc. next-- in the meantime, feel free to read over the other TODO about the safe string insertion.

- Small issue: although the language spec says that we can specify the Cypher version in queries, when I tested this (both via the integration tests as well as manually in my browser, the queries failed).
```cypher
CYPHER 9
MATCH (n)
RETURN n
```
failed, but
```cypher
CYPHER
MATCH (n)
RETURN n
```
was fine. Changing this is simple (just find-and-replace for all instances of CYPHER 9), but let's confirm that this is a legitimate issue and not just some setup issue on my computer first.

### DIFF
--- a/graphql_compiler/query_formatting/cypher_formatting.py
+++ b/graphql_compiler/query_formatting/cypher_formatting.py
@@ -25,11 +25,9 @@ def _safe_cypher_string(argument_value):
 
     # Using JSON encoding means that all unicode literals and special chars
     # (e.g. newlines and backslashes) are replaced by appropriate escape sequences.
-    # However, the quoted result is wrapped in double quotes, and $ signs are not escaped,
-    # so that would allow arbitrary code execution in Cypher.
-    # TODO: Leon
-    #  does the gremlin strategy work? alternatively, just wrap everything in backticks. But then
-    #  we need to escape backticks.
+    # Since the compiler only allows variables with upper or lower case letters (A-Z, a-z)
+    # or underscores (_), arbitrary code execution is not possible, so we don't need any
+    # other special handling for these strings.
     escaped_and_quoted = json.dumps(argument_value)
     return escaped_and_quoted
 
@@ -40,9 +38,25 @@ def _safe_cypher_decimal(argument_value):
     return 'toFloat(' + _safe_cypher_string(str(decimal_value)) + ')'
 
 
-def _safe_cypher_date_and_datetime(expected_type, param, argument_value):
-    # TODO: Leon
-    raise NotImplementedError
+def _safe_cypher_date_and_datetime(graphql_type, expected_python_types, value):
+    """Represent date and datetime objects as Cypher strings."""
+    # Python datetime.datetime is a subclass of datetime.date,
+    # but in this case, the two are not interchangeable.
+    # Rather than using isinstance, we will therefore check for exact type equality.
+    value_type = type(value)
+    if not any(value_type == x for x in expected_python_types):
+        raise GraphQLInvalidArgumentError(u'Expected value to be exactly one of '
+                                          u'python types {}, but was {}: '
+                                          u'{}'.format(expected_python_types, value_type, value))
+
+    # The serialize() method of GraphQLDate and GraphQLDateTime produces the correct
+    # ISO-8601 format that MATCH expects. We then simply represent it as a regular string.
+    try:
+        serialized_value = graphql_type.serialize(value)
+    except ValueError as e:
+        raise GraphQLInvalidArgumentError(e)
+
+    return _safe_cypher_string(serialized_value)
 
 
 def _safe_cypher_list(inner_type, argument_value):
--- a/graphql_compiler/query_formatting/cypher_formatting.py
+++ b/graphql_compiler/query_formatting/cypher_formatting.py
@@ -20,8 +20,9 @@ def _safe_cypher_string(argument_value):
         if isinstance(argument_value, bytes):  # should only happen in py3
             argument_value = argument_value.decode('utf-8')
         else:
-            raise GraphQLInvalidArgumentError(u'Attempting to convert a non-string into a string: '
-                                              u'{}'.format(argument_value))
+            raise GraphQLInvalidArgumentError(
+                u'Attempting to convert a non-string into a string: ' u'{}'.format(argument_value)
+            )
 
     # Using JSON encoding means that all unicode literals and special chars
     # (e.g. newlines and backslashes) are replaced by appropriate escape sequences.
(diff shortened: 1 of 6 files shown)


LWprogramming:
> Changing this is simple (just find-and-replace for all instances of CYPHER 9), but let's confirm that this is a legitimate issue and not just some setup issue on my computer first.

It turns out that version numbers _are_ supported, just not the ones expected. The Docker container in the integration tests for PR #369 is the most up-to-date image currently on Dockerhub (version 3.5.6).

Running `CYPHER 9.0 MATCH (n) RETURN n` yields this error: `Neo.ClientError.Statement.SyntaxError: Supported CypherVersion values are: 2.3, 3.1, 3.4, 3.5`.

If we run `CYPHER 3.5 MATCH (n) RETURN n`, the query actually does run without erroring, but everywhere I looked (in particular, the language spec) suggested that `CYPHER 9` is a valid version number, which is the reason for the confusion.

Given that 3.5 works, shall we change the queries to use that version number then?

pmantica1:
Please see:
https://github.com/Mats-SX/openCypher/blob/7e78a2a1e19ae6aed545a3bcae2a66941ca18038/cip/1.accepted/CIP2017-10-17-Cypher-9.adoc

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,135 @@
...
+
+def insert_arguments_into_cypher_query(compilation_result, arguments):
+    """Insert the arguments into the compiled Cypher query to form a complete query.
+
+    Args:
+        compilation_result: CompilationResult, compilation result from the GraphQL compiler.
+        arguments: Dict[str, Any], parameter name -> value, for every parameter the query expects.
+
+    Returns:
+        string, a Cypher query with  with inserted argument data.
```
pmantica1: Please fix typo.

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,134 @@
...
+        else:
+            raise GraphQLInvalidArgumentError(u'Attempting to convert a non-string into a string: '
+                                              u'{}'.format(argument_value))
+
+    # Using JSON encoding means that all unicode literals and special chars
+    # (e.g. newlines and backslashes) are replaced by appropriate escape sequences.
+    # However, the quoted result is wrapped in double quotes, and $ signs are not escaped,
+    # so that would allow arbitrary code execution in Cypher.
+    # TODO: Leon
+    #  does the gremlin strategy work? alternatively, just wrap everything in backticks. But then
```
pmantica1: This seems to be quite important for security purposes so I won't merge the branch until this is addressed. Could you please figure this out? Maybe ask Predrag when he is unoccupied. Though it could be faster to figure this out by yourself. The rest of the branch looks good. Also please do not pose question TODOs like this one in the future. It's really a question. Not a TODO.

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,134 @@
...
+######
+# Public API
+######
+
+
+def insert_arguments_into_cypher_query(compilation_result, arguments):
+    """Insert the arguments into the compiled Cypher query to form a complete query.
+
+    Args:
+        compilation_result: CompilationResult, compilation result from the GraphQL compiler.
```
ReaLNeroM: Make this consistent with https://github.com/kensho-technologies/graphql-compiler/blob/b6e0dcfdd507023a64da884c7afd6609ab16e7bb/graphql_compiler/query_formatting/match_formatting.py#L124

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,134 @@
...
+        arguments: Dict[str, Any], parameter name -> value, for every parameter the query expects.
```
ReaLNeroM: I've usually seen ```dict, str -> any, mapping x to y``` in docstrings. Is there a reason why this is different?

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,134 @@
...
+from ..compiler.helpers import strip_non_null_from_type
+from ..exceptions import GraphQLInvalidArgumentError
+from ..schema import GraphQLDate, GraphQLDateTime, GraphQLDecimal
+from .representations import coerce_to_decimal, represent_float_as_str, type_check_and_str
+
+
+def _safe_cypher_string(argument_value):
+    """Sanitize and represent a string argument in Cypher."""
+    if not isinstance(argument_value, six.string_types):
+        if isinstance(argument_value, bytes):  # should only happen in py3
```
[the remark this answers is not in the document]
LWprogramming: I'm a bit confused by what you mean-- if the version isn't py3, it will enter the else-block and raise a GraphQLInvalidArgumentError anyways.

graphql_compiler/query_formatting/cypher_formatting.py
LWprogramming: good call, will fix

graphql_compiler/query_formatting/cypher_formatting.py
LWprogramming: There doesn't seem to be any one format that's used everywhere-- for uniformity I'll change this so it matches the description in gremlin_ and match_formatting.py

LWprogramming: and add types to the comment as well.

graphql_compiler/query_formatting/cypher_formatting.py
LWprogramming: Went over it with Predrag. Cypher doesn't allow arbitrary code execution in its strings that way Groovy [does](http://www.groovy-lang.org/syntax.html#_double_quoted_string) so we're good to go. Dates and datetime formatting is fine as well.

graphql_compiler/query_formatting/cypher_formatting.py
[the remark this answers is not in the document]
pmantica1: The docstring is inaccurate in a couple of ways. So in python3 str objects are by default stored as Unicode code points while in python2 they are by default stored as bytes. However, in both versions, they can be encoded as both Unicode code points and strings. So feel free to fix the comment` # likely to only happen` in py2 all the similar parts in the repo.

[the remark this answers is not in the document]
LWprogramming: fixed

PR #369, opened 2019-06-24
Cypher support integration tests

LWprogramming:
Part of PR #366. This PR is just for the integration tests themselves.

- There's a TODO about the right client abstraction for the Cypher client, but otherwise this is ready to merge. Once the TODO is resolved and there are no other issues, this'll be ready for merge.


graphql_compiler/tests/test_data_tools/neo4j.py
```diff
@@ -0,0 +1,27 @@
...
+    """Return an Neo4j path for the specified database on the NEO4J_SERVER."""
+    template = 'bolt://{}/{}'
+    return template.format(NEO4J_SERVER, database_name)
+
+
+def get_test_neo4j_graph(graph_name, generate_data_func):
+    """Generate the test database and return the Neo4j client."""
+    client = Neo4jClient(graph_name)
+    generate_data_func(client)
+    return client
```
pmantica1: please add newline.

graphql_compiler/tests/test_compiler.py
```diff
@@ -422,7 +422,7 @@ def test_traverse_and_output(self):
         '''
         expected_sql = NotImplementedError
         expected_cypher = '''
-            CYPHER 9
+            CYPHER
```
pmantica1: Please make a separate branch where you change the cypher version.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -26,12 +27,15 @@
     test_backend.MYSQL,
     test_backend.SQLITE,
     test_backend.MSSQL,
+    test_backend.NEO4J,
+    # test_backend.REDISGRAPH,
```
pmantica1: This seems like a TODO. Mind adding it as one.

graphql_compiler/tests/integration_tests/integration_backend_config.py
```diff
@@ -27,6 +27,11 @@
     test_backend.ORIENTDB,
 }
 
+CYPHER_BACKENDS = {
+    test_backend.NEO4J,
+    # test_backend.REDISGRAPH,
```
pmantica1: same as below.

graphql_compiler/tests/conftest.py
```diff
@@ -50,6 +52,35 @@ def _init_graph_client(load_schema_func, generate_data_func):
     return graph_client
 
 
+@pytest.fixture(scope='session')
+def init_integration_neo4j_client():
+    """Return a client for an initialized db, with all test data imported."""
+    return _init_neo4j_client(generate_neo4j_integration_data)
+
+
+def _init_neo4j_client(generate_data_func):
```
pmantica1: What about making a retry decorator with most of the logic here?

graphql_compiler/tests/test_compiler.py
LWprogramming: #370

LWprogramming: will be undoing the change here to avoid merge conflict.

graphql_compiler/tests/conftest.py
LWprogramming: I like that idea-- will do something similar for graph_client as well (aka orientdb client)

LWprogramming: Also, I think it makes more sense to keep these two separate since there's no load_schema function for Neo4j but there is for Orientdb.

LWprogramming:
Note: the CI tests will fail here because I separated the integration tests from the implementation itself (in the cypher_support_compiler_changes branch). Once everything is merged into cypher_support, then things should be ok.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -0,0 +1,139 @@
...
+                             .format(cypher_step))
+
+    has_where_block = cypher_step.where_block is not None
+    is_optional_step = (
+        isinstance(cypher_step.step_block, Traverse) and
+        cypher_step.step_block.optional
+    )
+    if has_where_block and is_optional_step:
+        raise AssertionError(u'Received an illegal CypherStep containing an optional step together '
+                             u'with a "where" Filter block. This is a bug in the lowering code, as '
```
ReaLNeroM: Are you 100% certain the bug is always going to be due to the lowering code? 
You could have just replaced the second sentence with "The filtering should have been moved to the global operations section".

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -0,0 +1,139 @@
...
+            direction = cypher_step.step_block.direction
+            edge_name = cypher_step.step_block.edge_name
+
+        template_data['edge_type'] = edge_name
+
+        direction_lookup = {
+            'in': ('left_edge_mark', u'<'),
+            'out': ('right_edge_mark', u'>'),
+        }
+        key, value = direction_lookup[direction]
```
ReaLNeroM: Could you make the variable names clearer?

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -0,0 +1,139 @@
...
+        template_data[key] = value
+
+    if isinstance(cypher_step.step_block, Recurse):
+        template_data['quantifier'] = u'*0..%d' % cypher_step.step_block.depth
+
+    # Try to obey the Cypher style guidebook at least a little bit.
```
ReaLNeroM: Way too many words to communicate conformity to the guidebook.
"Comply with Cypher style guidebook on whitespace a bit.", perhaps?

LWprogramming: sounds good to me

graphql_compiler/compiler/emit_cypher.py
LWprogramming: Filters in optional locations get moved to the global operations section in the method `move_filters_in_optional_locations_to_global_operations` in `compiler/ir_lowering_cypher/__init__.py` so seeing a filter block in an optional section means that function didn't work correctly.

The `emit_code_from_ir` function gets called immediately after the lowering function is called, so we don't have a situation where code in the middle can mix it up either.

graphql_compiler/compiler/emit_cypher.py
LWprogramming: done

LWprogramming:
Fixed the merge conflict, which clears up the problems with the PR.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -80,6 +105,13 @@ def compile_and_run_query(cls, graphql_query, parameters, backend_name):
         elif backend_name in MATCH_BACKENDS:
             results = compile_and_run_match_query(
                 cls.schema, graphql_query, parameters, cls.orientdb_client)
+        elif backend_name in NEO4J_BACKENDS:
+            results = compile_and_run_neo4j_query(
+                cls.schema, graphql_query, parameters, cls.neo4j_client)
+            print('results ' + str(results))
```
obi1kenobi: Remove this print.

graphql_compiler/tests/test_data_tools/integration_data/create_redisgraph_integration.cypher
```diff
@@ -0,0 +1,16 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+# This data is almost identical to the neo4j integration test data except for the first line, which doesn't appear in
```
obi1kenobi: Add a reference to the Github issue you opened about this, so we know when we can remove this black magic code.

graphql_compiler/tests/test_data_tools/neo4j.py
```diff
@@ -0,0 +1,28 @@
...
+from neo4j import GraphDatabase
+
+
+NEO4J_SERVER = 'localhost'
+NEO4J_PORT = 7687
+NEO4J_USER = 'neo4j'
+NEO4J_PASSWORD = 'root'
+
+
+class Neo4jClient:
```
obi1kenobi: This won't work as you'd expect in Python 2, since it will create an "old-style" class. You should make it `Neo4jClient(object)` instead.

graphql_compiler/tests/test_data_tools/integration_data/create_neo4j_integration.cypher
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+# delete all leftover nodes from previous tests
+match (n) detach delete n
+create (:Animal {name: 'Animal 1', net_worth: toFloat('100'), uuid: 'cfc6e625-8594-0927-468f-f53d864a7a51'})
```
obi1kenobi: Instead of `toFloat()`, can we just set `100.0`, `200.0` etc.?

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -63,3 +63,36 @@ def compile_and_run_sql_query(schema, graphql_query, parameters, engine, metadat
...
+
+def compile_and_run_redisgraph_query(schema, graphql_query, parameters, redisgraph_client):
+    """Compiles and runs a Cypher query against the supplied graph client."""
+    converted_parameters = {
+        name: try_convert_decimal_to_string(value)
+        for name, value in six.iteritems(parameters)
+    }
+    compilation_result = graphql_to_cypher_redisgraph(schema, graphql_query, converted_parameters)
+    query = compilation_result.query
+    query = query.replace('CYPHER 3.5', '') # TODO Leon: hack-- need to remove version number
```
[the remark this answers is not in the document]
LWprogramming: pushed the commit to the other PR, will remove in my next commit to here.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -63,3 +63,36 @@ def compile_and_run_sql_query(schema, graphql_query, parameters, engine, metadat
...
+    result_set = redisgraph_client.query(query).result_set
+    # redisgraph gives us back bytes, but we want strings.
+    column_names = [column_name.decode('utf-8') for column_name in result_set[0]]
```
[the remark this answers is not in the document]
LWprogramming: It's partially documented.

Redisgraph does specify how the output is formatted:

https://oss.redislabs.com/redisgraph/result_structure/

In short: result_set is a list `[column_names, data, metadata]`

Column_names is a list of attributes for the nodes, data is a list of lists (each entry in data corresponds to a single node), and metadata is a list containing information such as how long the query took. 

------

As for the Redisgraph python client, it seems that the version I have locally returns the first two elements of result_set but inspecting the code shows that it differs from the version in the [Github repo](https://github.com/RedisGraph/redisgraph-py/blob/master/redisgraph/graph.py#L106). Either way, there's no documentation for the client itself.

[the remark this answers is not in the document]
LWprogramming: The README for the redisgraph python client does show looping through `[1:]` to access each record, though.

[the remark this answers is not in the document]
LWprogramming: Follow up-- after further experimentation, it appears that both the Redis and Redisgraph clients return their data in the two-element list format, and the `result_set` field of a `QueryResult` for the Redisgraph client is a list of lists, where the first element in `result_set` is a list of column names, and each subsequent item in `result set` is a list corresponding to a node's or edge's property.

For instance, 

```python
>>> rg.query('match (n)-[r]->(m) return n, r, m').result_set
[[b'n.uuid', b'n.net_worth', b'n.name', b'r', b'm.uuid', b'm.net_worth', b'm.name'], [b'ce4f0889-ecdb-4e27-8ffa-3140eb507549', None, b'Species 1', b'', b'660b4d07-37fe-4eeb-bb9f-fbc3845e35a9', None, b'Species 2']]

```

graphql_compiler/tests/conftest.py
```diff
@@ -30,25 +34,37 @@ def init_integration_orientdb_client():
     return _init_orientdb_client(load_schema, generate_orient_integration_data)
 
 
+@retry(tries=20, timeout=1)  # pylint: disable=no-value-for-parameter
```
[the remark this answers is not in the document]
LWprogramming: Noted. Turns out it was a misunderstanding about decorators in this case so it'll be removed.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -63,3 +63,36 @@ def compile_and_run_sql_query(schema, graphql_query, parameters, engine, metadat
...
+    rows = [[row_item.decode('utf-8') if type(row_item) in (bytes, bytearray) else row_item for row_item in row]
```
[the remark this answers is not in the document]
LWprogramming: cleaned this up, hopefully this is easier to read

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -26,12 +29,36 @@
...
+    test_backend.ORIENTDB,
+    test_backend.POSTGRES,
+    test_backend.MARIADB,
+    test_backend.MYSQL,
+    test_backend.SQLITE,
+    test_backend.MSSQL,
+    test_backend.NEO4J,
+])
+
+all_backends_but_cypher = parameterized.expand([
```
pmantica1: I am not a big fan of this idea. I prefer having a 
if backend==test_backend.NEO4J:
    return

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -63,3 +64,43 @@ def compile_and_run_sql_query(schema, graphql_query, parameters, engine, metadat
...
+    query = compilation_result.query
+    with neo4j_client.driver.session() as session:
+        results = session.run(query, parameters)
+    return results.data()
+
+
+def compile_and_run_redisgraph_query(schema, graphql_query, parameters, redisgraph_client):
+    """Compile and run a Cypher query against the supplied graph client."""
+    converted_parameters = {
+        name: try_convert_decimal_to_string(value)
```
obi1kenobi: There shouldn't be any `Decimal`-typed values in tests involving Cypher, no? I thought we decided that neither RedisGraph nor Neo4j supported such a data type.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -63,3 +64,43 @@ def compile_and_run_sql_query(schema, graphql_query, parameters, engine, metadat
...
+    column_names = result_set[0]
+    records = result_set[1:]
+
+    # redisgraph gives us back bytes, but we want strings.
+    decoded_column_names = [column_name.decode('utf-8') for column_name in column_names]
+    decoded_records = []
+    for record in records:
+        # decode if bytes, leave alone otherwise. For more info see here:
+        # https://oss.redislabs.com/redisgraph/result_structure/#top-level-members
+        decoded_record = [field.decode('utf-8') if type(field) in (bytes, bytearray) else field
```
obi1kenobi: nit: since this is a multi-line comprehension, let's stick to the multi-line format used elsewhere in the project:
```
foo = [
    bar if baz else quux
    for bar in bars
]
```

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -27,12 +30,36 @@
     test_backend.MYSQL,
     test_backend.SQLITE,
     test_backend.MSSQL,
+    test_backend.NEO4J,
+    test_backend.REDISGRAPH,
+])
+
+# TODO Leon this is really bad but it works for now
```
obi1kenobi: Yeah let's not do this. Why can't we generate the parameterization with toggle-able backends? Ideally, we'd like to be able to do things like `@test_against_all_backends(except_redisgraph=True)` or something like that, instead of writing one of these for every combination of backends we'd like to include or exclude.

LWprogramming:
> I get the feeling that it's trying to do a few too many things at once

Fair point! It does look like a lot because there are changes spread out over many files, but what it essentially is doing can be boiled down to "integration tests for Cypher".

This involves creating a file for each client (for Neo4j and RedisGraph), creating integration test data, the fixtures that allow the tests to use the client, the tests themselves, etc. but they all need to go together (hard to have tests without the client!) so I'm not sure how such a PR would be split up. Open to suggestions though!

As for the TODOs, I plan to review them and clean them up before merging. If I come across any that probably need further discussion, at that point I definitely agree it'd be good to make a separate PR for those issues.

graphql_compiler/tests/integration_tests/test_backends_integration.py
LWprogramming: Commit aedc4a9 does exactly that now :)

graphql_compiler/tests/integration_tests/test_backends_integration.py
LWprogramming: See commit aedc4a9 for fix

graphql_compiler/tests/integration_tests/integration_test_helpers.py
LWprogramming: Ah, right-- this a leftover from before we found out about that. Will remove

graphql_compiler/tests/integration_tests/integration_test_helpers.py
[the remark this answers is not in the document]
LWprogramming: Updated comment explaining formatting.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -12,31 +12,60 @@
...
 # Individual tests can supply the full @pytest.mark.usefixtures to override if necessary.
 integration_fixtures = pytest.mark.usefixtures(
+    'integration_neo4j_client',
     'integration_orientdb_client',
+    'integration_redisgraph_client',
     'sql_integration_data',
 )
 
 
+def almost_all_backends(excluded_backends):
```
LWprogramming: I'm open to renaming this, since it might not be super-clear from the name alone what this is doing. Maybe `use_almost_all_backends`?

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -12,31 +12,60 @@
...
     test_backend.SQLITE,
     test_backend.MSSQL,
-])
+    test_backend.NEO4J,
+    test_backend.REDISGRAPH,
+]
+
+# Store the test parametrization for running against all backends. Individual tests can customize
+# the list of backends to test against with the full @parametrized.expand([...]) decorator.
+all_backends = parameterized.expand(all_backends_list)
```
LWprogramming: We can also consider this a special case for the `almost_all_backends` decorator with `excluded_backends` as just an empty list. How do we feel about replacing `all_backends` with the `almost_all_backends` decorator? We could rename the `almost_all_backends` decorator as well to make things clearer.

graphql_compiler/tests/conftest.py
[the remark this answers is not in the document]
LWprogramming: Update: turns out my local lint.sh script doesn't complain but the Travis CI does-- that's why I originally put them there

Pipfile
```diff
@@ -32,6 +32,9 @@ graphql-core = ">=2.1,<3"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
+neo4j = "==1.7.4"
```
LWprogramming: General question: I notice that some of these other ones don't specify a particular version but rather a set of possible versions. How big a deal is this? I remember being told in some of the Github issues that some functionality that we rely on (e.g. the output format for RedisGraph) would change in future releases.

graphql_compiler/tests/conftest.py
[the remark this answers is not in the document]
LWprogramming: Hopefully final update: after fixing whatever Pipfile mishaps I was dealing with in commit 6f84837, now the `retry` decorator complaints are showing up locally as well. I'll add it back and leave a comment explaining its purpose.

graphql_compiler/tests/integration_tests/test_backends_integration.py
obi1kenobi: I'm hesitant to have two definitions. Also, `almost_all_backends` with no arguments seems like bad design -- it doesn't communicate that it actually doesn't exclude any backends. Why not have one definition, `all_backends` and allow an exclude to be passed to it? Then, if there are no args, it does the obvious thing, and if it's called with a kwarg called e.g. `except_backends` then it also does the obvious thing?

graphql_compiler/tests/conftest.py
```diff
@@ -30,25 +33,45 @@ def init_integration_orientdb_client():
     return _init_orientdb_client(load_schema, generate_orient_integration_data)
 
 
+# retry is a decorator that attempts to call the decorated function repeatedly, and it takes a
+# parameter "call". Call is the function being decorated, so we don't need to specify that in
+# the parameters that the retry decorator takes, so we can disable pylint's warning about
```
obi1kenobi: Can you share the error pylint raises here? This comment effectively states that pylint has buggy handling of decorators, which is definitely possible but bears verifying before we silence the error.

Also, `tries` is a positional argument, not a kwarg -- so let's pass it positionally.

LWprogramming: in PR 410 https://github.com/kensho-technologies/graphql-compiler/pull/410#discussion_r304635008

graphql_compiler/tests/integration_tests/test_backends_integration.py
LWprogramming: https://github.com/kensho-technologies/graphql-compiler/pull/411/files#r304643551

LWprogramming:
All the pull requests from this branch have been merged into master, so this is done.

PR #373, opened 2019-06-26
Cypher support emit cypher and some compiler tests

LWprogramming:
Part of PR #325 containing just Predrag's commits, specifically the test_compiler tests and emit_cypher.

I tried splitting this up into two branches, one for emit cypher and one for test_compiler, but couldn't because things are tangled together. For instance, commit `f35653c` modifies `cypher_query.py` (which has already been merged) and `emit_cypher.py` (which has not). Probably not worth the time to split them apart in this case.


obi1kenobi:
Closed this because we got into an unfortunate catch-22, where all of these are my commits, @LWprogramming is blocked on merging them, and yet he was unable to review the PR since he opened it. Reopened as #374.

PR #377, opened 2019-06-27
Add support for edge_counts_between_vertex_pairs

ReaLNeroM:
Okay, I lied. I did quite a few changes:

- add tests for new statistic. Check if it overrides estimates using get_class_count.
- The cardinality estimator now tries to use the statistic in the title if possible, which helps out when using really general edges like Entity->Entity, when we really want to traverse between two rare vertices.
- As a design decision, when traversing from vertex type A to vertex type B, the estimator first finds out the number of edges from A to B in a separate function, then divides by the number of vertices of type A. To make this work, class_count needs to multiply by the count of A vertices, which we'll later cancel out. This means we need to add `BirthEvent` class_counts to tests that didn't previously have them, even though it doesn't impact the result.
- I made all statistics besides get_class_count optional by assigning them default values of `None`, and reflected that in the test set.

### DIFF
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -82,6 +82,75 @@ def _get_parent_and_child_name_from_edge(schema_graph, child_location):
     return parent_name_from_edge, child_name_from_edge
 
 
+def _get_outbound_and_inbound_names_from_locations(query_metadata, child_location, parent_location):
+    """Get the out and in classnames of a location and its parent from the last edge information."""
+    edge_direction, _ = _get_last_edge_direction_and_name_to_location(child_location)
+    parent_name_from_location = query_metadata.get_location_info(parent_location).type.name
+    child_name_from_location = query_metadata.get_location_info(child_location).type.name
+    if edge_direction == INBOUND_EDGE_DIRECTION:
+        outbound_vertex_name = child_name_from_location
+        inbound_vertex_name = parent_name_from_location
+    elif edge_direction == OUTBOUND_EDGE_DIRECTION:
+        outbound_vertex_name = parent_name_from_location
+        inbound_vertex_name = child_name_from_location
+    return outbound_vertex_name, inbound_vertex_name
+
+
+def _probe_statistics_for_edge_count_between_vertex_pair(
+    statistics, query_metadata, child_location, parent_location
+):
+    """Probe statistics for the count of edges connecting two vertex classes."""
+    _, edge_name = _get_last_edge_direction_and_name_to_location(child_location)
+    outbound_vertex_name, inbound_vertex_name = _get_outbound_and_inbound_names_from_locations(
+        query_metadata, child_location, parent_location
+    )
+    probe_result = statistics.get_edge_count_between_vertex_pair(
+        outbound_vertex_name, inbound_vertex_name, edge_name
+    )
+    return probe_result
+
+
+def _estimate_edge_count_between_vertex_pair_using_class_count(
+    schema_graph, statistics, query_metadata, child_location, parent_location
+):
+    """Estimate the count of edges connecting two vertex classes via the class_count statistic.
+
+    Given a parent location of type A and a child location of type B, we estimate the number of AB
+    edges using class counts. If A and B are subclasses of C and D respectively, we only have access
+    to CD edges, so in general, we'll use (# of CD edges) / (# of C vertices), but since we're only
+    interested in edges to Bs, we'll scale this result by the fraction of Bs over Ds.
+
+    Args:
+        schema_graph: SchemaGraph object
+        statistics: Statistics object
+        query_metadata: QueryMetadataTable object
+        child_location: BaseLocation object
+        parent_location: BaseLocation object
+    """
+    _, edge_name = _get_last_edge_direction_and_name_to_location(child_location)
+    edge_counts = statistics.get_class_count(edge_name)
+
+    parent_name_from_location = query_metadata.get_location_info(parent_location).type.name
+    child_name_from_location = query_metadata.get_location_info(child_location).type.name
+    parent_name_from_edge, child_name_from_edge = _get_parent_and_child_name_from_edge(
+        schema_graph, child_location
+    )
+    # Scale edge_counts if child_location's type is a subclass of the edge's endpoint type.
+    if child_name_from_location != child_name_from_edge:
+        edge_counts *= (
+            float(statistics.get_class_count(child_name_from_location)) /
+            statistics.get_class_count(child_name_from_edge)
+        )
+    # Scale edge_counts if parent_location's type is a subclass of the edge's endpoint type.
+    if parent_name_from_location != parent_name_from_edge:
+        edge_counts *= (
+            float(statistics.get_class_count(parent_name_from_location)) /
+            statistics.get_class_count(parent_name_from_edge)
+        )
+
+    return edge_counts
+
+
 def _estimate_children_per_parent(
     schema_graph, statistics, query_metadata, parameters, child_location, parent_location
 ):
@@ -89,9 +158,7 @@ def _estimate_children_per_parent(
 
     Given a parent location of type A and child location of type B, assume all AB edges are
     distributed evenly over A vertices, so the expected number of child edges per parent vertex is
-    (# of AB edges) / (# of A vertices). If A and B are subclasses of C and D respectively, we only
-    have access to CD edges, so in general, we'll use (# of CD edges) / (# of C vertices), but since
-    we're only interested in edges to Bs, we'll scale this result by the fraction of Bs over Ds.
+    (# of AB edges) / (# of A vertices).
 
     Args:
         schema_graph: SchemaGraph object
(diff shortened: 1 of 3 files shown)


graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -82,57 +82,120 @@ def _get_parent_and_child_name_from_edge(schema_graph, child_location):
...
+
+
+def _estimate_edge_count_between_vertex_pair_using_class_count(
+    schema_graph, statistics, query_metadata, child_location, parent_location
+):
+    """Estimate the count of edges connecting two vertex classes via the class_count statistic.
+
+    Given a parent location of type A and a child location of type B, we estimate the number of AB
+    edges using class counts. If A and B are subclasses of C and D respectively, we only have access
+    to CD edges, so in general, we'll use (# of CD edges) / (# of C vertices), but since we're only
```
pmantica1: Shouldn't you multiply this by the number of A vertices too?

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -276,13 +241,9 @@ def test_fold(self):
             'Animal': 5,
             'Animal_BornAt': 7,
             'Animal_FedAt': 3,
+            'FeedingEvent': 11
```
pmantica1: You seem to be adding test data that doesn't change the tests. Why?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -159,7 +166,7 @@ def get_edge_count_between_vertex_pair(
             - int, the count of edges if the statistic exists
             - None otherwise
         """
-        statistic_key = (vertex_in_class_name, vertex_out_class_name, edge_class_name)
+        statistic_key = (vertex_out_class_name, vertex_in_class_name, edge_class_name)
```
pmantica1: Please update the docs.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
ReaLNeroM: I address this in my third bullet point (I didn't mention FeedingEvent, but it's the same reason).
It doesn't affect the test result since it gets cancelled out, but since it still gets used and it's a *required* statistic, we need to add this to our test data.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: The docs are correct, my code was just wrong :)

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Good catch!

graphql_compiler/cost_estimation/statistics.py
pmantica1: Actually, do you mind changing the order? (Having `vertex_in_class_name` first). I find it more intuitive to write first the source and then the destination.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
pmantica1: Oh ok got it.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: I spoke with @obi1kenobi about this exact thing. I fully agree with you, and this is what vertex_out and vertex_in reflects. 
You can think of it this way: vertex_out is where the edge goes out of, and vertex_in is where the goes into.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -82,57 +82,126 @@ def _get_parent_and_child_name_from_edge(schema_graph, child_location):
...
+
+def _estimate_edge_count_between_vertex_pair_using_class_count(
+    schema_graph, statistics, query_metadata, child_location, parent_location
+):
+    """Estimate the count of edges connecting two vertex classes via the class_count statistic.
+
+    Given a parent location of type A and a child location of type B, we estimate the number of AB
+    edges using class counts. If A and B are subclasses of C and D respectively, we only have
+    access to CD edges, so in general, we assume that the number of CD edges with A as an endpoint
+    are proportional to the fraction of A vertices over C vertices, and likewise for B and D.
```
LWprogramming: Thoughts on rewording this part to make our assumption more precise? One possible rewording below, but feel free to suggest changes. 

From 

"we assume that the number of CD edges with A as an endpoint are proportional to the fraction of A vertices over C vertices, and likewise for B and D."

to

"we assume that the edges between a vertex $c$ of type C and vertex $d$ of type D are independent of whether $c$ is of type $A$ and whether $d$ is of type $B$. In other words, we assume the fraction of edges from $c$ to $d$ such that $c$ is of type A is the same as the fraction of C vertices in general that are also of type A. Likewise, we assume the fraction of edges from $c$ to $d$ such that $d$ is of type B is the same as the fraction of D vertices in general that are also of type B."

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -82,57 +82,126 @@ def _get_parent_and_child_name_from_edge(schema_graph, child_location):
...
-    # Count the number of edges between child_location and parent_location type vertices.
     _, edge_name = _get_last_edge_direction_and_name_to_location(child_location)
-    # TODO(evan): If edge is recursed over, we need a more detailed statistic
-    # TODO(vlad): Use get_edge_count_between_vertex_pair statistic
     edge_counts = statistics.get_class_count(edge_name)
 
-    # Scale edge_counts if child_location's type is a subclass of the edge's endpoint type.
+    parent_name_from_location = query_metadata.get_location_info(parent_location).type.name
+    child_name_from_location = query_metadata.get_location_info(child_location).type.name
     parent_name_from_edge, child_name_from_edge = _get_parent_and_child_name_from_edge(
```
LWprogramming: Let's rename these variables to `child_` and `parent_base_classname` to make it clear that these names might be more general than the names we get from the locations themselves in `_get_outbound_and_inbound_names_from_locations` (e.g. `Entity` instead of `Bank`). Same for the `_get_parent_and_child_name_from_edge` method above.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -82,57 +82,126 @@ def _get_parent_and_child_name_from_edge(schema_graph, child_location):
...
+    edge_direction, _ = _get_last_edge_direction_and_name_to_location(child_location)
+    parent_name_from_location = query_metadata.get_location_info(parent_location).type.name
+    child_name_from_location = query_metadata.get_location_info(child_location).type.name
+    if edge_direction == INBOUND_EDGE_DIRECTION:
+        outbound_vertex_name = child_name_from_location
+        inbound_vertex_name = parent_name_from_location
+    elif edge_direction == OUTBOUND_EDGE_DIRECTION:
+        outbound_vertex_name = parent_name_from_location
+        inbound_vertex_name = child_name_from_location
+    return outbound_vertex_name, inbound_vertex_name
```
LWprogramming: nit: add an else-block here to check that edge_direction is one of inbound or outbound.

same for `_get_parent_and_child_name_from_edge` (the method above)

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: > Thoughts on rewording this part to make our assumption more precise? One possible rewording below, but feel free to suggest changes.
> 
> From
> 
> "we assume that the number of CD edges with A as an endpoint are proportional to the fraction of A vertices over C vertices, and likewise for B and D."
> 
> to
> 
> "we assume that the edges between a vertex $c$ of type C and vertex $d$ of type D are independent of whether $c$ is of type $A$ and whether $d$ is of type $B$. In other words, we assume the fraction of edges from $c$ to $d$ such that $c$ is of type A is the same as the fraction of C vertices in general that are also of type A. Likewise, we assume the fraction of edges from $c$ to $d$ such that $d$ is of type B is the same as the fraction of D vertices in general that are also of type B."

I like the suggestion of more explicitly stating the assumption. Since it's very easy to explain this hand-wavingly, I'll try expressing it in form of an example.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -151,6 +131,123 @@ def test_complex_traverse(self):
...
+        params = {}
+
+        count_data = {
+            'Entity': 19,
+            'Animal': 3,
+            'Event': 7,
+            'Entity_Related': 11
+        }
+        edge_count_data = {
+            ('Animal', 'Event', 'Entity_Related'): 2
```
obi1kenobi: Can we make the order be `from -- along edge --> to`?

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -617,13 +660,9 @@ def test_optional_then_filter(self):
...
             'FeedingEvent': 6,
+            'BirthEvent': 13
         }
-        statistics = LocalStatistics(
-            class_counts=count_data,
-            edge_count_between_vertex_pairs=dict(),
-            count_of_distinct_field_values=dict(),
-            histograms=dict()
-        )
+        statistics = LocalStatistics(count_data)
```
obi1kenobi: Can you split out the changes that remove all the empty `dict()` parameters into a separate PR? It will shave hundreds of lines that are not that relevant or difficult to review, and make it easier for us to focus on the actual changes of substance here.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: I've reworded it. I'd love to hear your thoughts on the current iteration.

LWprogramming: nit: "with A as an endpoint" -> "with an endpoint of class A"

everything else looks good to me!

ReaLNeroM: That's not the version I decided on, probably some git thing that I'll fix soon.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,99 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+def _estimate_edge_count_between_vertex_pair_using_class_count(
+    schema_graph, statistics, query_metadata, child_location, parent_location
+):
+    """Estimate the count of edges connecting two vertex classes via the class_count statistic.
+
+    Given a parent location of class A and a child location of class B, this function estimates the
+    number of AB edges using class counts. If A and B are subclasses of the edge's endpoint classes
+    which we'll name C and D respectively, we only have statistics for CD edges, so in general, we
+    estimate using the assumption that CD edges are distributed among all C and D vertices
+    independently of the vertex's class.  Using this assumption, in general we estimate the
```
LWprogramming: "independently of the vertex's class." -> "independent of whether or not the vertex of class C is also a member of class A, and likewise for D and B."

Alternatively, split up the whole sentence since it's a bit of a run-on:

If A and B are subclasses of the edge's endpoint classes (which we'll name C and D respectively), we only have statistics for CD edges. To make estimates, we assume that CD edges are distributed among all C and D vertices independent of whether or not the vertex of class C is also a member of class A, and likewise for D and B. 

Can also add this sentence if you want to clarify.

In other words, we assume whether there is an edge from a vertex of class C to a vertex of class D is independent of whether or not the vertex of class C is also of class A and likewise for D and B.

PR #379, opened 2019-06-28
Cypher support emit output tests

LWprogramming:
Tests emit output for Cypher.

In particular these tests differ from the tests for Gremlin and MATCH because the emit output tests were written before QueryMetadataTable came along, whereas Cypher does depend on this data structure to work.

I've included things not present in the other tests, such as the GraphQL query that would produce the sequence of IR blocks in the test.


graphql_compiler/tests/test_emit_output.py
```diff
@@ -243,8 +248,8 @@ def test_output_inside_optional_traversal(self):
                     BinaryComposition(
                         u'!=',
                         # HACK(predrag): The type given to OutputContextVertex here is wrong,
-                        # but it shouldn't cause any trouble since it has absolutely nothing to do
-                        # with the code being tested.
+                        # but it shouldn't cause any trouble since it has absolutely nothing to
```
obi1kenobi: Any particular reason for this change?

graphql_compiler/tests/test_emit_output.py
```diff
@@ -289,3 +294,239 @@ def test_datetime_output_representation(self):
...
+
+class EmitCypherTests(unittest.TestCase):
+    def setUp(self):
+        """Disable max diff limits for all tests."""
+        self.maxDiff = None
+
+    def test_simple_immediate_output(self):
+        # corresponds to:
+        # graphql_string = '''{
+        #     Foo {
```
obi1kenobi: As much as possible, I'd try to keep all tests (even these ones) in sync with the test schema -- that makes it more likely that the various test helpers will be able to be reused. Not doing that to start with this module was a mistake, and it's a bit late to rectify it now.

graphql_compiler/tests/test_emit_output.py
```diff
@@ -289,3 +294,239 @@ def test_datetime_output_representation(self):
...
+        #         name @output(out_name: "foo_name")
+        #     }
+        # }'''
+
+        base_location = Location(('Foo',))
+        base_name_location = base_location.navigate_to_field('name')
+        base_location_info = LocationInfo(
+            parent_location=None,
+            type=GraphQLObjectType(name='Foo', fields={'name': GraphQLString}),
```
obi1kenobi: This is an example of a place where having the test helpers available would help -- you wouldn't have to construct a new type object, which reduces the setup footprint of the test and makes it less likely that the test setup itself breaks in the future.

graphql_compiler/tests/test_emit_output.py
```diff
@@ -289,3 +294,239 @@ def test_datetime_output_representation(self):
...
+            coerced_from_type=None,
+            optional_scopes_depth=0,
+            recursive_scopes_depth=0,
+            is_within_fold=False,
+        )
+
+        ir_blocks = [
+            QueryRoot({'Foo'}),
+            MarkLocation(base_location),
+            GlobalOperationsStart(),  # necessary for Cypher. Filter/output blocks come after this.
```
obi1kenobi: It should exist for all backends -- please add it if it's missing. `ConstructResult` is also a global operation.

graphql_compiler/tests/test_emit_output.py
LWprogramming: Linter was complaining about line lengths

obi1kenobi: The old line length was 99, and the linter I believe is set to enforce line length under 100. Is this your local linter? If so, can you double-check your linter configuration?

LWprogramming: Hmm, that's strange-- when I ran the linter originally, it did complain, but after changing it back, it stopped complaining. 

My best guess as to what happened is that an extra space or two snuck in by accident which is why it got upset -> I change the line length -> then get rid of the extra whitespace. In any case, that's resolved!

graphql_compiler/tests/test_emit_output.py
LWprogramming: Will modify in a separate PR so this PR is easier to review (this one will cover only the new Cypher tests)

graphql_compiler/tests/test_emit_output.py
LWprogramming: done for these! Other test changes will come soon.

PR #380, opened 2019-06-28
Cypher support date datetime

LWprogramming:
Waiting on PR #368. Also need to implement safe date and datetime in cypher_formatting.py once that file gets merged in, anyways.


LWprogramming:
Oops, misclicked ready for review-- my intention was to get the neo4j client back before merging this just in case any issues came up. I did that locally and it worked but I want to see the ci tests pass here too, just to be safe.

graphql_compiler/compiler/expressions.py
```diff
@@ -227,6 +227,33 @@ def to_cypher(self):
         # Cypher has built-in support for variable expansion, so we'll just emit a variable
         # definition and rely on Cypher to insert the value.
         self.validate()
+
+        # The Neo4j client allows us to pass dates and datetimes directly as arguments. E.g. this
```
pmantica1: Are you raising an error if somebody tries to pass in a date or datetime parameter? Also, I am against having this much documentation here on how to pass in datetime parameters for the neo4j library. I think that the overarching idea is that when you pass in parameters to neo4j you have to use the correct python type according to this link:
https://neo4j.com/docs/api/python-driver/  This documentation can be included elsewhere, but this doesn't seem to like the right place.

LWprogramming: What do you mean by raising an error? Including date(time) parameters along with the query is perfectly ok.

I'm open to cutting it down and simply saying that we don't process those types specially here (for other query languages such as MATCH and Gremlin, that's not the case) and telling them to go look at the integration tests for examples of passing in parameters.

pmantica1: I meant to say that ff somebody tries to pass in a date(time) parameter for RedisGraph, you should raise an error if someone tries to insert a date(time) parameter into a cypher RedisGraph query.

pmantica1: Yeah pointing to the integration tests makes sense to me.

LWprogramming: Addressed both. How does this new comment look?

graphql_compiler/compiler/expressions.py
```diff
@@ -227,6 +227,19 @@ def to_cypher(self):
         # Cypher has built-in support for variable expansion, so we'll just emit a variable
         # definition and rely on Cypher to insert the value.
         self.validate()
+
+        # The Neo4j client allows us to pass dates and datetimes directly as arguments. See the
```
pmantica1: Looks good to me. Just one nit: date objects and datetime objects, (dates and datetimes are not types in python and we want to be as clear as possible that we are referring to actual python types).

LWprogramming: done, ready to go

PR #381, opened 2019-07-01
Implement Cypher support for Neo4j and Redisgraph

LWprogramming:
~PR #368 was merged into cypher support but we still need to merge it into master.~

~No new changes here-- this is from PR 368 which merged cypher_support_compiler_changes into cypher_support-- it's just that at the time, it was supposed to merge into the cypher_support rather than master and that never got changed.~

Implement Cypher support for Neo4j and Redisgraph. There are some significant differences between the way the two backends support Cypher, see [1](https://oss.redislabs.com/redisgraph/cypher_support/) and [2](https://oss.redislabs.com/redisgraph/known_limitations/) for starters, which is why we separate the two.


LWprogramming:
The Neo4j client does understand query parameters, for instance from the [tutorial](https://neo4j.com/developer/python/):

```python
result = tx.run("CREATE (a:Greeting) "
                        "SET a.message = $message "
                        "RETURN a.message + ', from node ' + id(a)", message=message)
```
where message is a variable. 

However, `insert_arguments_into_{cypher, sql, gremlin, match}_query` ends up being used in `./__init__.py` in its `graphql_to_cypher` function and `test_end_to_end.py`, rather than with the Neo4j client. Right now there are no tests for Cypher in test_end_to_end.py but it follows the same pattern.

---------------

Unsure if redisgraph-py also supports query parameters. The documentation doesn't mention it, but I'll be able to test it out soon after this gets merged.

LWprogramming:
Update: I've looked through the entirety of the [redisgraph python client](https://github.com/RedisGraph/redisgraph-py/) codebase and haven't found anything for inserting arguments.

However there may be a [hacky way](https://github.com/andymccurdy/redis-py/issues/1176) to get this to work-- I'll update this as I find out more.

LWprogramming:
Follow up re: Neo4j's client being able to handle parameters.

It's pretty straightforward to call run with parameters-- something like this would be valid, given a Neo4j client called `neo4j_client`:
```python
query = '''
MATCH (n:Animal)
WHERE n.net_worth = $net_worth
RETURN n
'''
parameters = {'net_worth': 100}
with neo4j_client.driver.session() as session:
    results = session.run(query, parameters)
return results.data()
```
    
The problem is, the client only supports certain types e.g. int, float, bool, string, dict but not Decimal (and we're passing in all our floats in as type decimal.Decimal in the parameters dictionary). Worse yet, it looks like there's no support for dates and datetimes in the function that handles all of this. 

For reference, `PackStreamDehydrator`'s `dehydrate` [method](https://github.com/neo4j/neo4j-python-driver/blob/e3a772376c4fba8fc4007817cb6147f2b3e815ec/neo4j/types/__init__.py#L97) seems to be doing the type checking. `session.run()` calls the `fix_parameters` function which calls the `dehydrate` method. 

What approach makes more sense in this situation, use their library and try to patch up the types they don't support, or do the whole thing ourselves using the cypher formatting file?

graphql_compiler/__init__.py
```diff
@@ -198,3 +199,74 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
...
+    Returns:
+        a CompilationResult object, containing:
+            - query: string, the resulting compiled and parameterized query string
+            - language: string, specifying the language to which the query was compiled
+            - output_metadata: dict, output name -> OutputMetadata namedtuple object
+            - input_metadata: dict, name of input variables -> inferred GraphQL type, based on use
+    """
+    compilation_result = compile_graphql_to_cypher(
+        schema, graphql_query, type_equivalence_hints=type_equivalence_hints)
+    return compilation_result
```
LWprogramming: I'm not thrilled with this approach, because all the other methods have this line instead:

```    
return compilation_result._replace(
```
But this one doesn't because in `query_formatting/common.py`, line 136, we call `insert_arguments_into_cypher_query_redisgraph` which is RedisGraph specific (since RedisGraph doesn't support query parameters at all). The condition checks that the language is Cypher, so if we called `insert_arguments_into_query` here then we'd essentially be manually interpolating the arguments in.

Open to discussing alternative ways to organize this code.

obi1kenobi: Might be best to discuss this together in front of a whiteboard, it's kind of an unfortunate situation.

obi1kenobi:
> What approach makes more sense in this situation, use their library and try to patch up the types they don't support, or do the whole thing ourselves using the cypher formatting file?

How about this: let's simply add a `Float`-typed field somewhere in the schema, and use that? For databases that don't support `Decimal`-typed fields, let's not use the field that is defined as `Decimal` at all -- pretend it doesn't exist.

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,152 @@
...
+    # Since the compiler only allows variables with upper or lower case letters (A-Z, a-z)
+    # or underscores (_), arbitrary code execution is not possible, so we don't need any
+    # other special handling for these strings.
+    escaped_and_quoted = json.dumps(argument_value)
+    return escaped_and_quoted
+
+
+def _safe_cypher_decimal(argument_value):
+    """Cypher doesn't support decimals, only ints and floats, so we'll raise an error here."""
+    raise NotImplementedError
```
obi1kenobi: Raise a value of the error type, not the type itself: `raise NotImplementedError()`, note the parens.

Also, I'd include an error message explaining the same thing that's in the docstring, otherwise someone facing this error has to dig through the code to figure out why they got it.

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+######
+# Public API
+######
+
+
+def insert_arguments_into_cypher_query_redisgraph(compilation_result, arguments):
+    """Insert the arguments into the compiled Cypher query to form a complete query.
+    This is only for Redisgraph because Neo4j's client can do this on its own.
```
obi1kenobi: docstrings should have a one-line headline, followed by one blank line and then any other text.

graphql_compiler/__init__.py
```diff
@@ -198,3 +199,74 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
     schema_graph = get_orientdb_schema_graph(schema_data, [])
     return get_graphql_schema_from_schema_graph(schema_graph, class_to_field_type_overrides,
                                                 hidden_classes)
+
+
+def graphql_to_cypher_redisgraph(schema, graphql_query, parameters, type_equivalence_hints=None):
```
obi1kenobi: Thoughts on making the name use `redisgraph_cypher` order instead of `cypher_redisgraph`? In my mind, `RedisGraph` is a modifier for `Cypher`, not the other way round.

graphql_compiler/__init__.py
```diff
@@ -198,3 +199,74 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
...
+            - output_metadata: dict, output name -> OutputMetadata namedtuple object
+            - input_metadata: dict, name of input variables -> inferred GraphQL type, based on use
+    """
+    compilation_result = compile_graphql_to_cypher(
+        schema, graphql_query, type_equivalence_hints=type_equivalence_hints)
+    return compilation_result._replace(
+        query=insert_arguments_into_query(compilation_result, parameters))
+
+
+def graphql_to_cypher_neo4j(schema, graphql_query, parameters, type_equivalence_hints=None):
```
obi1kenobi: Similarly here, thoughts on `neo4j_cypher` instead of `cypher_neo4j`?

LWprogramming:
Summary of changes:
- Remove graph_to_neo4j_cypher function and make a note of it in the README and in the docstring for the RedisGraph function
- Nits

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,155 @@
...
+    if not isinstance(argument_value, six.string_types):
+        if isinstance(argument_value, bytes):  # likely to only happen in py2
+            argument_value = argument_value.decode('utf-8')
+        else:
+            raise GraphQLInvalidArgumentError(u'Attempting to convert a non-string into a string: '
+                                              u'{}'.format(argument_value))
+
+    # Using JSON encoding means that all unicode literals and special chars
+    # (e.g. newlines and backslashes) are replaced by appropriate escape sequences.
+    # Since the compiler only allows variables with upper or lower case letters (A-Z, a-z)
```
pmantica1: Correct me if I am wrong, but I think that this is true for the name of the variable. Not for the value of the variable. So I don't see how this is relevant.

LWprogramming: I think I might've mixed up parameter names and parameter values when I wrote that comment.

For parameter names, it's enforced in helpers.py in `safe_variable_string`

README.md
```diff
@@ -1638,6 +1639,24 @@ print(schema_graph.get_unique_indexes_for_class('Animal'))
 
 For more information about the `SchemaGraph`, please see the class documentation. 
 
+
+### Cypher query parameters
+Redisgraph doesn't support query parameters, so we perform manual parameter interpolation in the
+graphql_to_redisgraph_cypher function. However, for Neo4j, we can use Neo4j's client to do
```
pmantica1: micro-nit: `graphql_to_redisgraph_cypher` (this way it appears highlighted as code in markdown)

README.md
```diff
@@ -1638,6 +1639,24 @@ print(schema_graph.get_unique_indexes_for_class('Animal'))
 
 For more information about the `SchemaGraph`, please see the class documentation. 
 
+
+### Cypher query parameters
+Redisgraph doesn't support query parameters, so we perform manual parameter interpolation in the
+graphql_to_redisgraph_cypher function. However, for Neo4j, we can use Neo4j's client to do 
+parameter interpolation on its own so that we don't reinvent the wheel.
+
+The function insert_arguments_into_query does so based on the query language, which isn't
```
pmantica1: `insert_arguments_into_query`

README.md
obi1kenobi: ^ What I think @pmantica1 is referring to is that you should use backticks ("`" characters) to wrap things that should appear in fixed-width font. This is the difference between graphql_to_redisgraph_cypher and `graphql_to_redisgraph_cypher`.

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,154 @@
...
+    # Unlike with Gremlin, unescaped dollar signs $ are not a problem when contained in a
+    # string literal in Cypher because they do not allow for arbitrary code execution.
+    escaped_and_quoted = json.dumps(argument_value)
+    return escaped_and_quoted
+
+
+def _safe_cypher_decimal(argument_value):
+    """Cypher doesn't support decimals, only ints and floats, so we'll raise an error here."""
+    raise NotImplementedError(u'Cypher doesn\'t support Decimals, only ints and floats. See '
+                              u'OpenCypher documentation. Argument value was {}'.
```
obi1kenobi: nit: I'd move the `.` to the next line, otherwise the syntax highlighting gets a bit confused and thinks that `format()` is a top-level function rather than a method on the string itself.

graphql_compiler/query_formatting/gremlin_formatting.py
```diff
@@ -147,7 +147,8 @@ def insert_arguments_into_gremlin_query(compilation_result, arguments):
 
     Args:
         compilation_result: a CompilationResult object derived from the GraphQL compiler
-        arguments: dict, mapping argument name to its value, for every parameter the query expects.
+        arguments: dict, str -> any, mapping argument name to its value, for every parameter the
+                    query expects.
```
obi1kenobi: `query` is off by one indentation space relative to `dict`, let's align them

graphql_compiler/query_formatting/match_formatting.py
```diff
@@ -122,7 +122,8 @@ def insert_arguments_into_match_query(compilation_result, arguments):
 
     Args:
         compilation_result: a CompilationResult object derived from the GraphQL compiler
-        arguments: dict, mapping argument name to its value, for every parameter the query expects.
+        arguments: dict, str -> any, mapping argument name to its value, for every parameter the
+                    query expects.
```
obi1kenobi: same indentation problem here

graphql_compiler/__init__.py
```diff
@@ -202,11 +202,15 @@ def get_graphql_schema_from_orientdb_schema_data(schema_data, class_to_field_typ
 
 
 def graphql_to_redisgraph_cypher(schema, graphql_query, parameters, type_equivalence_hints=None):
-    """Compile the GraphQL input using the schema into a Cypher query and associated metadata.
+    """Compile the GraphQL input using the schema into a Cypher query (for use specifically with
```
obi1kenobi: Unfortunately, the formatting rules about docstrings are pretty rigid on the headline always being a one-liner, so for the sake of not breaking with convention and tooling, let's try to play along.

How about: `Compile the GraphQL input into a RedisGraph Cypher query and associated metadata.` for the headline?

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -109,6 +109,14 @@ def _emit_with_clause_components(cypher_steps):
 
 def emit_code_from_ir(cypher_query, compiler_metadata):
     """Return a Cypher query string from a CypherQuery object."""
+    # According to the Cypher Query Language Reference [0], the standard Cypher version is Cypher
```
obi1kenobi: For clarity, don't break the line between `Cypher` and `9`.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -109,6 +109,14 @@ def _emit_with_clause_components(cypher_steps):
...
+    # 9 (page 196) and we should be able to specify the Cypher version in the query.
+    # Unfortunately, this turns out to be invalid in both Neo4j and RedisGraph-- Neo4j supports
+    # Cypher version 2.3, 3.4, and 3.5 [1] while Redisgraph doesn't support the syntax at all [2].
+    # When this does eventually get resolved, we can change `query_data` back to `[u'CYPHER 9']`
```
obi1kenobi: Nice writeup!

graphql_compiler/__init__.py
LWprogramming: Might be good to specify explicitly that it's only for RedisGraph?

`Compile the GraphQL input into a Cypher query (for RedisGraph only) and its metadata.` fits the line length rules. How's that sound? If it looks good, I'll update the other functions' headlines so they have a similar structure.

pmantica1: Can you make this headline shorter?

LWprogramming: `Compile the GraphQL input into a Cypher query (for RedisGraph only) and its metadata.` does end up being one line.

graphql_compiler/query_formatting/cypher_formatting.py
```diff
@@ -0,0 +1,154 @@
...
+# Public API
+######
+
+
+def insert_arguments_into_cypher_query_redisgraph(compilation_result, arguments):
+    """Insert the arguments into the compiled Cypher query to form a complete query.
+
+    This is only for Redisgraph because Neo4j's client can do this on its own.
+    Redisgraph doesn't support parameterized queries (see here:
+    https://github.com/RedisGraph/RedisGraph/issues/544#issuecomment-507963576).
```
obi1kenobi: Both `)` and `.` are valid URL arguments, so technically this URL is invalid. Consider restructuring the comment to avoid punctuation around the URL?

LWprogramming: fixed

PR #382, opened 2019-07-01
Refactoring changes to cost-estimation

ReaLNeroM:
- Change arguments of edge_counts_between_vertex_pairs.
- Make all statistics besides get_class_count optional, and reflect that in the test data.

### DIFF
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -60,50 +60,10 @@ class Statistics(object):
         """
         return None
 
-    @abstractmethod
-    def get_distinct_field_values_count(self, vertex_name, field_name):
-        """Return the count of distinct values the given vertex field has over all vertex instances.
-
-        This statistic helps estimate the result size of @filter directives used to restrict
-        values using equality operators like '=', '!=', and 'in_collection'.
-
-        Args:
-            vertex_name: str, name of a vertex.
-            field_name: str, name of a vertex field.
-
-        Returns:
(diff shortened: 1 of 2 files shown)


graphql_compiler/cost_estimation/statistics.py
```diff
@@ -101,22 +101,29 @@ def get_histogram(self, vertex_name, field_name):
...
                 count of edge instances of given class connecting instances of two vertex classes.
-            count_of_distinct_values: dict, (str, str) -> int, mapping vertex class name and field
-                name on that vertex class to the count of distinct values of the field for that
-                vertex class.
-            histograms: dict, (str, str) -> list[tuple(float, float, int)], mapping vertex class
-                name and field name on that vertex class to histogram.
+            count_of_distinct_values: optional dict, (str, str) -> int, mapping vertex class name
+                and field name on that vertex class to the count of distinct values of the field for
+                that vertex class.
+            histograms: optional dict, (str, str) -> list[tuple(float, float, int)], mapping vertex
```
obi1kenobi: I'm confused about the histogram format. Why is each data point a tuple of floats and ints? Doesn't that implicitly assume that we are histogramming over a float-typed field?

Please explain the data format of the histogram. Alternatively, since we aren't using histograms yet, and we haven't implemented anything there, just delete all mentions of histograms in the API. When we actually need them and have the time to implement them, then we can have the design discussion around them.

ReaLNeroM: I did the latter.

PR #388, opened 2019-07-03
Fix GraphQLScalarType object generation from schema text

pmantica1:


obi1kenobi:
What is the MATCH output here? Let's add this input to the test data file and add a test case to `test_ir_generation.py` and `test_compiler.py`.

README.md
```diff
@@ -1290,7 +1290,20 @@ is expected to correspond to a SQLAlchemy table object of the same name, case in
 schema type this could look like:
 
 ```python
-from sqlalchemy import MetaData, Table, Column, String
+from graphql import parse
```
obi1kenobi: I don't understand what these documentation changes have to do with the bug this PR originally set out to fix.

graphql_compiler/schema.py
```diff
@@ -281,6 +281,13 @@ def _parse_datetime_value(value):
     parse_literal=_unused_function,  # We don't yet support parsing Decimal objects in literals.
 )
 
+CUSTOM_SCALAR_TYPES = (
+    GraphQLDecimal,
+    GraphQLString,
```
obi1kenobi: `String` as a custom type? Could this be a typo?

graphql_compiler/tests/test_helpers.py
```diff
@@ -275,6 +276,20 @@ def get_schema():
     """Get a schema object for testing."""
     ast = parse(SCHEMA_TEXT)
     schema = build_ast_schema(ast)
+    # The schema text contains no information about how to parse or serialize non-builtin scalar
+    # types so we add this information manually. For more info see the link below:
+    # pylint: disable=max-line-length, protected-access
+    # https://stackoverflow.com/questions/47824603/graphql-custom-scalar-definition-without-graphql-tools
```
obi1kenobi: I think the link is probably not necessary, the description is sufficient here. Let's try to avoid the pylint suppressions if we can -- I think you can get around the protected access one by looking up type objects properly.

graphql_compiler/schema.py
pmantica1: Yes it is.

PR #389, opened 2019-07-03
Add query printing

pmantica1:


graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -189,6 +189,28 @@ def test_has_substring_precedence(self, backend_name):
...
+    def test_filter_on_date(self, backend_name):
+        graphql_query = '''
+        {
+            Animal {
+                name @output(out_name: "animal_name")
+                birthday @filter(op_name: "=", value: ["$birthday"])
+            }
+        }
+        '''
+        import datetime
```
LWprogramming: Do we want to import datetime locally like this or at the top?

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -189,6 +189,28 @@ def test_has_substring_precedence(self, backend_name):
...
+        if backend_name == test_backend.ORIENTDB:
```
LWprogramming: Is this meant for all SQL-based backends? I ask because Cypher support is coming soon so there will be other backends e.g. `test_backend.NEO4J` and `test_backend.REDISGRAPH`.

If you want to just run this test for all SQL-based backends, let's change this to check if the backend is not in SQL backends instead. 

If you want to exclude ORIENTDB in particular, then no need to change this.

PR #390, opened 2019-07-03
Renamed schema

qqi0O0:
The renaming part of renaming-and-merging-schemas. 

Before renaming, first gather information about the schema -- the name of the query type (to know what are the root fields to rename), and the set of scalars (to know what types to not rename). The set of directives is also gathered, but this is not used until the merging part.
Renaming happens in two steps -- first, rename types, including normal types, interfaces, enums, unions; then, rename root fields. 
The visitor for renaming types includes a function for every node class. Most of them do nothing and may be removed if preferred, they're currently included so that I know I did not forget about a type. There some types (anything that raises an AssertionError) I haven't seen and am not familiar with, and I would appreciate advice about what to do with them. 

Tests are not included in this PR, but they exist.


graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+        RenamedSchema
+
+    Raises:
+        SchemaError if input schema_string does not represent a valid schema without extensions,
+            or if there are conflicts between the renamed types or root fields
+    """
+    # Check that the input string is a parseable and valid schema.
+    try:
+        ast = parse(schema_string)
+        build_ast_schema(ast)
```
pmantica1: Does `build_ast_schema` do anything? I think it's only a getter function and does not mutate the ast.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
...
+class GetSchemaDataVisitor(Visitor):
+    """Gather information about the schema to aid any transforms."""
+    def __init__(self):
+        self.query_type = None  # str, name of the query type, e.g. 'RootSchemaQuery'
+        self.scalars = set()  # Set[str], names of scalar types
+        self.directives = set()  # Set[str], names of directives
+        self.has_extension = False
+        self.schema_data = None
+
+    def enter_TypeExtensionDefinition(self, node, *args):
```
pmantica1: What is a TypeExtension? I have never seen it.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class SchemaError(Exception):
+    pass
+
+
+SchemaData = namedtuple('SchemaData', ['query_type', 'scalars', 'directives', 'has_extension'])
+
+
+class GetSchemaDataVisitor(Visitor):
```
pmantica1: All this information, but the `has_extension` part is easily available in the `GraphQLSchema` so I am not sure this class is necessary.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+        rename_func: callable, used to rename fields of the query type
+        query_type: string, name of the query type, e.g. 'RootSchemaQuery'
+
+    Returns:
+        Dict[str, str], the new root field name to original root field name map
+
+    Raises:
+        SchemaError if rename causes root field name conflicts
+    """
+    visitor = RenameRootFieldsVisitor(rename_func, query_type)
```
pmantica1: RenameRootFieldsVisitor takes in 3 arguments but this gives it two. Also, this seems very similar to the function above. I am confused as to why there are not 1 function.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+            'InlineFragment', 'FragmentDefinition'
+        })
+
+    def _rename_name_add_to_record(self, node):
+        """Rename the value of the node, and add the name mapping to reverse_name_map.
+
+        Don't rename if the type is the query type, a scalar type, or a builtin type.
+
+        Args:
+            node: Name type Node
```
pmantica1: `Name type Node`?

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+
+        Raises:
+            SchemaError if the newly renamed node causes name conflicts with existing types,
+                scalars, or builtin types
```
pmantica1: nit: remove extra indentation?

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+    """Traverse a Document AST, editing the names of nodes."""
+    def __init__(self, rename_func, query_type, scalar_types):
+        self.rename_func = rename_func  # callable that takes string to string
+        self.reverse_name_map = {}  # Dict[str, str], from new name to original name
+        self.query_type = query_type  # str
+        self.scalar_types = frozenset(scalar_types)
+        self.builtin_types = frozenset({'String', 'Int', 'Float', 'Boolean', 'ID'})
+
+        # NOTE: behavior of Directive type is not completely clear
+        self.noop_types = frozenset({
```
pmantica1: Please create a test, (not in this pr), where you assert that the three sets are disjoint.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
...
+        self.schema_data = None
+
+    def enter_TypeExtensionDefinition(self, node, *args):
+        self.has_extension = True
+
+    def enter_OperationTypeDefinition(self, node, *args):
+        if node.operation == 'query':
+            self.query_type = node.type.name.value
+
+    def enter_ScalarTypeDefinition(self, node, *args):
```
pmantica1: Please add docstrings to all functions.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+        elif node_type in self.rename_types:
+            # Rename and put into record the name attribute of current node; continue traversal
+            self._rename_name_add_to_record(node.name)
+        elif node_type in self.unexpected_types:
+            # Node type unexpected in schema definition, raise error
+            raise SchemaError('Node type "{}" unexpected in schema AST'.format(node_type))
+        elif node_type == 'TypeExtensionDefinition':
+            raise SchemaError('Extension definition not allowed')
+        else:
+            # VariableDefinition, Variable, ObjectValue, ObjectField, InputObjectTypeDefinition
```
pmantica1: We don't use `InputObjectType` objects in the schema so we can ignore that. I think that`Variable` and `VariableDefinition` are only relevant for queries since I've never seen the `Variable` in schema generation. Not sure about `ObjectValue` but based on the name `ObjectField` might be referring to the field of `GraphQLObjectType` objects so I would double-check that it isn't important.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: I put it here as a way to check the validity of the input string -- if build_ast_schema doesn't raise errors, that says a lot about the legality of the structure of the input. It definitely feels inelegant and I should put a comment explaining the purpose of the line if nothing else. I'm not sure if there's a better way to check if the input is a good ast though.

graphql_compiler/schema_merging/utils.py
qqi0O0: Something like `extend RootSchemaQuery { something: something }`, some of these exist in the extension schema in Predrag's demo. They're not expected to exist in normal schemas I don't think.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: The input node is of class `Name` that inherits from class `Node`, maybe I should just say that the type is Name?

graphql_compiler/schema_merging/utils.py
qqi0O0: That's a good point, I'll use the information in GraphQLSchema for most of that. For extensions, I'll no longer check for them beforehand and raise an error if I run into one while traversing the ast.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: The first function renames types and the second renames root fields, which feel pretty separate so I put them into two functions. The one for renaming root fields doesn't need to know about the scalar types, RenameRootFieldsVisitor takes in 2 arguments (RenameSchemaVisitor, which I will rename to something better, takes in 3 arguments). I can get rid of both of these functions and combine them into `rename_schema` if that's better?

graphql_compiler/schema_merging/utils.py
qqi0O0: One downside of doing this though is that I need to go through transforming the ast into a GraphQLSchema, instead of directly working with the ast. It's not too much of a downside right here since I do build_ast_schema earlier anyway, but later in the merging step, I use the visitor to get schema data a bunch without ever turning the ast into a GraphQLSchema. So it might be useful to keep this.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: So `InputObjectType` can exist in schemas, but we don't use it/allow it in inputs? And `Variable` and `VariableDefinition` cannot exist in schemas? I'll try to figure out use of ObjectField and ObjectValue more.

graphql_compiler/schema_merging/rename_schema.py
obi1kenobi: With such a generic name, a somewhat longer explanation might be useful. The objective here is to be informative and useful even to an unfamiliar, casual reader of the code. To such a reader, `Name` type is not very informative, and `inherits from class Node` doesn't provide meaningful new information either because they are probably not familiar with the `Node` class.

In such a situation, explain what the significance of the object is, and how it is being operated on in this function. For example: `Name AST object, describing the name of an AST element such as a type, interface, or scalar.` plus some description of what will happen to it in this function.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+RenamedSchema = namedtuple(
+    'RenamedSchema', [
+        'schema_ast',  # type: Document, ast representing the renamed schema
+        'reverse_name_map',  # type: Dict[str, str], new type name to old type name
+        'reverse_root_field_map'  # type: Dict[str, str], new root field name to old root field
+    ]
+)
+
+
+def rename_schema(schema_string, rename_func=lambda name: name):
```
obi1kenobi: What's the advantage of taking a lambda over a dict here?

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+from graphql import build_ast_schema, parse
+from graphql.language.visitor import Visitor, visit
+
+from .utils import SchemaError, get_schema_data
+
+
+RenamedSchema = namedtuple(
+    'RenamedSchema', [
+        'schema_ast',  # type: Document, ast representing the renamed schema
+        'reverse_name_map',  # type: Dict[str, str], new type name to old type name
```
obi1kenobi: Nit: Rather than `new` and `old`, thoughts on using `renamed` and `original`? Being explicit and using precise terminology reduces the chance of misunderstandings and misuse.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+        'reverse_root_field_map'  # type: Dict[str, str], new root field name to old root field
```
obi1kenobi: Nit: What is a `root field`? I don't believe this is defined anywhere.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+def rename_schema(schema_string, rename_func=lambda name: name):
+    """Create a RenamedSchema, where types and query root fields are renamed using rename_func.
+
+    Args:
+        schema_string: string describing a valid schema that does not contain extensions
+        rename_func: callable that takes string to string, used to transform the names of
+                     types, interfaces, enums, and root fields. Defaults to identity
+
+    Returns:
+        RenamedSchema
```
obi1kenobi: Nit: I would go a bit more in-depth here, and explain what this object lets the user do.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+from graphql import build_ast_schema, parse
+from graphql.language.visitor import Visitor, visit
+
+from .utils import SchemaError, get_schema_data
+
+
+RenamedSchema = namedtuple(
```
obi1kenobi: I am a bit concerned about user confusion between this type and the built-in `GraphQLSchema` type. It would not be unreasonable for an unfamiliar user to assume that, since both objects claim to be schemas, they would be freely interchangeable and mutually compatible. This is of course not the case, so we risk having bugs here down the line.

I unfortunately don't have a concrete suggestion for a better name, but I'm happy to sit down and brainstorm one together if you'd like.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+    Returns:
+        RenamedSchema
+
+    Raises:
+        SchemaError if input schema_string does not represent a valid schema without extensions,
+            or if there are conflicts between the renamed types or root fields
+    """
+    # Check that the input string is a parseable and valid schema.
+    try:
+        ast = parse(schema_string)
```
obi1kenobi: Thoughts on accepting an AST directly (i.e. making the caller parse the string) instead here? Curious about what we might stand to gain and lose by switching.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+        build_ast_schema(ast)
+    except Exception as e:  # Can't be more specific -- see graphql/utils/build_ast_schema.py
+        raise SchemaError('Input schema does not define a valid schema.\n'
```
obi1kenobi: Two minor nits: Let's use an `u'foo'` explicit unicode string instead of `'foo'`, for the sake of Python 2. And let's avoid newlines in error messages, since sometimes they fail to get correctly represented through logs and other mechanisms.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def leave_Document(self, node, *args):
+        self.schema_data = SchemaData(query_type=self.query_type,
+                                      scalars=self.scalars,
+                                      directives=self.directives,
+                                      has_extension=self.has_extension)
+
+
+def get_schema_data(ast):
+    """Get schema data of input ast.
```
obi1kenobi: Let's put some more documentation in here. The generic `schema data` term doesn't really communicate what data is being collected. Furthermore, the function name, docstring headline, and return type description all just say `schema data` without telling the reader important things like:
- When would they want to call this function?
- What data does the function collect and return?
- In what situation might that data be useful?

In principle, I could replace `schema data` with `widget` everywhere here, and not gain or lose any informative content.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Args:
+        ast: Document
+
+    Return:
```
obi1kenobi: Nit: Per Google docstring style, `Returns`, not `Return`

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+from graphql.language.visitor import Visitor, visit
+
+
+class SchemaError(Exception):
+    pass
```
obi1kenobi: Nit: It would be preferable to document in a docstring line when this error could be raised. Consider the programmer who is presented with such an error. After reading the error message, and possibly the code surrounding the raising code, the next place they'd look would be the error type definition. They would greatly be aided if the docstring here told them things like what circumstances could produce this error.

If you choose to add a docstring here, the `pass` will become unnecessary and can be removed, since the docstring will provide the necessary "indented statement" that the Python parser requires.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: From what I can tell, it seems like `ObjectValue` and `ObjectField` only appear in things like query results. Schemas just have `FieldDefinition` and `ObjectTypeDefinition`. For now I'll categorize those as types that are unexpected in schemas. If this turns out to be wrong, we can fix it.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+SchemaData = namedtuple('SchemaData', ['query_type', 'scalars', 'directives', 'has_extension'])
```
obi1kenobi: Super micro nit: it's nice to get in the habit of using immutable types whenever mutability is not necessary. For example, that would mean using a tuple instead of a list here. In this case, it doesn't really matter too much one way or the other, so this code is totally fine. I also noticed that you used frozensets elsewhere in the code, so I know you already do this often. I encourage you to continue that, even if both sets of code "work fine", especially when there is no cognitive or visual overhead to one approach over the other -- in this case, it's just a question of `(` vs `[`.

graphql_compiler/schema_merging/utils.py
obi1kenobi: Once we've already called `build_ast_schema()` for its validation logic, we might as well keep the result and use it to simplify the rest of our code. I also think your idea of simply raising errors on extension types if/when they are encountered is a good one.

Overall, the simpler and more understandable we can make the code, the easier it will be to test and maintain in the long run. This is code that will be constantly called, in production, for many years to come. So let's do our future selves a favor 😃

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+        self.rename_types = frozenset({
+            'NamedType', 'ObjectTypeDefinition', 'InterfaceTypeDefinition',
+            'UnionTypeDefinition', 'EnumTypeDefinition'
+        })
+        self.unexpected_types = frozenset({
+            'OperationDefinition', 'SelectionSet', 'Field', 'FragmentSpread',
+            'InlineFragment', 'FragmentDefinition'
+        })
+
+    def _rename_name_add_to_record(self, node):
```
obi1kenobi: It would not be unreasonable for a reader who is otherwise unfamiliar with the code to read the function's name and docstring and on the basis of that assume that this function is intended to return something. Finding precise, unambiguous, and non-confusing names for things is one of the most difficult aspects of system design, and yet it's what differentiates complex systems that work and complex systems that don't work. Let's work together to find a better name for this function?

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+            'FieldDefinition', 'InputValueDefinition', 'EnumValueDefinition',
+            'DirectiveDefinition'
+        })
+        self.rename_types = frozenset({
+            'NamedType', 'ObjectTypeDefinition', 'InterfaceTypeDefinition',
+            'UnionTypeDefinition', 'EnumTypeDefinition'
+        })
+        self.unexpected_types = frozenset({
+            'OperationDefinition', 'SelectionSet', 'Field', 'FragmentSpread',
+            'InlineFragment', 'FragmentDefinition'
```
obi1kenobi: Nit: Imagine needing to add one more entry to one of these frozensets. The produced diff will show "removed one row, added one row" and will likely not make it obvious which entry was added.

As an alternative, consider the following format:
```
self.unexpected_types = frozenset({
    'Field',
    'FragmentDefinition',
    'FragmentSpread',
    'InlineFragment',
    'OperationDefinition',
    'SelectionSet',
})
```

This gets us the nice property that any one addition or removal is a single-line operation that affects only the entry being added or removed. Note the trailing comma after the last entry, which makes sure this is the case even for new entries that are at the end of the sequence.

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,232 @@
...
+    return visitor.reverse_field_map
+
+
+class RenameSchemaVisitor(Visitor):
+    """Traverse a Document AST, editing the names of nodes."""
+    def __init__(self, rename_func, query_type, scalar_types):
+        self.rename_func = rename_func  # callable that takes string to string
+        self.reverse_name_map = {}  # Dict[str, str], from new name to original name
+        self.query_type = query_type  # str
+        self.scalar_types = frozenset(scalar_types)
```
obi1kenobi: I really like the frozensets throughout this code!

graphql_compiler/schema_merging/rename_schema.py
```diff
@@ -0,0 +1,238 @@
...
+)
+
+
+def rename_schema(schema_string, rename_func=lambda name: name):
+    """Create a RenamedSchema, where types and query root fields are renamed using rename_func.
+
+    Args:
+        schema_string: string describing a valid schema that does not contain extensions or
+                       input object definitions
+        rename_func: callable that takes string to string, used to transform the names of
```
obi1kenobi: Not sure what value we get from setting a default value here. Isn't the identity renaming equivalent to just not calling this function at all?

graphql_compiler/schema_merging/rename_schema.py
pmantica1: From what I can tell, ignoring these 5 types for now seems to be the right thing to do and if it isn't we can fix it later. Furthermore, if it isn't the right to do, then we should be easily be able to tell from the tests.

graphql_compiler/schema_merging/utils.py
pmantica1: Ok then either option makes sense to me.

graphql_compiler/schema_merging/rename_schema.py
pmantica1: Nevermind I misread this.

graphql_compiler/schema_merging/utils.py
pmantica1: Also consider changing this to `something a bit more specific like `SchemaMergingError` if possible.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: Good point, fixed

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: By "root field" I meant "fields of the root type", aka fields of the query type. Where would be a good place to put this explanation? Here at the definition of RenamedSchema?

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: It felt to me that using a function instead of a dict to transform names is more convenient and clear in some ways, but I'm not completely sure which is better. If we use a dict, then we could take one of two approaches. 
One, the dict is required to contain all type names and field names. This would require the user to get all types of a schema, which can be inconvenient.
Two, the dict contains some subset of types and fields, and all other types and fields pass through without change. I feel like this approach could lead to errors, but I'm not completely sure.

In particular, I imagined this being used something like a namespace, by prefixing a string to all names of a schema. In that case, a func seems much more convenient than a dict.
If the user only wants to rename a few conflicting names, they can still define their rename_func to be something like: if input string is a key in this dict, return the value associated with that key, otherwise return the original string.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: Calling it with the identity function would indeed not modify the schema. Nonetheless it would create a RenamedSchema with the right type and field maps, that allows the schema to be later merged with other schemas. 

I'm not totally sure if this design is good -- a function will take in a OrderedDict mapping schema name to RenamedSchema, and output a MergedSchema.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: If the function takes in an AST directly, I'm still going to make a copy of it before doing anything else since in the next steps I modify the AST. I like the idea of passing in a string also because the string is immutable.

graphql_compiler/schema_merging/rename_schema.py
obi1kenobi: What do you think about prohibiting renaming of the query type altogether? I don't really see a use case that would require it, and I don't think it should matter what it is named so long as it doesn't conflict with any other type name. In that case, we won't ever need this value, since we can just look it up from the schema object itself.

The more functionality we can cut out without losing the core value proposition, the faster and better we'll be able to implement the rest.

graphql_compiler/schema_merging/rename_schema.py
obi1kenobi: Ah now I fully understand the conversation we had in person. I would not make the merging step be aware of the fact that it operates on renamed schemas. I think from its perspective, it should just be merging arbitrary schemas, and if they happen to be renamed, that would be handled by a separate step.

In other words, I think there's a primitive "merge schemas" function that is oblivious of the need for renaming. This function may then be used in a context where we know renaming happens. I think this approach is more easily understandable and testable.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: The description is definitely not good here -- I meant it as in, for example, fields of `RootSchemaQuery`. So if we have `RootSchemaQuery { Human: Human }`, this part renames the first `Human`

obi1kenobi: Ah I see, I definitely got confused. How about prohibiting divergences there as well, and enforcing that in the query type, the fields are named the same as the types they hold? Then I think we'd just be able to use the `reverse_name_map` directly.

Also, it might be good to mention in the comment next to `reverse_name_map` whether all types (even non-renamed ones) have entries, or types that aren't there are not renamed. I'd probably opt for the second, since we have some very large schemas.

qqi0O0: I was hesitant to enforce that the fields are named the same as the types, because GraphQL doesn't require this (and the basic GraphQL tutorial examples has query fields like `hero: Human`). Is this restriction enforced somewhere in graphql compiler?

obi1kenobi: I definitely understand the hesitation, it's very reasonable. We don't currently enforce this at the moment, because the compiler originally did not concern itself with how the schema was built, and simply assumed that it would be given a reasonable schema to work with. However, I wouldn't be surprised if having field names that don't match the type name on the query type of the schema causes some unexpected bugs to pop up in the compiler. So I think we should start enforcing it, first here and then probably in the rest of the compiler codebase as well.

graphql_compiler/schema_merging/utils.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from graphql.type.definition import GraphQLScalarType
+
+
+class SchemaError(Exception):
```
pmantica1: There is actually already a `SchemaError` defined in the compiler. It is in `schema_generation/exceptions.py`.

qqi0O0: Oh interesting, would it make sense to inherit from that `SchemaError` and make my own more specific errors?

qqi0O0: Or should I rename my base error to something else?

obi1kenobi: Nice catch, @pmantica1! Let's figure out the circumstances in which the built-in `SchemaError` is raised, and if they in spirit match the kinds of problems we may encounter, then subclass from it.

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: > Also, it might be good to mention in the comment next to `reverse_name_map` whether all types (even non-renamed ones) have entries, or types that aren't there are not renamed. I'd probably opt for the second, since we have some very large schemas.

In order to check for conflicts, I need to keep track of all names, even ones that are not changed, during the renaming step. At the end of that I'm going to have a map with all types, including non-renamed ones. Would it be better to keep that full dict and put it into the output? Or process it afterwards by removing all entries that didn't get renamed?

graphql_compiler/schema_merging/rename_schema.py
qqi0O0: Related question, should the output RenamedSchema contain the schema AST, or the schema string? It's easy to go back and forth between them with `parse` and `print_ast`, and there are benefits and downsides either way. Dealing with strings is nice since strings are immutable. Dealing with ASTs is nice since manipulations happen on ASTs, and not converting to strings in between means there's less waste converting back and forth.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -0,0 +1,135 @@
...
+    Includes all user defined scalars, as well as any builtin scalars used in the schema; excludes
+    builtin scalars not used in the schema.
+
+    Returns:
+        Set[str], set of names of scalars used in the schema
+    """
+    type_map = schema.get_type_map()
+    scalars = {
+        type_name
+        for type_name in type_map
```
obi1kenobi: micro nit: if you make this `for type_name, type_object in six.iteritems(type_map)` then you can avoid the `type_map[type_name]` lookup on the line below.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -0,0 +1,135 @@
...
+            field_name = node.name.value
+            type_node = node.type
+            # NamedType node may be wrapped in several layers of NonNullType or ListType
+            while not isinstance(type_node, NamedType):
+                type_node = type_node.type
+            queried_type_name = type_node.name.value
+            if field_name != queried_type_name:
+                raise SchemaStructureError(
+                    u'Query type\'s field name "{}" does not match corresponding queried type '
+                    'name "{}"'.format(field_name, queried_type_name)
```
obi1kenobi: another micro nit: you can add a `u` prefix before the single-quote here, and make the lines line up

graphql_compiler/schema_transformation/utils.py
```diff
@@ -0,0 +1,135 @@
...
+        schema: GraphQLSchema, representing the same schema as ast
+
+    Raises:
+        SchemaStructureError if the schema contains mutations, contains subscriptions, or some
+        query type field name does not match the type it queries.
+    """
+    if schema.get_mutation_type() is not None:
+        raise SchemaStructureError(u'Schema contains mutations.')
+    if schema.get_subscription_type() is not None:
+        raise SchemaStructureError(u'Schema contains subscriptions.')
```
obi1kenobi: I would add to the error messages something like: `Renaming schemas that contain X is currently not supported.` for the appropriate values of `X`.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -17,10 +19,41 @@ class SchemaStructureError(SchemaTransformError):
...
+    Args:
+        name: str
+
+    Raises:
+        InvalidNameError if the name doesn't consist of only alphanumeric characters and
+        underscores, starts with a numeric character, or starts with double underscores
+    """
+    if not isinstance(name, str):
+        raise InvalidNameError(u'Name "{}" is not a string.'.format(name))
+    pattern = re.compile(r'^[_a-zA-Z][_a-zA-Z0-9]*$')
```
obi1kenobi: Compiling a regex pattern is comparatively expensive, let's move this to the module-level scope so we only compile it once. We can make it have an underscored name as well, so it's clear that it is not to be imported.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -17,10 +19,41 @@ class SchemaStructureError(SchemaTransformError):
...
+    required by GraphQL), and doesn't start with double underscores as such types are reserved
+    for GraphQL internal use.
+    """
+
+
 class SchemaNameConflictError(SchemaTransformError):
     """Raised when renaming types or fields cause name conflicts."""
 
 
+def check_name_is_valid(name):
```
obi1kenobi: nit: `check_type_name_is_valid`, to be extra clear what kind of name we're talking about?

graphql_compiler/schema_transformation/utils.py
```diff
@@ -17,10 +19,41 @@ class SchemaStructureError(SchemaTransformError):
     """
 
 
+class InvalidNameError(SchemaTransformError):
+    """Raised if a type/field name is not valid.
+
+    This may be raised if the input schema contains invalid names, or if the user attempts to
+    rename a type/field to an invalid name. A name is considered valid if it consists of
+    alphanumeric characters and underscores and doesn't start with a numeric character (as
+    required by GraphQL), and doesn't start with double underscores as such types are reserved
```
obi1kenobi: micro nit: `such type names are reserved`

PR #391, opened 2019-07-10
Merge schemas

qqi0O0:
Function takes in an ordered dict of schema name to schema ast. 

On a high level, the function takes the list of definitions in each ast and appends it to the list of definitions in the merged ast, and similarly takes the list of root fields in each ast and appends it to the list of root fields in the merged ast. 

Some definitions, in particular the SchemaDefinition ("schema { query: RootSchemaQuery }") and the query type definition ("RootSchemaQuery { Human: Human }"), are skipped when appending definitions. Scalar definitions and directive definitions receive special treatment.

Before adding appending any definition, we check for name conflicts.

Tests are again not included in this PR, but tests exist.

### DIFF
--- a/graphql_compiler/schema_transformation/merge_schemas.py
+++ b/graphql_compiler/schema_transformation/merge_schemas.py
@@ -67,9 +67,9 @@ def merge_schemas(schemas_dict):
     schemas.
 
     Args:
-        schemas_dict: OrderedDict where keys are schema_identifiers, and values are
-                      ASTs (Documents) describing schemas. The ASTs will not be modified by
-                      this function
+        schemas_dict: OrderedDict[str, Document], where keys are names/identifiers of schemas,
+                      and values are ASTs describing schemas. The ASTs will not be modified
+                      by this funcion
 
     Returns:
         MergedSchemaDescriptor, a namedtuple that contains the AST of the merged schema,
@@ -77,11 +77,11 @@ def merge_schemas(schemas_dict):
         came from
 
     Raises:
-        SchemaStructureError if the schema does not have the expected form; in particular, if
+        - SchemaStructureError if the schema does not have the expected form; in particular, if
         the AST does not represent a valid schema, if any query type field does not have the
         same name as the type that it queries, if the schema contains type extensions or
         input object definitions, or if the schema contains mutations or subscriptions
-        SchemaNameConflictError if there are conflicts between the names of
+        - SchemaNameConflictError if there are conflicts between the names of
         types/interfaces/enums/scalars, or conflicts between the definition of directives
         with the same name
     """
@@ -90,11 +90,9 @@ def merge_schemas(schemas_dict):
 
     query_type = 'RootSchemaQuery'
     # NOTE: currently, the query type will always be named RootSchemaQuery
-    # could be changed so the user has an input, or by changed to always use the query type
-    # name in the first schema in the input, if desired
+    # This could be changed so the user specifies the name of the query type, or be changed
+    # to always use the query type name in the first input schema
     merged_schema_ast = basic_schema_ast(query_type)  # Document
-    merged_definitions = merged_schema_ast.definitions  # List[Node]
-    merged_query_type_fields = merged_definitions[1].fields  # List[FieldDefinition]
 
     name_id_map = {}  # Dict[str, str], name of type/interface/enum/union to schema id
     scalars = {'String', 'Int', 'Float', 'Boolean', 'ID'}  # Set[str], user defined + builtins
@@ -114,7 +112,7 @@ def merge_schemas(schemas_dict):
         cur_query_type = get_query_type_name(cur_schema)
 
         # Merge cur_ast into merged_schema_ast
-        # Concatenate new scalars, new directive, and all type definitions to merged_definitions
+        # Concatenate new scalars, new directive, and all type definitions to definitions list
         # Raise errors for conflicting scalars, directives, or types
         new_definitions = cur_ast.definitions  # List[Node]
         new_query_type_fields = None  # List[FieldDefinition]
@@ -141,7 +139,7 @@ def merge_schemas(schemas_dict):
                                                        print_ast(directives[new_name]))
                         )
                 # new directive
-                merged_definitions.append(new_definition)  # Add to AST
+                merged_schema_ast.definitions.append(new_definition)  # Add to AST
                 directives[new_name] = new_definition
 
             elif isinstance(new_definition, ast_types.ScalarTypeDefinition):
@@ -153,7 +151,7 @@ def merge_schemas(schemas_dict):
                         u'New scalar "{}" clashes with existing type.'.format(new_name)
                     )
                 # new, valid scalar
-                merged_definitions.append(new_definition)  # Add to AST
+                merged_schema_ast.definitions.append(new_definition)  # Add to AST
                 scalars.add(new_name)
 
             else:  # Generic type definition
@@ -166,7 +164,7 @@ def merge_schemas(schemas_dict):
                     raise SchemaNameConflictError(
                         u'New type "{}" clashes with existing type.'.format(new_name)
                     )
-                merged_definitions.append(new_definition)
+                merged_schema_ast.definitions.append(new_definition)
                 name_id_map[new_name] = cur_schema_id
 
         # Concatenate all query type fields
@@ -178,6 +176,7 @@ def merge_schemas(schemas_dict):
         # Note that as field names and type names have been confirmed to match up, and types
         # were merged without invalid names or name conflicts, query type fields can also be
         # merged without errors.
-        merged_query_type_fields.extend(new_query_type_fields)
+        query_type_index = 1  # Query type is the second entry in the list of definitions
+        merged_schema_ast.definitions[query_type_index].fields.extend(new_query_type_fields)
 
     return MergedSchemaDescriptor(schema_ast=merged_schema_ast, name_id_map=name_id_map)
--- a/graphql_compiler/schema_transformation/merge_schemas.py
+++ b/graphql_compiler/schema_transformation/merge_schemas.py
@@ -8,8 +8,8 @@ from graphql.language.printer import print_ast
 import six
 
 from .utils import (
-    InvalidTypeNameError, SchemaNameConflictError, SchemaStructureError,
-    check_ast_schema_is_valid, check_type_name_is_valid, get_query_type_name
+    SchemaNameConflictError, SchemaStructureError, check_ast_schema_is_valid,
+    check_type_name_is_valid, get_query_type_name
 )
 
 
--- a/graphql_compiler/schema_transformation/merge_schemas.py
+++ b/graphql_compiler/schema_transformation/merge_schemas.py
@@ -8,7 +8,8 @@ from graphql.language.printer import print_ast
 import six
 
 from .utils import (
-    SchemaNameConflictError, SchemaStructureError, check_ast_schema_is_valid, get_query_type_name
+    InvalidTypeNameError, SchemaNameConflictError, SchemaStructureError,
+    check_ast_schema_is_valid, check_type_name_is_valid, get_query_type_name
 )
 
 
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,183 @@
...
+
+MergedSchemaDescriptor = namedtuple(
+    'MergedSchemaDescriptor', (
+        'schema_ast',  # Document, AST representing the merged schema
+        'name_id_map',  # Dict[str, str], type name to id of the schema the type is from
+    )
+)
+
+
+def basic_schema_ast(query_type):
```
pmantica1: I think it's best if we remove this function and just put the code inside `merge_schemas`. I don't think any other function will use it and it requires a lot of specific documentation.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,183 @@
...
+        with the same name
+    """
+    if len(schemas_dict) == 0:
+        raise ValueError(u'Expected a nonzero number of schemas to merge.')
+
+    query_type = 'RootSchemaQuery'
+    # NOTE: currently, the query type will always be named RootSchemaQuery
+    # could be changed so the user has an input, or by changed to always use the query type
+    # name in the first schema in the input, if desired
+    merged_schema_ast = basic_schema_ast(query_type)  # Document
```
pmantica1: Not a huge fan of having three variables all point to parts of the same underlying python object.

pmantica1: I think that this makes the mutations to the AST a bit too implicit.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,183 @@
...
+    merged_definitions = merged_schema_ast.definitions  # List[Node]
+    merged_query_type_fields = merged_definitions[1].fields  # List[FieldDefinition]
```
pmantica1: Please use a variable instead of a magic constant: `query_type_index` instead of 1.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,183 @@
...
+        SchemaNameConflictError if there are conflicts between the names of
+        types/interfaces/enums/scalars, or conflicts between the definition of directives
+        with the same name
+    """
+    if len(schemas_dict) == 0:
+        raise ValueError(u'Expected a nonzero number of schemas to merge.')
+
+    query_type = 'RootSchemaQuery'
+    # NOTE: currently, the query type will always be named RootSchemaQuery
+    # could be changed so the user has an input, or by changed to always use the query type
```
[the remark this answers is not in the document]
qqi0O0: I meant those comments to be for a point of discussion (not comments that will remain in the final iteration of this code) -- right now, the query type will be `RootSchemaQuery` and that will contain all fields of query types in individual input schemas. I'm not sure if it's a useful feature for the user to be able to specify what the query type should be named to, or alternatively if I should take the query type name of the very first input schema and use that for the merged schema.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: What if I put an underscore in front? I'm not totally sure about the balance between having long functions, and having helpers that are for very specific situations and are used only once.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,182 @@
...
+                      by this funcion
+
+    Returns:
+        MergedSchemaDescriptor, a namedtuple that contains the AST of the merged schema,
+        and the map from names of types/query type fields to the id of the schema that they
+        came from
+
+    Raises:
+        - SchemaStructureError if the schema does not have the expected form; in particular, if
+        the AST does not represent a valid schema, if any query type field does not have the
```
pmantica1: micro nit: mind aligning the lines:
- Schema...
  the AST 
Not:
- Schema
the AST

qqi0O0: Autoformatting seems to have eaten any differences between these two. I'm assuming I want to add two spaces to all the other lines?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,170 @@
...
+from graphql.language.printer import print_ast
+import six
+
+from .utils import SchemaNameConflictError, check_ast_schema_is_valid, get_query_type_name
+
+
+MergedSchemaDescriptor = namedtuple(
+    'MergedSchemaDescriptor', (
+        'schema_ast',  # Document, AST representing the merged schema
+        'name_id_map',  # Dict[str, str], type name to id of the schema the type is from
```
pmantica1: nit: How about `name_to_schema_id`?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,170 @@
...
+
+            if (
+                isinstance(new_definition, ast_types.ObjectTypeDefinition) and
+                new_name == cur_query_type
+            ):  # query type definition
+                new_query_type_fields = new_definition.fields  # List[FieldDefinition]
+
+            elif isinstance(new_definition, ast_types.DirectiveDefinition):
+                if new_name in directives:  # existing directive
+                    if new_definition == directives[new_name]:  # definitions agree
```
pmantica1: Are you sure this won't lead to an error? I think this is testing equality by reference so if you parse a directive from two different schema strings then you might get an error no?

qqi0O0: I’m pretty sure it checks equality by value, not reference. I tested out checking equality between two node objects with identical structures before, and there’s also a test for deduplicating directives. It’d definitely be nice to check again to be sure though, if you want to confirm.

pmantica1: Yeah sorry you are right. I was thinking about GraphQLDirective objects.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+
+MergedSchemaDescriptor = namedtuple(
+    'MergedSchemaDescriptor', (
+        'schema_ast',  # Document, AST representing the merged schema
+        'name_to_schema_id',  # Dict[str, str], type name to id of the schema the type is from
+    )
+)
+
+
+def _basic_schema_ast(query_type):
```
obi1kenobi: Generally, it's good practice to name functions verb-like names (e.g. `_get_basic_schema_ast()`) instead of noun-like names (`_basic_schema_ast()`) -- this tells the reader what the function does when called.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+                fields=[],
+                interfaces=[],
+                directives=[],
+            ),
+        ]
+    )
+    return blank_ast
+
+
+def merge_schemas(schemas_dict):
```
obi1kenobi: Thoughts on avoiding the reverse Hungarian naming of the input argument here?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+    """Check that input schemas do not contain conflicting definitions, then merge.
+
+    Merged schema will contain all type, interface, enum, scalar, and directive definitions
```
obi1kenobi: Assuming the merged schema will also contain all unions defined in the input schemas as well? Also a grammar nit -- maybe add `The` at the beginning of the sentence?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+    from input schemas. The fields of its query type will be the union of the fields of the
+    query types of each input schema.
+
+    Args:
+        schemas_dict: OrderedDict[str, Document], where keys are names/identifiers of schemas,
```
obi1kenobi: Do we want to impose a constraint on what the identifiers may be here?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+        raise ValueError(u'Expected a nonzero number of schemas to merge.')
+
+    query_type = 'RootSchemaQuery'
+    merged_schema_ast = _basic_schema_ast(query_type)  # Document
+
+    name_to_schema_id = {}  # Dict[str, str], name of type/interface/enum/union to schema id
+    scalars = {'String', 'Int', 'Float', 'Boolean', 'ID'}  # Set[str], user defined + builtins
+    directives = {}  # Dict[str, DirectiveDefinition]
+
+    for cur_schema_id, cur_ast in six.iteritems(schemas_dict):
```
obi1kenobi: Let's avoid abbreviations like `cur`. Optimize for readability rather than typing speed, since code is written once and read many, many times.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+        cur_query_type = get_query_type_name(cur_schema)
+
+        # Merge cur_ast into merged_schema_ast
+        # Concatenate new scalars, new directives, and type definitions other than the query
+        # type to definitions list
+        # Raise errors for conflicting scalars, directives, or types
+        new_definitions = cur_ast.definitions  # List[Node]
+        new_query_type_fields = None  # List[FieldDefinition]
+        for new_definition in new_definitions:
+            if isinstance(new_definition, ast_types.SchemaDefinition):
```
obi1kenobi: Is there anything we'd like to validate on this definition before skipping to the next one?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+                                 u'found.'.format(cur_query_type))
+        # Note that as field names and type names have been confirmed to match up, and types
+        # were merged without name conflicts, query type fields can also be safely merged
+        query_type_index = 1  # Query type is the second entry in the list of definitions
+        merged_schema_ast.definitions[query_type_index].fields.extend(new_query_type_fields)
+
+    return MergedSchemaDescriptor(
+        schema_ast=merged_schema_ast,
+        name_to_schema_id=name_to_schema_id
+    )
```
obi1kenobi: Would it be possible to refactor this function a bit? There's a lot of complexity here, and it's quite long. I'm finding it difficult to keep track of everything going on in here.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -0,0 +1,173 @@
...
+                    continue
+                if new_name in name_to_schema_id:  # new scalar clashing with existing type
+                    raise SchemaNameConflictError(
+                        u'New scalar "{}" clashes with existing type.'.format(new_name)
+                    )
+                # new, valid scalar
+                merged_schema_ast.definitions.append(new_definition)  # Add to AST
+                scalars.add(new_name)
+
+            else:  # Generic type definition
```
obi1kenobi: Can we turn this into an `elif`, so we can positively check that we are indeed at a generic type definition, and turn the `else` case into an assertion error about unreachable code?

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: What kind of constraint do you think would be reasonable? I don't currently check that the schema_id is actually a string and I can check that, but besides that, do you want to impose a restriction on the length or the structure of the schema identifiers?

obi1kenobi: I'd say ASCII alphanumeric and underscore characters, and prohibit empty strings. Otherwise, we can have all sorts of problems later -- imagine someone using any of the following bad schema identifiers, and the problems that can happen as a result:
```python
bad_identifiers = [
    '',  # empty string
    '\b',  # backspace char
    '\n',
    '<script>alert("hello world")</script>',
    # also various flavors of Unicode special characters etc.
]
```

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: The SchemaDefinition looks something like this:
`SchemaDefinition(operation_types=[OperationTypeDefinition(operation='query', type=NamedType(name=Name(value='SchemaQuery')))], directives=[])`
`check_ast_schema_is_valid` already confirms that the SchemaDefinition only contains a query operation. It doesn't feel like there's a lot else to check. I supposed I could maybe check that the list of directives is empty (perhaps I should put that check into `check_ast_schema_is_valid`)?

obi1kenobi: It's probably fine as it is, I just wanted to confirm with you since you are a lot more familiar with the AST structure than me.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: Regex behaves a little bit funny, so a string like `'hello\n'` is also matched by `r'^[_a-zA-z0-9]+$'`. Is that fine?

PR #393, opened 2019-07-12
Rename schema tests

qqi0O0:


graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -0,0 +1,721 @@
...
+            }
+
+            input MessageInput {
+              content: String
+            }
+        ''')
+        with self.assertRaises(SchemaStructureError):
+            rename_schema(parse(schema_string), {})
+
+    def test_mutation_definition(self):
```
pmantica1: What do you think of splitting up schema validation and renaming since the validation seems so comprehensive? @obi1kenobi

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -0,0 +1,721 @@
...
+
+            directive @stitch(source_field: String!, sink_field: String!) on FIELD_DEFINITION
+
+            type SchemaQuery {
+              NewHuman: NewHuman
+              NewGiraffe: NewGiraffe
+            }
+        ''')
+        self.assertEqual(renamed_schema_string, print_ast(renamed_schema.schema_ast))
+        self.assertEqual({'NewCharacter': 'Character', 'NewGiraffe': 'Giraffe',
```
obi1kenobi: nit: this would be a bit more readable if each dict entry were on its own line: ```
{
    foo: 1,
    bar: 2,
}
```

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
obi1kenobi: Would probably be a bit easier to work with that way, sure.

qqi0O0: So splitting up input schema validation tests and renaming tests into two files?

qqi0O0: Or splitting up the code?

PR #394, opened 2019-07-13
Cost estimation nonunique field filters

ReaLNeroM:
- Support for distinct_field_values_count, so we can now predict result costs even when filters are done over non-unique properties!
- Added tests for non-unique filters, and changed old tests' docstrings so that they explicitly state that `distinct_field_values_count` is not provided.
- @reviewers: I'm debating whether to add a `*args=None` argument to the LocalStatistics `__init__` method after class_counts, so that every statistic after class_counts has to be passed as a keyword argument.

### DIFF
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -47,24 +47,29 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
 def _estimate_filter_selectivity_of_equality(
     schema_graph, statistics, location_name, filter_fields
 ):
-    """Calculate the selectivity of equality filter(s) at a given location."""
-    result_selectivity = Selectivity(kind=FRACTIONAL_SELECTIVITY, value=1.0)
+    """Calculate the selectivity of equality filter(s) at a given location.
+
+    Since there may be multiple equality filters at a given location, this function
+    """
+    all_selectivities = []
 
     unique_indexes = schema_graph.get_unique_indexes_for_class(location_name)
     if _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
         # TODO(evan): don't return a higher absolute selectivity than class counts.
-        result_selectivity = Selectivity(kind=ABSOLUTE_SELECTIVITY, value=1.0)
-    else:
-        # Assumption: each distinct field value is equally common.
-        statistics_result = statistics.get_distinct_field_values_count(
-            location_name, filter_fields[0]
-        )
+        all_selectivities.append(Selectivity(kind=ABSOLUTE_SELECTIVITY, value=1.0))
+
+    for field_name in filter_fields:
+        statistics_result = statistics.get_distinct_field_values_count(location_name, field_name)
 
         if statistics_result is not None:
-            result_selectivity = Selectivity(
+            # Assumption: all distinct field values are distributed evenly among vertex instances,
+            # so each distinct value occurs
+            # (# of current location vertex instances) / (# of distinct field values) times.
+            all_selectivities.append(Selectivity(
                 kind=FRACTIONAL_SELECTIVITY, value=1.0 / statistics_result
-            )
+            ))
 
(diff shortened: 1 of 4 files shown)


graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -90,20 +99,17 @@ def _get_filter_selectivity(
...
                 kind=ABSOLUTE_SELECTIVITY,
                 value=float(collection_size) * selectivity_per_entry_in_collection.value
             )
         elif _is_fractional(selectivity_per_entry_in_collection):
             result_selectivity = Selectivity(
-                kind=selectivity_per_entry_in_collection.kind,
+                kind=FRACTIONAL_SELECTIVITY,
                 value=min(float(collection_size) * selectivity_per_entry_in_collection.value,
                           1.0)
-                # The estimate may be above 1.0 in case of duplicates in the collection
```
obi1kenobi: Why delete this comment but preserve the `min` logic?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -54,8 +54,17 @@ def _estimate_filter_selectivity_of_equality(
     if _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
         # TODO(evan): don't return a higher absolute selectivity than class counts.
         result_selectivity = Selectivity(kind=ABSOLUTE_SELECTIVITY, value=1.0)
+    else:
+        # Assumption: each distinct field value is equally common.
+        statistics_result = statistics.get_distinct_field_values_count(
+            location_name, filter_fields[0]
```
obi1kenobi: This seems problematic. For example, I can't be sure since there's only a one-line docstring on this function, but I don't think we have a guarantee of ordering on the filter fields. This means that two runs with the the same input and statistics might produce two different estimates.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -86,34 +114,52 @@ def get_class_count(self, class_name):
...
     ):
-        """Return the count of edges connecting two vertices.
+        """Return the count of edges of the given class connecting vertex_source to vertex_target.
 
-        This statistic is optional, as the estimator can roughly predict this cost using
+        This statistic is optional, as the estimator can roughly predict this statistic using
         get_class_count(). In some cases of traversal between two vertices using an edge connecting
-        the vertices' superclasses, the estimates generated using get_class_count() may be off by
-        several orders of magnitude. In such cases, this statistic should be provided.
+        the vertices' superclasses, the predicitons made using get_class_count() may be off
```
obi1kenobi: Why change the wording here?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: You're right, this is very brittle. I'll use _combine_filter_selectivities instead,

> This seems problematic. For example, I can't be sure since there's only a one-line docstring on this function, but I don't think we have a guarantee of ordering on the filter fields. This means that two runs with the the same input and statistics might produce two different estimates.

You're right, this is very brittle. I'll use _combine_filter_selectivities instead,

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: My goal was to make it clearer that it's `your` job to ask for an estimate, this api function won't do it for you. I'll work on making this clearer.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: Reverted.

pmantica1:
Looks good to me! Approved. Also, let's not do *args=None. I don't think there is much value in adding it and it would introduce some complexity.

obi1kenobi:
@pmantica1 I don't see the `*args=None` you are referring to, assuming it got refactored out?

ReaLNeroM:
@obi1kenobi Oh I was just asking for reviewers' thoughts on it, I haven't put it in the PR itself.

PR #397, opened 2019-07-16
Split renaming and validation

qqi0O0:
All validation of input schema now lives inside util.py file.

Other changes: replace deepcopy with parse and print_ast for performance; indent some doc strings for readability

One test that checks type coverage will fail, since the sets like `check_name_validity_types` and `unexpected_types` and such have been removed.


pmantica1:
Please don't merge yet.

PR #398, opened 2019-07-16
Amend custom GraphQLScalarType objects in parsed schemas

### LINKED ISSUE #392: Provide user-facing functionality for fixing serialization/deserialization of custom scalars
When parsing the textual representation of a GraphQL schema into a schema object, custom scalars' serialization and deserialization code are not picked up since they are not part of the textual representation. This causes issues when working with custom scalar objects.

#388 fixes this problem for our test suite. We need to expose a similar mechanism that users of the package can use to resolve this problem for their use cases.

pmantica1:
Resolves #392.


graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+def amend_custom_scalar_types(schema, scalars):
+    """Amend the serialization, parsing and description of custom scalar types in the schema.
+
+       Information about the fields, (other than the name), of custom GraphQLScalarType objects is
```
obi1kenobi: The text that is both between commas and between parens seems kind of awkward. Maybe rephrase?

graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,26 @@
...
+       lost when GraphQLSchema objects are serialized into strings. This function uses the
+       original type definitions to mutate the schema and reconstructs the fields of custom
+       scalar types.
```
obi1kenobi: All of this is indented three spaces too far.

graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+def amend_custom_scalar_types(schema, scalars):
+    """Amend the serialization, parsing and description of custom scalar types in the schema.
+
+       Information about the fields, (other than the name), of custom GraphQLScalarType objects is
+       lost when GraphQLSchema objects are serialized into strings. This function uses the
+       original type definitions to mutate the schema and reconstructs the fields of custom
```
obi1kenobi: `reconstructs the fields of custom scalar types` seems somewhat misleading. Given that I filed the issue this is closing, I know what this means (the fields of the Python object representation of the scalar type), but a casual reader would think "GraphQL scalars don't have fields, by definition -- that's what object/interface types are for!" and get very confused.

Let's try to make the language more precise and framed in a way that will be helpful to the end user. They don't care about the fields of the Python object, they care about being able to serialize and deserialize their data.

graphql_compiler/schema_generation/utils.py
```diff
@@ -0,0 +1,26 @@
...
+       scalar types.
+
+       Args:
+           schema: GraphQLSchema object
```
obi1kenobi: Missing a big, important fact here: this input argument is mutated! I know the text above mentions it, but it's important to mention that in every place that a reader might reference.

README.md
```diff
@@ -1656,6 +1656,18 @@ compilation_result = compile_graphql_to_cypher(
...
+Information about the description, serialization and parsing of custom scalar type
+objects is lost when a GraphQL schema is parsed from a string. This causes issues when
+working with custom scalar type objects. In order to avoid these issues, one can use the code 
+snippet below to amend the definitions of the custom scalar types used by the compiler. 
+
+```python
+from graphql_compiler.schema import CUSTOM_SCALAR_TYPES
+from graphql_compiler.schema_generation.utils import amend_custom_scalar_types
+
+amend_custom_scalar_types(schema, CUSTOM_SCALAR_TYPES)
```
obi1kenobi: Make this `your_schema_object` or something like that. Make it extra obvious that this is something the user is intended to provide, not the `schema` module in `graphql_compiler`, or something that you forgot to include.

PR #404, opened 2019-07-17
Update Pipfile for Cypher backends

LWprogramming:
Add relevant libraries for integration tests later.


Pipfile
```diff
@@ -32,6 +32,9 @@ graphql-core = ">=2.1,<3"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
+neo4j = "==1.7.4"
```
LWprogramming: [question](https://github.com/kensho-technologies/graphql-compiler/pull/369/files#r304106727) from #369: 

> General question: I notice that some of these other ones don't specify a particular version but rather a set of possible versions. How big a deal is this? I remember being told in some of the Github issues that some functionality that we rely on (e.g. the output format for RedisGraph) would change in future releases.

obi1kenobi: This Pipfile is only used for the test setup, and does not influence what packages are considered compatible when actually installing the compiler package itself. That said, when adding a requirement here, you should:
- Consider what versions of the dependencies we'd like to support. For example, we generally want to support the version that was latest at the time of writing, and any minor/patch releases around it, but not a major version bump (first number) since that might contain breaking changes.
- Then, if the package is required when using the compiler, and not just while testing it, you should make sure that the `setup.py` of the compiler package has the correct requirement that you have decided on. If the package is only used in tests, skip this step.
- Finally, you should add the package to the Pipfile, in the `dev` section if only used in tests, and in the non-dev section otherwise, and make sure the version bound matches what you decided earlier.

obi1kenobi: So specifically for this line in particular, I'd make it `neo4j = ">=1.7.4,<2".

obi1kenobi: Are these packages necessary when simply using the compiler, or are they only used in tests? If the latter, move them to the dev section above.

LWprogramming: Good point-- done

Pipfile
```diff
@@ -15,6 +15,7 @@ flake8-quotes = "==1.0.0"
 frozendict = "==1.2"
 isort = "==4.3.4"
 mysqlclient = "==1.3.14"
+neo4j = ">=1.7.4,<2"
```
bojanserafimov: Can we pin to exact versions? It's dev-packages, so it can't be too restrictive.

obi1kenobi: Normally I'd agree, but this pipfile is only for the local setup and for the CI tests, and I have a gut feeling that we lose more than we gain by pinning to exact versions here. Say that 1.8 gets released and accidentally breaks something -- we'll never pick it up unless someone specifically updates the pipfile. If the pipfile is flexible then the next update to the lockfile will pick up the new version.

Ultimately, as far as I can tell, the difference doesn't seem that large one way or the other.

PR #405, opened 2019-07-17
Create integration data for Cypher backends

LWprogramming:
The data here is meant to match the data in the file
create_orientdb_integration.sql

### DIFF
--- a/graphql_compiler/tests/test_data_tools/integration_data/create_redisgraph_integration.cypher
+++ b/graphql_compiler/tests/test_data_tools/integration_data/create_redisgraph_integration.cypher
@@ -1,0 +1,17 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+# This data is almost identical to the neo4j integration test data except for the first line, which doesn't appear in
+# the neo4j data:
+create (n)
+# The Redisgraph DB doesn't exist until we create a node, but doesn't go away if we delete all the nodes in the graph.
+# Now we can remove all the old data that might be in the DB.
+# Refer to Github issue: https://github.com/RedisGraph/RedisGraph/issues/551
+match (n) delete n
+
+# The other difference is that Redisgraph doesn't support toFloat (and other casting functions) so those are removed.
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/test_data_tools/integration_data/create_redisgraph_integration.cypher
```diff
@@ -0,0 +1,17 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+# This data is almost identical to the neo4j integration test data except for the first line, which doesn't appear in
+# the neo4j data:
+create (n)
+# The Redisgraph DB doesn't exist until we create a node, but doesn't go away if we delete all the nodes in the graph.
+# Now we can remove all the old data that might be in the DB.
+# Refer to Github issue: https://github.com/RedisGraph/RedisGraph/issues/551
+match (n) delete n
+
+# The other difference is that Redisgraph doesn't support toFloat (and other casting functions) so those are removed.
```
pmantica1: Can't we add dates in redisgraph?

pmantica1: Also you don't reference `toFloat` in the neo4j tests but `DATE` instead so maybe use that?

LWprogramming: Unfortunately, no-- Dates and Datetimes are unsupported types in RedisGraph according to the Open Cypher docs, which RedisGraph is based on.

LWprogramming: toFloat is a casting function that we used to use in the Neo4j integration data, but was removed in favor of just using the literal value. For instance, what used to be `toFloat('100')` became `100.0` there.

However, since I forgot to leave a comment there, I didn't remember to change the comment here.

I'll avoid future confusion like this (needing to make changes in several locations) by make sure there are comments in both places as a reminder to change it everywhere. Longer-term, we could also attempt to generate integration test data for different backends from a python representation, which means we'd only need to leave comments in the data-generating code (that's not a super high priority right now, though).

PR #406, opened 2019-07-17
Refactor merge_schemas

qqi0O0:
Main function broken up into smaller pieces for each type of new definition.

Add validity check for schema identifiers.


graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -17,63 +20,29 @@
...
 
     Returns:
         MergedSchemaDescriptor, a namedtuple that contains the AST of the merged schema,
         and the map from names of types/query type fields to the id of the schema that they
-        came from
+        came from. Scalars and directives will not appear in the map, as the same set of
+        scalars and directives are expected to be defined in every schema.
 
     Raises:
+        - ValueError if the some schema identifier is not a nonempty string of alphanumeric
```
pmantica1: typo?

graphql_compiler/schema_transformation/utils.py
```diff
@@ -33,7 +34,31 @@ class InvalidTypeNameError(SchemaTransformError):
...
+          than alphanumeric characters and underscores
+    """
+    if not isinstance(identifier, str):
+        raise ValueError(u'Schema identifier "{}" is not a string.'.format(identifier))
+    if identifier == '':
+        raise ValueError(u'Schema identifier must be a nonempty string.')
+    illegal_characters = frozenset(identifier) - _alphanumeric_and_underscore
+    if illegal_characters:
+        raise ValueError(
+            u'Schema identifier contains illegal characters: {}'.format(illegal_characters)
```
pmantica1: Can you also include in the error message the full schema identifier?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -169,3 +127,148 @@ def merge_schemas(schemas_dict):
...
+    Args:
+        scalar: ScalarDefinition, an AST node representing the definition of a scalar
+        existing_scalars: Set[str], set of names of all existing scalars
+        existing_types: Set[str], set of names of all existing types
+        merged_schema_ast: Document, AST representing a schema
+
+    Raises:
+        - SchemaNameConflictError if there is an existing type those name conflicts with the
+          new scalar
+    """
```
pmantica1: I am against this much documentation for this function. It is easier for me to read the code than the documentation. I would leave the documentation as a one section headline.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -169,3 +127,148 @@ def merge_schemas(schemas_dict):
...
+    Args:
+        directive: DirectiveDefinition, an AST node representing the definition of a directive
+        existing_directives: Dict[str, DirectiveDefinition], mapping the name of each existing
+                             directive to the AST node defining it
+        merged_schema_ast: Document, AST representing a schema
+
+    Raises:
+        - SchemaNameConflictError if there is an existing directive with the same name, but
+          a different definition, as the input directive
+    """
```
pmantica1: Same as below.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -169,3 +127,148 @@ def merge_schemas(schemas_dict):
...
+        schema_id: str, the identifier of the schema that this type came from
+        existing_scalars: Set[str], set of names of all existing scalars
+        name_to_schema_id: Dict[str, str], mapping names of types to the identifier of the schema
+                           that they came from
+        merged_schema_ast: Document, AST representing a schema
+
+    Raises:
+        - SchemaNameConflictError if there is an existing type or scalar those name conflicts
+          with the new type's name
+    """
```
pmantica1: Same as above.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -82,84 +51,73 @@ def merge_schemas(schemas_dict):
...
+                ast_types.ObjectTypeDefinition,
+                ast_types.UnionTypeDefinition,
+            )):
+                _process_generic_type_definition(
+                    new_definition, current_schema_id, scalars, name_to_schema_id, merged_schema_ast
+                )
+
+            else:  # All definition types should've been covered
+                raise AssertionError(
+                    u'Missed definition type: "{}"'.format(type(new_definition).__name__)
```
pmantica1: I think it is nearly strictly better to just do `.format(new_definition)`. If the `__str__` method is overridden then stringifying the element is likely to produce useful information. If it is not overridden, then `__str__` defaults to `__repr__` which almost definitely produces at least the same amount of information as `type(new_definition).__name__`, (the default implementation includes the type name). Furthermore, `__repr__` method is often overridden by directive defintions so it'd be a good idea to use it.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -33,7 +34,33 @@ class InvalidTypeNameError(SchemaTransformError):
 
 
 class SchemaNameConflictError(SchemaTransformError):
-    """Raised when renaming types or fields cause name conflicts."""
+    """Raised when renaming or merging types or fields cause name conflicts."""
+
+
+_alphanumeric_and_underscore = frozenset(string.ascii_letters + string.digits + '_')
```
LWprogramming: We have something almost identical [here](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/helpers.py#L21)-- that version is python2 and 3 compatible too.

Might be a bit odd importing something from `..compiler`, so maybe move that definition into some higher-level directory?

graphql_compiler/schema_transformation/utils.py
```diff
@@ -33,7 +34,33 @@ class InvalidTypeNameError(SchemaTransformError):
...
+
+
+def check_schema_identifier_is_valid(identifier):
```
LWprogramming: Similar function [here](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/helpers.py#L147)

Granted, that function lets through other things (see line 150 of that file in particular) and it's not documented as well as this one is, but could be worth a shot modifying it to reuse that code?

graphql_compiler/schema_transformation/merge_schemas.py
LWprogramming: I actually found it easier to understand with the fleshed-out docstring, although I think it's valid to take out the paragraph right below the headline since the code is pretty readable below.

I think keeping the argument types (Args) and Raises section would be nice, and we can take out `# Add to AST` and `# Add to record of scalars` down below.

qqi0O0: I also rather like documenting the types of the arguments, because it's not super trivial to find out. I'm putting the args back for now, deleting some useless comments elsewhere.

pmantica1: Ok. This decision makes sense to me.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -169,3 +127,115 @@ def merge_schemas(schemas_dict):
...
+                          function
+        existing_types: Set[str], set of names of all existing types
+        merged_schema_ast: Document, AST representing a schema; modified by this function
+    """
+    scalar_name = scalar.name.value
+    if scalar_name in existing_scalars:
+        return
+    if scalar_name in existing_types:
+        raise SchemaNameConflictError(
+            u'New scalar "{}" clashes with existing type.'.format(scalar_name)
```
obi1kenobi: This is an error message that is absolutely correct, and yet would be quite frustrating to address as a user. Two things would make it better: telling the user what they need to do about it, and giving them the necessary information such that they can do what needs doing.

For "what they need to do about it", I am worried that a user that sees this error might incorrectly conclude something like "oh, I guess I can't merge these schemas then." They might be entirely unfamiliar with the fact that the compiler already has a tool that is able to rename types -- and a subtle nudge here toward using that tool would ensure they know about it and react appropriately. One additional tricky thing: should they rename the scalar, or the other type? This is a good place to give users the advice to rename the other (non-scalar) type, since the scalar may come up e.g. in query output metadata as well, which could be more risky.

Given that they now want to rename the type, they need to find out which type from which schema is the problem. The error message never actually tells us either of these facts. Inspecting the code would tell them that the types conflict because they have the same name (which doesn't necessarily have to have been the case), but it still won't tell them which schema the type came from. And we might be merging dozens of schemas, each with many thousands of types.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -169,3 +127,115 @@ def merge_schemas(schemas_dict):
...
+        schema_id: str, the identifier of the schema that this type came from
+        existing_scalars: Set[str], set of names of all existing scalars
+        name_to_schema_id: Dict[str, str], mapping names of types to the identifier of the schema
+                           that they came from; modified by this function
+        merged_schema_ast: Document, AST representing a schema; modified by this function
+    """
+    type_name = generic_type.name.value
+    if type_name in existing_scalars:
+        raise SchemaNameConflictError(
+            u'New type "{}" clashes with existing scalar.'.format(type_name)
```
obi1kenobi: Similar concerns about these error messages as for the one above.

graphql_compiler/schema_transformation/utils.py
obi1kenobi: The `six.text_type` part might be good to grab from there, but redefining this here as well is not awful -- it's a one-liner, and there's no semantic connection to the other code that would be aided by deduplicating these two lines.

graphql_compiler/schema_transformation/utils.py
obi1kenobi: It's probably not worth the effort -- they check for very different identifiers, and the logic is quite simple. The necessary top-level utils file is probably going to be more trouble than it's worth.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -33,7 +34,33 @@ class InvalidTypeNameError(SchemaTransformError):
...
+    """Check if input is a valid identifier, made of alphanumeric and underscore characters.
+
+    Args:
+        identifier: str
```
obi1kenobi: Might be nice to give the reader some intuition about what `identifier` we're talking about. The errors all mention `schema identifier` but the docstring doesn't really say what the identifier is supposed to be identifying.

In general, I'm not a fan of just defining the type of the argument without attempting to provide some intuition about where that value might be coming from / what it might be used for.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: I've edited several of the name conflict error messages to include more information as well as suggestions to rename_schema. For the directive clashing error though, I don't keep track of what previous schema(s) the old definition came from, and rename_schema doesn't rename directives anyway, so I haven't enhanced the error message. Is that fine?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -82,90 +51,200 @@ def merge_schemas(schemas_dict):
...
+                           that they came from
+        merged_schema_ast: Document, AST representing a schema; modified by this function
+    """
+    scalar_name = scalar.name.value
+    if scalar_name in existing_scalars:
+        return
+    if scalar_name in name_to_schema_id:
+        raise SchemaNameConflictError(
+            u'New scalar "{}" clashes with existing type "{}" in schema "{}". Consider '
+            u'renaming this type in schema "{}" using the tool rename_schema before merging '
```
pmantica1: nit: It is ambiguous what you mean by "this type".

graphql_compiler/schema_transformation/merge_schemas.py
obi1kenobi: It's probably okay for now, I think it's probably going to be a lot less common than conflicts between type names. If it becomes a problem down the line, we can fix it then.

PR #410, opened 2019-07-17
Create integration test fixtures

LWprogramming:

### DIFF
--- a/graphql_compiler/tests/conftest.py
+++ b/graphql_compiler/tests/conftest.py
@@ -15,6 +15,8 @@ from .test_data_tools.orientdb_graph import get_test_orientdb_graph
 from .test_data_tools.redisgraph_graph import get_test_redisgraph_graph
 from .test_data_tools.schema import load_schema
 
+GRAPH_NAME = 'animals'  # Name for integration test database
+
 
 # Pytest fixtures depend on name redefinitions to work,
 # so this check generates tons of false-positives here.
@@ -40,8 +42,7 @@ def init_integration_orientdb_client():
 @retry(20, timeout=1)  # pylint: disable=no-value-for-parameter
 def _init_orientdb_client(load_schema_func, generate_data_func):
     """Set up a database and return a client that can query the database."""
-    graph_name = 'animals'
-    orientdb_client = get_test_orientdb_graph(graph_name, load_schema_func, generate_data_func)
+    orientdb_client = get_test_orientdb_graph(GRAPH_NAME, load_schema_func, generate_data_func)
     return orientdb_client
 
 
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/conftest.py
```diff
@@ -30,25 +33,45 @@ def init_integration_orientdb_client():
     return _init_orientdb_client(load_schema, generate_orient_integration_data)
 
 
+# retry is a decorator that attempts to call the decorated function repeatedly, and it takes a
+# parameter "call". Call is the function being decorated, so we don't need to specify that in
+# the parameters that the retry decorator takes, so we can disable pylint's warning about
+# missing parameters here.
+@retry(tries=20, timeout=1)  # pylint: disable=no-value-for-parameter
```
LWprogramming: [From](https://github.com/kensho-technologies/graphql-compiler/pull/369/files#r304532811) PR #369:

> Can you share the error pylint raises here? This comment effectively states that pylint has buggy handling of decorators, which is definitely possible but bears verifying before we silence the error.

The error pylint gives is `E: 40, 1: No value for argument 'call' in function call (no-value-for-parameter)`

> Also, tries is a positional argument, not a kwarg -- so let's pass it positionally.

Good call-- done

graphql_compiler/tests/conftest.py
```diff
@@ -30,25 +33,48 @@ def init_integration_orientdb_client():
...
+    """Return a client for an initialized db, with all test data imported."""
+    return _init_neo4j_client(generate_neo4j_integration_data)
+
+
+# We can disable pylint's warning about missing parameters here because retry is a decorator. See
+# _init_orientdb_client function comment.
+@retry(20, timeout=1)  # pylint: disable=no-value-for-parameter
+def _init_neo4j_client(generate_data_func):
+    """Set up a database and return a client that can query the database."""
+    graph_name = 'animals'
```
LWprogramming: I'm open to defining `graph_name` somewhere else once instead of in every function (since we want the integration tests' database to be as similar as possible). Do we have a file to put this, or just leave this as a global variable in this file?

graphql_compiler/tests/conftest.py
```diff
@@ -30,25 +33,48 @@ def init_integration_orientdb_client():
     return _init_orientdb_client(load_schema, generate_orient_integration_data)
 
 
+# retry is a decorator that attempts to call the decorated function repeatedly, and it takes a
+# parameter "call". Call is the function being decorated, so we don't need to specify that in
+# the parameters that the retry decorator takes, so we can disable pylint's warning about
```
pmantica1: nit: so we don't need to specify which parameters the retry decorator takes,

graphql_compiler/tests/conftest.py
pmantica1: I can't think of a better place to put this than as a global variable here.

PR #411, opened 2019-07-17
Create the integration tests themselves for Cypher backends

LWprogramming:
Currently waiting on clients and data_tool to get merged into master.


graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -12,31 +12,60 @@
...
     test_backend.MARIADB,
     test_backend.MYSQL,
     test_backend.SQLITE,
     test_backend.MSSQL,
-])
+    test_backend.NEO4J,
+    test_backend.REDISGRAPH,
+]
+
+# Store the test parametrization for running against all backends. Individual tests can customize
```
LWprogramming: [From](https://github.com/kensho-technologies/graphql-compiler/pull/369#discussion_r303974011) PR #369:

> I'm hesitant to have two definitions. Also, almost_all_backends with no arguments seems like bad design -- it doesn't communicate that it actually doesn't exclude any backends. Why not have one definition, all_backends and allow an exclude to be passed to it? Then, if there are no args, it does the obvious thing, and if it's called with a kwarg called e.g. except_backends then it also does the obvious thing?

Sounds good to me. How about renaming `all_backends` -> `use_all_backends`? Then the decorator might look like `@use_all_backends(except_backends([test_backend.REDISGRAPH]))` which almost looks like natural language to me. We could also replace `except_backends` with `except_for_backends` in that case.

obi1kenobi: `use_all_backends` sounds fine to me. I'd make `except_backends` be a kwarg instead of a function of its own, so the call looks like `@use_all_backends(except_backends=[test_backend.REDISGRAPH])`.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -12,31 +12,60 @@
...
+def almost_all_backends(excluded_backends):
+    """Decorator for test functions to use a specific backend.
+
+    Args:
+        excluded_backends: List[str], a list of backend strings from test_backend.py to exclude in
+                           testing.
+
+    Returns:
+        function that expands tests. parameterized.expand() takes in a list of test parameters (in
+                  this case, backend strings specifying which backends to use for the test) and
```
obi1kenobi: de-indent this to the same level as `function`, not sure why this is indented so much

obi1kenobi:
I really like the detailed comments when excluding backends from tests! It will make it substantially more likely that we remember to un-exclude backends once the underlying limitations are resolved, and that makes me happy 👍

graphql_compiler/tests/integration_tests/test_backends_integration.py
LWprogramming: done-- now, if we use the decorator with no arguments it's written as `@use_all_backends()` which might look a bit weird, but the alternative (just `@use_all_backends` if there are no arguments) requires a [complicated structure](https://realpython.com/primer-on-python-decorators/#both-please-but-never-mind-the-bread) that probably wouldn't be very easy to read later on.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -12,31 +12,60 @@
...
 
 
+def use_all_backends(except_backends=()):
+    """Decorator for test functions to use specific backends.
+
+    By default, tests decorated with this function use all backends. However, some backends don't
+    support certain features, so it's useful to exclude certain backends for individual tests.
+
+    Args:
+        except_backends: Tuple[str], optional argument. Tuple of backend strings from
```
pmantica1: nit: You mention Tuple but pass in the docstring lists. I think this might lead to errors when we start using mypy.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -12,31 +12,60 @@
...
+                         test_backend.py to exclude in testing.
+
+    Returns:
+        function that expands tests. parameterized.expand() takes in a list of test parameters (in
```
pmantica1: The mentioning of parameterized.expand() is best written as a comment instead of a docstring. A docstring should explain what the function does not how it does it.

PR #412, opened 2019-07-18
Rename query

qqi0O0:
Code for renaming queries.

Currently, I don't check the validity of renamings or the validity of the output query, so it's potentially possible for the user to, for example, rename multiple types to the same thing. 

It feels to me that it's much easier and more reasonable to check for query validity when the corresponding schema is known. I also have the (possible incorrect) impression that queries are pretty small, so if a renaming seems to have gone wrong, it's reasonable for the user to take a look at the renamed query and notice issues.

Tests exist, but not in this PR


graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+from graphql.language.visitor import Visitor, visit
+
+from .utils import QueryStructureError
+
+
+def rename_query(ast, renamings):
+    """Translate all types and root fields (fields of query type) using renamings.
+
+    Most fields (fields of types other than the query type) will not be renamed. Any type
+    or field that does not appear in renamings will be unchanged.
```
pmantica1: Let's remove the first two sentences. The first sentence is not necessarily true. The second sentence is more of a technicality and I think is best expressed in the Args section as you have it right now.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+from copy import deepcopy
+
+from graphql.language import ast as ast_types
+from graphql.language.visitor import Visitor, visit
+
+from .utils import QueryStructureError
+
+
+def rename_query(ast, renamings):
+    """Translate all types and root fields (fields of query type) using renamings.
```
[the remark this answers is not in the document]
pmantica1: Also, I don't think that you rename all types.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Most fields (fields of types other than the query type) will not be renamed. Any type
+    or field that does not appear in renamings will be unchanged.
+    Roughly speaking, rename_query renames the same elements as rename_schema.
```
pmantica1: I think it'd be best that you just explained that after we rename a schema, we also need to rename the queries intended for that schema in this docstring block.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Args:
+        ast: Document, representing a valid query. It is assumed to have passed GraphQL's
+             builtin validation -- validate(schema, ast). ast not modified by this function
```
pmantica1: nit: The ast is not modified by this function. (It looks like a misspelling to not start a sentence with a capital letter).

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Raises:
+        - QueryStrutureError if the ast does not have the expected form; in particular, if the
+          AST contains Fragments, or if it contains an InlineFragment at the root level
+    """
+    # NOTE: There is a validation section in graphql-core that takes in a schema and a
+    # query ast, and checks whether the query is valid. This code assumes this validation
+    # step has been done on the input AST.
+    if len(ast.definitions) > 1:  # includes either multiple queries, or fragment definitions
+        raise QueryStructureError(
```
pmantica1: There are already exceptions regarding query validity in graphql_compiler/exceptions so I think it's best you either use one from there or put `QueryStructureError` there in order to have all exceptions of the same kind in one place.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def enter_SelectionSet(self, node, *args):
+        """Record that we entered another nested level of selections."""
+        self.selection_set_level += 1
+
+    def leave_SelectionSet(self, node, *args):
+        """Record that we left a level of selections."""
+        self.selection_set_level -= 1
+
+    def enter_Field(self, node, *args):
```
pmantica1: What other type of Field objects are there other than root fields?

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Returns:
+        Document, a new AST representing the renamed query
+
+    Raises:
+        - QueryStrutureError if the ast does not have the expected form; in particular, if the
+          AST contains Fragments, or if it contains an InlineFragment at the root level
+    """
+    # NOTE: There is a validation section in graphql-core that takes in a schema and a
+    # query ast, and checks whether the query is valid. This code assumes this validation
```
pmantica1: Instead of mentioning the validation functions here and in the docstring, I would mention what they validate instead.

graphql_compiler/schema_transformation/rename_query.py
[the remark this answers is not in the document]
qqi0O0: Do you mean that I don't rename types whose names don't appear in the dict?

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: For example, in query 
```
{
  Human {
    friend {
      name
    }
  }
}
```
friend and name are both fields but are not root vertex fields

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: I took a look at those errors and feel uncertain of what I should do. compiler_frontend.py checks for some of the same criteria that I do here (for example, not starting with type coercion) and raise GraphQLCompilationError if the check fails, but I'm not doing compilation here, just a transformation. So that error doesn't feel right even if it's raised under some of the same conditions. The other existing errors don't apply either. What are your thoughts?

graphql_compiler/schema_transformation/rename_query.py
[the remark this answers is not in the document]
pmantica1: Yes.

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: The initially intended use case is kind of backwards -- rename schema S to S' so that S' can be merged without conflicts, then query Q' written by the user in S' gets renamed to Q and can actually be run against a database. This is somewhat hard to explain though. What if I just say "This function is intended to be used in conjunction with rename_schema"?

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: The full list of validations is long (specified_rules  in https://github.com/graphql-python/graphql-core/blob/05a4c8222c0b65ff8df61564fe7f41efee060b4e/graphql/validation/rules/__init__.py), I'll mention a few of them as examples.

graphql_compiler/schema_transformation/rename_query.py
pmantica1: How about: 
"The intended use of this function is to fix queries that were invalidated because of a schema renaming by analogously renaming the elements in the query. "

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def rename_query(ast, renamings):
+    """Translate names of types and root vertex fields using renamings.
+
+    Besides root vertex fields (fields of the query type), no field will be renamed.
+
+    Args:
+        ast: Document, representing a valid query. It is assumed to have passed GraphQL's
+             builtin validation -- validate(schema, ast), in that it has the structure of a
```
pmantica1: If you are going to mention a function please be very specific about where this function is located.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,96 @@
...
+from graphql.language import ast as ast_types
+from graphql.language.visitor import Visitor, visit
+
+from ..exceptions import GraphQLValidationError
+
+
+def rename_query(ast, renamings):
+    """Translate names of types and root vertex fields using renamings.
+
+    Besides root vertex fields (fields of the query type), no other fields will be renamed.
```
obi1kenobi: I don't think this is sufficient -- what about `InlineFragment` AST objects, which we use to perform type coercions?

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        ast: Document, representing a valid query. It is assumed to have passed GraphQL's
+             builtin validation -- validate(schema, ast) in graphql/validation/validation.py --
+             in that it has the structure of a valid query, does not reference non-existent
+             types or fields, and passes type checks. The ast is not modified by this function
```
obi1kenobi: nit: in plain English (i.e. when not referring to a Python argument), let's make sure to always capitalize AST since it's an acronym.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from copy import deepcopy
+
+from graphql.language import ast as ast_types
+from graphql.language.visitor import Visitor, visit
+
+from ..exceptions import GraphQLValidationError
+
+
+def rename_query(ast, renamings):
```
obi1kenobi: Any reason not to take a `RenamedSchemaDescriptor` object? Then we have a somewhat stronger belief that the renamings were actually computed via a call to `rename_schema` and are therefore valid.

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: All types will be renamed, and only root vertex fields will be renamed. So for example, in 
```
{
  Character {
    friend {
      ... on Human {
        name
      }
    }
  }
}
```
Character will be renamed since it's a root vertex field, Human will be renamed since it's a type, friend and name will not be renamed since they're normal fields.

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: That's a good point. I started by thinking that rename_query would take in a `RenamedSchemaDescriptor`, but moved away from that idea for several reasons, but there are definitely arguments on both sides.

Taking in a `RenamedSchemaDescriptor` means that rename_query is probably being used in conjunction with rename_schema as expected, and it also allows validity checking (if necessary) since we have access to its schema.

One issue though is that I think users may be confused about which direction rename_query operates, if the input is a `RenamedSchemaDescriptor`. Suppose the user previous renamed schema S to S'. The intended use of rename_query will change query Q' written in terms of types in S' into a query Q written in terms of types in S. However, a user may think that rename_query instead changes Q to Q', in the same direction as rename_schema. Passing in a renamings dictionary doesn't cause this kind of confusion.

pmantica1: On the other hand, we only want someone to use the function the reversed_named map of `RenamedSchemaDescriptor`, so it is useful to establish this dependency.

pmantica1: However, we'll make this dependency clear when we add a section to the README.md about schema merging so I slightly prefer keeping it the way it is.

obi1kenobi: Here's what I'm concerned about. I believe that `renamings` here is actually a reverse name map – the opposite of the `renamings` used for schema renaming. I suspect _this_ is more likely to cause the directionality confusion @qqi0O0 mentioned, compared to taking in a `RenamedSchemaDescriptor`.

Option 1 is to make it impossible to misuse the code, by requiring a `RenamedSchemaDescriptor` object to be used, which essentially proves that the user called the right functions in the right order. Option 2 is to write instructions into the `README.md` and pray that people read it (which I suspect they won't – it's huge!).

I'm quite certain I prefer option 1 here.

graphql_compiler/schema_transformation/rename_query.py
obi1kenobi: It might be good to add a sentence into the docstring saying something to this effect. For example:

```
All type names (including ones in type coercions), as well as root vertex fields (fields of the query type) will be renamed. No other field names will be renamed.
```

graphql_compiler/schema_transformation/rename_query.py
qqi0O0: It's true that the `renamings` here is the reverse of the `renamings` used in schema renaming, for the use case that we have in mind. Nonetheless, if we pass in a dict here, it seems clear what things will get renamed to what -- the keys in `renamings` will get renamed to their corresponding values.
On the other hand, if we pass in a `RenamedSchemaDescriptor` here, it's not clear what will be renamed to what and what mapping of names will be used -- whether it's the same `renamings` dict that produced the renamed schema, or the `reverse_name_map` in the `RenamedSchemaDescriptor`. For that reason I think passing in a `RenamedSchemaDescriptor` is more potentially misleading.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -3,30 +3,43 @@
...
 
-def rename_query(ast, renamings):
-    """Translate names of types and root vertex fields using renamings.
+def rename_query(ast, renamed_schema_descriptor):
+    """Translate names of types using reverse_name_map of the input RenamedSchemaDescriptor.
 
-    Besides root vertex fields (fields of the query type), no other fields will be renamed.
+    The direction in which types and fields are renamed is opposite of the process that
+    produced the renamed schema descriptor. If a type X was renamed to Y in the schema, then
+    any occurances of type Y in the input query ast will be renamed to X.
```
obi1kenobi: typo, I believe: `occurance` -> `occurrence`

obi1kenobi:
Other than the one typo, looks great!

PR #413, opened 2019-07-18
Improve schema graph documentation

pmantica1:


README.md
```diff
@@ -1628,17 +1629,33 @@ schema_data = [record.oRecordData for record in schema_records]
...
 
+# Get all the subclasses of a class.
 print(schema_graph.get_subclass_set('Animal'))
 # {'Animal', 'Dog'}
 
+# Get all the outgoing edge classes of a vertex class.
+print(schema_graph.get_vertex_schema_element_or_raise('Animal').out_connections)
+# {'Animal_Eats', 'Animal_FedAt', 'Animal_LivesIn'}
+
+# Get the vertex classes allowed in the arrow side of an edge class.
```
qqi0O0: I don't particularly like calling it "the arrow side" since that terminology feels imprecise. Perhaps `Get the vertex classes allowed as the terminal vertex of an edge class`? Not totally sure. https://cs.stackexchange.com/questions/70928/name-for-the-starting-vertex-and-ending-vertex-of-a-directed-edge

pmantica1: Ok I addressed this comment. I changed it from out_connections to in_connections so that I could use the phrase the the "tail vertex" which I feel is as precise and clearer that "terminal vertex".

README.md
```diff
@@ -1600,16 +1600,17 @@ would enable the use of a `@fold` on the `adjacent_animal` vertex field of `Foo`
 ```
 
 ### SchemaGraph 
-The `SchemaGraph` is a utility class that encodes a database schema and allows for better schema 
-introspection than the GraphQL schema. The  `SchemaGraph` has three main advantages:
+When building a GraphQL schema from the database metadata, we first build a `SchemaGraph` from 
+the metadata and then, from the `SchemaGraph`, build the GraphQL schema. The `SchemaGraph` is also 
+an encoding of the underlying database schema, but it has three main advantages that make it a
```
obi1kenobi: `an encoding` might be confusing, I'd replace it with `a representation`

README.md
```diff
@@ -1628,17 +1629,32 @@ schema_data = [record.oRecordData for record in schema_records]
...
 
+# Get all the subclasses of a class.
 print(schema_graph.get_subclass_set('Animal'))
 # {'Animal', 'Dog'}
 
+# Get all the outgoing edge classes of a vertex class.
+print(schema_graph.get_vertex_schema_element_or_raise('Animal').out_connections)
+# {'Animal_Eats', 'Animal_FedAt', 'Animal_LivesIn'}
+
+# Get the vertex classes allowed as the tail vertex of an edge class.
```
obi1kenobi: Consider `tail vertex` -> `destination vertex` instead?

PR #414, opened 2019-07-19
Improve string validation error

pmantica1:
This PR aims to invalid string validation errors, specifically those relating to output and argument names, since those are the most likely ones to be encountered by the user.


graphql_compiler/compiler/helpers.py
```diff
@@ -161,7 +161,10 @@ def validate_safe_string(value):
 
     if not set(value).issubset(VARIABLE_ALLOWED_CHARS) and \
             value not in legal_strings_with_special_chars:
-        raise GraphQLCompilationError(u'Encountered illegal characters in string: {}'.format(value))
+        raise GraphQLCompilationError(u'Encountered illegal characters in string: {}. '
+                                      u'Strings are only allowed to have underscores, '
+                                      u'digits and, upper and lower case alphabetical letters.'
+                                      .format(value))
```
bojanserafimov: how about `legal_strings_with_special_chars`?

LWprogramming: Could be worth mentioning that variable in the error message. That way, if we end up changing `legal_strings_with_special_chars` later, we don't need to go find every single error message and update it or send users on a wild goose chase because of an outdated error message.

pmantica1: Can't figure out a better way to express this that starting with "In general, ..."

obi1kenobi: I think it might be time to have a separate validation function that allows `legal_strings_with_special_chars` to go through, because `validate_safe_string` is used in a lot of places now, and many of those places shouldn't allow those strings with special chars.

I'm assuming this fix is intended to improve the error message when validating output names?

pmantica1: To improve the error message when validating runtime argument, tagged parameter names and output names.

pmantica1: I don't think that it is worth the effort to separate validation right now. I think that the most important thing is to be as descriptive and precise as possible when raising validation errors for argument and output names since these are errors are among the most likely ones for the user to encounter.

graphql_compiler/compiler/helpers.py
```diff
@@ -161,15 +161,47 @@ def validate_safe_string(value):
...
-        raise GraphQLCompilationError(u'Encountered illegal characters in string: {}'.format(value))
+        raise GraphQLCompilationError(u'Encountered illegal characters in string: {}. Provided '
+                                      u'string is only allowed to have upper and lower case '
+                                      u'letters, digits and underscores.'
+                                      .format(value))
 
 
-def validate_output_name(value):
+def _validate_safe_argument_or_output_name(name, name_description):
+    """Ensure that the provided argument or output name does not have illegal characters."""
```
pmantica1: I added a name description in order for the user to be able to quickly tell whether it was a tagged parameter, runtime argument or output name that failed validation. I think that it's important that this information is communicated. However, I am unsure of how to best reflect this change in the code. Here is one way to do it, but I am open to suggestions.

graphql_compiler/compiler/helpers.py
```diff
@@ -140,36 +140,51 @@ def get_uniquely_named_objects_by_name(object_list):
...
 
     if value[0] in string.digits:
-        raise GraphQLCompilationError(u'String values cannot start with a digit: {}'.format(value))
+        raise GraphQLCompilationError(u'Encountered invalid {}: {}. It cannot start with a '
+                                      u'digit.'.format(value_description, value))
+
+    if not set(value).issubset(VARIABLE_ALLOWED_CHARS):
+        raise GraphQLCompilationError(u'Encountered illegal characters in {}: {}. It is only '
+                                      u'allowed to have upper and lower case letters, '
+                                      u'digits and underscores. '
```
obi1kenobi: Trailing space in the message. Also, let's refactor this to say which specific disallowed characters were present:
```
disallowed_chars = frozenset(value) - VARIABLE_ALLOWED_CHARS
if disallowed_chars:
    ...
```

graphql_compiler/compiler/helpers.py
```diff
@@ -140,36 +140,51 @@ def get_uniquely_named_objects_by_name(object_list):
...
+    """Ensure that the provided string is valid for use as a runtime argument name."""
+    validate_safe_string(name, 'runtime argument name')
+
 
-    if not set(value).issubset(VARIABLE_ALLOWED_CHARS) and \
-            value not in legal_strings_with_special_chars:
-        raise GraphQLCompilationError(u'Encountered illegal characters in string: {}'.format(value))
+def validate_tagged_argument_name(name):
+    """Ensure that provided string is valid for use as a tagged argument name"""
+    validate_safe_string(name, 'tagged argument name')
```
obi1kenobi: do not pass kwargs positionally

graphql_compiler/compiler/emit_match.py
```diff
@@ -111,10 +111,10 @@ def _represent_fold(fold_location, fold_ir_blocks):
     mark_name, _ = fold_location.get_location_name()
     base_location_name, _ = fold_location.base_location.get_location_name()
 
-    validate_safe_string(mark_name)
-    validate_safe_string(base_location_name)
-    validate_safe_string(edge_direction)
-    validate_safe_string(edge_name)
+    validate_safe_or_special_string(mark_name)
```
obi1kenobi: Many of these shouldn't be `safe_or_special`, they should just be `safe`. The mark name can never be `@rid`.

I realize that this is something the old code allowed, but now that we are making this in-depth change, leaving that in would just add to the confusion since it seems like an intentional decision to allow these special names.

PR #415, opened 2019-07-19
Fix bug in macro expansion with pro-forma fields

bojanserafimov:


graphql_compiler/tests/test_macro_expansion.py
```diff
@@ -38,6 +38,60 @@ def test_macro_edge_basic(self):
...
+                }
+            }
+        }'''
+        expected_args = {}
+
+        expanded_query, new_args = perform_macro_expansion(self.macro_registry, query, args)
+        compare_graphql(self, expected_query, expanded_query)
+        self.assertEqual(expected_args, new_args)
+
+    def test_macro_edge_pro_forma_fields_2(self):
```
pmantica1: How about `test_macro_edge_multiple_pro_forma_fields`?

graphql_compiler/tests/test_macro_expansion.py
```diff
@@ -38,6 +38,60 @@ def test_macro_edge_basic(self):
         compare_graphql(self, expected_query, expanded_query)
         self.assertEqual(expected_args, new_args)
 
+    def test_macro_edge_pro_forma_fields(self):
```
pmantica1: Can you make a comment explaining what a pro_format field here is and in the test below?

graphql_compiler/macros/macro_expansion.py
```diff
@@ -89,12 +89,20 @@ def _merge_selection_sets(selection_set_a, selection_set_b):
     # having an empty selection in a vertex field. We remove pro-forma fields if they are
     # no longer necessary.
     if len(merged_selection_dict) > 1:  # Otherwise we need a pro-forma field
-        merged_selection_dict = {
+        actual_fields = {
```
obi1kenobi: nit: Not a huge fan of `actual` here, how about `non_pro_forma_fields` instead?

PR #416, opened 2019-07-19
Improve invalid argument error

pmantica1:


graphql_compiler/compiler/filters.py
```diff
@@ -154,7 +157,14 @@ def _represent_argument(directive_location, context, argument, inferred_type):
         return (representation, non_existence_expression)
     else:
         # If we want to support literal arguments, add them here.
-        raise GraphQLCompilationError(u'Non-argument type found: {}'.format(argument))
+        raise GraphQLCompilationError(u'Invalid argument found: {}. The compiler supports only '
+                                      u'runtime arguments, which must begin with the $ character, '
+                                      u'and tagged arguments, which must begin with the % '
+                                      u'character. Literal arguments, (ex. 10, \'Kensho '
```
obi1kenobi: use `e.g.` instead of `ex.`, I think it's a more conventional

graphql_compiler/compiler/filters.py
```diff
@@ -154,7 +157,14 @@ def _represent_argument(directive_location, context, argument, inferred_type):
...
+                                      u'Technologies\', \'2018-01-01\'), are currently not '
+                                      u'supported. Please use runtime arguments and pass in the '
+                                      u'corresponding literal values as query parameters.'
+                                      u''.format(argument))
```
obi1kenobi: you can actually remove the `u''` from this, it will still parse correctly

graphql_compiler/compiler/filters.py
```diff
@@ -98,12 +98,12 @@ def _represent_argument(directive_location, context, argument, inferred_type):
               expression that will evaluate to True if the argument is skipped as optional and
               therefore not present, and False otherwise.
     """
-    # Regardless of what kind of variable we are dealing with,
-    # we want to ensure its name is valid.
     argument_name = argument[1:]
-    validate_safe_string(argument_name)
 
     if is_runtime_parameter(argument):
+        # We want to valid the argument name after we validated that it is not a literal argument
```
LWprogramming: Nit: `want to valid the argument name` -> `want to validate the argument name`

PR #417, opened 2019-07-19
Cost estimation documentation

ReaLNeroM:
@reviewers: Since we want an explanation of how this works, I'll wait until we start using readthedocs, where it's okay to delve into detail. As it stands, a 200-word explanation of how it works is not going to help.


README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
   Alternatively, the entire OrientDB class and all edges that may point to it may be omitted
   entirely from the GraphQL schema.
 
+## Query cost estimation
+
+Due to GraphQL's easy-to-use nature, queries that use a lot of memory and/or have a long execution
```
LWprogramming: Elaborate on why it's easy to write queries that take up a lot of space and time, i.e. bc we have arbitrary levels of nesting

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+time can happen often. For better user-friendliness, it's worthwhile to estimate a query's cost (in
```
LWprogramming: Nit: "For better user-friendliness," -> "To prevent queries from using too many resources" (too much memory and/or time? or too many resources?)

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+terms of time taken and memory used) using statistics without executing the query on the database.
+Since the time and memory required for query execution varies depending on the underlying database,
+as a rough proxy, the cost estimator estimates the number of result rows for the given query i.e.
```
LWprogramming: Nit: as a rough proxy, the cost estimator estimates the number of result rows for the given query i.e. the query's cardinality.

->

the cost estimator estimates the query's cardinality (i.e. the number of result rows for the given query) as a rough proxy for resource usage.

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+Before estimates can be generated, you need to provide statistics about the graph data. The
+applicable classes are in `graphql_compiler/cost_estimation/statistics.py`. Currently, two classes
+are implemented:
+- `Statistics`, an abstract base class that describes the interface the cost estimator expects for
+  statistics. This allows users to create their own implementations for the statistic class.
+- `LocalStatistics`, an implementation of `Statistics`, where the statistics are provided during
+  initialization, and can't be modified thereafter.
+
+Currently, three statistics are supported. `get_class_count()` must be provided for cardinality
+estimation. The other two are optional, as they can be roughly approximated.
```
LWprogramming: > they can be roughly approximated.

Worthwhile discussing how they can be approximated? Up to you.

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+
+### Required statistics
+
+- `get_class_count(class)`: 
+  Given the name of a vertex or edge class, return the number of objects of that type. If the given
+  class is an interface, the objects that inherit from this class should also be included. So if
+  `get_class_count('Entity')` is called, it should also include the number of 'Animal' objects,
```
LWprogramming: Nit: 'Animal' -> `Animal`, surround with backticks. same for `Food`.

Also good to say explicitly that they inherit from `Entity`.

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+as a rough proxy, the cost estimator estimates the number of result rows for the given query i.e.
+the query's cardinality.
+
+These estimates can be used to improve user-friendliness for query execution.
+For example, the estimates can be used to warn users about a query that asks for too many results.
+
+### Usage of query cost estimation
+
+To get estimates for the number of result rows for a given query and statistics object, use the
+`estimate_query_result_cardinality()` function in
```
LWprogramming: I think this is going to be really helpful! ^^ Pointing the user to the entry point is something I find really helpful

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+
+- `get_class_count(class)`: 
+  Given the name of a vertex or edge class, return the number of objects of that type. If the given
+  class is an interface, the objects that inherit from this class should also be included. So if
+  `get_class_count('Entity')` is called, it should also include the number of 'Animal' objects,
+  'Food' objects, etc.
+
+### Optional statistics
+
+- `get_vertex_edge_vertex_count(source vertex class, edge class, target vertex class)`:
```
LWprogramming: Nit: add missing underscores e.g. source_vertex_class here and below

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+  Given the names of two vertex classes and an edge class, return the number of edges of type `edge
+  class` whose source endpoint is of type `source vertex class`, and the destination endpoint is of
+  type `target vertex class`. For example, `get_vertex_edge_vertex_count('Animal', 'Entity_Related',
+  'BirthEvent')` should provide the number of 'Entity_Related' edges that start from a vertex of
```
LWprogramming: Nit: surround things with backticks as needed

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+  type 'Animal' and end at a vertex of type 'BirthEvent'.
+- `get_distinct_field_values_count(vertex class, field name)`:
+  Given a vertex class, and a property field on that vertex class, return how many distinct values  
+  the given property field has. E.g. If `get_distinct_field_values_count('Animal', 'name')` is
+  called, and there are 3 animals, two named 'Alice', and one called 'Bob', then the result should
```
LWprogramming: Nit: backticks again (you get the idea, these are just so it's easier to see where the missing backticks are :) )

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+### Usage of query cost estimation
+
+To get estimates for the number of result rows for a given query and statistics object, use the
+`estimate_query_result_cardinality()` function in
+`graphql_compiler/cost_estimation/cardinality_estimator.py`.
+
+### Statistics
+
+Before estimates can be generated, you need to provide statistics about the graph data. The
+applicable classes are in `graphql_compiler/cost_estimation/statistics.py`. Currently, two classes
```
LWprogramming: This is also helpful! The user can see where the statistics are computed-- do you want to discuss a bit about what sort of assumptions go into computing these statistics here?

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+  statistics. This allows users to create their own implementations for the statistic class.
+- `LocalStatistics`, an implementation of `Statistics`, where the statistics are provided during
+  initialization, and can't be modified thereafter.
+
+Currently, three statistics are supported. `get_class_count()` must be provided for cardinality
+estimation. The other two are optional, as they can be roughly approximated.
+
+### Required statistics
+
+- `get_class_count(class)`:
```
pmantica1: This can be best shown through a code example than described here.

README.md
pmantica1: same as above.

README.md
```diff
@@ -1105,6 +1110,59 @@ However, the following rules of thumb are useful to keep in mind:
...
+To get estimates for the number of result rows for a given query and statistics object, use the
+`estimate_query_result_cardinality()` function in
+`graphql_compiler/cost_estimation/cardinality_estimator.py`.
+
+### Statistics
+
+Before estimates can be generated, you need to provide statistics about the graph data. The
+applicable classes are in `graphql_compiler/cost_estimation/statistics.py`. Currently, two classes
+are implemented:
+- `Statistics`, an abstract base class that describes the interface the cost estimator expects for
```
pmantica1: Since LocalStatistics is the only implementation right now let's remove the Statistics base class mention from the README.md.

README.md
ReaLNeroM: I removed this part.

README.md
ReaLNeroM: Removed this.

README.md
ReaLNeroM: I don't think this should be in the README. There are many reasons: the estimator's implementation is only prototypical; the current implementation is well-documented, but it's not really simple; and most important of all: the user shouldn't need to care how it's implemented.

pmantica1: I disagree. If the user does not have an intuition for how it works, he or she will not trust it.

README.md
ReaLNeroM: I don't see how the underscores add anything of value?

pmantica1: Code presented in a README.md should still be syntactically correct. It currently isn't.

README.md
ReaLNeroM: The user doesn't (and shouldn't have to!) care about the assumptions, as long as it works.

README.md
ReaLNeroM: I'll also include an example. The current documentation on statistics can't hurt, can it?

pmantica1: It can hurt. The README.md should be as concise as possible. Furthermore, this section should fit the style of the rest of the README.md. I don't think that class documentation is mentioned anywhere else, (and for good reason).

README.md
ReaLNeroM: Included an example usage of this. Let me know what you think!

README.md
```diff
@@ -56,6 +56,11 @@ For a more detailed overview and getting started guide, please see
      * [\__typename](#__typename)
      * [\_x_count](#_x_count)
   * [The GraphQL schema](#the-graphql-schema)
+  * [Query cost estimation](#query-cost-estimation)
```
obi1kenobi: I would move this to under the `SQL` section. It's hard to talk about cost estimation when we haven't even described what the execution model is.

Also, consider using `Title Case`. I know we don't consistently use it, but we should -- all of these are section and subsection titles.

README.md
ReaLNeroM: In the readme, at no point do we describe how graphql compilation works (GraphQL->AST->IR->database language). I don't see why a user should understand how cost estimation works and how the IR is used, when they don't even know what IR is.

I've noticed we disagree on this. I'll speak with other members of the team, and see what's the best course of action.

LWprogramming: How about describing details like these in the comments, and in the README tell the user to read the comments in X file if they want to know more details about how the statistics are calculated? This way we don't put too much stuff in the README but it's still clear exactly where they can get an intuition of how things work.

README.md
LWprogramming: As we move to read the docs formatting, is it possible that all of this stuff will get moved to there (where conciseness matters less) and the README ends up much shorter than it is now? In that case maybe we don't need to worry about being concise as much.

README.md
ReaLNeroM: I'm not so sure. Creating your own implementation of Statistics requires implementing three functions, and LocalStatistics is not what you need.

pmantica1:
Can you please fix this branch so that it points to master?

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
...
+    }
+}
+'''
+params = {'name_params': ['Albert', 'Betty', 'Charlie']}
+
+# This dictionary describes the number of Animal vertices, Animal_ImportantEvent
+# edges, and FeedingEvent vertices.
+class_counts = {
+    'Animal': 6,
+    'Animal_ImportantEvent': 20,
```
pmantica1: I think it's best to put this class last and not in the middle to make it clear that is intrinsically different than the other two classes. (It refers to an edge not a vertex).

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
...
+vertex_edge_vertex_counts = {
+    ('Animal', 'Animal_ImportantEvent', 'FeedingEvent'): 10,
+}
+
+# We include statistics for class_counts and vertex_edge_vertex_counts.
+statistics_object = LocalStatistics(class_counts,
+                                    vertex_edge_vertex_counts=vertex_edge_vertex_counts)
+
+print(estimate_query_result_cardinality(schema_graph, statistics_object, graphql_query, params))
+# The estimator expects approximately 5.0 rows in the result.
```
pmantica1: If you don't explain the logic of how you got to the result, then saying that "The estimator expects approximately 5.0 rows in the result." doesn't convey any useful information to the reader.

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
...
+
+```python
+from graphql_compiler.cost_estimation.cardinality_estimator import estimate_query_result_cardinality
+from graphql_compiler.cost_estimation.statistics import LocalStatistics
+
+# The cost estimator needs the schema graph for information about the classes in the schema.
+# Get schema graph from hypothetical Animals database.
+schema_graph = obtain_schema_graph_from_animals_database()
+
+# This is the query and parameters that the estimator will produce a cardinality estimate for.
```
pmantica1: Make the comment here and the ones below more concise. Maybe aim to make them 1 line long because they should be as skimmable as possible. I would rewrite this as something along the lines of "Write GraphQL query".

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
   Alternatively, the entire OrientDB class and all edges that may point to it may be omitted
   entirely from the GraphQL schema.
 
+## Query Cost Estimation
```
pmantica1: The README.md lacks structure right now and there is no great place to put this section in, but this doesn't seem this is the right place. I would put it directly above the miscellaneous section for now.

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
...
+
+Some GraphQL queries executed over a database may use a lot of memory and/or have a long execution
+time. In such cases, it's better to warn the user, and suggest they revise their query. So it's
```
pmantica1: nit: You write "it's better to warn the user ...", but don't mention what it is better than. 

How about "it's better to warn the user and suggest they revise their query than to compile and run the query."?

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
   Alternatively, the entire OrientDB class and all edges that may point to it may be omitted
   entirely from the GraphQL schema.
 
+## Query Cost Estimation
+
+Some GraphQL queries executed over a database may use a lot of memory and/or have a long execution
```
pmantica1: nit: The problem is not that the queries use a lot of memory. The problem is that they use too much memory, (by some definition of "too much". Therefore, consider rewriting this to "Some GraphQL queries executed over a database use too much memory and/or take too long to run."

README.md
```diff
@@ -1105,6 +1108,89 @@ However, the following rules of thumb are useful to keep in mind:
...
+time. In such cases, it's better to warn the user, and suggest they revise their query. So it's
+worthwhile to estimate a query's cost (in terms of time taken and memory used) using statistics
```
pmantica1: I would delete this sentence since it is a bit repetitive and start mentioning the query cost estimator here.

PR #420, opened 2019-07-22
Add readthedocs directory and contributing info

pmantica1:
This is branch is essentially the result of running sphinx-quickstart
 from https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html and adding some documentation of how to interact with readthedocs in the GraphQL compiler.


CONTRIBUTING.md
```diff
@@ -100,4 +100,23 @@ Then, re-run
 ```
 pipenv sync --dev
 ```
-to install the relevant dependencies.
\ No newline at end of file
+to install the relevant dependencies.
```
pmantica1: This line seems identical to me that the line above. I don't understand why git is not marking them as equal.

pmantica1: I think it has to do with the issue that this file was originally missing a "newline" at the end.

LWprogramming: Talked to Vlad-- seems like he had a similar thing happen to him and he said it was a newline thing, yes.

CONTRIBUTING.md
```diff
@@ -100,4 +100,23 @@ Then, re-run
...
+
+## Read the Docs
```
pmantica1: I don't like the concept of modifying existing documentation while adding Read the Docs, but I don't know where else to put this.

obi1kenobi:
Should this be merging into `master` instead?

pmantica1:
> Should this be merging into `master` instead?

No, it should not.  I would like no changes to this documentation branch coming from updates to the master branch for now.  I also do not like that since we are constantly updating master with new commits, then we always have to run CI again before merging into master.

docs/source/conf.py
```diff
@@ -0,0 +1,52 @@
...
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'GraphQL Compiler'
+copyright = '2019-Present Kensho Technologies, LLC'
+author = 'Predrag Gruevski, Pedro Mantica, Amartya Shankha Biswas, and Jeremy Meulemans'
```
pmantica1: Bojan is not included in the authors.md file yet.

docs/source/conf.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'GraphQL Compiler'
+copyright = '2019-Present Kensho Technologies, LLC'
```
pmantica1: From the html file, this gets displayed as "2019-Present Kensho Technologies, LLC.", (with a dot at the end). If we add a dot after LLC here it will look like "2019-Present Kensho Technologies, LLC.."

docs/source/conf.py
obi1kenobi: @bojanserafimov ^ if you'd like to fix that omission, you're welcome to :)

docs/source/conf.py
obi1kenobi: Can we make the `Present` bit not be capitalized?

obi1kenobi: Add this comment into the file itself, please. Otherwise, when we look at this in 6 months, we'll try to "fix" it again.

PR #421, opened 2019-07-22
Update MATCH emit output tests

LWprogramming:
I decided to separate this from Cypher and Gremlin emit output tests because this isn't a direct find-and-replace job-- some tests seem to be giving output that doesn't make sense, so I'm correcting them here too.

* Implement the change requested [here](https://github.com/kensho-technologies/graphql-compiler/pull/379#discussion_r304515049) for MATCH

* Make these emit output tests use the test schema

* Correct any incorrect tests from before


obi1kenobi:
Seems fine so far. Happy to merge as-is if you think it's ready.

LWprogramming:
One more test for here (that is already in Gremlin but not here). Also will take a look at the failing CI, but I'll mark it when it's ready so you don't have to keep checking :)

graphql_compiler/tests/test_emit_output.py
```diff
@@ -3,10 +3,13 @@
 
 from graphql import GraphQLString
 
-from ..compiler import emit_gremlin, emit_match
+
```
obi1kenobi: this will make the linter upset

PR #422, opened 2019-07-22
Update gremlin emit output tests

LWprogramming:
* Address comment [here](https://github.com/kensho-technologies/graphql-compiler/pull/379#discussion_r304515049)

* Make these emit output tests use the test schema

* Add in missing tests (i.e. here we had a test for MATCH but not its counterpart for Gremlin even though both backends handle it).


obi1kenobi:
If you can fix the merge conflicts here, this is fine by me.

graphql_compiler/tests/test_emit_output.py
```diff
@@ -3,10 +3,13 @@
 
 from graphql import GraphQLString
 
-from ..compiler import emit_gremlin, emit_match
+
```
obi1kenobi: the linter will be upset at the two blank lines here

PR #424, opened 2019-07-23
Remove neo4j driver encryption to deal with driver issue

pmantica1:


obi1kenobi:
The new error is different, though -- it used to be an SSL error, and now it's just a plain connection failure.
```
ServiceUnavailable: Failed to read any data from server Address(host='localhost', port=7687) after connected
```

This could just mean that the tests were ready to run against Neo4j before the Neo4j process was ready to serve data. Let's try throwing some retries in there and see if the problem is at least diminished?

pmantica1:
> The new error is different, though -- it used to be an SSL error, and now it's just a plain connection failure.
> 
> ```
> ServiceUnavailable: Failed to read any data from server Address(host='localhost', port=7687) after connected
> ```
> 
> This could just mean that the tests were ready to run against Neo4j before the Neo4j process was ready to serve data. Let's try throwing some retries in there and see if the problem is at least diminished?

We already retry 20 times with one second between each retry. I told Leon to submit a PR removing neo4j integration tests until we find a solution to this issue.

LWprogramming:
Given that we've merged the neo4j client back in and we don't have any problems, we don't need this PR anymore right?

PR #427, opened 2019-07-23
Improve duplicate field error message

pmantica1:


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -173,7 +173,9 @@ def _get_fields(ast):
             # If we ever allow repeated field names,
             # then we have to change the Location naming scheme to reflect the repetitions
             # and disambiguate between Recurse and Traverse visits to a Location.
-            raise GraphQLCompilationError(u'Encountered repeated field name: {}'.format(name))
+            raise GraphQLCompilationError(u'Encountered repeated field name: {}. To use '
+                                          u'multiple directives on a field, write the other '
+                                          u'directives directly below the first one.'.format(name))
```
obi1kenobi: I think this is solving a real problem that is worth solving. We just have to be mindful that our solution doesn't cause more damage than good.

There are other situations where this error can happen, it's not just multiple directives. In principle, the following will raise this error as well:
```
{
    Animal {
        name
        name
        alias @output(out_name: "alias")
    }
}
```

If I were to write the above, and then see this error message in response, that would be very confusing.

Also, "directly below" is misleading and unnecessarily strict -- not all editors have "directly below" as an option (e.g. consider the default Flask-JSONRPC frontend). How about "adjacent to" instead?

bojanserafimov: This advice applies only to duplicate property fields, and is very misleading for duplicate vertex fields. Can you distinguish which one it is? Also, say `after` the first one, not `below`.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -173,7 +173,15 @@ def _get_fields(ast):
...
-            raise GraphQLCompilationError(u'Encountered repeated field name: {}'.format(name))
+            raise GraphQLCompilationError(u'Encountered repeated field name: {}. If you are '
+                                          u'attempting to specify multiple directives on a single '
+                                          u'property field, one way to do so is to follow the '
+                                          u'format: propertyField directive1 directive2 ...'
+                                          u'However, note that all whitespace sequences, '
+                                          u'(e.g. spaces, tabs and newline characters),  are '
+                                          u'treated in the same way as delimiters in GraphQL, '
+                                          u'so feel free to space out directives in a different '
+                                          u'fashion.'.format(name))
```
bojanserafimov: This is way too technical, just to explain that if the line is too long they can break it up.

graphql_compiler/compiler/compiler_frontend.py
bojanserafimov: Can you actually distinguish with an if statement and separate error message? I still think this message is adding confusion when there's duplicate vertex fields.

PR #429, opened 2019-07-23
Implement cross schema edges in merge_schemas

qqi0O0:
New tests covering the behavior of cross schema edges exist, but not in this PR


graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,6 +138,89 @@ def _get_basic_schema_ast(query_type):
     return blank_ast
 
 
+def _merge_single_schema(merged_schema_ast, name_to_schema_id, scalars, directives,
+                         current_schema_id, current_ast):
+    """Merge current_ast into merged_schema_ast, update all records accordingly.
+
+    Args:
+        merged_schema_ast: Document; modified by this function as current_ast is incorporated
```
pmantica1: nit: why the ";" ? (Specially since the later part is not a full sentence).

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,6 +138,89 @@ def _get_basic_schema_ast(query_type):
     return blank_ast
 
 
+def _merge_single_schema(merged_schema_ast, name_to_schema_id, scalars, directives,
+                         current_schema_id, current_ast):
+    """Merge current_ast into merged_schema_ast, update all records accordingly.
```
pmantica1: It seems to me that "update all records accordingly" isn't really saying anything.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,6 +138,89 @@ def _get_basic_schema_ast(query_type):
...
+
+    Args:
+        merged_schema_ast: Document; modified by this function as current_ast is incorporated
+        name_to_schema_id: Dict[str, str], mapping type name to id of the schema that the
+                           type is from; modified by this function
+        scalars: Set[str], names of all scalars in the merged_schema so far; potentially
```
pmantica1: Are you sure your use of ; is grammatically correct?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -248,3 +308,274 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+        scheam_ast: Document; modified by this function
+        name_to_schema_id: Dict[str, str], mapping type name to id of the schema that the
+                           type is from
+        scalars: Set[str], names of all scalars in the merged_schema so far
+        cross_schema_edges: List[CrossSchemaEdgeDescriptor], containing all edges connecting
+                            fields in multiple schemas to be added to the merged schema
+        query_type: str, name of the query type in the merged schema
+
+    Raises:
+        - SchemaNameConflictError if some cross schema edge name causes a name conflict with
```
pmantica1: nit: if any

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,19 +21,41 @@
 )
 
 
-def merge_schemas(schema_id_to_ast):
-    """Check that input schemas do not contain conflicting definitions, then merge.
+CrossSchemaEdgeDescriptor = namedtuple(
+    'CrossSchemaEdgeDescriptor', (
+        'edge_name',  # str, name used for the corresponding in and out fields
+        'outbound_side',  # FieldReference for the outbound (source) field
+        'inbound_side',  # FieldReference for the inbound (sink) field
```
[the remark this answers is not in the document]
pmantica1: Also let's stick to one terminology for variable names, functions and error messages. Either outbound/inbound or source/destination.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -62,61 +88,12 @@ def merge_schemas(schema_id_to_ast):
...
-            )):
-                _process_generic_type_definition(
-                    new_definition, current_schema_id, scalars, name_to_schema_id, merged_schema_ast
-                )
-            else:  # All definition types should've been covered
-                raise AssertionError(
-                    u'Missed definition type: "{}"'.format(type(new_definition).__name__)
-                )
+        current_ast = deepcopy(current_ast)
+        _merge_single_schema(merged_schema_ast, name_to_schema_id, scalars, directives,
```
pmantica1: nit: let's change name_to_schema_id to type_name_to_schema_id here and in other relevant places.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -248,3 +308,274 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+def _check_field_types_are_matching_scalars(type_name_to_definition, scalars, cross_schema_edge):
+    """Check that stitched fields in a cross schema edge are of the same scalar type.
+
+    It is also legal for fields to be of a NonNull wrapped scalar type.
+
+    Args:
+        type_name_to_definition: Dict[str, (Interface/Object/Union)TypeDefinition], mapping
+                                 name of types to their definitions
+        scalars: Set[str], names of all scalars in the merged_schema so far
+        cross_schema_edge: CrossSchemaEdge, the edge that we check the validity of
```
pmantica1: nit CrossSchemaEdge namedtuple. When you mention that a type is a namedtuple, you give the reader an intuition of how to interact with it.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,19 +21,41 @@
 )
 
 
-def merge_schemas(schema_id_to_ast):
-    """Check that input schemas do not contain conflicting definitions, then merge.
+CrossSchemaEdgeDescriptor = namedtuple(
+    'CrossSchemaEdgeDescriptor', (
+        'edge_name',  # str, name used for the corresponding in and out fields
+        'outbound_side',  # FieldReference for the outbound (source) field
```
pmantica1: nit: FieldReferenced namedtuple. (I know that it's a bit obvious that it is a namedtuple here, but just to be consistent with the comment I left below).

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,19 +21,41 @@
...
+)
+
+
+CrossSchemaEdgeDescriptor.__new__.__defaults__ = (False,)
+
+
+FieldReference = namedtuple(
+    'FieldReference', (
+        'schema_id',  # str, identifier for the schema of the field
+        'type_name',  # str, name of the type of the field
```
pmantica1: nit: How about `vertex_name`. `type_name` seems to vague to me.

graphql_compiler/schema_transformation/merge_schemas.py
pmantica1: What do you think of renaming _merge_single_schema to _accumulate_types ?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,19 +21,41 @@
...
 
-    Note that the output AST will share mutable objects with input ASTs.
-
     Args:
-        schema_id_to_ast: OrderedDict[str, Document], where keys are names/identifiers of schemas,
-                          and values are ASTs describing schemas. The ASTs will not be modified
-                          by this funcion
+        schema_id_to_ast: OrderedDict[str, Document], where keys are names/identifiers of
+                          schemas, and values are ASTs describing schemas. The ASTs will not
+                          be modified by this funcion
```
pmantica1: nit: function not funcion

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,19 +21,41 @@
...
+    'CrossSchemaEdgeDescriptor', (
+        'edge_name',  # str, name used for the corresponding in and out fields
+        'outbound_side',  # FieldReference for the outbound (source) field
+        'inbound_side',  # FieldReference for the inbound (sink) field
+        'out_edge_only',  # bool, defaults to False, whether to not add the in-edge
+    )
+)
+
+
+CrossSchemaEdgeDescriptor.__new__.__defaults__ = (False,)
```
pmantica1: I am against the choice of having out_edge_only default to False. I am in favor of user the having the choice between building a one-directional vs bidirectional edge through the out_edge_only parameter. However, I think that it important that this choice is as explicit as possible so that the user is conscious of what he or she is doing because adding unintended edges seems like a very plausible and detrimental bug.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -248,3 +308,274 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+    if field_type_names[0] != field_type_names[1]:  # fields return different scalars
+        raise InvalidCrossSchemaEdgeError(
+            u'The outbound and inbound fields of edge "{}" are of different types, '
+            u'"{}" and "{}"; they are expected to be of the same scalar type.'.format(
+                cross_schema_edge, field_type_names[0], field_type_names[1]
+            )
+        )
+
+
+def _add_edge_field(source_type_node, sink_type_node, source_field_name, sink_field_name,
```
pmantica1: I might be wrong but it doesn't seem that you are considering inheritance in this function. If an interface has an vertex, then all of the objects that implement it must also have it. If you declare an edge with an union type destination and out_field_only=False, then I would expect that each object in that union also gains a vertex field. Finally, I've never seen union with fields in a GraphQL schema, so I am kind of wary of adding field definitions for them in the AST.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -62,61 +88,12 @@ def merge_schemas(schema_id_to_ast):
...
-                ast_types.UnionTypeDefinition,
-            )):
-                _process_generic_type_definition(
-                    new_definition, current_schema_id, scalars, name_to_schema_id, merged_schema_ast
-                )
-            else:  # All definition types should've been covered
-                raise AssertionError(
-                    u'Missed definition type: "{}"'.format(type(new_definition).__name__)
-                )
+        current_ast = deepcopy(current_ast)
```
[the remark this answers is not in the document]
qqi0O0: Since now the addition of cross schema edges actually modify the input ASTs, some form of copying is needed. The options are to deepcopy at the beginning, or to make shallow copies whenever needed, which is far faster but somewhat more complex. I asked Predrag which route to take and he said to start with deepcopy and optimize later.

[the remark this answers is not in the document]
qqi0O0: I do think it's probably necessary to change to doing shallow copies though since merge_schemas can take in huge schemas. Will work on that.

graphql_compiler/schema_transformation/merge_schemas.py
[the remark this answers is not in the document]
qqi0O0: Outbound/inbound and source/sink are actually different concepts, confusingly enough. If we're currently adding a field to the outbound side, the outbound side is the source and the inbound side is the sink. If we're currently adding a field to the inbound side, the inbound side is the source and the outbound side is the sink. That's how the stitch directive works at least.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -248,3 +308,274 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+                    edge_name, direction):
+    """Add one direction of the specified edge as a field of the source type.
+
+    Args:
+        source_type_node: (Interface/Object/Union)TypeDefinition, where a new field representing
+                          one direction of the edge will be added; modified by this function
+        sink_type_node: (Interface/Object/Union)TypeDefinition, representing the other end of
+                        the edge
+        source_field_name: str, name of the source side field that will be stitched
+        sink_field_name: str, name of the sink side field that will be stitched
```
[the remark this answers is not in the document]
qqi0O0: That's a fair point, but the `@stitch` directive uses the terminology of source_field and sink_field. See `get_extension_schema` in experimental/schema_splitting/splitting_demo.py

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: That's a good point, having the default value makes it less obvious that there is a choice. I'll make that change.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: Ah interesting, I didn't think very carefully about special treatment for adding an edge to an Interface or a Union. So we don't want to add edges to Union types, and when we add edges to Interface types we want to add that same edge to every concrete Object inheriting the Interface? I'll work on that.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -248,3 +308,274 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+        ('out', cross_schema_edge.outbound_side),
+        ('in', cross_schema_edge.inbound_side),
+    ):
+        type_name = field_reference.type_name
+        field_name = field_reference.field_name
+
+        fields = type_name_to_definition[type_name].fields  # List[FieldDefinition]
+        field_type = None
+        for field in fields:
+            if field.name.value == field_name:
```
qqi0O0: This is bugged, will fix

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: I wanted to express that `name_to_schema_id`, `scalars`, and `directives` are all potentially getting modified to reflect information in the new schema ast. I can just remove that part of the sentence if it's not useful.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: There are a lot of places in this file where `type` or `type_name` is used, and `vertex` isn't used anywhere yet. I've considered changing to using the graphql compiler's flavor of terminology, but I think a lot of variable names and documentation would need to be changed together. I'll work on this after resolving more immediate concerns.

qqi0O0: I will change the comment here for clarify for now.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: Very possibly it's not grammatical. I'll change the ; to periods and change sentence fragments to full sentences.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -48,79 +69,33 @@ def merge_schemas(schema_id_to_ast):
           same name as the type that it queries, if the schema contains type extensions or
           input object definitions, or if the schema contains mutations or subscriptions
         - SchemaNameConflictError if there are conflicts between the names of
-          types/interfaces/enums/scalars, or conflicts between the definition of directives
-          with the same name
+          types/interfaces/enums/scalars or fields, or conflicts between the definition of
+          directives with the same name
+        - InvalidCrossSchemaEdgeError if some cross schema edge provided lies within one schema,
```
LWprogramming: or if edges clash. For instance, let `cross_schema_edges` contain a 
```python
CrossSchemaEdgeDescriptor(
    edge_name = 'example_edge',
    outbound_side = FieldReference(
        schema_id = 'first',
        type_name = 'Human',
        field_name = 'id',
        ),
    inbound_side = FieldReference(
        schema_id = 'second',
        type_name = 'Person',
        field_name = 'identifier',
    ),
    out_edge_only=False,
)
```
Then `merge_schema` will attempt to create a new field for `Person` named `in_example_edge`. If Person already has a field `in_example_edge`, we have a clash.

Another way to generate a clash is if we do something like this but there are multiple types with the same edge name (e.g. type `City` might have an edge named `city_in` to both `Country` and `Region` types).

^^ It's up to you how much detail you want to include here, but I'd explicitly give an example of how a clash can happen since I was a bit confused by what clashing meant at first while reading through the tests.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -48,79 +69,33 @@ def merge_schemas(schema_id_to_ast):
...
-        # Since query_type was taken from the schema built from the input AST, the query type
-        # should never be not found.
-        if new_query_type_fields is None:
-            raise AssertionError(u'Query type "{}" field definitions unexpected not '
-                                 u'found.'.format(current_query_type))
-        # Note that as field names and type names have been confirmed to match up, and types
-        # were merged without name conflicts, query type fields can also be safely merged.
-        query_type_index = 1  # Query type is the second entry in the list of definitions
-        merged_schema_ast.definitions[query_type_index].fields.extend(new_query_type_fields)
+    _add_cross_schema_edges(merged_schema_ast, type_name_to_schema_id, scalars,
```
LWprogramming: Are we missing the [type equivalence hints](https://github.com/kensho-technologies/graphql-compiler/pull/429/files#diff-3931d7bec621ab228f4814115c364a90R330) argument?

graphql_compiler/schema_transformation/merge_schemas.py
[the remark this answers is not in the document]
qqi0O0: I think this optimization will probably live in another PR though.

graphql_compiler/schema_transformation/merge_schemas.py
LWprogramming: never mind-- was viewing an out-of-date file

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,14 +151,100 @@ def _get_basic_schema_ast(query_type):
...
 def _process_directive_definition(directive, existing_directives, merged_schema_ast):
     """Compare new directive against existing directives, update records and schema.
 
     Args:
         directive: DirectiveDefinition, an AST node representing the definition of a directive
         existing_directives: Dict[str, DirectiveDefinition], mapping the name of each existing
-                             directive to the AST node defining it; modified by this function
-        merged_schema_ast: Document, AST representing a schema; modified by this function
+                             directive to the AST node defining it. It is modified by this
+                            function
```
LWprogramming: nit: indentation +1 space

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +315,330 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+    if such a union type is specified by type_equivalence_hints.
+
+    New vertex fields will be added to not only each vertex specified by the cross schema
+    edge, but to all of their subclass vertices as well.
+
+    For examples demonstrating the above behaviors, see tests in test_merge_schemas.py that
+    involve subclasses.
+
+    Args:
+        scheam_ast: Document. It is modified by this function
```
LWprogramming: nit `schema_ast`

graphql_compiler/schema_transformation/merge_schemas.py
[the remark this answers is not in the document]
pmantica1: Ok makes sense.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +315,330 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+    # Error if the type is nonexistent (includes if type is an enum or scalar)
+    # TODO: Consider improving error message if type is a union or enum or scalar
+    if type_name not in type_name_to_definition:
+        raise InvalidCrossSchemaEdgeError(
+            u'Type "{}" specified in the field reference "{}" is not found '
+            u'in the merged schema.'.format(type_name, field_reference)
+        )
+
+    # Error if the type is in a wrong or nonexistent schema
+    if type_name_to_schema_id[type_name] != schema_id:
```
LWprogramming: check if type_name is in type_name_to_schema_id at all?

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +315,330 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+        sink_field_name: str, name of the sink side field that will be stitched
+        edge_name: str, name of the edge that will be used to name the new field
+        direction: str, either 'in' or 'out'
+
+    Raises:
+        - SchemaNameConflictError if the new cross schema edge name causes a name conflict with
+          existing fields, or fields created by previous cross schema edges
+    """
+    type_fields = source_type_node.fields
+    new_edge_field_name = direction + '_' + edge_name
```
LWprogramming: worth validating that `direction in ('in', 'out')`

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: type_name_to_definition.keys() is a subset of type_name_to_schema_id.keys(), so the previous if statement actually ensures type_name is in type_name_to_schema_id. 

The naming is definitely not good here. type_name_to_definition only contains objects and interfaces, while type_name_to_schema_id also contains unions and enums in addition. Not sure how to change variable names to succinctly describe these differences.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +315,330 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+                                inbound_side.field_name, outbound_side.field_name, edge_name, 'in')
+
+
+def _check_cross_schema_edge_is_valid(type_name_to_definition, type_name_to_schema_id, scalars,
+                                      cross_schema_edge):
+    """Check that the edge crosses schemas and has valid field references of correct types.
+
+    Args:
+        type_name_to_definition: Dict[str, (Interface/Object/Union)TypeDefinition], mapping
+                                 name of types to their definitions
```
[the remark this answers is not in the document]
qqi0O0: Good catch, I forgot to change the documentation upon changing the code to disallow edges to unions.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -1,38 +1,72 @@
...
 )
 
 
-def merge_schemas(schema_id_to_ast):
-    """Check that input schemas do not contain conflicting definitions, then merge.
+CrossSchemaEdgeDescriptor = namedtuple(
+    'CrossSchemaEdgeDescriptor', (
+        'edge_name',  # str, name used for the corresponding in and out fields
+        'outbound_side',  # FieldReference namedtuple for the outbound (source) field
+        'inbound_side',  # FieldReference namedtuple for the inbound (sink) field
```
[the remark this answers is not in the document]
qqi0O0: Thanks for catching the bad comments implying out/in are the same as source/sink. local/remote does feel more clear than source/sink to me, we can ask for some opinions on whether it's worth to change the terminology.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: I'll add a little more to note the fact that field name conflicts can be due to fields created by cross schema edges.

graphql_compiler/schema_transformation/subclass.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2017-present Kensho Technologies, LLC.
+from graphql.type.definition import GraphQLObjectType, GraphQLUnionType
+import six
+
+
+# NOTE: taken from bojan's macro system branch
+
+
+def _add_transitive_closure(graph):
```
obi1kenobi: All of this is fine for now, but in the immediate future we may move away from type equivalence hints and toward schema graph, since that API is stabilized and available. It's fine to save that for the next PR, just giving you a heads-up.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,14 +152,100 @@ def _get_basic_schema_ast(query_type):
...
+
+    # Concatenate all query type fields.
+    # Since query_type was taken from the schema built from the input AST, the query type
+    # should never be not found.
+    if new_query_type_fields is None:
+        raise AssertionError(u'Query type "{}" field definitions unexpected not '
+                             u'found.'.format(current_query_type))
+    # Note that as field names and type names have been confirmed to match up, and types
+    # were merged without name conflicts, query type fields can also be safely merged.
+    query_type_index = 1  # Query type is the second entry in the list of definitions
```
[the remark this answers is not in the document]
qqi0O0: This is guaranteed to be the case because _get_basic_schema_ast builds a nearly blank schema whose second entry is the query type. The merged schema AST starts off being that basic schema, and all future additions just append more type definitions to the end. There is a doc string in _get_basic_schema_ast that says the second entry is the query type, but I'll add another comment here explaining the reason since it's easy to miss.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +316,348 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+        type_name = field_reference.type_name
+        field_name = field_reference.field_name
+
+        fields = type_name_to_definition[type_name].fields  # List[FieldDefinition]
+        field_type = None
+        for field in fields:
+            if field.name.value == field_name:
+                field_type = field.type
+                while isinstance(field_type, ast_types.NonNullType):  # strip NonNull
+                    field_type = field_type.type
```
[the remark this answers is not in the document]
qqi0O0: For `NonNullType` in the language/ast.py file, it's `.type`. For `GraphQLNonNullType` in the type/definition.py file, it's `.of_type`. So I can't directly use that helper here.

I can make a new helper somewhere instead of inlining.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +316,348 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+
+        if isinstance(field_type, ast_types.ListType):
+            raise InvalidCrossSchemaEdgeError(
+                u'The {}bound field of cross schema edge "{}" gives a list, while it '
+                u'should be a single scalar'.format(
+                    direction, cross_schema_edge
+                )
+            )
+        elif isinstance(field_type, ast_types.NamedType):
+            if field_type.name.value not in scalars:
```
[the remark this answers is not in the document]
qqi0O0: It also includes all five built in scalars.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -238,13 +316,348 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+                    )
+                )
+        else:  # since NonNull is stripped, field_type can only be ListType or NamedType
+            raise AssertionError(
+                u'Field has missed type "{}"'.format(type(field_type).__name__)
+            )
+
+        field_type_names.append(field_type.name.value)
+
+    if field_type_names[0] != field_type_names[1]:  # fields return different scalars
```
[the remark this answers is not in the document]
qqi0O0: field_type_names is appended to twice, once in each iteration of the for loop 
    for direction, field_reference in (
        (OUTBOUND_EDGE_DIRECTION, cross_schema_edge.outbound_field_reference),
        (INBOUND_EDGE_DIRECTION, cross_schema_edge.inbound_field_reference),
    )
I'm not totally sure if it's worth doing an assertion here checking that the length of this list is 2. It's a good point to give the two values descriptive names though, I will do that.

graphql_compiler/ast_manipulation.py
```diff
@@ -95,3 +95,15 @@ def get_only_selection_from_ast(ast, desired_error_type):
                                      .format(ast_name))
 
     return selections[0]
+
+
+def get_ast_with_non_null_stripped(ast):
+    """Strip a NonNullType layer around the ast if there is one, return the underlying ast."""
```
obi1kenobi: nit: `ast` -> `AST` when writing English rather than Python :)

graphql_compiler/ast_manipulation.py
```diff
@@ -95,3 +95,15 @@ def get_only_selection_from_ast(ast, desired_error_type):
...
+    if isinstance(ast, NonNullType):
+        stripped_ast = ast.type
+        if isinstance(stripped_ast, NonNullType):
+            raise AssertionError(u'NonNullType may not be wrapped around another NonNullType, '
```
obi1kenobi: nit: Hitting this error message would be somewhat confusing – the user's reaction would be "of course it can be, I just gave you an AST that does that". I would reword this to emphasize that for this code, this situation was _unexpected_, since that is obviously true.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -333,3 +343,373 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
     merged_schema_ast.definitions.append(generic_type)
     type_name_to_schema_id[type_name] = schema_id
+
+
+def _add_cross_schema_edges(schema_ast, type_name_to_schema_id, scalars, cross_schema_edges,
+                            type_equivalence_hints, query_type):
+    """Add cross schema edges into the schema AST.
+
+    Each cross schema edge will be incorporated into the schema by adding vertex fields
+    with a @stitch directive to relevant vertex types. New fields will be named out_
```
obi1kenobi: nit: consider quotes or backticks around `out_` and `in_`, and clarifying how the "new fields" relate to the cross-schema edges. Also consider hyphenating `cross-schema`.

How about something like:
The new fields corresponding to the added cross-schema edges will have names constructed from the edge name, prefixed with "out_" on the edge's outbound side, and "in_" on the edge's inbound side.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -333,3 +343,373 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+    or in_ concatenated with the edge name.
+
+    The type of the new field will either be the type of the opposing vertex specified in
+    the cross schema edge, or the equivalent union type of the type of the opposing vertex
+    if such a union type is specified by type_equivalence_hints.
```
obi1kenobi: This is great! I didn't think of this -- nice catch!

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -333,3 +343,373 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+
+    New vertex fields will be added to not only each vertex specified by the cross schema
+    edge, but to all of their subclass vertices as well.
+
+    For examples demonstrating the above behaviors, see tests in test_merge_schemas.py that
+    involve subclasses.
+
+    Args:
+        schema_ast: Document. It is modified by this function
```
obi1kenobi: Some more explanation would be helpful here, because I think there are some (currently implicit) expectations of this schema AST -- it's not "just any" schema AST.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -333,3 +343,373 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+    _check_field_reference_is_valid(
+        type_name_to_definition, type_name_to_schema_id, union_type_names, outbound_field_reference
+    )
+    _check_field_reference_is_valid(
+        type_name_to_definition, type_name_to_schema_id, union_type_names, inbound_field_reference
+    )
+
+    if outbound_field_reference.schema_id == inbound_field_reference.schema_id:  # not cross schema
+        raise InvalidCrossSchemaEdgeError(
+            u'Edge "{}" does not cross schemas.'.format(cross_schema_edge)
```
obi1kenobi: nit: Let's tell the user which `schema_id` and which type in that schema it was defined on, so they don't have to look through thousands of lines of schema? Also, we should tell the user why not crossing schemas is a problem -- because we're trying to define cross-schema edges here.

graphql_compiler/schema_transformation/merge_schemas.py
obi1kenobi: Updating the docstring to capture this subtlety would be the next best thing. Naming is hard :)

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -333,3 +343,373 @@ def _process_generic_type_definition(generic_type, schema_id, existing_scalars,
...
+        )
+
+
+def _scalars_match(scalar_name1, scalar_name2):
+    """Return True if the two input scalars are considered to be the same, False otherwise.
+
+    For now, two input scalars are considered to be the same if they're the same scalar, if
+    one is ID and the other is String, or if one is ID and the other is Int. This may be
+    extended in the future.
+    """
```
obi1kenobi: nit: If you don't mind, could you add a full docstring with `Args` and `Returns` sections? Our general rule of thumb is that if the description doesn't fit on one line, it needs those sections.

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: `cross_schema_edge` will be printed out with all relevant information (schema_id of both sides, types and fields of both sides) since it's a namedtuple with nice builtin representation, so I don't think it's necessary to additionally repeat this information. I'll add in explanations for why not crossing schemas is bad.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -461,9 +467,10 @@ def _check_cross_schema_edge_is_valid(type_name_to_definition, type_name_to_sche
 
     Args:
         type_name_to_definition: Dict[str, (Interface/Object)TypeDefinition], mapping
-                                 name of types to their definitions
-        type_name_to_schema_id: Dict[str, str], mapping type name to the id of the schema that the
-                                type is from
+                                 name of Interfae and Object types to their definitions
```
obi1kenobi: typo: `Interfae`

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -504,15 +512,16 @@ def _check_field_reference_is_valid(type_name_to_definition, type_name_to_schema
...
```
obi1kenobi: typo: `Interfae`, possibly a bad find-and-replace?

qqi0O0: A bad copy paste of Args description, sorry about that

graphql_compiler/schema_transformation/merge_schemas.py
obi1kenobi: @pmantica1 is on vacation, and I believe this has been adequately addressed in the two weeks since this change request, so in the interest of progress I'll dismiss this review and take on responsibility for fixing any follow-up concerns @pmantica1 might have when he returns.

PR #430, opened 2019-07-24
Cross schema edge tests

qqi0O0:


graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -456,23 +479,755 @@ def test_clashing_directives(self):
...
+            }
+
+            type Person {
+              identifier: String
+            }
+        ''')
+        self.assertEqual(merged_schema_string, print_ast(merged_schema.schema_ast))
+
+    def test_multiple_fields_cross_schema_edge_descriptor(self):
+        multiple_fields_schema = dedent('''\
```
LWprogramming: Can we move this string to `input_schema_strings.py`, or is there some potential issue I've overlooked?

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -456,23 +479,755 @@ def test_clashing_directives(self):
...
+                            type_name = 'Person',
+                            field_name = 'identifier',
+                        ),
+                        out_edge_only=False,
+                    ),
+                ]
+            )
+
+    def test_invalid_edge_type_in_wrong_schema(self):
+        with self.assertRaises(InvalidCrossSchemaEdgeError):
```
LWprogramming: Doesn't this test overlap with `test_invalid_edge_nonexistent_type`? Seems like in both cases, we're looking for a nonexistent type in a schema.

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -456,23 +479,755 @@ def test_clashing_directives(self):
...
+                            type_name = 'Person',
+                            field_name = 'name',
+                        ),
+                        out_edge_only=False,
+                    ),
+                ]
+            )
+
+    def test_invalid_edge_clash_with_existing_field(self):
+        clashing_field_schema = dedent('''\
```
LWprogramming: same question as above, https://github.com/kensho-technologies/graphql-compiler/pull/430/files#r306861443

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -456,23 +479,755 @@ def test_clashing_directives(self):
...
+                            type_name = 'Person',
+                            field_name = 'identifier',
+                        ),
+                        out_edge_only=False,
+                    ),
+                ]
+            )
+
+    def test_invalid_edge_field_not_scalar_type(self):
+        not_scalar_field_schema = dedent('''\
```
LWprogramming: string to ISS

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -456,23 +479,755 @@ def test_clashing_directives(self):
...
+            }
+
+            type Person {
+              identifier: String!
+              in_example_edge: [Human] @stitch(source_field: "identifier", sink_field: "id")
+            }
+        ''')
+        self.assertEqual(merged_schema_string, print_ast(merged_schema.schema_ast))
+
+    def test_non_null_scalar_mismatch_normal_scalar(self):
```
LWprogramming: add invalid to the function name, since this raises an exception

pmantica1:
Can you please split this up somehow? Also, since you have so many tests it would be useful if you included a testing summary.

qqi0O0:
This is too large for one PR. The tests are already broken up into classes, actual PRs will be adding one class at a time.

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
qqi0O0: The actual execution of these two invalid cases ends up being the same (same flow, raises same error) through the code, but it could have potentially been the case (or perhaps will be the case in the future if the implementation changes) that types are stored and checked for differently, and not finding a schema has a different execution flow then finding a schema and not finding the specified type.

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
qqi0O0: This confusion is probably due to poor naming, since two different things are both called fields here: one, the two ends of the edge specified in the `@stitch` directive ('id', 'identifier', etc), which is how it's used in test_invalid_edge_nonexistent_field; two, the newly added fields that describe the edge ('out_example_edge', etc), which is how it's used in test_invalid_edge_clash_with_existing_field.

I think this can be fixed by calling the first case 'property field' and the second case 'vertex field'. Will change.

qqi0O0:
Merging classes of tests individually, no need for this PR.

PR #431, opened 2019-07-25
Assert that arguments is a dictionary

pmantica1:


LWprogramming:
I don't have any problem with this change in particular, but I'm wondering why we type check this argument and not others throughout the codebase? Is there a particular pattern to what we type check and what we don't, or is it just ad hoc?

bojanserafimov:
This doesn't look like the right place. Macro expansion happens before it, and will break with some weird `KeyError` first if it is passed a list instead of dict.

obi1kenobi:
Agreed with @bojanserafimov. Let's find a better place.

PR #434, opened 2019-07-25
Separate class_counts into vertex and edge counts

ReaLNeroM:
I've split class_counts into a vertex_count and edge_count statistic. This makes the code clearer. 

The downside of this is that now every test's counts had to be modified manually, resulting in a large diff. 

@reviewers: I've git grepped for "class_count" and "class count", so I've made sure traces of class_counts do not exist anymore.

### DIFF
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -116,7 +116,7 @@ def _probe_statistics_for_vertex_edge_vertex_count(
     return probe_result
 
 
-def _estimate_vertex_edge_vertex_count_using_class_count(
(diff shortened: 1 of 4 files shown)


obi1kenobi:
Are you sure this is always a net improvement in code clarity?

I'm especially curious about situations where these metrics have to be
collected from a database that is not natively a graph (e.g. SQL) and
therefore the "edge vs vertex" distinction is only semantic. For example,
look into how many-to-many relations are represented in SQL.

pmantica1:
I agree with Predrag. The SchemaGraph holds all classes in one dictionary. I don't see that much benefit from breaking convention. Also part of what Predrag meant to say junction tables https://en.wikipedia.org/wiki/Associative_entity can be represented as edges in SQL.

PR #437, opened 2019-07-26
[placeholder, do not review] Implement fold support

LWprogramming:
Don't review this PR-- this is just a placeholder because I wanted to have these changes somewhere besides my own computer.

I'll use this as a placeholder to contain all of my changes and when it's ready for review, I'll split functionality and tests into separate PRs. Until then, I'll work entirely in the `cypher_fold` branch so the CI tests can run.

Currently known bugs:

- [ ] Invalid field_type
    
    * test_filter_within_fold_scope

    * Test_filter_on_fold_scope

- [ ] Same node getting named different things

    * test_optional_and_fold
    
    * test_fold_and_optional

Other things to check:

- [ ] Check potential edge case: folds at different levels. e.g.

```
{
        Animal {
            in_Animal_ParentOf @fold {
                out_Animal_ParentOf {
                }
            }
            out_Animal_ParentOf {
                in_Animal_ParentOf @fold {
                    name @output(out_name: "spouse_and_self_names_list")
                }
            }
        }
    }
```


LWprogramming:
Design decision that I'm open to discussing:

For the GraphQL query
```graphql
{
        Animal {
            out_Animal_ParentOf @fold {
            }
        }
    }
```
we get the following Cypher query:
```cypher
            MATCH (Animal___1:Animal)
            MATCH (Animal___1)-[:Animal_ParentOf]->(Animal___1___out_Animal_ParentOf:Animal)
            WITH
              Animal___1 AS Animal___1,
              collect(Animal___1___out_Animal_ParentOf.name) AS collected_Animal___1___out_Animal_ParentOf_name
            RETURN
              Animal___1.name AS `animal_name`,
              collected_Animal___1___out_Animal_ParentOf_name AS `child_names_list`
```
Note that in the WITH clause we're collecting the name field rather than the vertices themselves. This means we need to collect every single field of that vertex separately (if we want to output multiple fields from that folded scope, e.g.  `Animal___1___out_Animal_ParentOf.uuid` as well). 

Alternatively, we could only collect the vertices in the WITH clause and extract the relevant fields in RETURN with a list comprehension, which would reduce the number of times we call collect (potentially making the query a bit easier to read?) but we'd end up needing to build all those lists anyways as well as the additional list of vertices to boot.

obi1kenobi:
Cool. So what does this buy us over the alternative? Any risks we should be
mindful of?

LWprogramming:
All the changes in this PR have been merged into master except for certain edge cases with optional and fold in certain places in the same query. Since the implementation has diverged from what we have here, I'll close this down and make a new draft PR just for that edge case.

PR #438, opened 2019-07-26
Cost estimation statistics

ReaLNeroM:
I've included the final version of the statistics class. While two new statistics have been added, tests and functionality for them will be in a separate PR for easier reviewing.

### DIFF
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -45,14 +45,13 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
 
 
 def _get_filter_selectivity(
-    schema_graph, lookup_class_counts, filter_info, parameters, location_name
+    schema_graph, statistics, filter_info, parameters, location_name
 ):
     """Calculate the selectivity of an individual filter at a given location.
 
     Args:
         schema_graph: SchemaGraph object
-        lookup_class_counts: function, string -> int, that accepts a class name and returns the
-                             total number of instances plus subclass instances
+        statistics: Statistics object
         filter_info: FilterInfo object, filter on the location being filtered
         parameters: dict, parameters with which query will be executed
         location_name: string, type of the location being filtered
@@ -106,14 +105,13 @@ def _combine_filter_selectivities(selectivities):
 
 
 def adjust_counts_for_filters(
-    schema_graph, lookup_class_counts, filter_infos, parameters, location_name, counts
+    schema_graph, statistics, filter_infos, parameters, location_name, counts
 ):
     """Adjust result counts for filters on a given location by calculating selectivities.
 
     Args:
         schema_graph: SchemaGraph object
-        lookup_class_counts: function, string -> int, that accepts a class name and returns the
-                             total number of instances plus subclass instances
+        statistics: Statistics object
         filter_infos: list of FilterInfos, filters on the location being filtered
         parameters: dict, parameters with which query will be executed
         location_name: string, type of the location being filtered
@@ -124,7 +122,7 @@ def adjust_counts_for_filters(
     """
     selectivities = [
         _get_filter_selectivity(
-            schema_graph, lookup_class_counts, filter_info, parameters, location_name
+            schema_graph, statistics, filter_info, parameters, location_name
         )
         for filter_info in filter_infos
     ]
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -51,9 +51,9 @@ class Statistics(object):
         several orders of magnitude. In such cases, this statistic should be provided.
 
         Args:
-            vertex_source_class_name: str, vertex class name.
-            edge_class_name: str, edge class name.
-            vertex_target_class_name: str, vertex class name.
+            vertex_source_class_name: str, vertex class name defined in the GraphQL schema.
+            edge_class_name: str, edge class name defined in the GraphQL schema.
+            vertex_target_class_name: str, vertex class name defined in the GraphQL schema.
 
         Returns:
             - int, count of edges of class edge_class with the two vertex classes as its endpoints
@@ -70,7 +70,7 @@ class Statistics(object):
         values using equality operators like '=', '!=', and 'in_collection'.
 
         Args:
-            vertex_name: str, name of a vertex.
+            vertex_name: str, name of a vertex defined in the GraphQL schema.
             field_name: str, name of a vertex field.
 
         Returns:
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -8,7 +8,7 @@ import six
 @six.python_2_unicode_compatible
 @six.add_metaclass(ABCMeta)
 class Statistics(object):
-    """Abstract class for statistics regarding GraphQL types, interfaces, and fields.
+    """Abstract class for statistics on classes and fields for all backend instances.
 
     For the purposes of query cardinality estimation, we need statistics to provide better
     cardinality estimates when operations like edge traversal or @filter directives are used.
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -39,7 +39,6 @@ class Statistics(object):
         """
         raise NotImplementedError()
 
-    @abstractmethod
     def get_vertex_edge_vertex_count(
         self, vertex_source_class_name, edge_class_name, vertex_target_class_name
     ):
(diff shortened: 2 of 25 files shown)


graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+from abc import ABCMeta, abstractmethod
+
+from frozendict import frozendict
+import six
+
+
+@six.python_2_unicode_compatible
+@six.add_metaclass(ABCMeta)
+class Statistics(object):
+    """Abstract class for statistics regarding GraphQL objects.
```
bojanserafimov: I don't think `regarding GraphQL objects` is clear. It seems to me that it has something to do with python objects from the `graphql-core` library.

The statistics we gather are all about `the current state of a backend instance`. That's how I'd define it. Thoughts?

ReaLNeroM: It's not ideal, since we need statistics about every class in the schema, not about a specific backend. I don't care if half the backend is SQL and the other half is something else, I need statistics about all databases in one Statistics object.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Raises:
+            AssertionError, if the count statistic for the given vertex/edge class does not exist.
+        """
+        raise NotImplementedError()
+
+
+class LocalStatistics(Statistics):
+    """Provides statistics using ones given at initialization."""
+    def __init__(
```
obi1kenobi: Any reason not to make this be on one line?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+    @abstractmethod
+    def get_class_count(self, class_name):
+        """Return how many vertex or edge instances have, or inherit, the given class name.
+
+        Args:
+            class_name: str, either a vertex class name or an edge class name defined in the
+                        GraphQL schema.
+
+        Returns:
+            - int, count of vertex or edge instances having, or inheriting, the given class name.
```
obi1kenobi: remove the leading `-`

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+    cardinality estimates when operations like edge traversal or @filter directives are used.
+    All statistics except get_class_count() are optional, so if the statistic doesn't exist, a
+    value of None should be returned.
+    """
+    def __str__(self):
+        """Return a human-readable unicode representation of the Statistics object."""
+        raise NotImplementedError()
+
+    def __repr__(self):
+        """Return a human-readable str representation of the Statistics object."""
```
obi1kenobi: remove the `str`

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@six.python_2_unicode_compatible
+@six.add_metaclass(ABCMeta)
+class Statistics(object):
+    """Abstract class for statistics on classes and fields for all backend instances.
+
+    For the purposes of query cardinality estimation, we need statistics to provide better
+    cardinality estimates when operations like edge traversal or @filter directives are used.
+    All statistics except get_class_count() are optional, so if the statistic doesn't exist, a
```
obi1kenobi: Based on this class, there are no statistics other than `get_class_count` -- nothing else is defined.

If there should be other statistics, I would define their methods and then make them return `None` as you say they should.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        raise NotImplementedError()
+
+
+class LocalStatistics(Statistics):
+    """Provides statistics using ones given at initialization."""
+    def __init__(
+        self, class_counts
+    ):
+        """Initializes statistics with the given data.
```
obi1kenobi: use imperative

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,71 @@
...
+            - int, count of vertex or edge instances having, or inheriting, the given class name.
+
+        Raises:
+            AssertionError, if the count statistic for the given vertex/edge class does not exist.
+        """
+        raise NotImplementedError()
+
+
+class LocalStatistics(Statistics):
+    """Provides statistics using ones given at initialization."""
```
obi1kenobi: Use imperative and consider rephrasing. "Given at initialization" is vague and misses the important detail -- the data is all in-memory and served from a dict, rather than e.g. being remotely stored and accessed as necessary.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: We'll add some more juicy statistics that make this more than one line easily.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Decided on ```Abstract class for statistics regarding the data in a backend instance.```

graphql_compiler/cost_estimation/statistics.py
obi1kenobi: If you define methods but then mark them `@abstractmethod` then they aren't truly optional. Did you try instantiating a class that doesn't implement those and relies on the default implementation of `return None`? You'll see that operation fail because of the `@abstractmethod`, because the default implementation is not usable.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,126 @@
...
+class LocalStatistics(Statistics):
+    """Statistics class that receives all statistics at initialization, storing them in-memory."""
+    def __init__(
+        self, class_counts, vertex_edge_vertex_counts=None,
+        distinct_field_values_counts=None
+    ):
+        """Initialize statistics with the given data.
+
+        Args:
+            class_counts: dict, str -> int, mapping vertex/edge class name to class count.
```
LWprogramming: Do we explicitly define anywhere what class count is? I hadn't looked at this code in a while and thought for a second class count was the number of classes in the schema, but then realized it actually means number of vertices or edges of that class.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -45,14 +45,13 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
-    schema_graph, lookup_class_counts, filter_info, parameters, location_name
+    schema_graph, statistics, filter_info, parameters, location_name
 ):
     """Calculate the selectivity of an individual filter at a given location.
 
     Args:
         schema_graph: SchemaGraph object
-        lookup_class_counts: function, string -> int, that accepts a class name and returns the
-                             total number of instances plus subclass instances
+        statistics: Statistics object
```
LWprogramming: why does this function require a statistics object? it doesn't seem to be used anywhere

LWprogramming: follow up: filter selectivity tests might not need an empty statistics object passed in to them either if we remove this.

OTOH if it'll serve a purpose at some later point then we can leave a note here saying that

ReaLNeroM: Answering this question is not the goal of this PR.
Luckily, there's a happy ending: once functionality for the 3rd statistic gets added, this parameter will be necessary.

graphql_compiler/cost_estimation/statistics.py
ReaLNeroM: Agreed. Class_counts was a bit of a holdover term that cost estimation used before I even started working on this, so I've been a bit reluctant to change it.
I've changed this line to 
```
            class_counts: dict, str -> int, mapping vertex/edge class name to count of
                instances of that class.
```
Thanks for your feedback!

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: Okay, I've dug through `.pylintrc` and found this:
```
    # unused-argument,         # needs some cleanup and per-line suppression,
                               # buggy / unclear how to suppress only a single function
```

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -0,0 +1,125 @@
...
+    """Statistics class that receives all statistics at initialization, storing them in-memory."""
+    def __init__(
+        self, class_counts, vertex_edge_vertex_counts=None,
+        distinct_field_values_counts=None
+    ):
+        """Initialize statistics with the given data.
+
+        Args:
+            class_counts: dict, str -> int, mapping vertex/edge class name to count of
+                instances of that class.
```
pmantica1: Please fix indentation.

PR #439, opened 2019-07-26
Cost estimation vertex edge vertex and refactoring

ReaLNeroM:
Added tests and refactoring was done for the vertex edge vertex counts statistic.
A lot of refactoring was done over `cardinality_estimator.py` as well.


pmantica1:
Can you rebase this pr to point to master instead?

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,99 @@ def _get_last_edge_direction_and_name_to_location(location):
     return edge_direction, edge_name
 
 
-def _get_parent_and_child_name_from_edge(schema_graph, child_location):
+def _get_parent_and_child_base_classnames_from_edge(schema_graph, child_location):
```
pmantica1: nit:`_get_base_parent_and_child_class_names_from_edge` reads better to me.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,99 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+        parent_base_classname = edge_element.base_out_connection
+        child_base_classname = edge_element.base_in_connection
     elif edge_direction == OUTBOUND_EDGE_DIRECTION:
-        parent_name_from_edge = edge_element.base_in_connection
-        child_name_from_edge = edge_element.base_out_connection
-    return parent_name_from_edge, child_name_from_edge
+        parent_base_classname = edge_element.base_in_connection
+        child_base_classname = edge_element.base_out_connection
+    else:
+        raise AssertionError(u'Expected edge direction to be either inbound or outbound.'
```
pmantica1: Why not "Expected edge direction to be either 'in' or 'out'." ? It communicates the same message and is just as precise.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -131,6 +131,163 @@ def test_complex_traverse(self):
...
+            schema_graph, statistics, graphql_input, params
+        )
+
+        # Vertex_edge_vertex_data tells us that no Entity_Related edges connect Animals and Events,
+        # so the result set is empty.
+        expected_cardinality_estimate = 0.0
+        self.assertAlmostEqual(expected_cardinality_estimate, cardinality_estimate)
+
+    @pytest.mark.usefixtures('snapshot_orientdb_client')
+    def test_traversal_in_reverse_direction_provided_both_statistics(self):
```
pmantica1: nit: inbound instead of reverse to consistent with the docstring.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,99 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+    if parent_name_from_location != parent_base_classname:
+        edge_counts *= (
+            float(statistics.get_class_count(parent_name_from_location)) /
+            statistics.get_class_count(parent_base_classname)
+        )
+
+    return edge_counts
 
 
 def _estimate_children_per_parent(
```
pmantica1: This function should be named something more akin to `_estimate_parent_child_edges` because that is exactly what it is doing. Estimating the number of children vertices per parent vertex is slightly different.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -89,41 +171,32 @@ def _estimate_children_per_parent(
 
     Given a parent location of type A and child location of type B, assume all AB edges are
```
pmantica1: This docstring can be a bit more general now that we are using vertex_edge_counts. (We do not need to rely on the uniformity assumption).

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Edge direction is something the user specifies, but only implicitly. For example, we don't have a field called `edge direction: ` which the user fills with either `in` or `out`. If the error message were `either 'in' or 'out'`, the user would get the impression that they've missed an `edge direction: ` field somewhere.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Refactoring this function is not the goal of this PR. 

If you'd like, create a PR for this change. I think it's going to be worthwhile to have someone else interacting with the cost_estimator, as I won't be here forever.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: We are still using an assumption for uniformity. Since there's a lot of assumptions involving uniformity, I'll describe the current one. 
We find the number of AB edges (this is where vertex_edge_vertex_counts is used). Then, we assume each A vertex has an equal number of AB edges, and we finally get that for one A vertex, there are approximately `(# of AB edges) / (# of A vertices)` edges.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
ReaLNeroM: Good point!

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Rewrote it as `_get_base_classnames_of_parent_and_child_from_edge`. Let me know what you think.

graphql_compiler/cost_estimation/cardinality_estimator.py
pmantica1: Ok nevermind it is correct as it is.

graphql_compiler/cost_estimation/cardinality_estimator.py
pmantica1: Because of the time CI integration tests take, it is not practical to be making PRs for tiny nits like this one. I don't see how submitting a PR for this nit would make me any more experienced in dealing with the query cost estimator. Even though this is not the primary goal of this pr, I don't see the harm of making this change and it's related to this PR because you are changing the function's docstring.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,103 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+    else:
+        raise AssertionError(u'Expected edge direction to be either inbound or outbound.'
+                             u'Found: edge {} with direction {}'.format(edge_name, edge_direction))
+    return outbound_vertex_name, inbound_vertex_name
+
+
+def _probe_statistics_for_vertex_edge_vertex_count(
+    statistics, query_metadata, child_location, parent_location
+):
+    """Probe for the count of edges that connect parent_location and child_location vertices."""
```
obi1kenobi: What does `probe` mean here? It feels a bit vague, and the docstring uses the same word so it's hard to figure out what this function actually does.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,103 @@ def _get_last_edge_direction_and_name_to_location(location):
...
-    return parent_name_from_edge, child_name_from_edge
+        parent_base_classname = edge_element.base_in_connection
+        child_base_classname = edge_element.base_out_connection
+    else:
+        raise AssertionError(u'Expected edge direction to be either inbound or outbound.'
+                             u'Found: edge {} with direction {}'.format(edge_name, edge_direction))
+    return parent_base_classname, child_base_classname
+
+
+def _get_outbound_and_inbound_names_from_locations(query_metadata, child_location, parent_location):
```
obi1kenobi: nit: Any particular reason for the `child_location, parent_location` order of arguments? So long as we're consistent across all functions that take those args, I guess it's fine, though that would not have been the order I would have assumed a priori.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,103 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+    B. In the general case, we estimate the statistic as
+    (# of AB edges) = (# of CD edges) * (# of A vertices) / (# of C vertices) *
+                                        (# of B vertices) / (# of D vertices).
+
+    Args:
+        schema_graph: SchemaGraph object
+        statistics: Statistics object
+        query_metadata: QueryMetadataTable object
+        child_location: BaseLocation object
+        parent_location: BaseLocation object
```
obi1kenobi: Please actually describe each of these arguments. What you have written here are type hints, not documentation.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,17 +69,103 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+
+    Returns:
+        float, estimate for (# of AB edges)
```
obi1kenobi: Please use precise language rather than relying on symbols, unless you've defined the symbol previously in the docstring. Replace `#` with `number` etc.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Merely due to consistency. 
I'd be happy to change this in another PR once this gets merged. How does that sound?

obi1kenobi: Sure. Or you could add something about "refactoring" to the title and save yourself a round-trip through the CI bot :) All the same to me.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -36,7 +36,7 @@ def _get_all_original_child_locations(query_metadata, start_location):
 
     Args:
         query_metadata: QueryMetadataTable object
-        start_location: Location object, where we're looking for children
+        start_location: Location object, where we're looking for child locations
```
ReaLNeroM: This sentence sounds really bad. I had to change it.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Since refactoring is now a part of this PR, I've changed the function name to `_estimate_edges_to_children_per_parent`.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Decided on `query` instead of `probe`.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,76 +69,174 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+        statistics, query_metadata, parent_location, child_location
+    )
+
+    if edge_counts is None:
+        edge_counts = _estimate_vertex_edge_vertex_count_using_class_count(
+            schema_graph, statistics, query_metadata, parent_location, child_location
         )
-        # pylint: enable=old-division
 
+    parent_name_from_location = query_metadata.get_location_info(parent_location).type.name
```
pmantica1: You pass in the query metadata table as arguments to your helper functions and extract the parent and child names multiple times. I think you would have an easier time documenting the code if you extracted the parent and child names here, (as well as the other query metadata you need) and passed those as arguments.

ReaLNeroM: But we also need child_location for other reasons later inside those functions! We'll need to have a `child_location` argument and a `child_vertex_class` argument, and we'll have to assert that `child_vertex_class` is THE vertex class of child location, so we actually add more code.

pmantica1: I am talking specifically about 
_estimate_vertex_edge_vertex_count_using_class_count and 
_query_statistics_for_vertex_edge_vertex_count

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,76 +69,174 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+    """Query statistics for the count of edges connecting parent and child_location vertices.
 
-    Given a parent location of type A and child location of type B, assume all AB edges are
-    distributed evenly over A vertices, so the expected number of child edges per parent vertex is
-    (# of AB edges) / (# of A vertices). If A and B are subclasses of C and D respectively, we only
-    have access to CD edges, so in general, we'll use (# of CD edges) / (# of C vertices), but since
-    we're only interested in edges to Bs, we'll scale this result by the fraction of Bs over Ds.
+    Given a parent location and a child location, there are three constraints on each edge directly
+    connecting the two:
+    1. The edge class must be the same as the target location's last traversed edge.
```
pmantica1: This is not as much a constraint as a specification of which edge class we will be using in this function. Also, the other two constraints are already mentioned in the args section of the docstring.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: Resolved offline.

graphql_compiler/cost_estimation/cardinality_estimator.py
ReaLNeroM: I'm unsure how to communicate the first constraint as a specification elsewhere. Could you help out?

The args section says the following:
```
parent_location: BaseLocation, corresponding to the location the edge traversal begins from.
child_location: BaseLocation, child of parent_location corresponding to the location the
                edge traversal ends at.
```
This communicates the constraint in a very implicit fashion. Just by reading the quoted part, I wouldn't get that it's a constraint on the edge. Explicit is better than implicit.

obi1kenobi: I'd agree with @ReaLNeroM, explicit is better than implicit and the args text is nice but insufficient. The extra few lines are worth it here.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,76 +69,174 @@ def _get_last_edge_direction_and_name_to_location(location):
...
-    parent_name_from_edge, child_name_from_edge = _get_parent_and_child_name_from_edge(
-        schema_graph, child_location
-    )
+    parent_name_from_location = query_metadata.get_location_info(parent_location).type.name
     child_name_from_location = query_metadata.get_location_info(child_location).type.name
-    if child_name_from_edge != child_name_from_location:
-        # False-positive bug in pylint: https://github.com/PyCQA/pylint/issues/3039
-        # pylint: disable=old-division
+    parent_base_class_name, child_base_class_name = \
+        _get_base_class_names_of_parent_and_child_from_edge(schema_graph, child_location)
```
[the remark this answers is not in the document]
obi1kenobi: +1 to @pmantica1, any reason not to use parens as he suggested? The Google Style Guide says to avoid `\` whenever possible and prefer parens instead.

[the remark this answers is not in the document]
obi1kenobi: I agree with @pmantica1, any reason not to follow the Google Style guide and avoid `\` by using parens?

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,76 +69,174 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+
+
+def _estimate_edges_to_children_per_parent(
+    schema_graph, statistics, query_metadata, parameters, parent_location, child_location
+):
+    """Estimate the count of edges per parent_location that connect to child_location vertices.
+
+    Given a parent location of type A and child location of type B, assume all AB edges are
+    distributed evenly over A vertices, so the expected number of child edges per parent vertex is
+    (# of AB edges) / (# of A vertices).
```
obi1kenobi: I think it would be better to use the same notation as above near line 147, `number of AB edges` instead of `# of AB edges`.

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -69,76 +69,174 @@ def _get_last_edge_direction_and_name_to_location(location):
...
+        statistics: Statistics object.
+        query_metadata: QueryMetadataTable object.
+        parameters: dict, parameters with which query will be executed.
+        parent_location: BaseLocation, corresponding to the location the edge traversal begins from.
+        child_location: BaseLocation, child of parent_location corresponding to the location the
+                        edge traversal ends at.
+
+    Returns:
+        float, expected number of edges per parent_location vertex that connect to child_location
+               vertices.
```
obi1kenobi: nit: I'd indent this to under `float` instead of under `expected`, since this section does not have names followed by a `:`

graphql_compiler/cost_estimation/cardinality_estimator.py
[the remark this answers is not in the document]
ReaLNeroM: I recall fixing this, but it's not in my local push history nor in the repo. I fully agreed with Pedro, must've been a mistake on my part.

PR #440, opened 2019-07-26
Pre cross schema edge refactor

qqi0O0:
Changes in preparation for the actual code involving cross schema edges. 

No logic involving adding cross schema edges is added. merge_schemas now has extra arguments, and _accumulate_types now lives in its own function. Empty lists are added everywhere to tests, due to the merge_schemas's new argument.

### DIFF
--- a/graphql_compiler/schema_transformation/merge_schemas.py
+++ b/graphql_compiler/schema_transformation/merge_schemas.py
@@ -224,9 +224,13 @@ def _accumulate_types(merged_schema_ast, type_name_to_schema_id, scalars, direct
     if new_query_type_fields is None:
         raise AssertionError(u'Query type "{}" field definitions unexpected not '
                              u'found.'.format(current_query_type))
+
     # Note that as field names and type names have been confirmed to match up, and types
     # were merged without name conflicts, query type fields can also be safely merged.
-    query_type_index = 1  # Query type is the second entry in the list of definitions
+
+    # Query type is the second entry in the list of definitions of the merged_schema_ast,
+    # as guaranteed by _get_basic_schema_ast
+    query_type_index = 1
     merged_schema_ast.definitions[query_type_index].fields.extend(new_query_type_fields)
 
 
--- a/graphql_compiler/schema_transformation/merge_schemas.py
+++ b/graphql_compiler/schema_transformation/merge_schemas.py
@@ -99,8 +99,8 @@ def merge_schemas(schema_id_to_ast, cross_schema_edges, type_equivalence_hints=N
 
     for current_schema_id, current_ast in six.iteritems(schema_id_to_ast):
         current_ast = deepcopy(current_ast)
-        _accumulate_types(merged_schema_ast, type_name_to_schema_id, scalars, directives,
-                          current_schema_id, current_ast)
+        _accumulate_types(merged_schema_ast, query_type, type_name_to_schema_id, scalars,
+                          directives, current_schema_id, current_ast)
 
     return MergedSchemaDescriptor(
         schema_ast=merged_schema_ast,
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -15,24 +16,55 @@
...
     )
 )
 
 
-def merge_schemas(schema_id_to_ast):
-    """Check that input schemas do not contain conflicting definitions, then merge.
+CrossSchemaEdgeDescriptor = namedtuple(
+    'CrossSchemaEdgeDescriptor', (
+        'edge_name',  # str, name used for the corresponding in and out fields
+        'outbound_field_reference',  # FieldReference namedtuple for the outbound field
```
pmantica1: How about source_field_reference instead?

graphql_compiler/schema_transformation/utils.py
```diff
@@ -36,7 +36,21 @@ class InvalidTypeNameError(SchemaTransformError):
...
+    CrossSchemaEdgeDescriptor provided when merging schemas has an edge name that causes a
+    name conflict with an existing field.
+    """
+
+
+class InvalidCrossSchemaEdgeError(SchemaTransformError):
+    """Raised when a CrossSchemaEdge provided when merging schemas is invalid.
+
+    This may be raised if the provided CrossSchemaEdge refers to nonexistent types, types in
+    wrong/nonexistent schemas, or nonexistent fields.
```
pmantica1: wrong schemas? Maybe be a bit more clear here?

graphql_compiler/schema_transformation/merge_schemas.py
qqi0O0: I think this goes back to the source/sink vs out/in issue again, here I believe we want out/in rather than source/sink since we're describing the edge as a whole, not from the perspective of one end.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,14 +145,104 @@ def _get_basic_schema_ast(query_type):
...
+    # should never be not found.
+    if new_query_type_fields is None:
+        raise AssertionError(u'Query type "{}" field definitions unexpected not '
+                             u'found.'.format(current_query_type))
+
+    # Note that as field names and type names have been confirmed to match up, and types
+    # were merged without name conflicts, query type fields can also be safely merged.
+
+    # Query type is the second entry in the list of definitions of the merged_schema_ast,
+    # as guaranteed by _get_basic_schema_ast
```
obi1kenobi: I'm still a bit hesitant about this. Can we at least pass in the name of the query type, and assert that the element at index 1 has the expected name? I'd feel a lot better about having at least some sanity checking here.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -161,14 +145,104 @@ def _get_basic_schema_ast(query_type):
...
+    # Concatenate all query type fields.
+    # Since query_type was taken from the schema built from the input AST, the query type
+    # should never be not found.
+    if new_query_type_fields is None:
+        raise AssertionError(u'Query type "{}" field definitions unexpected not '
+                             u'found.'.format(current_query_type))
+
+    # Note that as field names and type names have been confirmed to match up, and types
+    # were merged without name conflicts, query type fields can also be safely merged.
+
```
obi1kenobi: nit: I'd make this a blank comment line, just a properly-indented `#` to signify that the two comments are related.

graphql_compiler/schema_transformation/merge_schemas.py
obi1kenobi: nit: `_get_basic_schema_ast()` to indicate it's a function and not just a variable

PR #442, opened 2019-07-30
Add get transitive closure

pmantica1:


pmantica1:
This is an algorithmically faster version of the `_add_transitive_closure` function in the `macro_system` branch, (O(N^3) vs O(N+E), where N is the number of vertices and E is the number of transitive edges). It also does not mutate the input.

graphql_compiler/global_utils.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import OrderedDict
+from itertools import chain
+
+import six
+
+
+def _get_dag_with_keys_in_reverse_topological_order(dag):
+    """Return an OrderedDict representing a DAG with keys ordered in reversed topological order.
```
qqi0O0: Might be useful to briefly explain what reversed topological order is. 

Probably also nice to mention that the output may not be unique.

graphql_compiler/global_utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Args:
+        dag: dict, string -> set of strings, mapping each vertex in the directed acyclic graph to
+             the set of vertices it has a directed edge to.
+
+    Returns:
+        OrderedDict, str -> str, representing a directed acyclic graph (DAG) with keys ordered in
```
qqi0O0: Do you mean str -> set of strings?

graphql_compiler/global_utils.py
```diff
@@ -0,0 +1,82 @@
...
+            traversal ordered in the reverse order they were visited.
+        """
+        if vertex_name in processed_vertices:
+            return []
+
+        if vertex_name in current_trace:
+            raise AssertionError(
+                'Encountered self-reference in dependency chain of {}'.format(vertex_name))
+
+        vertex_list = []
```
qqi0O0: A more descriptive name could be helpful, perhaps something like vertices_in_reverse_traversal_order (maybe that's too long)?

graphql_compiler/global_utils.py
```diff
@@ -0,0 +1,82 @@
...
+             the set of vertices it has a directed edge to.
+
+    Returns:
+        dict, string -> set of strings, representing the transitive closure of the directed
+        acyclic graph.
+    """
+    reversed_toposorted_dag = _get_dag_with_keys_in_reverse_topological_order(dag)
+    transitive_closure = dict()
+    for vertex_name, adjacent_vertices in six.iteritems(reversed_toposorted_dag):
+        transitive_closure[vertex_name] = adjacent_vertices
```
qqi0O0: This creates an alias with the input sets, and the next line modifies the input. You could clarify that this function modifies the input in the doc string, but perhaps you want to make a copy of adjacent_vertices here.

graphql_compiler/global_utils.py
```diff
@@ -0,0 +1,82 @@
...
+        transitive_closure[vertex_name].update(set(chain.from_iterable(
+            transitive_closure[adjacent_vertex_name]
+            for adjacent_vertex_name in adjacent_vertices
+        )))
```
qqi0O0: After updating, you can freeze the adjacent_vertices set for some more nice immutability.

graphql_compiler/global_utils.py
ReaLNeroM: I think reversed topological order is fairly obvious, given that the reader knows what a topological sort is: the reversed order of a valid topological order. I think this is fine.
I do agree that we need to point out the toposort is not unique. We could just reword it to `ordered in a reversed topological order`, and it'll be correct.

graphql_compiler/global_utils.py
pmantica1: Freezing the adjacent_vertices set would be out of the scope. I will freeze the transitive_closure[vertex_name] set though.

graphql_compiler/global_utils.py
pmantica1: Will create a copy.

PR #443, opened 2019-08-01
Remove add transitive closure

pmantica1:
I remove the `_add_transitive_closure` function because it has O(n^3) runtime complexity, (where is the sum of the number of object types and interface types), while having no effect. More details in the added comment.

### DIFF
--- a/graphql_compiler/compiler/subclass.py
+++ b/graphql_compiler/compiler/subclass.py
@@ -45,7 +45,7 @@ def compute_subclass_sets(schema, type_equivalence_hints=None):
         else:
             raise AssertionError(u'Unexpected type {}'.format(type(equivalent_type)))
 
-    # Note that the inheritance structure in the GraphQL schema is already transitive. Union type
+    # Note that the inheritance structure in the GraphQL schema is already transitive. Union types
     # encompass all of the object type subclasses of their equivalent object type and cannot
     # encompass other union types. Interface types are implemented by all their object type
     # subclasses and cannot be implemented by other interface types.
--- a/graphql_compiler/compiler/subclass.py
+++ b/graphql_compiler/compiler/subclass.py
@@ -46,6 +46,7 @@ def compute_subclass_sets(schema, type_equivalence_hints=None):
             raise AssertionError(u'Unexpected type {}'.format(type(equivalent_type)))
 
     # The inheritance structure in the GraphQL schema is flat. Union type encompass all of the
-    # subclasses of their equivalent object type and interface types are implemented by all their
-    # subclasses. Therefore, subclass_sets is already a transitive closure.
+    # object type subclasses of their equivalent object type and cannot encompass other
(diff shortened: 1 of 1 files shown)


graphql_compiler/compiler/subclass.py
```diff
@@ -37,7 +27,8 @@ def compute_subclass_sets(schema, type_equivalence_hints=None):
     # A class is a subclass of itself
     subclass_set = {
         classname: {classname}
-        for classname in six.iterkeys(schema.get_type_map())
+        for classname, graphql_type in six.iterkeys(schema.get_type_map())
+        if isinstance(graphql_type, (GraphQLInterfaceType, GraphQLObjectType))
```
pmantica1: According to the docstring above:
"A class here means an object type or interface."

graphql_compiler/compiler/subclass.py
```diff
@@ -54,6 +45,8 @@ def compute_subclass_sets(schema, type_equivalence_hints=None):
         else:
             raise AssertionError(u'Unexpected type {}'.format(type(equivalent_type)))
 
-    # If B subclasses A, and C subclasses B, then C subclasses A
-    _add_transitive_closure(subclass_set)
+    # The inheritance structure in the GraphQL schema is flat. Union type encompass all of the
+    # object type subclasses of their equivalent object type and cannot encompass other
+    # union types. Interface types are implemented by all their object type subclasses and cannot
```
bojanserafimov: Consider: A is equivalent to union_A, that contains B, which is equivalent to union_B, which contains C.

graphql_compiler/compiler/subclass.py
```diff
@@ -54,6 +45,8 @@ def compute_subclass_sets(schema, type_equivalence_hints=None):
         else:
             raise AssertionError(u'Unexpected type {}'.format(type(equivalent_type)))
 
-    # If B subclasses A, and C subclasses B, then C subclasses A
-    _add_transitive_closure(subclass_set)
+    # The inheritance structure in the GraphQL schema is flat. Union type encompass all of the
```
bojanserafimov: It's not flat. There's inheritance.

graphql_compiler/compiler/subclass.py
pmantica1: Then in the GraphQL schema union_A should already encompass C. Suppose that we had an edge E from B to union_A. If union_A did not encompass C already, then I there would be no way to use transverse E from B to C, but there should be because C is a subclass of A.

bojanserafimov: ok

pmantica1: I'll add this in the comment discussion in the comment for future reference.

graphql_compiler/compiler/subclass.py
pmantica1: Fixed to "transitive".

PR #444, opened 2019-08-02
Upgrade linter versions to the newest versions that still install under both Python 2 and Python 3.7

obi1kenobi:
We picked up some bugfixes along the way, so I fixed the lint they detected, and also seemingly a bug in pylint, which I reported: https://github.com/PyCQA/pylint/issues/3039


obi1kenobi:
We run linting under Python 3.6, but unfortunately the way that pipenv is currently set up requires the linters to successfully install under all environments, and `pylint==1.9.5` is marked as not supporting Python 3.7. Similarly, `pylint>=2` are Python 3-only, so we have to use `pylint==1.9.4` as the closest version that isn't blacklisting Python 3.7. I suspect that version doesn't _actually support_ 3.7, but since the development environment (and by extension the lint environment) is using 3.6, that doesn't really matter here.

TL;DR: My solution is ugly, but it works well enough to not be worth spending additional time on at the moment.

Pipfile
```diff
@@ -4,29 +4,33 @@ url = "https://pypi.org/simple"
...
+# Linters and other development tools
+bandit = ">=1.5.1,<2"
+coveralls = ">=1.5.1,<2"
+flake8 = ">=3.6.0,<4"
+flake8-print = ">=3.1.0,<4"
+flake8-quotes = ">=1.0.0,<2"
+isort = ">=4.3.4,<5"
+parameterized = ">=0.6.1,<1"
+pydocstyle = ">=2.1.1,<3"
+pylint = "==1.9.4"  # v1.9.5 is marked as "python_requires < 3.7", update this after we're Py3+ only
```
pmantica1: Shouldn't this be "python_requires > 3.7" ? (That being the reason why it doesn't work with python2.)

obi1kenobi: No. `pylint>=2` is `python_requires >= 3.3`, and `pylint==1.9.5` is `python_requires < 3.7` -- it explicitly does not support Python 3.7 and is 3.6-and-below only.

obi1kenobi: Updated the PR title to clarify.

pmantica1: Ok got it. Misunderstood what this meant.

PR #445, opened 2019-08-02
Add macro directive and error definitions, and a few small helpers.

obi1kenobi:
Continuing the work of splitting up #356 into manageable and reviewable pieces.

### DIFF
--- a/graphql_compiler/schema.py
+++ b/graphql_compiler/schema.py
@@ -2,6 +2,7 @@
 from collections import OrderedDict
 from datetime import date, datetime
 from decimal import Decimal
+from itertools import chain
 
 import arrow
 from graphql import (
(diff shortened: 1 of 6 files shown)


graphql_compiler/schema.py
```diff
@@ -355,3 +357,22 @@ def insert_meta_fields_into_existing_schema(graphql_schema):
...
+    }
+
+    directive_names = {
+        directive.name
+        for directive in directives
+    }
+
+    nondefault_directives_found = directive_names - expected_directive_names
+    if nondefault_directives_found:
+        raise AssertionError(u'Unexpected non-default directives found: {}'.format(
```
pmantica1: This error message is unclear. "Unexpected non-default directives" makes it seem as if we accept non-default directives.

PR #447, opened 2019-08-05
Update the specification of fold to account for the _x_count meta field.

obi1kenobi:
The `@fold` specification wasn't updated to account for the addition of `_x_count`, so the compiler was accepting queries that were technically prohibited by the spec.

An example of such a query was one that did not `@output` anything from the `@fold`, such as the following one which returns the names of all animals that have more than `count` children whose names contain the substring `substr`.
```
{
  Animal {
    out_Animal_ParentOf {
      _x_count @filter(op_name: ">=", value: ["$count"])
      name @filter(op_name: "has_substring", value: ["$substr"])
    }
  }
}
```

It also appears that the `README` had a bunch of trailing whitespace which my editor stripped automatically. We should look into adding a linter for this, to ensure we don't accidentally merge such changes again.


README.md
```diff
@@ -294,37 +294,66 @@ of that property field.
...
+- The `_x_count` meta field may only appear at the innermost scope of a `@fold` marked scope.
+- Marking the `_x_count` meta field with an `@output` produces an integer value corresponding to the
+  number of results within that fold.
+- Marking for `@output` any field other than the `_x_count` meta field produces a list of results,
+  where the number of elements in that list is equal to the value of the `_x_count` meta field,
+  if it were output.
+- If multiple fields (other than `_x_count`) are marked `@output`, the resulting output lists
+  are parallel: the `i`th element of each such list is the value of the corresponding field of
+  the `i`th element of the `@fold`, for some fixed order of elements. The order of elements
+  within the output of a `@fold` is not specified and not guaranteed to be stable between multiple
```
LWprogramming: "The order of elements within the output of a `@fold` is not specified"

I found this particular wording potentially confusing, since we just said "for some fixed order of elements" in the previous sentence. What if we replaced "not specified" with "fixed only for that query's particular run"?

README.md
```diff
@@ -294,37 +294,66 @@ of that property field.
...
+  runs of the same query, or even between result sets in the same run.
```
LWprogramming: Confused on what this means: `even between result sets in the same run.` Is this saying that, for example, if you have two fold scopes in the same query, one might be in alphabetical order but that doesn't mean they all have to be in alphabetical order? 

i.e. whatever ordering we have only applies to folded fields from the same vertex, whereas all bets are off for folded fields from different vertices.

README.md
obi1kenobi: Yeah, agreed – it's one of the things I was hoping to get some help wordsmithing. For a given `@fold` in a given result set, for a particular execution of the query, the order is fixed. The order is not known to the user ahead of time, and may change across runs of the query, and in different result sets of a particular execution of a given query.

README.md
obi1kenobi: Close – see above. A `result set` at the beginning of the README is defined to basically be a "row" of returned results. I should probably avoid the word `run` since it's not defined, but the idea is that the same `@fold` on two different rows could produce results whose order cannot be in any way linked to each other.

Would love some help wordsmithing on both of these.

README.md
```diff
@@ -294,37 +294,66 @@ of that property field.
...
   within a scope marked `@fold`.
+- The `_x_count` meta field may only appear at the innermost scope of a `@fold` marked scope.
+- Marking the `_x_count` meta field with an `@output` produces an integer value corresponding to the
+  number of results within that fold.
+- Marking for `@output` any field other than the `_x_count` meta field produces a list of results,
+  where the number of elements in that list is equal to the value of the `_x_count` meta field,
+  if it were output.
+- If multiple fields (other than `_x_count`) are marked `@output`, the resulting output lists
+  are parallel: the `i`th element of each such list is the value of the corresponding field of
+  the `i`th element of the `@fold`, for some fixed order of elements. The order of elements
```
MichaelaShtilmanMinkin: Are all values in the list required to exist? If not, how does this get handled?

README.md
```diff
@@ -294,37 +294,66 @@ of that property field.
...
+  within the output of a `@fold` is not specified and not guaranteed to be stable between multiple
+  runs of the same query, or even between result sets in the same run.
 - Use of type coercions or `@filter` at or within the vertex field marked `@fold` is allowed.
-  Only data that satisfies the given type coercions and filters is returned by the `@fold`.
-- If the compiler is able to prove that the type coercion in the `@fold` scope is actually a no-op,
+  The order of operations is conceptually as follows:
```
MichaelaShtilmanMinkin: Why `conceptually` and not literally? I'm assuming this is specific for a reason and I'm wondering why we're being abstract.

README.md
```diff
@@ -294,37 +294,66 @@ of that property field.
...
+marking it `@output` will produce an integer value. See the [\_x_count](#_x_count) section for
+more details.
+
 #### Example Use
 ```graphql
 {
     Animal {
         out_Animal_ParentOf @fold {
+            _x_count @output(out_name: "child_count")
```
MichaelaShtilmanMinkin: It may be nice to include type coercion here since it is listed explicitly as allowed

README.md
```diff
@@ -1310,7 +1368,7 @@ for a possible option to resolve such naming discrepancies.
 An end-to-end example including relevant GraphQL schema and SQLAlchemy engine preparation follows.
 
 This is intended to show the setup steps for the SQL backend of the GraphQL compiler, and
-does not represent best practices for configuring and running SQLAlchemy in a production system. 
+does not represent best practices for configuring and running SQLAlchemy in a production system.
```
MichaelaShtilmanMinkin: no need for `,` before `and`

README.md
obi1kenobi: Good call.

README.md
obi1kenobi: Split into separate sentences.

README.md
obi1kenobi: Right, this is because we're defining a specification – all of the behaviors in it are abstract since they describe the _expectation_ for a particular operation's effects. Backends are free to implement the actual execution however they see fit, so long as the observable effects match the specification.

In other words, backends are allowed to cheat and not do what the "conceptually" in the spec says they need to do, but only if they can't possibly get caught cheating. Many optimizations rely on this "if a tree falls in the forest but nobody is there to see it" kind of reasoning.

README.md
obi1kenobi: Not sure I understand the question – how would a value not exist? Wouldn't that be in violation of the type's definition in the schema?

README.md
```diff
@@ -294,37 +294,66 @@ of that property field.
...
 - All `@output` fields within a `@fold` traversal must be present at the innermost scope.
   It is invalid to expand vertex fields within a `@fold` after encountering an `@output` directive.
 - `@tag`, `@recurse`, `@optional`, `@output_source` and `@fold` may not be used anywhere
   within a scope marked `@fold`.
+- The `_x_count` meta field may only appear at the innermost scope of a `@fold` marked scope.
+- Marking the `_x_count` meta field with an `@output` produces an integer value corresponding to the
+  number of results within that fold.
+- Marking for `@output` any field other than the `_x_count` meta field produces a list of results,
+  where the number of elements in that list is equal to the value of the `_x_count` meta field,
+  if it were output.
```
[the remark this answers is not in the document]
obi1kenobi: changed to `selected for output`

obi1kenobi:
@MichaelaShtilmanMinkin thanks for the review, I pushed some changes based on your feedback. If you have a moment, would you mind looking over the questions @LWprogramming raised and seeing if you have similar concerns or any ideas on how to improve those parts? Wordsmithing is not my forte.

README.md
LWprogramming: To be honest, what you wrote in that comment sounds pretty clear to me ^^

README.md
LWprogramming: How about

The order of elements within the output of a `@fold` is **only guaranteed to be stable within each result set. There is no ordering guarantee for two different result sets, even if the two result sets come from the same `@fold` in the same query.**

obi1kenobi: Neat, thanks for the suggestion. I tweaked that a bit to turn it into this:
```
The order of elements within the output of a `@fold` is only guaranteed to be stable
within each individual result set. There is no guarantee of consistent ordering across 
the same `@fold` in two or more result sets.
```

Let me know what you think.

obi1kenobi: Actually rewrote it some more, to incorporate the comment from above.

LWprogramming: looks good to me!

PR #293, opened 2019-05-16
Edge macro readme

bojanserafimov:

PR #448, opened 2019-08-05
Macro edge documentation

obi1kenobi:
Direct move of the changes from #356 into here. Haven't had a chance to review the comments on the contents of this file that were in #356, so not ready for review yet.


obi1kenobi:
Going to spruce this one up a bit with the latest validation rules and some
examples from the tests.

README.rst
```diff
@@ -1962,6 +1968,344 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+    # Prints out the following arguments:
+    # {'animal_name': 'Hedwig'}
+
+Advanced macro edges use cases
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When defining macro edges, one may freely use other compiler query functionality,
+such as :code:`@recurse`, :code:`@filter`, :code:`@tag`, and so on. Here is a more complex
+macro edge definition that relies on such more advanced features to define an edge
+that connects :code:`Animal` vertices to their siblings who are both older an have a
```
pmantica1: typo: and have a higher net worth

obi1kenobi: Thanks, fixed!

README.rst
```diff
@@ -1962,6 +1968,344 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+    macro_registry = create_macro_registry(your_graphql_schema_object)
+
+To retrieve the GraphQL schema object with all its macro-based additions, use
+the :code:`get_schema_with_macros` function:
+
+.. code:: python
+
+    from graphql_compiler.macros import get_schema_with_macros
+
+    graphql_schema = get_schema_with_macros(macro_registry)
```
pmantica1: I think that you should show at one point what an GraphQL object type with macros looks like.  (Not the entire schema because that would be too much text).

README.rst
```diff
@@ -1962,6 +1968,344 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+  the users' GraphQL queries. This is generally not achievable with SQL.
+
+- Since macro expansion does not interact in any way with the underlying data system, it works
+  seamlessly with all databases and even on schemas stitched together from multiple databases.
+  In contrast, not all databases support SQL-like :code:`VIEW` functionality.
+
+For now, the compiler supports one type of macro, **macro edges**. We hope to add more types
+of macros in the future.
+
+Macro registry
```
pmantica1: I would start by giving an example of a macro edge as soon as possible. Like in https://github.com/kensho-technologies/graphql-compiler/pull/293/files

I would not talk about pro-forma fields in that example. I would probably have a small section to talk about the peculiarity of using macros with pro-forma fields.

README.rst
```diff
@@ -1962,6 +1968,344 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+            name @filter(op_name: "=", value: ["$animal_name"])
+            out_Animal_ParentOf {
+                out_Animal_ParentOf {
+                    name @output(out_name: "grandchild_name")
+                }
+            }
+        }
+    }
+
+If operations on animals' grandchildren are common in our specific use case, we may want to avoid
```
pmantica1: This is way too much text. Please write this section more concisely.

README.rst
```diff
@@ -1962,6 +1968,344 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
 for each table and to use only SQL types allowed for the dialect specified in the
 :code:`SQLAlchemySchemaInfo`.
 
+Macro system
+------------
+
+In many real-life situations, the database schema does not fit the user's mental model of the data.
```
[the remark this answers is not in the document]
obi1kenobi: Two questions: 
- So why not skip ahead to the macro edges section in that case?
- Imagine you are a new user of the compiler, and you're just poking around the docs. If you see a section about macros and think "this is just SQL views reinvented", do you think you'd make it _all the way to the bottom of the entire macro system section_ before you decide this is reinventing the wheel and stop reading?

README.rst
obi1kenobi: Good point. Unfortunately, currently the `get_schema_with_macros()` function is slightly broken because of #521 so the macro edge directive won't show up, and the example will either be lying or be not that useful.

Perhaps in a follow-up PR, I can extract a new function (e.g. `get_schema_ast_with_macros()`) that is called by `get_schema_with_macros()`. This new function would return a schema AST, which when printed would contain the `@macro_edge` directives in the appropriate places and would therefore be suitable for an example. I've made #588 for this purpose.

README.rst
obi1kenobi: Trimmed it down. Let me know what you think.

README.rst
pmantica1: Even if the field is not marked by a macro edge directive, just seeing how a new field gets added to the schema would help the user understand macros. There is also an `:emphasize-lines:` option from sphinx that would fit in nicely here so it might be worthwhile to add a TODO for that if you proceed with this approach.

I would rather not talk about the ast because we don't talk about the ast anywhere else in the documentation and some users might not even know what an ast is. Also asts are not necessarily the easiest concept to understand/explain.

README.rst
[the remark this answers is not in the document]
pmantica1: 1. You wouldn't skip to the macro edges section in this case because a new reader assumes that the headline section is important to read in order to understand how the components of the macro system work. 
2. That is a fair point. 

I guess that my bigger issue with this section is that it puts together two related, but different subjects: what is the macro system and why did we build it. How about writting it as follows? Give a short introduction of what the macro system is just below the "Macro system" header. Then in a "Motivation" section explain why we built it. That way the user can skim over the motivation section. That way the user the user will get a clear signal that this section is going to talk about the motivation of the system and choose to skip it if he wants to.

README.rst
```diff
@@ -1962,6 +1968,341 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+                    name @output(out_name: "grandchild_name")
+                }
+            }
+        }
+    }
+
+If operations on animals' grandchildren are common in our use case, we may wish that
+an edge like :code:`out_Animal_GrandparentOf` had existed and saved us some repetitive typing.
+
+One of our options is to materialize such an edge in the underlying database itself.
```
pmantica1: I think that a lot of this text belongs in the motivation section I talked about earlier.

README.rst
```diff
@@ -1962,6 +1968,341 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+  is of type :code:`[Animal]` (we know this from our schema), the compiler will automatically infer
+  that the :code:`out_Animal_GrandparentOf` macro edge also points to vertices
+  of type :code:`Animal`.
+
+- The :code:`uuid` within the inner :code:`out_Animal_ParentOf` scope is a "pro-forma" field -- it
+  is there simply to satisfy the GraphQL parser, since per the GraphQL specification, each pair of
+  curly braces must reference at least one field. The named field has no meaning in this definition,
+  and the user may choose to use any field that exists within that pair of curly braces.
+  The preferred convention for pro-forma fields is to use the field of GraphQL type :code:`ID`
+  which is the primary key of the given type in the underlying database.
```
pmantica1: The ID type does not need to be the primary key in the underlying database. In fact, the ID type has a python string type and in the cases where the primary key is an int, it can't be a primary key.

README.rst
```diff
@@ -1962,6 +1968,341 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+    #         }
+    #     }
+    # }
+
+    print(new_args)
+    # Prints out the following arguments:
+    # {'animal_name': 'Hedwig'}
+
+Advanced macro edges use cases
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
pmantica1: Maybe add  "Macro edges with advanced directives" and "Macro edges with runtime parameters" subsections.

README.rst
```diff
@@ -1962,6 +1968,341 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+
+If operations on animals' grandchildren are common in our use case, we may wish that
+an edge like :code:`out_Animal_GrandparentOf` had existed and saved us some repetitive typing.
+
+One of our options is to materialize such an edge in the underlying database itself.
+However, this causes denormalization of the database -- there are now two places where
+an animal's grandchildren are written down -- requiring additional storage space,
+and introducing potential for user confusion and data inconsistency between the two representations.
+
+Another option is to introduce a non-materialized view within the database that *makes it appear*
```
pmantica1: Same as above.

README.rst
obi1kenobi: Can we revisit this when we move to ReadTheDocs? I don't want to make too many purely-structural changes right before we change the structure entirely.

README.rst
obi1kenobi: Removed the distracting `ID` type reference. A well-formed GraphQL schema will have the primary key be typed `ID` for compliance with GraphQL principles, but the macro system can be more lenient.

README.rst
obi1kenobi: For what it's worth, based on chatting with a number of GraphQL users at a meetup a couple of weeks ago, it seems that ~nobody reads GraphQL schemas manually. Based on that, while I agree that more examples would be helpful, I don't think this is a high value-add relative to the time we'd spend on it.

In my opinion, it might be better to fix the schema-generation API to have an AST-based option, potentially move to ReadTheDocs in the meantime, and then update the docs to something even better. I think this is good enough in the meantime.

README.rst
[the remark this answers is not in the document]
obi1kenobi: > a new reader assumes that the headline section is important to read in order to understand how the components of the macro system work.

It is important. It describes what macros fundamentally aim to accomplish: "reshape how you present your data without actually moving or changing it." In my mind, there are two kinds of readers: ones that don't know what macros are meant to be (and should read this section), and ones that understand the idea already and want to know the details of macro edges (and should skip ahead).

@JulianGoetz any advice?

README.rst
JulianGoetz: I agree with pmantica here on an early example. Perhaps you can add a sentence in the first paragraph before, "As a result," that says, "Imagine if you were.. xyz" or "Imagine trying to .. abc"

README.rst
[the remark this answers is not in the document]
JulianGoetz: My advice was to create a bit of a hybrid of what you both were saying. Rather than separate out problem statement from definition, start with a quick one-sentence definition that establishes the "What" immediately, then moves onto the "Why" and then the "How." The current format as exists on (I believe) Predrag's local is intuitive and effective in my opinion. In short, I think you're both right, we don't want to drop into the Why before establishing the What, but we also don't want to divorce the definition and motivation.

README.rst
```diff
@@ -1962,6 +1968,347 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+the :code:`get_schema_with_macros` function:
+
+.. code:: python
+
+    from graphql_compiler.macros import get_schema_with_macros
+
+    graphql_schema = get_schema_with_macros(macro_registry)
+
+Schema for defining macros
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
pmantica1: Since I am not sure what a user would do with the macro-definition schema
.

README.rst
```diff
@@ -1962,6 +1968,347 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+                    name @output(out_name: "grandchild_name")
+                }
+            }
+        }
+    }
+
+If operations on animals' grandchildren are common in our use case, we may wish that
+an edge like :code:`out_Animal_GrandparentOf` had existed and saved us some repetitive typing.
+
+One of our options is to materialize such an edge in the underlying database itself.
```
pmantica1: I think these two paragraphs here are a bit repetitive. I feel that the point they are making are already mostly made by the introduction to the macro system. Consider deleting them.

README.rst
```diff
@@ -1962,6 +1968,347 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+and introducing potential for user confusion and data inconsistency between the two representations.
+
+Another option is to introduce a non-materialized view within the database that *makes it appear*
+that such an edge exists, and query this view via the GraphQL compiler. While this avoids some
+of the drawbacks of the previous approach, not all databases support non-materialized views.
+Also, querying users are not always able to add views to the database, and may require additional
+permissions on the database system.
+
+Macro edges give us the opportunity to define a new :code:`out_Animal_GrandparentOf` edge without
+involving the underlying database systems at all. We simply state that such an edge
```
pmantica1: If you follow the suggestion above, please remember to modify this paragraph accordingly.

README.rst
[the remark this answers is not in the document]
obi1kenobi: You actually can't define macros at all without it -- they won't validate since the macro directives aren't in the usual schema. I consider that a pretty critical detail :) Perhaps I should make that clear in this section.

README.rst
```diff
@@ -1962,6 +1968,347 @@ However, if specifying :code:`Table` objects manually, please make sure to inclu
...
+data structure that tracks all currently available macros, the resulting GraphQL schema
+(accounting for macros), and any additional metadata needed by the compiler.
+The compiler then leverages this registry to expand queries that rely on macros,
+rewriting them into equivalent queries that do not contain any macros and therefore
+reflect the actual underlying data model.
+
+This makes macros somewhat similar to SQL's idea of non-materialized views,
+though there are some key differences:
+
+- SQL views require database access and special permissions; databases are
```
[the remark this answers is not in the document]
obi1kenobi: Happy to go with whatever you and @JulianGoetz decide on. I have no strong preference.

README.rst
obi1kenobi: Again deferring to @JulianGoetz since this is his specialty and not mine.

README.rst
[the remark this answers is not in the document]
JulianGoetz: I think the bullets are fine. Unless of course this is required knowledge. I, for one, do not know what "SQL's idea of non-materialized views" means, and I'm not sure it is needed. Therefore, a lower hierarchy of bullets compared to headers or bolded messages makes sense to me.

README.rst
JulianGoetz: resolved

README.rst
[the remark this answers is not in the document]
obi1kenobi: I don't think it's required knowledge.

README.rst
JulianGoetz: I am in favor of repeating the important ideas in this case, especially if they appear, as they do, while rephrasing. 

If needed, can add, "As mentioned previously,.." or "As discussed earlier,..".

obi1kenobi: I'm leaning toward not adding "as mentioned previously" or anything to that effect, since in the move to ReadTheDocs, these sections will probably end up on different pages and different URLs, and adding a link that points to a roundabout reference to the same idea (see rest of my comment below) doesn't seem worth it.

Also, in my opinion the macro system only says the same thing if you already have a handle on what's going on here, and are able to piece together the implications of the macro system comparison to SQL on your own. I don't think users in that situation are reading the documentation, because they already understand everything they need.

PR #449, opened 2019-08-05
Split query

qqi0O0:
Initial implementation of split query 

A second step is creating a query plan from these query nodes, but that
part is in another file and another PR. This step splits queries and
adds property fields used in the stitch if needed, but don't add new
directives.

The idea is to visit the AST, each time stripping off one layer of children, then recursively repeating the process on the children.

The result is a tree of SubQueryNodes. Later on, this class will have a method `reroot_tree` implemented, so that the order of execution can be changed.

### DIFF
--- a/graphql_compiler/schema_transformation/split_query.py
+++ b/graphql_compiler/schema_transformation/split_query.py
@@ -16,8 +16,12 @@ from .utils import SchemaStructureError, try_get_ast, try_get_ast_and_index
 QueryConnection = namedtuple(
     'QueryConnection', (
         'sink_query_node',  # SubQueryNode
-        'source_field_path',  # List[Union[int, str]]
-        'sink_field_path',  # List[Union[int, str]]
+        'source_field_path',
+        # List[Union[int, str]], the attribute names or list indices used to access the property
+        # field used in the stitch, starting from the root of the source AST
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,430 @@
...
+
+        parent_field_name, child_field_name = self._try_get_stitch_fields(node)
+
+        # Check for existing field in parent
+        parent_field, parent_field_index = try_get_ast_and_index(
+            parent, parent_field_name, ast_types.Field
+        )
+        # Process parent field, and get parent field path
+        if parent_field is None:  # No existing source property field
+            # Change current field to stitch's source property field
```
qqi0O0: This is currently bugged -- if there is a normal vertex field before this cross schema vertex field, then this step will cause a property field to come after a vertex field in the selection, which is not allowed. Working on the nicest way to solve this issue.

qqi0O0: Solved by adding new property field between all existing property fields and existing vertex fields.

qqi0O0:
Running into issues dealing with Nones left over by removed branches, doing a rework, closing for now.

PR #451, opened 2019-08-06
Tests for rename_query

qqi0O0:
Add tests for rename query.


qqi0O0:
> Two suggestions:
> 
> * Consider splitting up the "valid input" test cases from the "raise an error" test cases, perhaps in separate files. That way it's easier for the reader to understand the expectations for a given test, without having to dig too much into the code.
> * For the "valid input" test cases, make sure to use queries that the compiler will actually accept and compile. At the moment, it seems like some of the queries wouldn't compile, for example because they don't mark anything with `@output`. This may cause the tests to break unnecessarily in the future, for example if we add more validation checks earlier in the query processing.

Should I split the tests into two files or into two classes in the same file? The same question applies to tests for merge_schema (currently I separated those out into four classes in a single large file)

obi1kenobi:
I have a mild preference for separate files, just to maximally avoid
ambiguity, but either is fine really.

obi1kenobi:
If you can make the linter happy, this is fine by me.

PR #454, opened 2019-08-07
AST traversal and rewriting helpers for macro edges

obi1kenobi:
Refactor of the `helpers.py` file that @bojanserafimov wrote in #356. Highlights:
- Split up the original helpers file into separate files, each focusing on one area among AST traversal, AST rewriting, and name generation.
- For the AST rewriting code, tweaked the functions to make them all follow the same "return the input object if making no changes" contract that is commonplace in structural sharing implementations.
- More (and more consistent) documentation.


obi1kenobi:
Also, to make review easier, the first commit is a straight-up copy of the `helpers.py` file from #356, and all the changes I made are in the subsequent commit(s). So if you are already familiar with the file in question from #356, you can review from the second commit onward.

PR #456, opened 2019-08-07
is_null and is_not_null filters

chewselene:
Adding new filtering options for `is_null` and `is_not_null`.

Ideally, we would like to eliminate `value` from `is_null`/`is_not_null` filters since is it not used. 

e.g.
```
graphql_query = '''{
        Animal {
            net_worth @filter(op_name: "is_null")
        }
    }'''
```

However, this is outside the scope of this PR and will be done separately.


graphql_compiler/compiler/expressions.py
```diff
@@ -894,12 +895,18 @@ def to_match(self):
         negated_regular_operator_format = '(NOT (%(left)s %(operator)s %(right)s))'
         # pylint: enable=unused-variable
 
+        # Comparing null to a value does not make sense.
+        if self.left == NullLiteral:
+            raise AssertionError(u'The left expression cannot be a NullLiteral! Received operator '
+                                 u'{} and right expression {}.'.format(self.operator, self.right))
```
bojanserafimov: There's a lot of checks and complexity here. Would it be easier to use a `UnaryTransformation` for `is_null` and `is_not_null`? They are after all unary `is_null : Any -> Bool`.

bojanserafimov:
Looks good overall :) Left one comment.

I'm personally ok with the `value: []` that's required for these filters. It would break uniformity if some filters take a list of arguments, and some don't. And we'd have to actively go against the GraphQL spec to achieve that.

graphql_compiler/compiler/expressions.py
obi1kenobi: They are, but I think we'd get more value from using the same representation as the other null checks. For example, the null checks generated from lowering `ContextFieldExistence` expressions all use `BinaryComposition` already, and I don't think there's a lot of value in refactoring all that code to use unary transforms instead.

obi1kenobi:
> And we'd have to actively go against the GraphQL spec to achieve that.

I actually think it might be possible to do it without breaking the spec, if we define the filter directive as:
```
directive @filter(op_name: String!, value: [String!]) on FIELD | INLINE_FRAGMENT
```
Note that the definition of `value` here loses a `!`: from `[String!]!` to `[String!]`. That would make `value` not be required.

graphql_compiler/compiler/expressions.py
```diff
@@ -940,21 +947,29 @@ def to_gremlin(self):
...
-            u'not_contains': (u'contains', negated_dotted_operator_format),
-            u'intersects': (u'intersect', intersects_operator_format),
-            u'has_substring': (u'contains', dotted_operator_format),
-        }
+        # Comparing null to a value does not make sense.
+        if self.left == NullLiteral:
+            raise AssertionError(
+                u'The left expression cannot be a NullLiteral! Received operator '
+                u'{} and right expression {}.'.format(self.operator, self.right))
+        else:
```
obi1kenobi: nit: since the `raise` breaks control flow already, I'd remove the `else` here since I feel that might make the code more readable – and it will definitely make Git not show these lines as modified.

graphql_compiler/compiler/expressions.py
bojanserafimov: I'm not convinced. In person?

bojanserafimov: Resolved offline.

graphql_compiler/tests/test_compiler.py
```diff
@@ -3336,6 +3337,76 @@ def test_has_edge_degree_op_filter_with_fold(self):
...
                         expected_cypher)
 
+    def test_is_null_op_filter(self):
+        test_data = test_input_data.is_null_op_filter()
+
+        expected_match = '''
+            SELECT
+                Animal___1.name AS `name`
+            FROM  (
+                MATCH  {{ class: Animal,
```
obi1kenobi: nit: mind putting this on its own line, and using single spaces between `FROM` and `(`, and `MATCH` and `{{`? That way it'll match the structure of the other MATCH test cases.

graphql_compiler/tests/test_compiler.py
```diff
@@ -3336,6 +3337,76 @@ def test_has_edge_degree_op_filter_with_fold(self):
...
+                        expected_cypher)
+
+    def test_is_not_null_op_filter(self):
+        test_data = test_input_data.is_not_null_op_filter()
+
+        expected_match = '''
+            SELECT
+                Animal___1.name AS `name`
+            FROM  (
+                MATCH  {{ class: Animal,
```
obi1kenobi: same nit as above

obi1kenobi:
> I initially approved, but then I noticed it's probably a good idea to add the new filters to the README? This way we can avoid a state on master where the documentation doesn't match the implementation.

Yeah, good call. Let's do that.

README.md
```diff
@@ -842,6 +844,37 @@ no valid result sets can be produced, and the resulting query will return no res
...
+#### Example Use
+```graphql
+{
+    Animal {
+        name @output(out_name: "animal_name")
+        color @filter(op_name: "is_null", value: [])
+    }
+}
+```
+This returns one row for every `Animal` that has a color defined.
```
bojanserafimov: I think you have the `is_null` and `is_not_null` swapped :)

README.md
```diff
@@ -53,6 +53,8 @@ For a more detailed overview and getting started guide, please see
      * [not_contains](#not_contains)
      * [intersects](#intersects)
      * [has_edge_degree](#has_edge_degree)
+     * [is_null](#is_null)
+     * [is_not_null](#is_null)
```
bojanserafimov: The link on `is_not_null` is pointing to `is_null`

bojanserafimov: A good way to self-review your README changes is to view the file itself on github, and click around, see if it renders correctly.

obi1kenobi:
Coveralls appears down (getting an error message on its homepage), so going to merge this without coveralls approval.

PR #457, opened 2019-08-08
Implement full-path name generation for FoldScopeLocation

LWprogramming:
Useful method for #437, allowing us to generate unique names for all vertices along the path for a fold scope.

### DIFF
--- a/graphql_compiler/compiler/emit_cypher.py
+++ b/graphql_compiler/compiler/emit_cypher.py
@@ -1,6 +1,7 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 """Convert lowered IR basic blocks to Cypher query strings."""
 from graphql_compiler.compiler.helpers import Location
+
 from .blocks import Fold, QueryRoot, Recurse, Traverse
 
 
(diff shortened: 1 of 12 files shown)


graphql_compiler/compiler/helpers.py
```diff
@@ -510,6 +510,24 @@ def get_first_folded_edge(self):
         first_folded_edge_direction, first_folded_edge_name = self.fold_path[0]
         return first_folded_edge_direction, first_folded_edge_name
 
+    def get_full_path_location_name(self):
```
obi1kenobi: Would it harm anything if this approach (or a similar one) were made to be the standard `get_location_name()` implementation for this type, instead of having a separate method for it? The comment on lines 489-495 may have been true when written, but it doesn't seem to be true anymore given the need for this new function, so it may be time to revisit.

LWprogramming: This is specifically useful for when we need to traverse while within a fold scope, i.e. the fold directive isn't at the innermost scope. For gremlin and match we only need to name the vertex at the end of the path, so it's easy to get away with using a different naming system-- but if we replace `get_location_name()` with this implementation and modify the compiler tests, it shouldn't be too bad (since it's just a naming change). 

Could be a worthwhile thing to do after fold is implemented, to get consistent naming for the different backends in test_compiler.py

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -1,8 +1,31 @@
...
+    # For other query languages like MATCH and Gremlin, the fold directive doesn't require
+    # us to name all the intermediate vertices on the path, which is why this is Cypher-specific.
+    base_location = fold_scope_location.base_location
+    base_query_path = base_location.query_path  # the path traversed so far
+    fold_path = fold_scope_location.fold_path  # the path specified at or within the folded scope.
+    full_path = base_query_path + tuple(u'_'.join(edge_name) for edge_name in fold_path)
+    location = Location(full_path, field=base_location.field,
+                        visit_counter=base_location.visit_counter)
+    if base_location.field is not None:
+        raise ValueError(u'Expected base_location\'s field to be None since this method is used to'
```
obi1kenobi: `to` and `traverse` on the next line don't have a space between them, so when Python concatenates the pieces together, you'll get `totraverse` as one word. Same as `afield` below.

PR #458, opened 2019-08-08
Generate CypherStep from IR blocks within fold scopes

LWprogramming:
Instead of just mapping each FoldScopeLocation to the list of the IR blocks from within that scope, now we map the FoldScopeLocations to the list of CypherStep objects that are generated from the IR blocks.

Each CypherStep corresponds to one additional traversal within the fold scope.


graphql_compiler/compiler/cypher_query.py
```diff
@@ -47,7 +48,9 @@
 
 def _get_all_supertypes_of_exact_type(query_metadata_table, exact_type):
     """Return the set of all supertypes of the given exact type."""
-    # TODO(predrag): Plumb the SchemaGraph work through here.
+    # TODO(predrag): Plumb the SchemaGraph work through here. When/ if this implementation changes,
+    # we may also need to modify how CypherStep objects for fold directives are constructed in
+    # _generate_CypherStep_list_from_ir_blocks based on the comment there.
```
obi1kenobi: `_generate_CypherStep_list_from_ir_blocks` seems like it was the victim of a find-and-replace, since the capitalization doesn't match the function's actual name

graphql_compiler/compiler/cypher_query.py
```diff
@@ -102,6 +105,52 @@ def _make_query_root_cypher_step(query_metadata_table, linked_location, current_
...
+                        .format(fold_scope_ir_blocks_dict,
+                                type(fold_scope_ir_blocks_dict).__name__))
+    for fold_scope_location in fold_scope_ir_blocks_dict:
+        folded_cypher_steps = []  # list of CypherStep objects corresponding to this fold scope.
+        linked_location = fold_scope_location.base_location
+        step_block = Fold(fold_scope_location)  # for first CypherStep object in this fold scope.
+        fold_scope_ir_blocks = [step_block] + fold_scope_ir_blocks_dict[fold_scope_location]
+        query_metadata_table = None  # In _make_cypher_step, this is needed for only two
+        # things: making a query root CypherStep and getting all the supertypes of the type
+        # specified in the coercion block. However, this is not a root block because we're
```
obi1kenobi: Any reason not to just make `Fold` be a root block? I'm not sure I understand what's going on here, and I'm concerned about setting `query_metadata_table = None` when we could have possibly just passed in the "real" metadata table object.

graphql_compiler/compiler/cypher_query.py
```diff
@@ -102,6 +105,52 @@ def _make_query_root_cypher_step(query_metadata_table, linked_location, current_
...
 
+def _generate_cypherstep_list_from_ir_blocks(fold_scope_ir_blocks_dict):
+    """Generate CypherStep object list for each FoldScopeLocation's list of IR blocks.
+
+    Args:
+        fold_scope_ir_blocks_dict: dict of FoldScopeLocation ->
+                                   list of IR blocks corresponding to that @fold scope.
+
+    Returns:
+        dict of FoldScopeLocation -> list of CypherStep objects corresponding to that @fold scope.
```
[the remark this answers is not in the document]
LWprogramming: good point, would remove the need to type check too

graphql_compiler/compiler/cypher_query.py
LWprogramming: Right now Python complains about it being unhashable, which means that if we mutate it anywhere, we can't use it as a dictionary key in a CypherQuery's `fold` field. If it's immutable, we could probably just use the Fold block's FoldScopeLocation's hash as its hash?

`_make_cypher_step` requires parameters `query_metadata_table`, `linked_location`, `current_step_blocks`. 

Query metadata table is used for two things: making a QueryRoot block (which doesn't apply) and computing `step_types`. However, `_get_all_supertypes_of_exact_type` doesn't use this parameter anyways, so essentially for working with folds, query metadata table doesn't seem to serve any purpose.

Linked location is where you're coming from in the traversal

Current step blocks is the list of blocks that we'll generate a CypherStep with

graphql_compiler/compiler/cypher_query.py
```diff
@@ -102,6 +105,45 @@ def _make_query_root_cypher_step(query_metadata_table, linked_location, current_
...
+            # MarkLocation is the last block needed for the next CypherStep object.
+            continue
+        # The complete list of IR blocks need not end with a MarkLocation block, e.g. if you
+        # have a Backtrack block at the end. That's fine; Backtrack is unused here anyways.
+        cypher_step = _make_cypher_step(query_metadata_table, linked_location,
+                                        current_step_ir_blocks)
+        folded_cypher_steps.append(cypher_step)
+
+        linked_location = block.location
+        current_step_ir_blocks = []
```
LWprogramming: note: assert that all leftover blocks are just unused blocks like Backtrack

PR #459, opened 2019-08-08
Implement fold list comprehension to return values

LWprogramming:
Purpose of this is to align lists if we're collecting multiple fields, e.g. if we have [(a, 1), (b, 2)] then this PR (along with the PR for generating WITH clauses, coming soon) ensures that the lists are indeed parallel and we don't end up with lists [a, b] and [2, 1] for instance.


graphql_compiler/compiler/expressions.py
```diff
@@ -704,8 +704,24 @@ def to_gremlin(self):
...
+        mark_name = u'collected_' + self.fold_scope_location.get_full_path_location_name()
+        validate_safe_string(mark_name)
+
+        template = u'[x IN %(mark_name)s | x.%(field_name)s]'
+        template_data = {
+            'mark_name': mark_name,
+        }
+
+        if field_name == COUNT_META_FIELD_NAME:
+            raise NotImplementedError()  # TODO Leon-- _x_count.
```
obi1kenobi: I'd remove the `TODO`, I think the `NotImplementedError` is clear enough about the fact that there's missing functionality.

graphql_compiler/compiler/expressions.py
```diff
@@ -704,8 +704,24 @@ def to_gremlin(self):
...
+        else:
+            template_data['field_name'] = field_name
+
+        return template % template_data
```
obi1kenobi: I'd use `.format()` instead of `%` formatting, for consistency with the other `to_cypher()` methods, e.g.: https://github.com/kensho-technologies/graphql-compiler/pull/459/files#diff-473d4ca274648f3e3c6f016992306395R535

graphql_compiler/compiler/expressions.py
```diff
@@ -704,8 +704,19 @@ def to_gremlin(self):
         raise NotImplementedError()
 
     def to_cypher(self):
-        """Not supported yet."""
-        raise NotImplementedError()
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        _, field_name = self.fold_scope_location.get_location_name()
+        mark_name = u'collected_' + self.fold_scope_location.get_full_path_location_name()
```
obi1kenobi: I think we decided to implement this differently and that code has been merged into `master`, so I think this PR needs an update.

LWprogramming:
There have been some slight changes to the code and it exists in #463 as well, so I'll combine the changes there and close this PR.

PR #460, opened 2019-08-11
Macro edge descriptor definition, construction, and validation

obi1kenobi:
Same idea as #454. Taking a manageable chunk from #356, copying it as-is in the first commit, then doing a bit of refactoring and cleanup in subsequent commits to get it into a reviewable, mergeable state.

### DIFF
--- a/graphql_compiler/macros/macro_edge/validation.py
+++ b/graphql_compiler/macros/macro_edge/validation.py
@@ -1,0 +1,275 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+from copy import copy
+from itertools import chain
+
+from graphql.language.ast import (
+    Argument, Directive, Document, Field, InlineFragment, Name, OperationDefinition, SelectionSet,
+    StringValue
+)
+from graphql.validation import validate
+
+from ...ast_manipulation import (
+    get_ast_field_name, get_human_friendly_ast_field_name, get_only_selection_from_ast
+)
+from ...compiler.compiler_frontend import ast_to_ir
+from ...exceptions import GraphQLInvalidMacroError
+from ...query_formatting.common import ensure_arguments_are_provided
+from ...schema import VERTEX_FIELD_PREFIXES, FoldDirective, is_vertex_field_name
+from .ast_rewriting import remove_directives_from_ast
+from .ast_traversal import get_directives_for_ast
+from .directives import (
+    DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION, DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION,
+    MACRO_EDGE_DIRECTIVES, MacroEdgeDefinitionDirective, MacroEdgeTargetDirective
+)
+
+
+def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
+    """Raise errors if the macro uses the macro directives incorrectly or is otherwise invalid."""
+    if ast.directives:
+        directive_names = [directive.name.value for directive in ast.directives]
+        raise GraphQLInvalidMacroError(
+            u'Unexpectedly found directives at the top level of the GraphQL input. '
+            u'This is not supported. Directives: {}'.format(directive_names))
+
+    if ast.variable_definitions is not None:
+        raise GraphQLInvalidMacroError(
+            u'Unexpectedly found variable definitions at the top level of the GraphQL input. '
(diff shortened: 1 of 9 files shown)


graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -0,0 +1,249 @@
...
+            u'The provided macro edge name "{}" has the same name as an existing field on the '
+            u'"{}" GraphQL type or interface. This is not allowed, please choose a different name.'
+            .format(macro_edge_name, class_name))
+
+
+def _get_minimal_query_ast_from_macro_ast(macro_ast):
+    """Get a query that should successfully compile to IR if the macro is valid."""
+    ast_without_macro_directives = remove_directives_from_ast(macro_ast, {
+        directive.name
+        for directive in MACRO_EDGE_DIRECTIVES
```
pmantica1: nit: MACRO_EDGE_DIRECTIVES to DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -0,0 +1,249 @@
...
+                u'macro edge definition GraphQL.'.format(directive_definition.name))
+
+        if len(macro_data) > 1:
+            raise GraphQLInvalidMacroError(
+                u'Required macro edge directive "@{}" was unexpectedly present more than once in '
+                u'the supplied macro edge definition GraphQL. It was found {} times.'
+                .format(directive_definition.name, len(macro_data)))
+
+
+def _validate_macro_ast_directives(ast, inside_fold_scope=False):
```
pmantica1: I would change this to
`_validate_use_of_non_required_macro_directives`

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -0,0 +1,249 @@
...
+                                lead to incorrect output queries being generated.
+                                *****
+
+    Returns:
+        tuple (class name for macro, name of macro edge, MacroEdgeDescriptor),
+        where the first two values are strings and the last one is a MacroEdgeDescriptor object
+    """
+    macro_directives = get_directives_for_ast(ast)
+
+    _validate_macro_ast_with_macro_directives(schema, ast, macro_directives)
```
pmantica1: I think this function could be much more readable if the validation helper functions would have better names. I recommend a number of ways to change these functions so that it is more clear what is going on.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -0,0 +1,249 @@
...
+        for directive in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION
+    })
+
+    for directive in ast.directives:
+        name = directive.name.value
+        if name not in names_of_allowed_directives:
+            raise GraphQLInvalidMacroError(u'Unexpected directive name found: {} {}'
+                                           .format(name, directive))
+
+        if name == FoldDirective.name:
```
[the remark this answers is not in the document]
obi1kenobi: I think the last example in the `@fold` section of the README is a use case for `@fold` in a macro edge: https://github.com/kensho-technologies/graphql-compiler#fold

Folds don't always have to output data, and the way we'd use them in macro edges would be in that non-outputting capacity since the `@macro_edge_target` can't be inside a scope marked `@fold`.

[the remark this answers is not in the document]
pmantica1: Oh ok makes sense.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -59,11 +64,11 @@ def _validate_macro_ast_with_macro_directives(schema, ast, macro_directives):
...
 
 
-def _validate_macro_ast_directives(ast, inside_fold_scope=False):
-    """Check that the macro is using non-macro directives properly.
+def _validate_non_required_macro_definition_directives(ast, inside_fold_scope=False):
+    """Check that the macro is using non-required macro edge definition directives properly.
 
     Restrictions on use of directives:
-    - @output and @output_source are disallowed
+    - @macro_edge, @output and @output_source are disallowed
```
obi1kenobi: I don't think `macro_edge` is necessary here, since per its definition you can't use it outside of a `FIELD_DEFINITION` location (i.e. in schema definitions), otherwise the built-in validation `validate()` call will fail. No need to remove it once it's here, just a bit redundant.

pmantica1: Noted.

PR #462, opened 2019-08-12
Rework of split_query

qqi0O0:
This PR is very large. I'd appreciate larger structural comments here. I'll break this down into smaller pieces when actually merging changes.

Tests exists, but not in this PR.

The structure of the code is rather complex:
Overall, we start with a basic SubQueryNode with the full tree, then use `_split_query_one_level` which creates children (but not grandchildren) SubQueryNodes, and repeat `_split_query_one_level` on these children in a DFS pattern, until there are no more splits.
In `_split_query_one_level`, we again use a DFS pattern of traversal.
When visiting a node, there are two general cases -- the AST needs to be split off at the node, or the AST doesn't need to be split off at the node.
In the second case, we recursively visit all the child selections of the node.
The `_split_query_ast_one_level_recursive` function returns an AST node, in a way that's similar to some of the functions recently added from the macros branch. If there are no changes, the exact input object is returned. If some changes down the line occurred, a shallow copy with a new selection set is returned. The tricky case is the return value if a cross schema vertex Field is split off -- then the property field used by the vertex field will be returned.


graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+        directives_from_existing_field = parent_field.directives
+        if directives_from_existing_field is not None:
+            new_field.directives.extend(directives_from_existing_field)
+
+    # Transfer directives from edge
+    if directives_from_edge is not None:
+        for directive in directives_from_edge:
+            if directive.name.value == u'output':  # output is illegal on edge field
+                raise GraphQLValidationError(
+                    u'Directive "{}" is not allowed on an edge field, as @output directives '
```
obi1kenobi: I believe these are called `vertex field` in the spec: https://github.com/kensho-technologies/graphql-compiler#definitions

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+                                  schema_ast: Document representing the merged schema
+                                  schema: GraphQLSchema representing the merged schema
+                                  type_name_to_schema_id: Dict[str, str], mapping name of each
+                                                          type to the id of the schema it came
+                                                          from
+
+    Returns:
+        Tuple[SubQueryNode, frozenset[str]]. The first element is the root of the tree of
+        QueryNodes. Each node contains an AST representing a part of the overall query,
+        targeting aspecific schema. The second element is the set of all intermediate output
```
LWprogramming: nit: aspecific -> a specific

graphql_compiler/schema_transformation/utils.py
```diff
@@ -263,3 +284,84 @@ def check_ast_schema_is_valid(ast):
...
+                    )
+            else:
+                past_property_fields = True
+
+
+def check_query_is_valid_to_split(schema, query_ast):
+    """Check the query is valid for splitting.
+    In particular, ensure that the query validates against the schema, does not contain
+    unsupported directives, and that in each selection, all property fields occur before all
+    vertex fields.
```
LWprogramming: Nit: Add a newline after this before `Args:`. Same for before `Raises:`

graphql_compiler/schema_transformation/utils.py
```diff
@@ -263,3 +284,84 @@ def check_ast_schema_is_valid(ast):
...
+        )
+
+
+class CheckQueryIsValidToSplitVisitor(Visitor):
+    """Check the query only has supported directives, and its fields are correctly ordered."""
+    # This is very restrictive for now. Other cases (e.g. tags not crossing boundaries) are
+    # also ok, but temporarily not allowed
+    supported_directives = frozenset(('filter', 'output', 'optional'))
+
+    def enter_Directive(self, node, *args):
```
LWprogramming: nits: capitalization, let's make this `enter_directive` and likewise for other functions.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -263,3 +284,84 @@ def check_ast_schema_is_valid(ast):
...
+            return True
+        else:
+            return False
+    else:
+        raise AssertionError(
+            u'Input selection "{}" is not of type Field or InlineFragment.'.format(selection)
+        )
+
+
+class CheckQueryIsValidToSplitVisitor(Visitor):
```
LWprogramming: Would `QuerySplitValidVisitor` or some more concise version of this work? Had to read through the code to figure out what objects of this class does.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -263,3 +284,84 @@ def check_ast_schema_is_valid(ast):
...
+                if past_property_fields:
+                    raise GraphQLValidationError(
+                        u'In the selections {}, the property field {} occurs after a vertex '
+                        u'field or a type coercion statement, which is not allowed, as all '
+                        u'property fields must appear before all vertex fields.'.format(
+                            node.selections, field
+                        )
+                    )
+            else:
+                past_property_fields = True
```
LWprogramming: Found this potentially confusing since `past` can either mean we've already seen a property field (which was my original interpretation of this and I was confused because seeing property fields doesn't mean we can't have any more), or it could mean we've gone _past_ any property fields and should expect to see no more. 

How about renaming this to be `seen_vertex_field` or some variant of that?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+        child_type_name = type_coercion_inline_fragment.type_condition.name.value
+        child_selection_set = type_coercion_inline_fragment.selection_set
+
+    return child_type_name, child_selection_set.selections
+
+
+def _get_edge_to_stitch_fields(merged_schema_descriptor):
+    """Get a map from type/field of each cross schema edge, to the fields that the edge stitches.
+
+    This is necessary only because graphql currently doesn't process schema directives correctly.
```
LWprogramming: nit: capitalize `graphql`

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+        - GraphQLValidationError if the query doesn't validate against the schema, contains
+          unsupported directives, or some property field occurs after a vertex field in some
+          selection
+        - SchemaStructureError if the input merged_schema_descriptor appears to be invalid
+          or inconsistent
+    """
+    check_query_is_valid_to_split(merged_schema_descriptor.schema, query_ast)
+
+    # If schema directives are correctly represented in the schema object, type_info is all
+    # that's needed to detect and address stitching fields. However, before this issue is
```
LWprogramming: Could be worth specifying that the issue is that GraphQL currently doesn't process schema directives correctly. I found that from reading `_get_edge_to_stitch_fields`'s comments but didn't know just from reading this. Your call if you want to include it or not

graphql_compiler/schema_transformation/utils.py
LWprogramming: resolved offline

~why would python let you do this~

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+                                  and a map from names of types to their schema ids
+
+    Returns:
+        Dict[Tuple(str, str), Tuple(str, str)], mapping (type name, edge field name) to
+        (source field name, sink field name) used in the @stitch directive, for each cross
+        schema edge
+    """
+    edge_to_stitch_fields = {}
+    for type_definition in merged_schema_descriptor.schema_ast.definitions:
+        if isinstance(type_definition, (
```
LWprogramming: Should it be possible for `type_definition` to not be one of these two types? If not, what should happen if you get one that's not? If it shouldn't be possible, maybe raise an error instead of this conditional.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+    else:
+        return output_directive.arguments[0].value.value  # Location of value of out_name
+
+
+class IntermediateOutNameAssigner(object):
+    """Used to generate and keep track of out_name of @output directives."""
+    def __init__(self):
+        """Create assigner with empty records."""
+        self.intermediate_output_names = set()
+        self.intermediate_output_count = 0
```
LWprogramming: Could we replace this with `len(self.intermediate_output_names)` wherever it's used, or is it possible for this field to diverge from the length of the names set?

graphql_compiler/schema_transformation/split_query.py
qqi0O0: It's possible -- for instance, union type, scalars, directive definitions. We just ignore those, as they don't contain vertex fields (or any fields at all).

graphql_compiler/schema_transformation/split_query.py
qqi0O0: My feeling is that it doesn't matter a that much how this is represented. This is a tentative way of generating intermediate output names. I made this object to assign and keep track of names, so that it's easy to change. In its currently iteration, the size of the set is equal to the count, but this could easily change I think.

qqi0O0: My feeling is that it doesn't matter a that much how this is represented. This is a tentative way of generating intermediate output names. I made this object to assign and keep track of names, so that it's easy to change. In its currently iteration, the size of the set is equal to the count, but this could easily change I think.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,642 @@
...
+        'sink_query_node',  # SubQueryNode
+        'source_field_out_name',
+        # str, the unique out name on the @output of the the source property field in the stitch
+        'sink_field_out_name',
+        # str, the unique out name on the @output of the the sink property field in the stitch
+    )
+)
+
+
+class SubQueryNode(object):
```
LWprogramming: Your call: would `QuerySubtree` be a clearer name for this?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -39,6 +47,165 @@ def __init__(self, query_ast):
...
+    for type_definition in merged_schema_descriptor.schema_ast.definitions:
+        if isinstance(type_definition, (
+            ObjectTypeDefinition, InterfaceTypeDefinition
+        )):
+            for field_definition in type_definition.fields:
+                stitch_directive = try_get_ast_by_name_and_type(
+                    field_definition.directives, u'stitch', Directive
+                )
+                if stitch_directive is not None:
+                    source_field_name = stitch_directive.arguments[0].value.value
```
obi1kenobi: This is actually buggy in a very subtle and non-obvious way: directive arguments are passed by name, not positionally – so I could in principle pass `@stitch(sink_field: "foo", source_field: "bar")` and cause this code to load them backwards.

The `compiler_frontend.py` file works around this by passing `directive.arguments` to `get_uniquely_named_objects_by_name` and then loading argument values by name. I'd suggest that here as well.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -39,6 +47,165 @@ def __init__(self, query_ast):
...
+
+    Returns:
+        Dict[Tuple(str, str), Tuple(str, str)], mapping (type name, vertex field name) to
+        (source field name, sink field name) used in the @stitch directive, for each cross
+        schema edge
+    """
+    edge_to_stitch_fields = {}
+    for type_definition in merged_schema_descriptor.schema_ast.definitions:
+        if isinstance(type_definition, (
+            ObjectTypeDefinition, InterfaceTypeDefinition
```
obi1kenobi: Would this work if the stitch directives are provided as part of type/interface extension definitions?

For example:
```graphql
extend type Foo {
    out_Foo_Bar: [Bar] @stitch(source_field: "id", sink_field: "uuid")
}
```

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -559,3 +726,54 @@ def assign_and_return_out_name(self):
...
+        if type_name in self.type_name_to_schema_id:  # It may be a scalar, and thus not found
+            current_type_schema_id = self.type_name_to_schema_id[type_name]
+            prior_type_schema_id = self.query_node.schema_id
+            if prior_type_schema_id is None:  # First time checking schema_id
+                self.query_node.schema_id = current_type_schema_id
+            elif current_type_schema_id != prior_type_schema_id:
+                # A single query piece has types from two schemas -- merged_schema_descriptor
+                # is invalid: an edge field without a @stitch directive crosses schemas,
+                # or type_name_to_schema_id is wrong
+                raise SchemaStructureError(
```
obi1kenobi: Very nice! Would it be a lot of work to figure out one example of such a type that came in from a different schema without a stitch? If it's a lot of work then don't worry about it, but if it's relatively simple I think it would make a scary user-facing error seem quite a bit less scary.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: The code in its current form wouldn't work with that. I can extend it if desired, but something to note is that `merge_schemas` does not allow type extensions in the input schemas, so we don't expect there to be type extensions in the merged schema AST.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: Ah good point. The current iteration of `merge_schemas` always creates `@stitch` directives in the order of source to sink, but you're definitely right that I should check arguments by name to be safe.

On that note, should I make a test where the `@stitch` arguments are switched? It's a little awkward to make a schema for that test, since I'll need to create a MergedSchemaDescriptor by hand instead of by `merge_schemas`

obi1kenobi: Probably a good idea to have a test, even if it requires making a descriptor by hand. It's the sort of thing where (now that we've spotted the potential issue) we don't expect problems in the future, but if a future refactor caused a regression then we'd lose a lot of time in debugging. Better to invest a bit of time now in a good test instead.

graphql_compiler/schema_transformation/split_query.py
obi1kenobi: That's okay then, don't worry about it. We can change it when we support them end-to-end.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: I think it's not hard to write a validator where you pass in a schema AST and a map of type name to schema id, and the validator spits out fields that cross schemas but don't have a `@stitch`, if any. I don't think it's easy to do right here though with the information we have.

PR #463, opened 2019-08-12
Implement fold-related clauses for Cypher

LWprogramming:
This PR handles generating WITH and RETURN clauses. We could split this up into
two PRs but since the changes aren't super huge and it doesn't really
make sense to have the two separate (because then we get incorrect
queries), I combined this into one PR.

Tests for these coming soon.


graphql_compiler/compiler/emit_cypher.py
```diff
@@ -68,14 +75,23 @@ def _emit_code_from_cypher_step(cypher_step):
             u'%(left_edge_mark)s-[:%(edge_type)s%(quantifier)s]-%(right_edge_mark)s' +
             step_vertex_pattern
         )
-        linked_location_name, _ = cypher_step.linked_location.get_location_name()
+        if isinstance(cypher_step.linked_location, FoldScopeLocation):
```
obi1kenobi: Perhaps a good opportunity for a helper function here? This code exists at least twice, once around line 55 and once here.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -108,11 +124,16 @@ def _emit_with_clause_components(cypher_steps):
...
-        for cypher_step in cypher_steps
-    }
+    result = []
+    location_names = set()
+    for cypher_step in cypher_steps:
+        location = cypher_step.as_block.location
+        if isinstance(location, Location):
+            location_name = location.get_location_name()[0]
+        else:
+            # must be a FoldScopeLocation because as_block is a MarkLocation block.
```
obi1kenobi: How sure are you about this? If you are very sure, then assert it: `elif isinstance(location, FoldScopeLocation):` + `else: raise AssertionError(<message about unreachable code here>)`. If you aren't very sure, then don't claim it in a comment :)

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
     return result
 
 
+def _emit_with_clause_components_for_current_fold_scope(current_fold_scope_cyphersteps):
+    """Emit the component strings for Cypher WITH clause for current fold."""
+    # At this point we should already have emitted all component strings from previous CypherSteps,
+    # whether or not they were in fold scopes or not.
```
obi1kenobi: this doesn't quite flow: `whether or not` + `fold scopes or not`

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
...
+    result = []
+
+    vertex_names = {}
+    for cypher_step in current_fold_scope_cyphersteps:
+        fold_scope_location = cypher_step.as_block.location
+        full_vertex_name = fold_scope_location_full_path_name(fold_scope_location)
+        vertex_names[u'collect(' + full_vertex_name + ')'] = u'collected_' + full_vertex_name
```
obi1kenobi: Maybe another good opportunity for a helper function: `collected_' + fold_scope_location_full_path_name(fold_scope_location)` exists both here and on line 135. That means that if we change the naming of this variable, we have to find both places in the code and e.g. modify two constants that happen to be set to the same literal value but are not otherwise related to each other.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
...
+
+    # Sort the locations, to ensure a deterministic order.
+    for index, collect_call in enumerate(sorted(vertex_names)):
+        # haha get it? collect_call because we're calling the function collect()?
```
obi1kenobi: I don't think this comment is contributing to the code clarity or maintainability, I'd remove it.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
...
+                      {'collect_call': collect_call, 'collected_name': vertex_names[collect_call]})
+    return result
+
+
+def _emit_fold_scope(cypher_query):
+    """Return a Cypher query pattern match expression for each fold scope in cypher_query."""
+    query_data = []
+    previous_fold_scope_cypher_steps = []
+    for fold_scope_location in cypher_query.folds:
+        current_fold_scope_cyphersteps = cypher_query.folds[fold_scope_location]
```
obi1kenobi: `cypher_steps` instead of `cyphersteps` for consistency with the rest of the code

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
     return result
 
 
+def _emit_with_clause_components_for_current_fold_scope(current_fold_scope_cyphersteps):
+    """Emit the component strings for Cypher WITH clause for current fold."""
+    # At this point we should already have emitted all component strings from previous CypherSteps,
```
obi1kenobi: What are these `component strings` mentioned here? I don't think I understand what this comment is referring to.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
     return result
 
 
+def _emit_with_clause_components_for_current_fold_scope(current_fold_scope_cyphersteps):
```
obi1kenobi: `cyphersteps` -> `cypher_steps` for consistency with the rest of the code

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
...
+            query_data.append(_emit_code_from_cypher_step(cypher_step))
+        # Now create the WITH clause
+        query_data.append(u'WITH')
+        # First for all non-fold-scope CypherSteps
+        query_data.extend(_emit_with_clause_components(cypher_query.steps))
+
+        # And then for all fold scope CypherSteps not in this particular fold scope
+        if previous_fold_scope_cypher_steps:
+            query_data.append(u',')
+        query_data.extend(_emit_with_clause_components(previous_fold_scope_cypher_steps))
```
obi1kenobi: Would something like this reduce the amount of duplication and special-casing in this code?

```
with_clause_steps = chain(cypher_query.steps, previous_fold_scope_cypher_steps)
query_data.extend(_emit_with_clause_components(with_clause_steps))
```

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +148,64 @@ def _emit_with_clause_components(cypher_steps):
...
+        # haha get it? collect_call because we're calling the function collect()?
+        if index > 0:
+            result.append(u',')
+        result.append(u'\n  %(collect_call)s AS %(collected_name)s' %
+                      {'collect_call': collect_call, 'collected_name': vertex_names[collect_call]})
+    return result
+
+
+def _emit_fold_scope(cypher_query):
+    """Return a Cypher query pattern match expression for each fold scope in cypher_query."""
```
obi1kenobi: It might not be a bad idea to give an example on what kind of structure this function generates. I already knew what to expect since we've discussed this already, but it still took me several minutes of digging into the code to make sense of it.

graphql_compiler/compiler/expressions.py
```diff
@@ -641,8 +642,25 @@ def to_gremlin(self):
                              u'in Gremlin, so this is a bug. This function should not be called.')
 
     def to_cypher(self):
-        """Not implemented yet."""
-        raise NotImplementedError()
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        _, field_name = self.fold_scope_location.get_location_name()
+        mark_name = u'collected_' +\
```
obi1kenobi: Per the Google Python style guide, please use parens instead of `\` whenever possible.

```
mark_name = (
    u'collected_' + 
    emit_cypher.fold_scope_location_full_path_name(self.fold_scope_location)
)
```

Although if you write that helper function I suggested above, this would just be a call to that helper directly rather than a third separate implementation of the same naming convention.

graphql_compiler/compiler/emit_cypher.py
LWprogramming: These are slightly different since the condition for which we want the full path name is slightly different. Earlier it depends on if the current step is in a fold scope, but here we want the linked location and if this is the first step in a fold scope, the linked location would be a Location.

I'm open to passing in the cypher_step and the boolean for whether or not to use the `FoldScopeLocation` function though since the logic besides that is the same.

graphql_compiler/compiler/emit_cypher.py
LWprogramming: Good idea-- caught another example in expressions.py too

graphql_compiler/compiler/emit_cypher.py
LWprogramming: Each thing (vertex, list) that gets passed through a WITH clause is a component of the clause, hence component string.

Changed to clarify

graphql_compiler/compiler/emit_cypher.py
LWprogramming: TIL about chain, thanks!

graphql_compiler/compiler/emit_cypher.py
obi1kenobi: I'm not sure I understand. What I was envisioning was something like:

```python
def new_helper_function_with_better_name(location):
    if isinstance(location, FoldScopeLocation):
        return fold_scope_location_full_path_name(step_location)
    elif isinstance(location, Location):
        location_name, _ = location.get_location_name()
        return location_name

    raise AssertionError(<unreachable code error message>)
```

Then, around line 59-66 you'd do
```python
step_location_name = new_helper_function_with_better_name(cypher_step.as_block.location)
```
and around lines 85-91 you'd do
```python
linked_location_name = new_helper_function_with_better_name(cypher_step.linked_location)
```

Is this not equivalent to what we have here?

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +159,120 @@ def _emit_with_clause_components(cypher_steps):
...
+    RETURN
+      Animal___1.name AS `animal_name`,
+      [x IN collected_Animal__out_Animal_ParentOf___1 | x.name] AS `child_names_list`,
+      [x IN collected_Animal__out_Animal_ParentOf___1 | x.uuid] AS `child_uuids_list`,
+      [x IN collected_Animal__in_Animal_ParentOf___1 | x.name] AS `parent_names_list`,
+      [x IN collected_Animal__in_Animal_ParentOf___1 | x.uuid] AS `parent_uuids_list`
+
+    For each fold scope, the structure generally works as follows:
+
+    First, traverse to the vertex just outside the fold scope. Each traversal is mandatory here, so
```
obi1kenobi: For the enumeration that starts on line 206, introduce some sort of bullet point symbol so that it's clear where each bullet point starts and ends. I'd suggest `-`.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +159,120 @@ def _emit_with_clause_components(cypher_steps):
...
+    it through the WITH clause as itself (i.e. `V AS V`). For each vertex at or inside a
+    previously-seen fold scope, return the name given to the list. For vertices at or inside the
+    current fold scope, collect the vertices in the WITH clause as well, which materializes the list
+    and guarantees that the order of elements within the output of a @fold is stable within each
+    result set. Corresponds to:
+        WITH
+          Animal___1 AS Animal___1,
+          collect(Animal__out_Animal_ParentOf___1) AS collected_Animal__out_Animal_ParentOf___1
+
+    This process is repeated for other fold scopes in the query. Corresponds to:
```
obi1kenobi: Nice, this helps a lot. For bonus points, use a numbered list instead of bullet points, and then tie the comments in the code into the numbered list by making them use matching numbering. A more hard-core approach (which might be overkill here, but just mentioning it for inspiration) exists in the `_compile_vertex_ast()` function in `compiler_frontend.py`.

graphql_compiler/compiler/expressions.py
```diff
@@ -641,8 +642,20 @@ def to_gremlin(self):
                              u'in Gremlin, so this is a bug. This function should not be called.')
 
     def to_cypher(self):
-        """Not implemented yet."""
-        raise NotImplementedError()
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        _, field_name = self.fold_scope_location.get_location_name()
+        mark_name = emit_cypher.collected_vertex_list_name(
```
obi1kenobi: `expressions.py` is calling into `emit_cypher` in two places here. On one hand, it's doing so from the `to_cypher()` method so maybe it's okay. On the other, it still feels kind of icky. Is there a place to put these two functions where we won't have to do this?

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -27,6 +29,11 @@ def _get_full_path_location_name(fold_scope_location):
     return step_location_name
 
 
+def collected_vertex_list_name(full_path_name):
```
obi1kenobi: Functions are verbs, not nouns – let's name them as such.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -1,11 +1,13 @@
...
-from graphql_compiler.compiler.helpers import Location
+from itertools import chain
+
+from graphql_compiler.compiler.helpers import FoldScopeLocation, Location
 
 from .blocks import Fold, QueryRoot, Recurse, Traverse
 
 
-def _get_full_path_location_name(fold_scope_location):
+def fold_scope_location_full_path_name(fold_scope_location):
```
obi1kenobi: Functions are verbs, not nouns – let's name them as such.

graphql_compiler/compiler/emit_cypher.py
LWprogramming: ah, you're right-- the fact that I checked if it was a fold step earlier threw me off

graphql_compiler/compiler/emit_cypher.py
LWprogramming: I like the numbering idea!

graphql_compiler/compiler/expressions.py
LWprogramming: I originally had the functions in `emit_cypher` since that has us write out `emit_cypher.function_name()`, so it's less likely someone goes and uses it for non-Cypher things in the future.

OTOH functions like `fold_scope_location_full_path_name` aren't necessarily Cypher-specific-- maybe at some later point it would be useful for new backends as well. Open to moving any non-Cypher-specific functions to `helpers.py`.

That doesn't solve the question of what to do with `collected_vertex_list_name` because that _is_ Cypher-specific-- in that case I'm not sure it would make sense in `helpers.py` because it's specific to one language. In that case we could probably keep `collected_vertex_list_name` where it is.

LWprogramming: follow-up point about `fold_scope_location_full_path_name`: PR #464 also uses this function in ir lowering so I definitely can see it going somewhere other than `emit_cypher` bc now it's crossing from emit to IR lowering and that shouldn't happen

graphql_compiler/compiler/cypher_helpers.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+"""Helper functions for Cypher, such as generating unique vertex names within fold scopes."""
+from graphql_compiler.compiler.helpers import FoldScopeLocation, Location
```
obi1kenobi: In library packages, prefer relative imports instead: `from .helpers import <stuff>`. Python import rules are complex and relative imports tend to break less often in libraries.

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +118,130 @@ def _emit_with_clause_components(cypher_steps):
     return result
 
 
+def _emit_with_clause_components_for_current_fold_scope(current_fold_scope_cypher_steps):
+    """Emit a list of strings, one for each vertex or list passing through a WITH clause."""
+    result = []
+
+    vertex_names = {}
+    for cypher_step in current_fold_scope_cypher_steps:
+        if not(isinstance(cypher_step, CypherStep)):
```
obi1kenobi: nit: the parens after `not` are not necessary, I'd remove them

graphql_compiler/compiler/emit_cypher.py
```diff
@@ -127,10 +118,130 @@ def _emit_with_clause_components(cypher_steps):
...
+        full_vertex_name = cypher_helpers.get_fold_scope_location_full_path_name(
+            fold_scope_location)
+        collected_name = cypher_helpers.get_collected_vertex_list_name(full_vertex_name)
+        vertex_names[u'collect(' + full_vertex_name + ')'] = collected_name
+
+    # Sort the locations, to ensure a deterministic order.
+    for index, collect_call in enumerate(sorted(vertex_names)):
+        if index > 0:
+            result.append(u',')
+        result.append(u'\n  %(collect_call)s AS %(collected_name)s' %
```
obi1kenobi: minor nit: prefer `.format()` instead of `%` formatting unless there's a specific reason not to do that. MATCH uses `%` formatting because curly braces are everywhere and `.format()` gets really annoying when you have to escape all those braces. Here it doesn't seem there's a strong reason to use `%` so I'd go with `.format`

graphql_compiler/compiler/cypher_helpers.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_fold_scope_location_full_path_name(fold_scope_location):
+    """Return a unique name with the full traversal path to this FoldScopeLocation."""
+    # HACK(Leon): Get a unique name for each vertex in a fold traversal in Cypher.
+    # For FoldScopeLocation objects, get_location_name() only uses the first edge in the traversal,
+    # which doesn't work for Cypher because we need to explicitly label every intermediate vertex
+    # along that path.
+    # For other query languages like MATCH and Gremlin, the fold directive doesn't require
+    # us to name all the intermediate vertices on the path, which is why this is Cypher-specific.
+    if not (isinstance(fold_scope_location, FoldScopeLocation)):
```
obi1kenobi: You technically don't need these parens either, but I'll stop nitpicking and merge this.

PR #464, opened 2019-08-12
Fix IR lowering for filtering and fold scopes

LWprogramming:
Previously, if we attempted to filter on or inside a fold scope, we'd
run into problems because the IR lowering would attempt to change the
LocalField within the filter to a FoldedContextField. However,
FoldedContextField's field type must be of type GraphQLList (because
it's for after the fold has happened).

In our case, we want to filter using the field belonging to the vertex
itself, pre-fold. We also can't use ContextField because that requires
its location be a Location type object, not a FoldScopeLocation object.

I've written a special class specifically to take care of this scenario.
Tests for these coming soon!


graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -89,16 +92,89 @@ def remove_mark_location_after_optional_backtrack(ir_blocks, query_metadata_tabl
...
+    def to_match(self):
+        """Should never be called, since this is Cypher-specific."""
+        raise NotImplementedError()
+
+    def to_cypher(self):
+        """Return a unicode object with the Cypher representation of this expression."""
+        self.validate()
+
+        _, field_name = self.fold_scope_location.get_location_name()
+        mark_name = emit_cypher.fold_scope_location_full_path_name(self.fold_scope_location)
```
obi1kenobi: The fact that we have to reach from the Cypher lowering code into the Cypher emit code seems to suggest that this helper function is in the wrong place. Let's put it in a file where we don't have to break abstraction boundaries to import it.

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -89,16 +92,89 @@ def remove_mark_location_after_optional_backtrack(ir_blocks, query_metadata_tabl
     return new_ir_blocks
 
 
+class CypherFoldScopeContextField(Expression):
+    """An expression for a field captured in a @fold scope for non _x_count filtering."""
```
obi1kenobi: I think a few updates and more bits of information would be nice here.
- I think `non _x_count filtering` is too restrictive as a description here. `_x_count` seems disallowed altogether in the `validate()` function, so it's not just about filtering on it.
- I'd explain how this is different from a regular `FoldedContextField`. Why do we need a separate class for it?
- Bonus points if you come up with a name that suggests the above difference immediately, so that readers of the code don't have to reach for the docstring to understand the distinction – especially if it helps avoid the confusion when they notice that both regular `FoldedContextField` objects and these new objects are used simultaneously.

LWprogramming: updated: the name is still really long but it's clearer that it's for pre-folding work

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -89,6 +92,88 @@ def remove_mark_location_after_optional_backtrack(ir_blocks, query_metadata_tabl
...
+        if not isinstance(self.fold_scope_location, FoldScopeLocation):
+            raise TypeError(u'Expected FoldScopeLocation fold_scope_location, got: {} {}'.format(
+                type(self.fold_scope_location), self.fold_scope_location))
+
+        if self.fold_scope_location.field is None:
+            raise ValueError(u'Expected FoldScopeLocation at a field, but got: {}'
+                             .format(self.fold_scope_location))
+
+        if self.fold_scope_location.field == COUNT_META_FIELD_NAME:
+            raise TypeError(u'Expected fold_scope_location field to not be the _x_count meta-field '
```
obi1kenobi: Very nice!

PR #466, opened 2019-08-13
Tests for invalid cross schema edges when merging schemas

qqi0O0:
The last class of tests for merge_schemas.

### DIFF
--- a/graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
+++ b/graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
@@ -10,7 +10,7 @@ import six
 from ...schema_transformation.merge_schemas import (
     CrossSchemaEdgeDescriptor, FieldReference, merge_schemas
 )
-from ...schema_transformation.utils import SchemaNameConflictError
+from ...schema_transformation.utils import InvalidCrossSchemaEdgeError, SchemaNameConflictError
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -954,6 +954,513 @@ def test_id_match_int(self):
         self.assertEqual(merged_schema_string, print_ast(merged_schema.schema_ast))
 
 
+class TestMergeSchemasInvalidCrossSchemaEdges(unittest.TestCase):
+    def test_invalid_edge_within_single_schema(self):
+        with self.assertRaises(InvalidCrossSchemaEdgeError):
+            merge_schemas(
+                OrderedDict([
```
obi1kenobi: Is passing in a dict of size 1 ever allowed? If not, we might unintentionally be testing something other than the intended target of the test – it's possible that `merge_schemas` just has a `if len(schema_dict) < 2: raise InvalidCrossSchemaEdgeError()` and that's how it passes this test.

qqi0O0: Good point. A dict of size 1 is allowed, but I'll change the test to make sure there's no ambiguity in the source of the error.

PR #468, opened 2019-08-13
Implementing projecting schema to only contain a subset of types

qqi0O0:
Also removed all edges going to removed types, and all 'implements X' where X is a removed interface.

Tests exist, not in this PR.


graphql_compiler/schema_transformation/restrict_schema.py
```diff
@@ -0,0 +1,140 @@
...
+from ..ast_manipulation import get_ast_with_non_null_and_list_stripped
+from .utils import SchemaStructureError, get_query_type_name, get_scalar_names
+
+
+def restrict_schema(schema_ast, types_to_keep):
+    """Return new AST containing only a subset of types.
+
+    In addition to all types named in types_to_keep, the query type will also be kept. All user
+    defined scalars and enums will be kept.
+    All property fields will be kept, and only vertex fields that go to kept types will be
```
obi1kenobi: A few more things to consider, document, and test:
- If I specify that I'd like to keep an interface, but not keep any of its implementations, is that a `SchemaStructureError`?
- If I specify that I'd like to keep an implementation of an interface but not the interface itself, is that an error, or is that okay?
- If I specify that I'd like to keep all implementations of an interface, but not the interface itself, is that an error, or is that okay?
- Do I have to specify that I'd like to keep union types, or are they kept and updated automatically? If I have to specify them, what happens if I don't?
- If I put things in `types_to_keep` that either are random strings that don't correspond to anything, or correspond to a scalar/enum/the root query type, is that an error?

There might be more considerations like the above.

graphql_compiler/schema_transformation/restrict_schema.py
```diff
@@ -0,0 +1,140 @@
...
+from copy import copy
+
+from graphql import build_ast_schema
+from graphql.language.visitor import REMOVE, Visitor, visit
+
+from ..ast_manipulation import get_ast_with_non_null_and_list_stripped
+from .utils import SchemaStructureError, get_query_type_name, get_scalar_names
+
+
+def restrict_schema(schema_ast, types_to_keep):
```
obi1kenobi: In the foreseeable future, we might have a similar function that restricts _which fields_ are visible on certain types, so it might be good to make sure the naming is ready for that. Maybe `restrict_schema_types` instead, so that `restrict_schema_fields` can be added later?

graphql_compiler/schema_transformation/restrict_schema.py
```diff
@@ -0,0 +1,140 @@
...
+    query_type_name = get_query_type_name(schema)
+    scalar_names = get_scalar_names(schema)
+
+    visitor = RestrictSchemaVisitor(
+        types_to_keep, query_type_name, scalar_names
+    )
+    restricted_schema = visit(schema_ast, visitor)
+
+    try:
+        build_ast_schema(schema_ast)
```
obi1kenobi: Should this be passing `restricted_schema` instead?

graphql_compiler/schema_transformation/restrict_schema.py
```diff
@@ -0,0 +1,140 @@
...
+    except Exception as e:  # Can't be more specific, build_ast_schema throws Exceptions
+        raise SchemaStructureError(u'The resulting schema is invalid. Message: {}'.format(e))
+
+    # Note that it is possible for some types in the restricted schema to be unreachable
+    return restricted_schema
+
+
+class RestrictSchemaVisitor(Visitor):
+    """Remove types that are not explicitly kept, and fields go these types."""
+    def __init__(self, types_to_keep, query_type, scalars):
```
obi1kenobi: I believe one of our linters is going to express a strong desire for a blank line between the class docstring and the init function.

graphql_compiler/schema_transformation/restrict_schema.py
```diff
@@ -0,0 +1,140 @@
...
+    kept.
+
+    Args:
+        schema_ast: Document, representing the schema we're using to create a new schema with
+                    fewer types. It is not modified by this function
+        types_to_keep: Set[str], the set of names of types that we want to keep in the output
+                       schema
+
+    Returns:
+        Document, representing the schema that is derived from the input schema AST, but only
```
obi1kenobi: Any thoughts on what we'd do with `type_equivalence_hints`? It might be good to include a helper function that is able to transform `type_equivalence_hints` on the unrestricted schema into the hints on the newly-generated restricted schema.

graphql_compiler/schema_transformation/restrict_schema.py
```diff
@@ -0,0 +1,140 @@
...
+    return restricted_schema
+
+
+class RestrictSchemaVisitor(Visitor):
+    """Remove types that are not explicitly kept, and fields go these types."""
+    def __init__(self, types_to_keep, query_type, scalars):
+        """Create a visitor for removing types and fields.
+
+        Args:
+            types_to_keep: Set[str], the set of names of types that we want to keep in the schema
```
pmantica1: Are the types only union, object and interface types? If so please document this here.

graphql_compiler/schema_transformation/restrict_schema.py
qqi0O0: The currently version of the code works in a fairly straightforward way in my opinion, in that exactly those things you specified are kept.
So it's legal to: keep an interface and none of its implementations, keep an implementation and not its interface, keep all implementations of an interface but not the interface itself. In the latter two cases, the 'implements X' is just removed from the implementations that are kept.
The user needs to specify unions in order to keep them. If the user doesn't put the union into `types_to_keep`, then the union will be removed, along with all edges to that union.
Putting random strings or scalar/enum/query type in `types_to_keep` doesn't do anything. It's a little like how the input dictionary into `merge_schemas` works, random strings/scalars/other things that don't get used are just ignored.

graphql_compiler/schema_transformation/restrict_schema.py
qqi0O0: Yes you're right, that's a mistake.

graphql_compiler/schema_transformation/restrict_schema.py
obi1kenobi: Just writing down notes from in-person discussion so I don't forget them.
- A concern was preserving an interface but not all of its implementations, since then a user could query for all instances of that interface and get back objects that are of types that don't exist in the schema.
- Another question was around `types_to_keep` containing names that don't exist in the schema, and whether that should do nothing, log to error/`warnings.warn`, or raise an error and blow up entirely.

pmantica1:
@qqi0O0 Do you mind submitting the branch with the tests if you still have it by any chance?

qqi0O0:
> @qqi0O0 Do you mind submitting the branch with the tests if you still have it by any chance?

Yeah, I can make a draft PR for it. Do note that these tests test the behavior that I was expecting this code to have before chatting with Predrag, so some tests test bad behavior and should be changed.

LWprogramming:
closing for same reason as #834 and #828, the bulk of it has been merged in

Thanks Qi Qi! :)

PR #470, opened 2019-08-14
Sort FoldScopeLocations to ensure predictable output order

LWprogramming:
for multiple folds-- was causing a Python2 test to fail earlier but not Python3, oddly enough.


obi1kenobi:
Python 2 dicts don't have a defined iteration order, but Python 3 dicts are ordered by default.

PR #471, opened 2019-08-14
Class definitions and a few more helpers for split_query

qqi0O0:
Definitions for several classes, plus three helpers that are fairly straightforward and not very closely tied to the complexity of the main code.

### DIFF
--- a/graphql_compiler/schema_transformation/split_query.py
+++ b/graphql_compiler/schema_transformation/split_query.py
@@ -1,0 +1,112 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+from copy import copy
+
+from graphql.language import ast as ast_types
+from graphql.language.visitor import TypeInfoVisitor, Visitor, visit
+from graphql.utils.type_info import TypeInfo
+
+from ..ast_manipulation import get_only_query_definition
+from ..compiler.helpers import strip_non_null_and_list_from_type
+from ..exceptions import GraphQLValidationError
+from ..schema import FilterDirective, OptionalDirective, OutputDirective
+from .utils import (
+    SchemaStructureError, check_query_is_valid_to_split, is_property_field_ast,
+    try_get_ast_by_name_and_type
+)
+
+
+QueryConnection = namedtuple(
+    'QueryConnection', (
+        'sink_query_node',  # SubQueryNode
+        'source_field_out_name',
+        # str, the unique out name on the @output of the the source property field in the stitch
+        'sink_field_out_name',
+        # str, the unique out name on the @output of the the sink property field in the stitch
+    )
+)
+
+
+class SubQueryNode(object):
+    def __init__(self, query_ast):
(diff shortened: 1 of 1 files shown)


graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,102 @@
...
+        sink_field_out_name=child_field_out_name,
+    )
+    new_query_connection_from_child = QueryConnection(
+        sink_query_node=parent_query_node,
+        source_field_out_name=child_field_out_name,
+        sink_field_out_name=parent_field_out_name,
+    )
+    # Add QueryConnections
+    parent_query_node.child_query_connections.append(new_query_connection_from_parent)
+    child_query_node.parent_query_connection = new_query_connection_from_child
```
obi1kenobi: Maybe assert that the child didn't already have a parent query? Bonus points if you can figure out a convenient and efficient way to assert that the parent didn't already have this child as a child too.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+from graphql.language import ast as ast_types
```
obi1kenobi: Wow you wrote this renamed import in a lot of places, huh :)

qqi0O0: Yeah... I can go change all of these at some point. There are possibly merged code with this kind of import too.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -0,0 +1,118 @@
...
+                child_query_node.parent_query_connection
+            )
+        )
+    if any(
+        query_connection_from_parent.sink_query_node is child_query_node
+        for query_connection_from_parent in parent_query_node.child_query_connections
+    ):
+        raise AssertionError(
+            u'The input parent query node already has the child query node in a child query '
+            u'connection.'
```
obi1kenobi: Would be nice to output both the parent and child nodes in both of these assertion error messages, it would greatly help debugging.

obi1kenobi:
I'll merge this in the interest of saving some time, but I do think we should update the error messages to include the info on the basis of which we decided to raise errors.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: Just outputting these nodes isn't super helpful since it'll just print SubQueryNode object at some location. I could include their ASTs in the error message?

obi1kenobi: Right. However, decorating the class with `@six.python_2_unicode_compatible` and then defining `__str__` / `__repr__` methods would be a reasonable fix. The `Location` class in `helpers.py` is a good example of how this would work.

I think the ASTs would be good to have too, but I'd also like the other information, like `schema_id` etc.

PR #473, opened 2019-08-14
Code for splitting query one level recursively

qqi0O0:
This code has most of the complexity of split query. The overall flow is somewhat similar to the structural sharing code in macro systems.

There are two base cases for recursion: 
A property field -- nothing to recurse on, return the input object.
A cross schema vertex field -- in this case, we want to cut off the vertex field, and add in (if necessary) the property field used in the stitch, with an `@output` added if necessary. In this case, return a Field object representing the property field. There's some extra complexity due to the fact that this property field may already exist, in which case we copy over any existing directives on that field. We also need to create a child QueryNode when a branch is cut off.

In the recursive case, we repeat the process on child selections. There's some complexity here as well, since we can't just append the new, possibly changed child selection to our new list of selections. For instance, take the query
```
{
  Human {
    friend {
      stuff
    }
    out_Human_Person {
      stuff
    }
  }
}
```
where out_Human_Person is a cross schema vertex field, stitching on field `id` on Human and field ID on Person.
If we're at the `Human` node, the function is going to be recursively called on its child selections, including `out_Human_Person`. The function called on `out_Human_Person` is going to return a property field representing `id`. We need to insert the new field in the appropriate place in selections, instead of just appending it. In the case that `id` already existed as a field, we need to replace the field rather than insert a new one.

### DIFF
--- a/graphql_compiler/schema_transformation/split_query.py
+++ b/graphql_compiler/schema_transformation/split_query.py
@@ -77,11 +77,20 @@ def _split_query_ast_one_level_recursive(
     """
     type_info.enter(ast.selection_set)
     selections = ast.selection_set.selections
-    if try_get_inline_fragment(selections) is not None:
+
+    type_coercion = try_get_inline_fragment(selections)
+    if type_coercion is not None:
         # Case 1: type coercion
-        new_selections = _split_query_ast_one_level_recursive_type_coercion(
-            query_node, selections, type_info, edge_to_stitch_fields, name_assigner
+        type_info.enter(type_coercion)
+        new_type_coercion = _split_query_ast_one_level_recursive(
+            query_node, type_coercion, type_info, edge_to_stitch_fields, name_assigner
         )
+        type_info.leave(type_coercion)
+
+        if new_type_coercion is type_coercion:
+            new_selections = selections
+        else:
+            new_selections = [new_type_coercion]
     else:
         # Case 2: normal fields
         new_selections = _split_query_ast_one_level_recursive_normal_fields(
@@ -96,45 +105,6 @@ def _split_query_ast_one_level_recursive(
         return new_ast
     else:
         return ast
-
-
-def _split_query_ast_one_level_recursive_type_coercion(
-    query_node, selections, type_info, edge_to_stitch_fields, name_assigner
-):
-    """One case of splitting query, selections contains a single inline fragment.
-
-    _split_query_ast_one_level_recursive will be called recursively on the single selection.
-
-    Args:
-        query_node: SubQueryNode, whose list of child query connections may be modified to
-                    include new children
-        selections: List[InlineFragment], containing a single inline fragment (type coercion)
-        type_info: TypeInfo, used to get information about the types of fields while traversing
-                   the query AST
-        edge_to_stitch_fields: Dict[Tuple(str, str), Tuple(str, str)], mapping
-                               (type name, vertex field name) to
-                               (source field name, sink field name) used in the @stitch directive
-                               for each cross schema edge
-        name_assigner: IntermediateOutNameAssigner, object used to generate and keep track of
-                       names of newly created @output directives
-
-    Returns:
-        List[InlineFragment], containing a single inline fragment, used to replace the
-        list of selections in the SelectionSet one level above. If no changes were made, the
-        exact input list object will be returned
-    """
-    type_coercion = try_get_inline_fragment(selections)
-
-    type_info.enter(type_coercion)
-    new_type_coercion = _split_query_ast_one_level_recursive(
-        query_node, type_coercion, type_info, edge_to_stitch_fields, name_assigner
-    )
-    type_info.leave(type_coercion)
-
-    if new_type_coercion is not type_coercion:
-        return [new_type_coercion]
-    else:
-        return selections
 
 
 def _split_query_ast_one_level_recursive_normal_fields(
@@ -180,7 +150,7 @@ def _split_query_ast_one_level_recursive_normal_fields(
 
     made_changes = False
 
-    # First, collection all property fields, but don't make any changes to them yet
+    # First, collect all property fields, but don't make any changes to them yet
     property_fields_map, vertex_fields = _split_selections_property_and_vertex(selections)
 
     # Second, process cross schema fields. This will modify our record of property fields, and
--- a/graphql_compiler/schema_transformation/split_query.py
+++ b/graphql_compiler/schema_transformation/split_query.py
@@ -1,7 +1,6 @@
 # Copyright 2019-present Kensho Technologies, LLC.
-from collections import namedtuple, OrderedDict
+from collections import OrderedDict, namedtuple
 from copy import copy
-import six
 
 from graphql.language.ast import (
     Argument, Directive, Document, Field, InlineFragment, InterfaceTypeDefinition, Name,
@@ -9,6 +8,7 @@ from graphql.language.ast import (
 )
 from graphql.language.visitor import TypeInfoVisitor, Visitor, visit
 from graphql.utils.type_info import TypeInfo
+import six
 
 from ..ast_manipulation import get_only_query_definition
 from ..compiler.helpers import strip_non_null_and_list_from_type
@@ -47,19 +47,15 @@ class SubQueryNode(object):
 
 
 def _split_query_ast_one_level_recursive(
-    query_node, ast, type_info, edge_to_stitch_fields, intermediate_out_name_assigner
+    query_node, ast, type_info, edge_to_stitch_fields, name_assigner
 ):
     """Return an AST node with which to replace the input AST in the selections that contain it.
 
-    This function examines the selections of the input AST, and divides them into three sets:
-    property fields, intra-schema vertex fields, and cross-schema vertex fields.
-
-    Each cross-schema vertex field will have its branch removed from the output AST, and made
-    into its own separate query AST. The parent and child property fields used in
-    the stitch will be added to the parent and child ASTs. @output directives will be added
-    to the parent and child property fields, if not already present.
-
-    The function will be called recursively on each intra-schema vertex field.
+    This function examines the selections of the input AST, and recursively calls either
+    _split_query_ast_one_level_recursive_type_coercion or
+    _split_query_ast_one_level_recursive_normal_fields
+    depending on whether the selections contains a single InlineFragment or a number of normal
+    fields.
 
     Args:
         query_node: SubQueryNode, whose list of child query connections may be modified to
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+    - If the input AST starts with a cross schema vertex field, the output will be a Field object
+      representing the property field used in the stitch
+      - If a property field of the expected name already exists in the selections one level
+        above (in parent_selections), the new Field will contain any existing directives of
+        this field
+    - Otherwise, the process will be repeated on child selections (if any) of the input AST
+      - If no child selection is modified, the exact same object as the input AST will be returned
+      - If some child selection is modified, a copy of the AST with the new selections will be
+        returned
+        - If a modified child selection is a property field
```
obi1kenobi: `If a` -> `If any` would make this more precise, I believe.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+          - If a property field of the same name already exists, the new field will replace the
+            existing field
```
obi1kenobi: If I understand correctly, this could use the addition of `, while preserving any directives from the existing field`

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+          - Otherwise, the property field will be inserted into selections after the last existing
+            property field
+        - Otherwise, the modified selection will be appended to selections
```
obi1kenobi: Something I suggested to @LWprogramming yesterday might be helpful here as well. It might be a good idea to number the various steps in this process, and reference those step numbers in the code so that the reader can easily keep track of which piece of code is doing which part of the described process. Otherwise, it's very easy to get lost, both in the description and in the code.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+
+    Args:
+        query_node: SubQueryNode, whose list of child query connections may be modified to include
+                    children
+        ast: type Node, the AST that we are trying to split into child components. It is not
```
obi1kenobi: Presumably this `Node` here is only things like `Field, InlineFragment, OperationDefinition` and not all subclasses of `Node`, right?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+             modified by this function
+        parent_selections: List[Node], containing all property fields (and possibly some vertex
+                           fields) in the level of selections that contains the input ast. It
+                           is not modified by this function
+        type_info: TypeInfo, used to get information about the types of fields while traversing
+                   the query ast
+        edge_to_stitch_fields: Dict[Tuple(str, str), Tuple(str, str)], mapping
```
obi1kenobi: Do we have a descriptor object that contains this data already and that we can pass around instead?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+        is unchanged from the input AST, the exact same object will be returned
+    """
+    # Check if there is a split here. If so, split AST, make child query node, return property
+    # field. If not, recurse on all child selections (if any)
+    if isinstance(ast, Field):
+        parent_type_name = type_info.get_parent_type().name
+        edge_field_name = ast.name.value
+        if (parent_type_name, edge_field_name) in edge_to_stitch_fields:
+            # parent_field_name and child_field_name are names of property fields in the stitch
+            parent_field_name, child_field_name = \
```
obi1kenobi: nit: avoid backslashes, per the Google Python style guide that we broadly follow: http://google.github.io/styleguide/pyguide.html#32-line-length

> Make use of Python’s implicit line joining inside parentheses, brackets and braces. If necessary, you can add an extra pair of parentheses around an expression.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+            # Create child query node around ast
+            child_query_node, child_output_name = _get_child_query_node_and_out_name(
+                ast, type_info, child_field_name, intermediate_out_name_assigner
+            )
+            # Create and add QueryConnections
+            _add_query_connections(
+                query_node, child_query_node, parent_output_name, child_output_name
+            )
+            # Return parent property field used in the stitch, to be added in selections above
+            return parent_property_field
```
obi1kenobi: Since this ends in a `return`, it might be a good idea to extract everything after the `if` on line 91 into a function, so that we get something nice and simple and fairly obvious like:
```python
if (parent_type_name, edge_field_name) in edge_to_stitch_fields:
    return foo()
```

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+
+    # No split here
+    if ast.selection_set is None:  # Property field, nothing to recurse on
```
obi1kenobi: Should this be a call to `is_property_field_ast` instead?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+        )
+
+        if new_selection is not selection:
+            made_changes = True
+            if is_property_field_ast(new_selection):
+                # If a property field is returned and is different from the input, then this is
+                # a property field used in stitching. If no existing field has this name, insert
+                # the new property field to end of property fields. If some existing field has
+                # this name, replace the existing field with the returned field
+                new_selections = _replace_or_insert_property_field(new_selections, new_selection)
```
obi1kenobi: Might it be easier to do this by just keeping track of the names of property fields we've already seen? This wholesale reassignment of `new_selections` is making me a bit nervous.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+        child_field_name: str. If no field of this name currently exists as a part of the root
+                          selections of the input AST, a new field will be created in the AST
+                          contained in the output child query node
+        intermediate_out_name_assigner: IntermediateOutNameAssigner, which will be used to
+                                        generate an out_name, if it's necessary to create a
+                                        new @output directive
+
+    Returns:
+        Tuple[SubQueryNode, str], the child sub query node wrapping around the input ast, and
+        the out_name of the @output directive uniquely identifying the field used or stitching
```
obi1kenobi: typo: `field used or stitching` seems to be missing an `f`

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+    child_property_field = _get_property_field(child_selections, child_field_name, [])
+    # Add @output if needed, record out_name
+    child_output_name = _get_out_name_optionally_add_output(
+        child_property_field, intermediate_out_name_assigner
+    )
+    # Get new child_selections
+    child_selections = _replace_or_insert_property_field(child_selections, child_property_field)
+    # Wrap around
+    # NOTE: if child_type_name does not actually exist as a root field (not all types are
+    # required to have a corresponding root vertex field), then this query will be invalid
```
obi1kenobi: Nice catch! Is there some validation we can perform when merging schemas to ensure this is impossible at splitting time?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+                )
+            elif directive.name.value == OptionalDirective.name:
+                if try_get_ast_by_name_and_type(
+                    new_field.directives, OptionalDirective.name, Directive
+                ) is None:
+                    # New optional directive
+                    new_field.directives.append(directive)
+            elif directive.name.value == FilterDirective.name:
+                new_field.directives.append(directive)
+            else:
```
obi1kenobi: Could `@fold` or `@output_source` make it to here?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+        ast: type Node, the AST that, if split off into its own Document, would have the the
+             type (named the same as the root vertex field) and root selections that this
+             function outputs
+        type_info: TypeInfo, used to check for the type that fields lead to
+
+    Returns:
+        Tuple[str, List[Union[Field, InlineFragment]]], name and selections of the root
+        vertex field of the input AST if it were made into its own separate Document
+    """
+    child_type = type_info.get_type()  # GraphQLType
```
obi1kenobi: It doesn't seem like `type_info` is actually used in this function outside of this line. Perhaps we can just send in the `child_type` value directly, and not make this function worry about how we actually know the type?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
...
+            u'corresponding to the field "{}" under type "{}" cannot be '
+            u'found.'.format(ast.name.value, type_info.get_parent_type())
+        )
+    child_type_name = strip_non_null_and_list_from_type(child_type).name
+
+    child_selection_set = ast.selection_set
+    # Adjust for type coercion
+    if (
+        child_selection_set is not None and
+        len(child_selection_set.selections) == 1 and
```
obi1kenobi: What happens if someone passes an illegal query like:

```
{
    Foo {
        out_Foo_Bar {
            name
            ... on Baz {
                name
            }
        }
    }
}
```
or
```
{
    Foo {
        out_Foo_Bar {
            ... on Baz {
                name
            }
            ... on Quux {
                name
            }
        }
    }
}
```

Is this caught by any validation prior to this step?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,392 @@ def __init__(self, query_ast):
...
+             modified by this function
+        child_field_name: str. If no field of this name currently exists as a part of the root
+                          selections of the input AST, a new field will be created in the AST
+                          contained in the output child query node
+        intermediate_out_name_assigner: IntermediateOutNameAssigner, which will be used to
+                                        generate an out_name, if it's necessary to create a
+                                        new @output directive
+
+    Returns:
+        Tuple[SubQueryNode, str], the child sub query node wrapping around the input ast, and
```
obi1kenobi: nit: `ast` -> `AST` since "input ast" doesn't seem to be referring to a function parameter

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +39,340 @@ def __init__(self, query_ast):
         # List[SubQueryNode], the queries that depend on the current query
 
 
+def _split_query_ast_one_level_recursive(
+    query_node, ast, parent_selections, type_info, edge_to_stitch_fields,
+    intermediate_out_name_assigner
+):
+    """Return a Node to replace the input AST by in the selections one level above.
```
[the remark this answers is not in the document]
qqi0O0: Changed the doc strings to be more clear, and changed mentions of Node (which is a class - the base AST component class) to be more specific.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: The MergedSchemaDescriptor currently doesn't contain this data. Right now, this dict is computed once from the ATS of the input MergedSchemaDescriptor then passed around.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: If all object and interface types have a corresponding root vertex field, then this is impossible. But I'm not sure if that's true or should be enforced.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: Currently, the validation step only allows output, filter, and optional to be in the AST.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: I'll fix this by adding in validation that inline fragments are alone in their scope.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,392 @@ def __init__(self, query_ast):
...
+        edge_to_stitch_fields: Dict[Tuple(str, str), Tuple(str, str)], mapping
+                               (type name, vertex field name) to
+                               (source field name, sink field name) used in the @stitch directive
+                               for each cross schema edge
+        intermediate_out_name_assigner: IntermediateOutNameAssigner, object used to generate
+                                        and keep track of names of newly created @output
+                                        directives
+    """
+    # Split selections into three kinds: property fields, cross schema vertex fields, normal
+    # vertex fields/type coercions
```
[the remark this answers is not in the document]
qqi0O0: You're right, I didn't do a special case before because I thought it's possible to have type coercions coexist with other fields and type coercions. Will do a special case now.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,392 @@ def __init__(self, query_ast):
...
+    )
+    # Create and add QueryConnections
+    _add_query_connections(
+        query_node, child_query_node, parent_output_name, child_output_name
+    )
+    # Add or replace the new property field
+    property_fields_map[parent_property_field.name.value] = parent_property_field
+
+
+def _split_selections_property_and_vertex(selections):
```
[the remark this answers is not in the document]
qqi0O0: Interesting, that function splits based on the name of the field, this function splits based on whether or not the field has further selections.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,392 @@ def __init__(self, query_ast):
...
+        _process_cross_schema_field(
+            query_node, cross_schema_field, property_fields_map, child_type_name,
+            parent_field_name, child_field_name, intermediate_out_name_assigner
+        )
+        type_info.leave(cross_schema_field)
+
+    # Then, process intra schema edges by recursing on them
+    new_intra_schema_fields = []
+    for intra_schema_field in intra_schema_fields:
+        type_info.enter(intra_schema_field)
```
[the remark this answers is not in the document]
qqi0O0: I probably won't have time to get to this PR, unfortunately. For now I've refactored the code a bit, so that it's hopefully easier to see the pairs of enter/leave.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,478 @@ def __init__(self, query_ast):
...
+):
+    """Return an AST node with which to replace the input AST in the selections that contain it.
+
+    This function examines the selections of the input AST, and recursively calls either
+    _split_query_ast_one_level_recursive_type_coercion or
+    _split_query_ast_one_level_recursive_normal_fields
+    depending on whether the selections contains a single InlineFragment or a number of normal
+    fields.
+
+    Args:
```
obi1kenobi: nit: this function returns stuff, but there's no `Returns:` section

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,478 @@ def __init__(self, query_ast):
...
+        # Case 1: type coercion
+        new_selections = _split_query_ast_one_level_recursive_type_coercion(
+            query_node, selections, type_info, edge_to_stitch_fields, name_assigner
+        )
+    else:
+        # Case 2: normal field
+        new_selections = _split_query_ast_one_level_recursive_normal_fields(
+            query_node, selections, type_info, edge_to_stitch_fields, name_assigner
+        )
+    type_info.leave(ast.selection_set)
```
obi1kenobi: Very clean, I like it!

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,478 @@ def __init__(self, query_ast):
...
+    property_fields_map, vertex_fields = _split_selections_property_and_vertex(selections)
+    intra_schema_fields, cross_schema_fields = _split_vertex_fields_intra_and_cross_schema(
+        vertex_fields, parent_type_name, edge_to_stitch_fields
+    )
+
+    made_changes = False
+    if len(cross_schema_fields) > 0:
+        made_changes = True
+
+    # First, no changes will be made to property fields
```
obi1kenobi: I'd move this to right above line 184 and phrase it as `First, collect all property fields but don't make any changes to them yet` or something to that effect

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,478 @@ def __init__(self, query_ast):
...
+    # Second, process cross schema fields. This will modify our record of property fields, and
+    # create child SubQueryNodes attached to the input SubQueryNode
+    for cross_schema_field in cross_schema_fields:
+        type_info.enter(cross_schema_field)
+        child_type = type_info.get_type()
+        if child_type is not None:
+            child_type_name = strip_non_null_and_list_from_type(child_type).name
+        else:
+            raise AssertionError(
+                u'The query may be invalid against the schema, causing TypeInfo to lost track '
```
obi1kenobi: typo: `causing TypeInfo to lost track` I think `lost -> lose`?

For bonus points, cross-schema field and maybe the query node at which this happened? Hitting this error would be quite scary and all the debug output we can provide would be helpful.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -34,6 +46,478 @@ def __init__(self, query_ast):
...
+        Tuple[SubQueryNode, str], the child sub query node wrapping around the input AST, and
+        the out_name of the @output directive uniquely identifying the field used for stitching
+        in this sub query node
+    """
+    # Get type and selections of child AST, taking into account type coercions
+    child_selection_set = ast.selection_set
+    if (
+        len(child_selection_set.selections) == 1 and
+        isinstance(child_selection_set.selections[0], InlineFragment)
+    ):
```
obi1kenobi: Consider making a helper function for this check, something like `_has_single_fragment_selection` maybe -- possibly even better if it returns the fragment selection AST so you can avoid the `[0]` below. It seems to be repeated a few times.

Maybe you could even use the `_get_inline_fragment(ast)` function from `compiler_frontend.py`?

graphql_compiler/schema_transformation/split_query.py
obi1kenobi: I think it's probably a good idea to enforce that if a given cross-schema edge is attempted to be defined, that its endpoint types and their subtypes exist in the corresponding schemas' root query types.

graphql_compiler/schema_transformation/split_query.py
qqi0O0: `_get_inline_fragment` does the right thing, although it raises an GraphQLCompilationError which I don't like very much -- would prefer GraphQLValidationError. You're definitely right I'm repeating myself here somewhat, I'll use that function for now.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -122,11 +123,7 @@ def _split_query_ast_one_level_recursive_type_coercion(
         list of selections in the SelectionSet one level above. If no changes were made, the
         exact input list object will be returned
     """
-    if len(selections) != 1:
-        raise AssertionError(u'Selections {} should only contain one element'.format(selections))
-    type_coercion = selections[0]
-    if not isinstance(type_coercion, InlineFragment):
-        raise AssertionError(u'Selection {} should be an inline fragment.'.format(type_coercion))
+    type_coercion = try_get_inline_fragment(selections)
```
obi1kenobi: Seems like here we implicitly expect `type_coercion` to not be `None`.

But when we call this function, we already run `try_get_inline_fragment(selections)`, though we just check the result for `None`. So how about we do a small refactor and accept a `type_coercion_ast` argument here, and pass the `try_get_inline_fragment(selections)` result from the caller into here. That way we don't have to worry about accidentally getting a bad `selections` object. We could then return only the new (or old) type coercion, and handle the selections-magic in the caller.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -181,35 +178,36 @@ def _split_query_ast_one_level_recursive_normal_fields(
...
-    intra_schema_fields, cross_schema_fields = _split_vertex_fields_intra_and_cross_schema(
-        vertex_fields, parent_type_name, edge_to_stitch_fields
-    )
-
     made_changes = False
-    if len(cross_schema_fields) > 0:
-        made_changes = True
 
-    # First, no changes will be made to property fields
+    # First, collection all property fields, but don't make any changes to them yet
```
obi1kenobi: typo: `collection -> collect`?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -98,45 +107,6 @@ def _split_query_ast_one_level_recursive(
         return ast
 
 
-def _split_query_ast_one_level_recursive_type_coercion(
```
obi1kenobi: Very nice!

graphql_compiler/schema_transformation/split_query.py
qqi0O0: There's very little code happening in the type coercion case, so I just got rid of the extra function and now do stuff directly in the caller.

PR #474, opened 2019-08-14
Cypher fold and optional

LWprogramming:
PR for handling a weird edge case where vertices get named the wrong thing when optional is at the same scope level as fold. For instance: see `test_optional_and_fold` in `test_compiler.py`

### DIFF
--- a/graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
+++ b/graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
@@ -7,7 +7,8 @@ from ..blocks import CoerceType, Filter, Fold, MarkLocation, Recurse, Traverse
 from ..compiler_entities import Expression
 from ..expressions import BinaryComposition, ContextField, LocalField, NullLiteral
 from ..helpers import (
-    FoldScopeLocation, get_only_element_from_collection, is_graphql_type, validate_safe_string
+    FoldScopeLocation, Location, get_only_element_from_collection, is_graphql_type,
+    validate_safe_string
 )
 from ..ir_lowering_common.common import merge_consecutive_filter_clauses
 from ..ir_lowering_common.location_renaming import (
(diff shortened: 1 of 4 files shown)


LWprogramming:
to be specific: every time an optional traversal returns to its original location, the `visit_counter` for that `Location` gets incremented. This means that for a given `@fold` directive, the `visit_counter` is 1 + the number of optional traversals at the same level in the same scope.

For example, right now if we compiled this query:

```python
graphql_input = '''{
    Animal {
            in_Animal_ParentOf {
                name @output(out_name: "a")
            }
        }
        out_Animal_OfSpecies @fold {
            in_Animal_OfSpecies {
                name @output(out_name: "b")
            }
        }
        out_Animal_ParentOf {
            out_Animal_ParentOf @optional {
                name @output(out_name: "c")
            }
        }
        out_Animal_LivesIn @fold {
            in_Animal_LivesIn {
                name @output(out_name: "d")
            }
        }
    }
}'''
```

will have all its fold-scope vertices with `visit_counter` = 2 = 1 +  1 for the `in_Animal_ParentOf` optional traversal, but not for the `out_Animal_ParentOf` optional traversal.

obi1kenobi:
Don't you love it when there's a good input test suite already? :)

Be careful, though:
> This means that for a given @fold directive, the visit_counter is 1 + the number of optional traversals at the same level in the same scope.
This is not always true. Check the `optional_traverse_after_mandatory_traverse` input case for an example with two traversals (one required, one optional) where the visit counter ends up set to 3 instead of 2. If we were to add a `@fold` after the optional traversal, your statement claims that the visit counter would be 2, but it would actually be 3.

obi1kenobi:
This seems fine. What's left until it's ready for review?

LWprogramming:
It's ready! Was just waiting for the tests to pass

graphql_compiler/compiler/ir_lowering_cypher/ir_lowering.py
```diff
@@ -272,3 +273,47 @@ def move_filters_in_optional_locations_to_global_operations(cypher_query, query_
...
+                            .format(type(block_location), block_location))
+
+    for block in ir_blocks:
+        if not isinstance(block, MarkLocation):
+            new_ir_blocks.append(block)
+            continue
+        if block.location in locations_marked:
+            continue
+        location = get_block_base_location(block.location)
+        location.visit_counter = 1
```
obi1kenobi: This mutates the location object, which is a bad idea. I suspect that because of sharing of location objects, this code happens to accidentally pass tests, but in general this wouldn't work because it doesn't rewrite locations used within `Expression` objects, e.g. in `Filter` or `ConstructResult` blocks.

PR #475, opened 2019-08-14
Estimates for inequality filters on UUID

ReaLNeroM:
Since the upcoming query pagination feature requires better estimates when using inequality ('>', '>=' etc.) filters on UUIDs from the cost estimator, I'm releasing this PR.

This PR adds functionality for estimating filter selectivities of inequality filters on UUID fields. So now, a 'between' filter on fields called 'uuid' will be estimated. 

We also have to take on a new assumption. Let the number of valid UUIDs be `u`, and the number of UUIDs passing through the filter be `f`. Then the field selectivity will be `max(0, f/u)`.

VERY IMPORTANT NOTE: Since we've assumed that all filters are independent of one another, we generally compute the product of each filter's selectivity. What does this mean for this PR? It means that a `between` filter could have a different result from a pair of `>=` and `<=` filters. 

Relaxing the independence assumption between filters would require a large refactoring effort. I'm wary of modifying much.

I've decided to include tests in a separate PR, as the amount of edge cases make want to test this out very liberally, which would cause a lot of diff lines.


graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+        raise AssertionError(u'Cost estimator found unsupported integer inequality operator {}'
+                             .format(filter_operator))
+
+    # If the interval of values passing through the filters is empty, no results will be
+    # returned. This happens if the interval's upper bound is smaller than the lower bound.
+    if query_interval[0] > query_interval[1]:
+        field_selectivity = Selectivity(kind=ABSOLUTE_SELECTIVITY, value=0.0)
+        return field_selectivity
+
+    if not domain_interval[0] <= query_interval[0] <= query_interval[1] <= domain_interval[1]:
```
ReaLNeroM: This is provably impossible to trigger, but I want to leave it in in case this code gets changed later. Thoughts?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -15,6 +17,11 @@
 ABSOLUTE_SELECTIVITY = 'absolute'
 FRACTIONAL_SELECTIVITY = 'fractional'
 
+INEQUALITY_OPERATORS = frozenset(['<', '<=', '>', '>=', 'between'])
+
+MIN_UUID_INT = 0
+MAX_UUID_INT = 2**128-1
```
LWprogramming: where do these numbers come from?

obi1kenobi: UUIDs are 128-bit numbers: https://en.wikipedia.org/wiki/Universally_unique_identifier

ReaLNeroM: https://en.wikipedia.org/wiki/Universally_unique_identifier
I couldn't find a `MAX_UUID` constant, so I had to make do with harcoding it.
Once we decide to implement histograms (mentioned in a TODO), we can just use histograms instead to inform ourselves about min and max values, and forgo these harcoded variables.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
     return False
 
 
+def _try_obtain_parameter_value(argument_name, parameters):
```
obi1kenobi: nit: prefixing the name with `try` is a common existing convention for functions where `None` or `False` is returned upon failure, and I think this would be confusing to the reader. I'd suggest trying to find a different name.

Here is an example function that uses a `try` prefix per the convention I mentioned: https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=netframework-4.8

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+    """Return value of argument in parameters dict or raise ValueError."""
+    parameter_name = get_parameter_name(argument_name)
+    if parameter_name not in parameters:
```
LWprogramming: Could we just attempt to access `parameters[parameter_name]` and let Python raise a `KeyError` if the parameter name isn't found? Would eliminate the need for this function

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+        if field_name == 'uuid':
+            # Instead of working with UUIDs, we convert each occurence of UUID to its corresponding
+            # integer representation.
+            uuid_domain = (MIN_UUID_INT, MAX_UUID_INT)
+
+            parameter_values = tuple(
+                _try_obtain_parameter_value(filter_argument, parameters)
+                for filter_argument in filter_info.args
+            )
+            parameter_values_as_integers = tuple(
```
LWprogramming: could we convert these two tuples into a single tuple? e.g. chaining the function calls.

ReaLNeroM: It starts looking like badly-written Lisp really quick though: ` _convert_uuid_string_to_int(parameters[get_parameter_name(filter_argument)])`.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+        filter_operator: str, describing the inequality filter operation being performed.
+
+    Returns:
+        Selectivity object, describing the selectivity of the integer inequality filter.
+    """
+    # Inequality filters on integers can also be represented by an interval for the subset of values
+    # that pass through the filter. By representing the filter as an interval, we avoid considering
+    # the operators separately.
+    query_interval = (domain_interval[0], domain_interval[1])
+    if filter_operator == '>':
```
obi1kenobi: All the`[0]` and `[1]` here are making this code quite difficult to read and understand. I'd suggesting splitting the code into two smaller functions, one that handles binary operators like `>/>=/</<=`, and one for `between` (and potentially other ternary operators in the future). Then, you can do `parameter_values[0]` with an assert in the first case, and `low_value, high_value = parameter_values` for the second case.

With that structure, the reader only has to verify that the `min` and `max` calls contain the same thing on the left-hand side of the `=` as they do inside the `min/max` call.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+        filter_info: FilterInfo object, filter on the location being filtered
+        parameters: dict, parameters with which query will be executed
+        location_name: string, type of the location being filtered
+
+    Returns:
+        Selectivity object, the selectivity of a specific inequality filter at a given location.
+    """
+    filter_operator = filter_info.op_name
+    if filter_operator not in INEQUALITY_OPERATORS:
+        raise ValueError(u'Inequality filter selectivity estimator received a filter'
```
obi1kenobi: `filter` and `with` don't have any whitespace between them, so this will end up saying `received a filterwith non-inequality`

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: Good point.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +51,138 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+                         u'with non-inequality filter operator {}: {} {}'
+                         .format(filter_operator, filter_info, location_name))
+
+    all_selectivities = []
+    for field_name in filter_info.fields:
+        field_selectivity = Selectivity(kind=FRACTIONAL_SELECTIVITY, value=1.0)
+
+        # TODO(vlad): Improve inequality estimation by implementing histograms.
+        if field_name == 'uuid':
```
obi1kenobi: I'd also add a `HACK(vlad)` note here so we know this is hardcoded for now.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: I've since deleted this function, since Leon mentioned how we didn't need this.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -12,9 +13,24 @@
                 # fraction of results that pass the filter
 ))
 
+# IntegerInterval is used to denote a continuous non-empty interval of integers.
+IntegerInterval = namedtuple('IntegerInterval', (
+    'lower_bound',    # optional int, inclusive lower bound of integers in the interval.
+                      # Intervals that do not have a lower bound denote this by using None.
+    'upper_bound',    # optional int, inclusive upper bound of integers in the interval.
+                      # Intervals that do not have an upper bound denote this by using None.
+))
```
bojanserafimov: I like this definition. You could have used a tuple, but then you'd have to document what it means everywhere :)

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -26,6 +42,58 @@ def _is_fractional(selectivity):
     return selectivity.kind == FRACTIONAL_SELECTIVITY
 
 
+def _get_intersection_of_IntegerIntervals(interval_a, interval_b):
```
bojanserafimov: There's a lot of complexity and room for error in the implementation. Is it simpler to define `stronger_lower_bound` and `sronger_upper_bound` functions, and return `(stronger_lower_bound(a.lower_bound, b.lower_bound), stronger_upper_bound(a.upper_bound, b.upper_bound)`, making sure to return `None` if the resulting interval is empty? Feel free to argue if it doesn't look simpler to you.

obi1kenobi: Let's stick to `snake_case` for naming. You could even make this function be named `_get_intersection_of_intervals` and even that would be okay.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +112,216 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
     return False
 
 
+def _convert_uuid_string_to_int(uuid_string):
+    """Return the integer representation of a UUID string."""
+    return UUID(uuid_string).int
+
+
+def _is_IntegerInterval_empty(interval):
+    """Return True if the given interval is empty."""
```
bojanserafimov: This is a bit confusing. An `IntegerInterval` represents a non-empty range, so of course it's not empty. I see a couple of ways to resolve this:
1. Rephrase this function as a validation, that checks if an `IntegerInterval` is valid. I'm not a fan of this, because the name would become more obscure. Also you're not using it for that purpose.
2. Remove the restriction that an interval has to be non-empty, and feel free to pass around (5, 3) as a valid interval containing 0 integers. I'm also not a fan because you create a foot-gun. Someone might mistakenly use `upper_bound - lower_bound + 1` as a size formula.
3. Make a `make_integer_interval(lower_bound, upper_bound)` that returns `None` if the interval would be empty.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -130,6 +408,13 @@ def _get_filter_selectivity(
                 # The estimate may be above 1.0 in case of duplicates in the collection
                 # so we make sure the value is <= 1.0
             )
+    elif filter_info.op_name in INEQUALITY_OPERATORS:
+        # TODO(vlad): Since we assume each filter is independent, we don't consider the correlation
+        #             inequality filters often have. badly. For example, a 'between' filter and a
```
obi1kenobi: I don't get the `badly` here. Update for grammar?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +112,216 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+
+    intersection = _get_intersection_of_IntegerIntervals(domain_interval, query_interval)
+
+    # If the interval of values passing through the filters is empty, no results will be
+    # returned. This happens if the interval's upper bound is smaller than the lower bound.
+    if intersection is None:
+        field_selectivity = Selectivity(kind=ABSOLUTE_SELECTIVITY, value=0.0)
+        return field_selectivity
+
+    if not domain_interval[0] <= intersection[0] <= intersection[1] <= domain_interval[1]:
```
bojanserafimov: you can use `.lower_bound`

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +112,216 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+    The larger the intersection relative to the domain interval, the higher the selectivity. For
+    example, if the intersection is half as big as the domain interval, this will correspond to a
+    Fractional Selectivity with a selectivity factor of 0.5 (i.e. roughly 50% of elements pass
+    through the filter).
+
+    Args:
+        domain_interval: IntegerInterval namedtuple, describing the finite interval of integers
+                         being filtered.
+        parameter_values: List[int], describing the parameters for the inequality filter.
+        filter_operator: str, describing the inequality filter operation being performed.
```
bojanserafimov: copy-paste the arg specs to all the helper functions. It really isn't clear what they mean just by their names.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: Thanks for the list, it was very helpful. I've decided on 3).

graphql_compiler/cost_estimation/filter_selectivity_utils.py
ReaLNeroM: Thanks for the suggestion!

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -44,6 +60,359 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
+    Args:
+        parameter_values: List[int], describing the parameters for the inequality filter.
+        filter_operator: str, describing the binary inequality filter operation being performed.
+
+    Returns:
+        - IntegerInterval namedtuple, non-empty interval of values that pass through the filter.
+        - None if the interval is empty.
+
+    Raises:
+        - AssertionError if the given operator is unsupported.
```
bojanserafimov: I'm not sure if others would agree, but I think documenting assertion errors doesn't make much sense:
1. An assertion error is raised when the preconditions aren't met. You're not required to document undefined behavior.
2. Nobody needs this documentation. Nobody should be catching assertion errors.
3. If I had to document every assertion error I raise, I might be discouraged to make assertions. Python code I've seen most often has fewer assertions that I'd like it to.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -130,6 +499,15 @@ def _get_filter_selectivity(
                 # The estimate may be above 1.0 in case of duplicates in the collection
                 # so we make sure the value is <= 1.0
             )
+    elif filter_info.op_name in INEQUALITY_OPERATORS:
+        # TODO(vlad): Since we assume each filter is independent, we don't consider the correlation
+        #             inequality filters often have. For example, a 'between' filter and an
+        #             equivalent pair of '>=' and '<=' are estimated differently. Consult the
+        #             FilterSelectivityTests/test_inequality_filters_on_uuid function for further
+        #             information.
```
bojanserafimov: Just checking if I get this: if you have a uuid filter `> median - 1` and another `< median + 1`, the correct result is `fractional selectivity 1/2^128`, but since you process and combine selectivities separately, you'd get `fractional selectivity 1/4`.

I'm not arguing we should fix it. It's definitely a good idea to leave it as a TODO for now.

ReaLNeroM: Yep, exactly. Fixing this requires revising one of our assumptions _filter selectivities combine independently_.

PR #485, opened 2019-08-15
Helpers and example schemas for testing split_query

qqi0O0:
See the draft PR on split_query tests to get a sense of how the helpers `_check_query_node_structure` and `_get_intermediate_outputs_set` are used.

### DIFF
--- a/graphql_compiler/tests/schema_transformation_tests/test_split_query.py
+++ b/graphql_compiler/tests/schema_transformation_tests/test_split_query.py
@@ -3,6 +3,7 @@ from collections import namedtuple
 import unittest
 
 from graphql import print_ast
+import six
 
 
 # The below namedtuple is used to check the structure of SubQueryNodes in tests
--- a/graphql_compiler/tests/schema_transformation_tests/test_split_query.py
+++ b/graphql_compiler/tests/schema_transformation_tests/test_split_query.py
@@ -1,15 +1,8 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from collections import namedtuple
-from textwrap import dedent
 import unittest
 
-from graphql import parse, print_ast
-
-from ...exceptions import GraphQLValidationError
-from ...schema_transformation.split_query import split_query
-from .example_schema import (
-    basic_merged_schema, interface_merged_schema, three_merged_schema, union_merged_schema
-)
+from graphql import print_ast
 
 
 # The below namedtuple is used to check the structure of SubQueryNodes in tests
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/schema_transformation_tests/test_split_query.py
```diff
@@ -0,0 +1,63 @@
...
+        # List[Tuple[ExampleQueryNode, str, str]]
+        # child example query node, parent out name, child out name
+    )
+)
+
+
+class TestSplitQuery(unittest.TestCase):
+    def _check_query_node_structure(self, root_query_node, root_example_query_node):
+        """Check root_query_node has no parent and has the same structure as the example input."""
+        self.assertIs(root_query_node.parent_query_connection, None)
```
obi1kenobi: I think `assertIsNone` is also available and probably a better choice here

graphql_compiler/tests/schema_transformation_tests/test_split_query.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _check_query_node_structure_helper(self, query_node, example_query_node):
+        """Check query_node has the same structure as example_query_node."""
+        # Check AST and id of the parent
+        self.assertEqual(print_ast(query_node.query_ast), example_query_node.query_str)
+        self.assertEqual(query_node.schema_id, example_query_node.schema_id)
+        # Check number of children matches
+        self.assertEqual(len(query_node.child_query_connections),
+                         len(example_query_node.child_query_nodes_and_out_names))
+        for i in range(len(query_node.child_query_connections)):
```
obi1kenobi: Consider instead doing something like `for i, (child_query_connection, expected_data) in enumerate(six.moves.zip(query_node.child_query_connections, example_query_node.child_query_nodes_and_out_names)):`

Then you get to avoid much of the `[i]` and destructuring below. Also consider assigning the values that get passed to `six.moves.zip` to some local variables, just to save on some typing and line complexity.

graphql_compiler/tests/schema_transformation_tests/test_split_query.py
```diff
@@ -0,0 +1,63 @@
...
+            # Check child and parent connections
+            child_query_connection = query_node.child_query_connections[i]
+            child_query_node = child_query_connection.sink_query_node
+            child_example_query_node, parent_out_name, child_out_name = \
```
obi1kenobi: Use parens instead of backticks, here and elsewhere in the PR.

PR #486, opened 2019-08-16
Change documentation to use restructuredtext format

pmantica1:
I am adding a restructured text version of the README.md that I automatically generated using a library called pandoc. The conversion is not perfect, but it is done pretty well. To see what it looks like run: 
cd docs
make html 
google-chrome build/html/README.html 

I am adding this file because it will come in handy when copy-pasting in sections of the README.rst into readthedocs. I can also delete sections of the README.rst file when adding them into readthedocs to keep track of what is left, (though we do not necessarily want to add all of the REAME.md content into readthedocs).  

At this point it is worth mentioning my plan from this point forward for adding readthedocs:
1. This PR 
2. PR with an outline of how readthedocs will be structured. 
3. PRs filling outline through mostly copy-paste material.
4. Release readthedocs 

I do not include the restructured text versions of the CONTRIBUTING.md and the  TROUBLESHOOTING.md files. These files are related to the development of the compiler and they should not be included in readthedocs. Instead, they should be kept as is and referenced by the finalized README.md file.

The README.md file is the same one as the one referenced by the commit hash in the CONTRIBUTING.md file since it has not had any changes since that commit.  

I remove the GraphQL syntax highlighting because there is no GraphQL lexer in pygments which is the library that Sphinx uses for syntax highlighting.


obi1kenobi:
I don't think we should have two READMEs, because who will keep them in sync?

You mentioned it's not perfect. Can you work to make it as good as you can, so then we can make it the default one and remove the other one?

pmantica1:
I think that we should have a temporary RST version of the README while still keeping the README.md for now. I think that you might understand better why if I elaborate a bit more on my plan to add readthedocs. 

1. This PR where I add an automatically generated version of the README in restructured text. This 
    PR is too large to verify that the conversion was done completely correctly. Instead, when we make 
    smaller adding the chunks of it to readthedocs, I'll verify that the restructured text is correctly 
    formatted added there.    
2. PR with an outline of how readthedocs will be structured.
3. PRs filling outline through mostly copy-paste material. Every time I add something into the 
     readthedocs, I'll remove the corresponding part from the README.rst copy in order to keep track 
     of what is left to add. 
4. Once there is no more left to add from the README.rst file, I'll do a diff from the markdown file in 
     the commit that is annotated on the CONTRIBUTING.md file to see what other things are left to 
     add in readthedocs.
6. "Atomically" release readthedocs and  delete the temporary README.rst file and most of the 
     README.md. In the README.md file, and we'll keep a short introduction and links to 
    CONTRIBUTING.md, TROUBLESHOOTING.md and readthedocs.

pmantica1:
This PR now converts all of the markdown files to restructuredtext so that we can more easily copy-paste documentation into the readthedocs directory. We might not text from all of the files into readthedocs, but I choose to convert all of the files to restructuredtext so that we have only 1 text format for documentation. 

After this PR, I will submit a PR outlining the structure of readthedocs.

pmantica1:
In terms of quality assurance I've done the following: 
1. Manually verified that all of the hyperlinks work.
2. Gone through converted README.rst several times and fixed any imperfections in the conversion I could find. 
3. Visually compared the README.md and the README.rst and verified that they rendered in the same way. 
4. Gone through all of of the .rst files once and fix any imperfections I could find. 

One drawdown of this conversion is that there is no way to highlight graphql syntax in rst. 

Also, I was not able to figure out after half an hour to an hour search how to render add code highlighting in hyperlinks. For instance, I couldn't find how to highlight `type_equivalence_hints` as code in the rst hyperlink below. 
`Optional type_equivalence_hints` compilation parameter <#optional-type_equivalence_hints-parameter>`__

However, there are very few instances, (probably less than 5), where we would like to highlight code in a hyperlink so this shouldn't be a huge issue.

AUTHORS.rst
```diff
@@ -0,0 +1,7 @@
+Authors
+=======
+
+- Predrag Gruevski (github: obi1kenobi)
+- Amartya Shankha Biswas (github: amartyashankha)
+- Jeremy Meulemans (github: jmeulemans)
+- Pedro Mantica (github: pmantica1)
```
obi1kenobi: @bojanserafimov if you want to add yourself here, go for it. Same goes for @qqi0O0, @LWprogramming, and @ReaLNeroM – you have all made substantial contributions to this project, and have earned your spot on this list.

setup.py
```diff
@@ -42,15 +42,15 @@ def find_name():
 
 def find_long_description():
     """Return the content of the README.md file."""
```
obi1kenobi: `README.md` in the docstring should be updated.

CHANGELOG.rst
```diff
@@ -0,0 +1,345 @@
+Changelog
+=========
+
+Current development version
+---------------------------
+
+.. _v1110:
```
obi1kenobi: I'm worried about these hyperlinks. It seems like `v11.1.0` and `v1.11.0` and `v1.1.10` might collide. Thoughts?

AUTHORS.rst
obi1kenobi: Though probably wait until this is merged, it's already a 5000-line PR :)

setup.py
pmantica1: Addressed.

CHANGELOG.rst
pmantica1: There are no hyperlinks to these titles so I am not sure why the autoconverter marked them as explicit hyperlink titles. I removed them. Also, there seems to be a bug in the autoconverter since _v.11.1.0 is a valid reference name.

pmantica1:
I will add a guide on the restructuredtext features that we use in one of the upcoming PRs by the way.

obi1kenobi:
Can you work with @jecdohmann and help him convert his documentation updates to the RST format? Unfortunately merging this will have caused a merge conflict in his PR, and I'd like to make sure we don't leave him hanging on this.

PR #487, opened 2019-08-16
Improve validation error message

pmantica1:
In this PR I aim to improve the validation error message, (mostly by providing the name of the invalid argument). However, this function is part of the public API and changing this function's interface in this PR involves a major version bump according to the rules described by semantic versioning.

One solution I was thinking of was to make the `name` parameter optional. Then if the inputted name is None, then change the format of the error message to:
"Expected argument type {}. Got {}." 

I am submitting this PR to ask for opinions on the matter above.

### DIFF
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -22,7 +22,7 @@ from .sql_formatting import insert_arguments_into_sql_query
 ######
 
 
-def validate_argument_type(expected_type, value):
+def validate_argument_type(expected_type, value, name, is_inner_value=False):
     """Ensure the value has the expected type and is usable in any of our backends, or raise errors.
 
     Backends are the database languages we have the ability to compile to, like OrientDB MATCH,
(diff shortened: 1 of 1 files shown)


pmantica1:
Please do not merge this PR. It was intended to be a draft PR, but I accidentally marked it as ready for review when I requested a review from ob1kenobi.

graphql_compiler/query_formatting/common.py
```diff
@@ -32,39 +32,49 @@ def validate_argument_type(expected_type, value):
     Args:
         expected_type: GraphQLType we expect. All GraphQLNonNull type wrappers are stripped.
         value: object that can be interpreted as being of that type
+        name: optional string. If it is a string, it is the name of the argument and will be used to
```
pmantica1: I decided to make this parameter optional in order to order to make this change backwards compatible since any non-backward compatible change to a public api function requires a major version bump, (and according to lines 20-22 this function is part of the public API).

obi1kenobi:
Can you give some examples of where you see this set of changes making a difference? I especially don't understand the use of `is_inner_value`.

pmantica1:
For that PR I can remove the `is_inner_value` since I agree it doesn't add much value, but pointing out the name of the invalid argument can help the user avoid the hassle of figuring out is invalid.

graphql_compiler/query_formatting/common.py
```diff
@@ -34,47 +34,43 @@ def validate_argument_type(expected_type, value, name=None, is_inner_value=False
...
         name: optional string. If it is a string, it is the name of the argument and will be used to
               provide more descriptive error messages.
-        is_inner_value: bool, whether this value corresponds to an element in a list argument.
     """
-    # TODO(pmantica1): make name parameter required in major version bump.
-    if name is not None:
-        if is_inner_value:
-            error_message_prefix = u'Invalid type for element in list argument {}.'.format(name)
-        else:
+    def _get_invalid_type_error_message(expected_python_type_name):
```
obi1kenobi: Why put this function definition right here, as opposed to making it a private function at module scope? Leaking `name` in by accident like this seems unnecessary and confusing.

graphql_compiler/query_formatting/common.py
```diff
@@ -34,47 +34,43 @@ def validate_argument_type(expected_type, value, name=None, is_inner_value=False
...
     if GraphQLString.is_same_type(stripped_type):
         if not isinstance(value, six.string_types):
-            raise GraphQLInvalidArgumentError(invalid_type_error_message.format('string'))
+            raise GraphQLInvalidArgumentError(_get_invalid_type_error_message('string'))
     elif GraphQLID.is_same_type(stripped_type):
         # IDs can be strings or numbers, but the GraphQL library coerces them to strings.
         # We will follow suit and treat them as strings.
         if not isinstance(value, six.string_types):
-            raise GraphQLInvalidArgumentError(invalid_type_error_message.format('string'))
+            raise GraphQLInvalidArgumentError(_get_invalid_type_error_message('string'))
```
obi1kenobi: Seems like we always raise an error with the output of that function – why not make the function raise the error as well?

graphql_compiler/query_formatting/common.py
```diff
@@ -21,8 +21,20 @@
 # Public API
 ######
 
+def _raise_invalid_type_error(expected_python_type_name, value, name):
```
obi1kenobi: nit: `name, value` seems like maybe a more intuitive argument order to me

graphql_compiler/query_formatting/common.py
```diff
@@ -21,8 +21,20 @@
...
+    """Raise a GraphQLInvalidArgumentError that states that the argument type is invalid."""
```
obi1kenobi: Please write a full docstring here. It's far from obvious that `name` could be `None`.

pmantica1:
Since next version bump is going to be a major version bump, I am going to refactor some of the changes here.

graphql_compiler/query_formatting/common.py
```diff
@@ -21,50 +21,51 @@
...
 
+def _raise_invalid_type_error(name, expected_python_type_name, value):
+    """Raise a GraphQLInvalidArgumentError that states that the argument type is invalid."""
+    raise GraphQLInvalidArgumentError(u'Invalid type for argument {}. Expected {}. Got value {} of '
+                                      u'type {}.'.format(name, expected_python_type_name, value,
+                                                         type(value).__name__))
 
-def validate_argument_type(expected_type, value):
+
+def validate_argument_type(name, expected_type, value):
```
pmantica1: The next version bump is going to be a major version bump. We might as well make the name parameter required here.

pmantica1:
@qqi0O0 Do you mind submitting the branch with the tests if you still have it by any chance?

PR #492, opened 2019-08-16
Implement make_query_plan, which adds filter directives

qqi0O0:
The print_query_plan part isn't done super carefully, it's partly copied from the prototype code. If we take query str 
```
            {
              Animal {
                out_Animal_Creature {
                  age @output(out_name: "age")
                }
              }
            }
```
and do `print_query_plan(make_query_plan(split_query(parse(query_str), merged_schema_descriptor)))`, we get 
```
Execute in schema named "first":
{
  Animal {
    uuid @output(out_name: "__intermediate_output_0")
  }
}

        Execute in schema named "second":
        {
          Creature {
            age @output(out_name: "age")
            id @output(out_name: "__intermediate_output_1") @filter(op_name: "in_collection", value: ["$__intermediate_output_0"])
          }
        }
        

Join together outputs as follows: [OutputJoinDescriptor(output_names=('__intermediate_output_0', '__intermediate_output_1'))]

Remove the following outputs at the end: frozenset({'__intermediate_output_0', '__intermediate_output_1'})
```
just to show what it's like.

### DIFF
--- a/graphql_compiler/schema_transformation/make_query_plan.py
+++ b/graphql_compiler/schema_transformation/make_query_plan.py
@@ -1,0 +1,274 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+from copy import copy
+
+from graphql import print_ast
+from graphql.language import ast as ast_types
+
+from ..ast_manipulation import get_only_query_definition
+from ..exceptions import GraphQLValidationError
+from ..schema import FilterDirective, OutputDirective
+
+
+SubQueryPlan = namedtuple(
+    'SubQueryPlan', (
+        'query_ast',  # Document, representing a piece of the overall query with directives added
+        'schema_id',  # str, identifying the schema that this query piece targets
+        'parent_query_plan',  # SubQueryPlan, the query that the current query depends on
+        'child_query_plans',  # List[SubQueryPlan], the queries that depend on the current query
+    )
+)
+
+
+OutputJoinDescriptor = namedtuple(
+    'OutputJoinDescriptor', (
+        'output_names',  # Tuple[str, str], (parent output name, child output name)
+        # May be expanded to have more attributes, e.g. is_optional, describing how the join
+        # should be made
+    )
+)
+
+
+QueryPlanDescriptor = namedtuple(
+    'QueryPlanDescriptor', (
+        'root_sub_query_plan',  # SubQueryPlan
+        'intermediate_output_names',  # frozenset[str], names of outputs to be removed at the end
+        'output_join_descriptors',
+        # List[OutputJoinDescriptor], describing which outputs should be joined and how
+    )
+)
+
+
+def make_query_plan(root_sub_query_node, intermediate_output_names):
+    """Return a QueryPlanDescriptor, whose query ASTs have @filters added.
+
+    For each parent of parent and child SubQueryNodes, a new @filter directive will be added
+    in the child AST. It will be added on the field whose @output directive has the out_name
+    equal to the child's out name as specified in the QueryConnection. The newly added @filter
+    will be a 'in_collection' type filter, and the name of the local variabe is guaranteed to
+    be the same as the out_name of the @output on the parent.
+
+    ASTs contained in the input node and its children nodes will not be modified.
+
+    Args:
+        root_sub_query_node: SubQueryNode, representing the base of a query split into pieces
+                             that we want to turn into a query plan
+
+    Returns:
+        QueryPlanDescriptor namedtuple, containing a tree of SubQueryPlans that wrap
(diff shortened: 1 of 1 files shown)


graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -0,0 +1,274 @@
...
+)
+
+
+def make_query_plan(root_sub_query_node, intermediate_output_names):
+    """Return a QueryPlanDescriptor, whose query ASTs have @filters added.
+
+    For each parent of parent and child SubQueryNodes, a new @filter directive will be added
+    in the child AST. It will be added on the field whose @output directive has the out_name
+    equal to the child's out name as specified in the QueryConnection. The newly added @filter
+    will be a 'in_collection' type filter, and the name of the local variabe is guaranteed to
```
obi1kenobi: typo: `variabe`

graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -0,0 +1,274 @@
...
+        output_join_descriptors=output_join_descriptors,
+    )
+
+
+def _make_query_plan_recursive(sub_query_node, sub_query_plan, output_join_descriptors):
+    """Recursively copy the structure of sub_query_node onto sub_query_plan.
+
+    For each child connection contained in sub_query_node, create a new SubQueryPlan for
+    the corresponding child SubQueryNode, add appropriate @filter directive to the child AST,
+    and attach the new SubQueryPlan to the list of children of the input sub query plan.
```
obi1kenobi: consider `sub-query` instead of `sub query` for clarity

graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -0,0 +1,274 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+from copy import copy
+
+from graphql import print_ast
+from graphql.language import ast as ast_types
```
obi1kenobi: heh I guess you wrote a lot of code that had this import line? :)

graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -0,0 +1,274 @@
...
+    Returns:
+        Field, InlineFragment, or OperationDefinition, identical to the input ast except
+        with an @filter added at the specified field if such a field is found. If no changes
+        were made, this is the same object as the input
+    """
+    if not isinstance(ast, (
+        ast_types.Field, ast_types.InlineFragment, ast_types.OperationDefinition
+    )):
+        raise AssertionError(
+            u'Input ast is of type "{}", which should not be a selection.'
```
obi1kenobi: nit: `ast -> AST`

graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -0,0 +1,274 @@
...
+    )
+
+
+def print_query_plan(query_plan_descriptor):
+    """Return a string describing query plan."""
+    query_plan_str = u''
+    plan_and_depth = _get_plan_and_depth_in_dfs_order(query_plan_descriptor.root_sub_query_plan)
+
+    for query_plan, depth in plan_and_depth:
+        line_separation = u'\n' + u' ' * 8 * depth
```
obi1kenobi: Let's use 4 spaces by default instead, and ideally make this configurable via a kwarg.

I'd make the function signature be `def print_query_plan(query_plan_descriptor, indentation_depth=4):`

graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -0,0 +1,274 @@
...
+        query_str = query_str.replace(u'\n', line_separation)
+        query_plan_str += query_str
+
+    query_plan_str += u'\n\n'
+    query_plan_str += u'Join together outputs as follows: '
+    query_plan_str += str(query_plan_descriptor.output_join_descriptors) + u'\n\n'
+    query_plan_str += u'Remove the following outputs at the end: '
+    query_plan_str += str(query_plan_descriptor.intermediate_output_names) + u'\n'
+
+    return query_plan_str
```
obi1kenobi: nit: All this string concatenation is actually quite expensive since it uses Shlemiel the Painter's algorithm: https://www.joelonsoftware.com/2001/12/11/back-to-basics/

It would be better to assemble all the substrings we'd like to join together into a list, and then just call `''.join(query_plan_components)` with the list at the end.

graphql_compiler/schema_transformation/make_query_plan.py
qqi0O0: Oops, I meant to fix this before pushing and then forgot

PR #493, opened 2019-08-16
Add SQLAlchemySchemaInfo class

bojanserafimov:
Merging my sql backend one piece at a time. This is the best starting point.


graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+
+# Complete schema information sufficient to compile GraphQL queries to SQLAlchemy
+SQLAlchemyMetadata = namedtuple('SQLAlchemyMetadata', (
+    'schema',  # GraphQLSchema
+    'tables',  # dict mapping every graphql type in the schema to a sqlalchemy table
```
pmantica1: nit: GraphQLObjectType

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -0,0 +1,13 @@
...
+    'joins',   # dict mapping every graphql type in the schema to:
```
pmantica1: nit: GraphQLObjectType

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -0,0 +1,13 @@
...
+               #    dict mapping edge fields at that type to a dict with keys:
```
pmantica1: I think that we refer to edge fields as vertex fields in the README.md so let's follow suit here.

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -0,0 +1,13 @@
...
+               #        from_column_name: string, column name in the from_table to join on
```
pmantica1: from_table makes it seems as though from_table is a python variable though it is not. (Same for to_table).

graphql_compiler/schema/sqlalchemy_schema.py
bojanserafimov: It's a string. Clarified.

pmantica1: Apologies. I meant to say that it does not map every graphql type in the schema to a sqlalchemy table, but instead every graphql **object**  type to a sqlalchemy table.

obi1kenobi: Not everything is a GraphQL object, there are also interfaces and enums etc. to worry about. I think GraphQL type is fine here.

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -0,0 +1,17 @@
...
+
+# Complete schema information sufficient to compile GraphQL queries to SQLAlchemy
+#
+# It describes the tables that correspond to each type, and gives instructions on how
+# to perform joins for each vertex field. The property fields on each type are implicitly
+# mapped to columns with the same name on the corresponding table.
+SQLAlchemySchemaInfo = namedtuple('SQLAlchemySchemaInfo', (
+    'schema',  # GraphQLSchema
+    'tables',  # dict mapping every graphql type name in the schema to a sqlalchemy table
+    'joins',   # dict mapping every graphql type name in the schema to:
```
obi1kenobi: Maybe `joins` -> `junctions` or `junction tables`, just to avoid confusion?

graphql_compiler/schema/sqlalchemy_schema.py
pmantica1: Technically, the GraphQLScalarType class is a subclass of the GraphQLType class. Therefore, this comment currently implies that we intend to map every graphql scalar type in the schema to a sqlalchemy table.

obi1kenobi: Yes. But `GraphQLType` and `GraphQL type` (which I wrote) aren't the same thing. By not pointing to a specific type by exact name, I think the latter is asking the user to exercise some caution and common sense. `GraphQLObjectType` is less precise here, and unless there's a great option we're all missing, I think we should go with the least-bad option and move on to the meat of the changes instead of spending more time on this.

PR #495, opened 2019-08-16
developing "starts_with" filter op

jecdohmann:
I added functionality for starts_with - I am not certain about the 'expected_gremlin', 'expected_sql', or 'expected_cypher' in tests.test_compiler.test_starts_with_op_filter, so please let me know if that's the right syntax for those dialects and I will add a more comprehensive test suite

For the gremlin backend I had to add some strings to the list of special strings during validation: '/^' and '.*/' because those are used as literals to construct the gremlin equivalent of starts_with

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -2,7 +2,7 @@
 
 ## Issues starting MySQL, PostgreSQL, or redis server with docker-compose
 
-If you have any trouble starting the MySQL/PostgreSQL database or the redis server, make sure any database service is not already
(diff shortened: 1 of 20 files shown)


TROUBLESHOOTING.md
```diff
@@ -1,17 +1,19 @@
...
 ```bash
 brew services stop mysql
 brew services stop postgresql
+brew services stop redis-server
 ```
 or on Ubuntu with:
 ```bash
 service mysql stop
 service postgresql stop
+service redis-server stop
```
pmantica1: Can you remove this change above? (I have no idea how it is appearing here).

PR #497, opened 2019-08-19
Create SQLAlchemySchemaInfo used for testing

bojanserafimov:
Comment if you notice issues with the approach. Don't read into details, things will change. The test fails.

### DIFF
--- a/graphql_compiler/schema/sqlalchemy_schema.py
+++ b/graphql_compiler/schema/sqlalchemy_schema.py
@@ -20,17 +20,28 @@ from graphql.type.definition import GraphQLInterfaceType, GraphQLObjectType, Gra
 # - Builtin types like __Schema, __Type, etc. don't need corresponding tables.
 # - Builtin fields like _x_count do not need corresponding columns.
 SQLAlchemySchemaInfo = namedtuple('SQLAlchemySchemaInfo', (
-    'schema',     # GraphQLSchema
-    'tables',     # dict mapping every graphql object type or interface type name in the schema to
-                  # a sqlalchemy table
-    'junctions',  # dict mapping every graphql object type or interface type name in the schema to:
-                  #    dict mapping every vertex field name at that type to a dict with keys:
-                  #        from_column_name: string, column name to join from
-                  #        to_column_name: string, column name to join to
+    # GraphQLSchema
+    'schema',
+
+    # dict mapping every graphql object type or interface type name in the schema to
+    # a sqlalchemy table
+    'tables',
+
+    # dict mapping every graphql object type or interface type name in the schema to:
+    #    dict mapping every vertex field name at that type to a dict with keys:
+    #        from_column_name: string, column name to join from
+    #        to_column_name: string, column name to join to
+    'junctions',
+
+    # dict mapping every graphql object type or interface type name in the schema to:
+    #    dict mapping every list-valued property field name to a dict with keys:
+    #        junction_name: string, previously specified junction that provides the set contents
+    #        value_column: string, the column in the junction target table that has the elements
+    'set_valued_fields',
 ))
 
 
-def make_sqlalchemy_schema_info(schema, tables, junctions, validate=True):
+def make_sqlalchemy_schema_info(schema, tables, junctions, set_valued_fields, validate=True):
     """Make a SQLAlchemySchemaInfo if the input provided is valid.
 
     Args:
@@ -44,6 +55,13 @@ def make_sqlalchemy_schema_info(schema, tables, junctions, validate=True):
         validate: Optional bool (default True), specifying whether to validate that the given
                   input is valid for creation of a SQLAlchemySchemaInfo. Consider not validating
                   to save on performance when dealing with a large schema.
+
+        set_valued_fields: dict mapping graphql object and interface type names in the schema to:
+                  dict mapping every list-valued property field name to a dict with keys:
+                      junction_name: string, previously specified junction that provides the
+                                     set contents
+                      value_column: strine, the column in the junction target table that has
+                                    the elements
 
     Returns:
         SQLAlchemySchemaInfo containing the input arguments provided
@@ -75,10 +93,23 @@ def make_sqlalchemy_schema_info(schema, tables, junctions, validate=True):
                                 raise AssertionError(u'No junction was specified for vertex '
                                                      u'field {} on type {}'
                                                      .format(field_name, type_name))
+                        elif isinstance(field_type.type, GraphQLList):
+                            if field_name not in set_valued_fields.get(type_name, {}):
+                                raise AssertionError(u'Set valued property field {} on type {} '
+                                                     u'has no implementation.'
+                                                     .format(field_name, type_name))
+                            set_implementation = set_valued_fields[type_name][field_name]
+                            junction = set_implementation['junction']
+                            value_column = set_implementation['value_column']
+                            if junction not in junctions.get(type_name, {}):
+                                raise AssertionError(u'Implementation for set valued property '
+                                                     u'field {} on type {} references junction {}, '
+                                                     u'which is not specified.'
+                                                     .format(field_name, type_name, junction))
                         else:
                             if field_name not in builtin_fields and field_name not in table.c:
                                 raise AssertionError(u'Table for type {} has no column '
                                                      u'for property field {}'
                                                      .format(type_name, field_name))
 
-    return SQLAlchemySchemaInfo(schema, tables, junctions)
+    return SQLAlchemySchemaInfo(schema, tables, junctions, set_valued_fields)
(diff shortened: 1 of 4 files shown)


graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,92 @@
...
+        validate: Optional bool (default True), specifying whether to validate that the given
+                  input is valid for creation of a SQLAlchemySchemaInfo. Consider not validating
+                  to save on performance when dealing with a large schema.
+
+    Returns:
+        SQLAlchemySchemaInfo containing the input arguments provided
+    """
+    if validate:
+        types_mapped_to_tables = (GraphQLInterfaceType, GraphQLObjectType, GraphQLUnionType)
+        builtin_types = {
```
obi1kenobi: For the sake of future-proofing this, I think we should mark all types prefixed with `__` as built-in, since the entire `__` (double-underscore) prefix is reserved for GraphQL internal use.

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,92 @@
...
+                    if type_name not in tables:
+                        raise AssertionError(u'Table for type {} not found'.format(type_name))
+                    table = tables[type_name]
+                    if not isinstance(table, sqlalchemy.Table):
+                        raise AssertionError(u'Table for type {} has wrong type {}'
+                                             .format(type_name, type(table)))
+
+                    # Check existence of all fields
+                    for field_name, field_type in six.iteritems(graphql_type.fields):
+                        is_vertex_field = (field_name.startswith('out_') or
```
obi1kenobi: Consider `from schema import is_vertex_field_name` instead.

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,92 @@
...
+                                           field_name.startswith('in_'))
+                        if is_vertex_field:
+                            if field_name not in junctions[type_name]:
+                                raise AssertionError(u'No junction was specified for vertex '
+                                                     u'field {} on type {}'
+                                                     .format(field_name, type_name))
+                        else:
+                            # TODO(bojanserafimov): We have no SQL representation of list types yet,
+                            #                       so they are excused.
+                            # XXX should I just remove list types from the schema?
```
obi1kenobi: Maybe raise `NotImplementedError()` on them for now? I wouldn't remove them altogether since I believe at least Postgres would allow list-typed columns, and we don't necessarily (yet) want to preclude supporting that in the future.

graphql_compiler/tests/test_helpers.py
```diff
@@ -281,6 +283,174 @@ def get_schema():
...
+            sqlalchemy.Column('eats', sqlalchemy.String(36), nullable=True),
+            schema='db_1.schema_1'
+        ),
+        'UniquelyIdentifiable': sqlalchemy.Table(
+            'UniquelyIdentifiable',
+            sqlalchemy_metadata,
+            sqlalchemy.Column('uuid', sqlalchemy.String(36), primary_key=True),
+            schema='db_1.schema_1'
+        ),
+        'Union__BirthEvent__Event__FeedingEvent': sqlalchemy.Table(
```
pmantica1: Why do you need to map union types to SQLAlchemy tables? I understand that you need union types in order to be able to have tests for sql that are almost identical to the orientdb tests. However, I don't understand why you chose to have a concrete backend implementation of union types.

bojanserafimov: Good question. I think it was just a convenience, so I can reuse the same tests. I will remove them for now and try to make the tests work without them when I get to that.

graphql_compiler/schema/sqlalchemy_schema.py
bojanserafimov: I still want to be able to create an incomplete schema info, with array fields missing, so that I can run tests that don't rely on array fields.

bojanserafimov: The way you'd implement aliases in MSSQL is to use a foreign key to a table of aliases. I'll implement it that way here, and raise `NotImplementedError()` when I need to deal with it later.

bojanserafimov:
Should the SQL flavor (Postgres, MSSQL, etc.) be part of the `SQLAlchemySchemaInfo`? I'm leaning yes because we will never have to do cross-flavor queries (it's not possible), and it will enable us to write flavor-specific passes. Thoughts?

obi1kenobi:
Yeah, I think that's reasonable. The config is already specific to the
particular provider instance, and at that point we might as well include
the SQL flavor since we clearly have one in mind.

graphql_compiler/schema/sqlalchemy_schema.py
bojanserafimov: See latest commit

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,115 @@
...
+    #    dict mapping every vertex field name at that type to a dict with keys:
+    #        from_column_name: string, column name to join from
+    #        to_column_name: string, column name to join to
+    'junctions',
+
+    # dict mapping every graphql object type or interface type name in the schema to:
+    #    dict mapping every list-valued property field name to a dict with keys:
+    #        junction_name: string, previously specified junction that provides the set contents
+    #        value_column: string, the column in the junction target table that has the elements
+    'set_valued_fields',
```
bojanserafimov: In Postgres, you can use `sqlalchemy.ARRAY` instead of this. I'll clarify this when I add the SQL flavor to the `SQLAlchemySchemInfo`

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,111 @@
...
+
+from . import is_vertex_field_name
+
+
+# Describes the intent to join two tables on equality of a column of theirs.
+# The resulting join expression could be something like:
+# JOIN origin_table.from_column = destination_table.to_column
+#
+# The type of join (inner vs left, etc.) is not specified.
+# The tables are not specified.
```
bojanserafimov: Not specified because you don't execute the join on the actual table objects, but on aliases of those tables.

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,111 @@
...
+DirectJoinDescriptor = namedtuple('DirectJoinDescriptor', (
+    'from_column',  # The column in the source table we intend to join on.
+    'to_column',    # The column in the destination table we intend to join on.
+))
+
+
+# TODO(bojanserafimov): Define a JunctionJoinDescriptor
```
pmantica1: Can you briefly mention what is a JunctionJoinDescriptor and why need it if we already have a DirectJoinDescriptor?

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,111 @@
...
+
 
 # Complete schema information sufficient to compile GraphQL queries to SQLAlchemy
 #
-# It describes the tables that correspond to each type, and gives instructions on how
-# to perform joins for each vertex field. The property fields on each type are implicitly
-# mapped to columns with the same name on the corresponding table.
+# It describes the tables that correspond to each type (object type, interface type or union type),
```
pmantica1: I don't think union types should be mapped to tables. (I think you only mention objects and interfaces below).

graphql_compiler/schema/sqlalchemy_schema.py
```diff
@@ -1,17 +1,111 @@
...
 from collections import namedtuple
 
+from graphql.type.definition import GraphQLInterfaceType, GraphQLObjectType
+import six
+import sqlalchemy
+
+from . import is_vertex_field_name
+
+
+# Describes the intent to join two tables on equality of a column of theirs.
```
pmantica1: "equality of column of theirs" makes it sound as if though the columns need to be named the same.

bojanserafimov: Can you suggest a rewording? Can I use the phrase "foreign key" without implying a foreign key constraint must be there?

graphql_compiler/schema/sqlalchemy_schema.py
bojanserafimov: Removed the TODO. When you have a need for a `JunctionJoinDescriptor` you'll figure out where to put it.

graphql_compiler/schema/sqlalchemy_schema.py
pmantica1: How about "Describes the intent to join two tables using the specified columns."?

PR #498, opened 2019-08-19
Query pagination skeleton

ReaLNeroM:
Using cost estimation and AST manipulation, we can effectively paginate arbitrary queries. I've created a directory `query_pagination` that has a skeleton for this functionality. 

The user interface is the following: the user calls `paginate_query` with their query, parameters, and relevant information for pagination. We then modify their query and parameters, generating two pairs of query and parameters. These two query/parameters' union is guaranteed to be equivalent to the user's original query, but with a twist: the first query we provide is estimated to return roughly a page of data. The user can then do whatever with these two queries. If they want to generate more than one query, they call `paginate_query` with the latter of the two queries we've provided.

Currently, the design behind pagination is the following:
1. The user provides us with a query and parameters to paginate and desired number of results per page.
2. We discover the number of pages this query will generate using the cost estimator's new function `estimate_number_of_pages`. If there's only one page expected for this query, we just return the user's query.
3. We add (or use the user's!) filters for pagination via `query_pagination/query_parameterizer/generate_parameterized_queries`. Note that the filters we add are not hydrated i.e. they do not have parameter values. We will generate them in the following step.
4. Using the pagination filters from the previous step, statistics, and knowledge about the user's parameters, we generate parameters for the two queries via `query_pagination/parameter_generator/hydrate_parameters_of_parameterized_query`.
5. We now have two queries, and their parameters, and return them to the user.

What's left now: Step 3 and Step 4 are currently not implemented in this PR. I do have working versions of them, so which is why I believe this skeleton works. I'll refine them a bit more, and will launch them in a separate PR.

@ reviewers: I've included a test in `tests/snapshot_tests/test_query_pagination` for the skeleton which may or may not be subject to change.

### DIFF
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -309,7 +309,6 @@ def estimate_number_of_pages(schema_graph, statistics, graphql_query, params, pa
         raise AssertionError(u'Received negative estimate {} for cardinality of query {}: {}'
                              .format(result_size, graphql_query, params))
 
-
     # Since using a // b returns the fraction rounded down, we instead use (a + b - 1) // b, which
     # returns the fraction value rounded up, which is the desired functionality.
     num_pages = int((result_size + page_size - 1) // page_size)
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -293,11 +293,14 @@ def estimate_number_of_pages(schema_graph, statistics, graphql_query, params, pa
 
     Returns:
         int, estimated number of pages if the query were executed.
+
+    Raises:
+        ValueError if page_size is below 1.
     """
     if page_size < 1:
-        raise AssertionError(u'Could not estimate number of pages for query {}'
-                             u' with page size lower than 1: {} {}'
-                             .format(graphql_query, page_size, params))
+        raise ValueError(u'Could not estimate number of pages for query {}'
+                         u' with page size lower than 1: {} {}'
+                         .format(graphql_query, page_size, params))
 
     result_size = estimate_query_result_cardinality(
         schema_graph, statistics, graphql_query, params
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -286,7 +286,7 @@ def estimate_number_of_pages(schema_graph, statistics, graphql_query, params, pa
 
     Args:
         schema_graph: SchemaGraph instance.
-        statistics: Statistics object
+        statistics: Statistics object.
         graphql_query: str, valid GraphQL query to be estimated.
         params: dict, parameters for the given query.
         page_size: int, desired number of result rows per page.
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -294,11 +294,23 @@ def estimate_number_of_pages(schema_graph, statistics, graphql_query, params, pa
     Returns:
         int, estimated number of pages if the query were executed.
     """
+    if page_size < 1:
+        raise AssertionError(u'Could not estimate number of pages for query {}'
+                             u' with page size lower than 1: {} {}'
+                             .format(graphql_query, page_size, params))
+
     result_size = estimate_query_result_cardinality(
         schema_graph, statistics, graphql_query, params
     )
+    if result_size < 0.0:
+        raise AssertionError(u'Received negative estimate {} for cardinality of query {}: {}'
+                             .format(result_size, graphql_query, params))
+
 
     # Since using a // b returns the fraction rounded down, we instead use (a + b - 1) // b, which
     # returns the fraction value rounded up, which is the desired functionality.
     num_pages = int((result_size + page_size - 1) // page_size)
+    if num_pages == 0:
+        num_pages = 1
+
     return num_pages
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -298,7 +298,7 @@ def estimate_number_of_pages(schema_graph, statistics, graphql_query, params, pa
         schema_graph, statistics, graphql_query, params
     )
 
-    # Since using a // b Returns the fraction rounded down, we instead use (a + b - 1) // b, which
+    # Since using a // b returns the fraction rounded down, we instead use (a + b - 1) // b, which
     # returns the fraction value rounded up, which is the desired functionality.
     num_pages = int((result_size + page_size - 1) // page_size)
     return num_pages
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -272,3 +272,33 @@ def estimate_query_result_cardinality(
     expected_query_result_cardinality = root_counts * results_per_root
 
     return expected_query_result_cardinality
+
+
+def estimate_number_of_pages(schema_graph, statistics, graphql_query, params, page_size):
+    """Estimate how many pages of results will be generated for a given query.
+
+    Using the cardinality estimator, we generate an estimate for the query result cardinality i.e.
+    the number of result rows, then divide (rounding up) by the page_size to get the approximate
+    number of pages that the query will produce.
+    For example, if a query were estimated to return 12000 result rows, and the desired page size is
+    5000, then the query can be divided into ceil(12000/5000)=3 pages, each with a result size below
+    (or equal to) 5000 results.
+
+    Args:
+        schema_graph: SchemaGraph instance.
+        statistics: Statistics object
+        graphql_query: str, valid GraphQL query to be estimated.
+        params: dict, parameters for the given query.
+        page_size: int, desired number of result rows per page.
+
+    Returns:
+        int, estimated number of pages if the query were executed.
+    """
+    result_size = estimate_query_result_cardinality(
+        schema_graph, statistics, graphql_query, params
+    )
+
+    # Since using a // b Returns the fraction rounded down, we instead use (a + b - 1) // b, which
+    # returns the fraction value rounded up, which is the desired functionality.
+    num_pages = int((result_size + page_size - 1) // page_size)
+    return num_pages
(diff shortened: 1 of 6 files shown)


graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -272,3 +272,33 @@ def estimate_query_result_cardinality(
...
+        int, estimated number of pages if the query were executed.
+    """
+    result_size = estimate_query_result_cardinality(
+        schema_graph, statistics, graphql_query, params
+    )
+
+    # Since using a // b returns the fraction rounded down, we instead use (a + b - 1) // b, which
+    # returns the fraction value rounded up, which is the desired functionality.
+    num_pages = int((result_size + page_size - 1) // page_size)
+    return num_pages
```
bojanserafimov: Let's make sure this never returns 0.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+from graphql.language.printer import print_ast
+
+from graphql_compiler.cost_estimation.cardinality_estimator import estimate_number_of_pages
+from graphql_compiler.query_pagination.parameter_generator import (
+    hydrate_parameters_of_parameterized_query
+)
+from graphql_compiler.query_pagination.query_parameterizer import generate_parameterized_queries
+
+
+PaginationQueries = namedtuple(
```
bojanserafimov: Add a description of what this is. It's only obvious from the name if you've been working on it for a while.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+    'PaginationQueries',
+    (
+        'next_page_query',          # Document, AST for the query that along with
```
bojanserafimov: Is it more intuitive to call it `next` or `first`? I see how `next` relates to the python generator vocabulary, but in context here, I think `first and continuation` is a clearer partition of all results than `next and continuation`.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+                                    # original query, or None if the original query was not split
+                                    # into two queries.
+        'continuation_parameters',  # dict or None, parameters with which to execute the
+                                    # continuation query, or None if the continuation query was not
+                                    # generated.
+    )
+)
+
+
+def split_into_next_page_query_and_continuation_query(
```
bojanserafimov: As a user of this module, it's not clear to me whether I should use `split_into_next_page_query_and_continuation_query`, or `paginate_query`. Both are public methods in the `__init__.py` module, and their documentations are very similar (I still don't know what's the difference. Is it just the last parameter? Do they do the same thing?)

graphql_compiler/query_pagination/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+    The first query when executed will return approximately page_size results, while the second
+    query will return the rest of the results. In case the given query is estimated to return less
+    than a page of data, the query will not be split.
+    Since the cost estimator may underestimate or overestimate the actual number of pages, you
+    should expect the actual number of results to be within two orders of magnitude of the estimate.
+
+    Args:
+        schema_graph: SchemaGraph instance.
+        statistics: Statistics object.
+        query_ast: Document, AST of the GraphQL query that is being paginated.
```
bojanserafimov: I think we'll always start with a graphql string, then parse it, then pass the AST in. Might be worth changing this to take in a string, which will also make the function more user-friendly, as a public facing method in `__init__.py`.

graphql_compiler/query_pagination/__init__.py
bojanserafimov: Some problems:
- The name is not the best, and I don't have a better idea.
- This class exists solely to better document the return value of a function, but in the `Returns` section of that function you end up explaining all over again what this means.
- You can't have `continuation_query` be `None`, but `continuation_parameters` to actually exist.

To me it seems more natural to group a query and parameters in one namedtuple (maybe call it `QueryWithParams`), and return a pair of `QueryWithParams` in your function, where the second one is optional.

graphql_compiler/query_pagination/parameter_generator.py
```diff
@@ -0,0 +1,35 @@
...
+        'continuation_parameters',  # dict, parameters with which to execute the parameterized
+                                    # continuation query.
+    ),
+)
+
+
+def hydrate_parameters_of_parameterized_query(
+    schema_graph, statistics, parameterized_pagination_queries, num_pages
+):
+    """Generate parameters combining the user's parameters and parameters for paging.
```
bojanserafimov: why is this combining called hydration? Can you point me to a definition of the word that applies in this context?

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+
+RESERVED_PARAMETER_PREFIX = '_paged_'
+
+ParameterizedPaginationQueries = namedtuple(
```
bojanserafimov: document what it is please. In what way is it parameterized?

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -0,0 +1,57 @@
...
+                                    # results when hydrated with pagination parameters.
+        'pagination_filters',       # List[PaginationFilter], filters usable for pagination.
+        'user_parameters',          # dict, parameters that the user has defined for other filters.
+    ),
+)
+
+PaginationFilter = namedtuple(
+    'PaginationFilter',
+    (
+        'vertex_field',                 # str, vertex class to which the property field belongs to.
```
bojanserafimov: `vertex_class` and `vertex_field` are different things. The name of a `vertex_field` is usually a name of an edge class.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -0,0 +1,57 @@
...
+        'property_field',               # str, name of the property field filtering is done over.
+        'next_page_query_filter',       # Directive, filter directive with '<' operator usable
+                                        # for pagination in the next page query.
```
bojanserafimov: What next page query? That's outside of the scope of this type. You can set up this context in the overall documentation of this namedtuple.

graphql_compiler/query_pagination/__init__.py
ReaLNeroM: If we don't use `PaginationQueries`, and instead just use a tuple with two `QueryWithParams`, the user will have to index by integers: `result[0]` for the next page query, and `result[1]` for the continuation query. In my eyes, `result.next_page_query` and `result.continuation_query` is bound to raise less eyebrows months on after this code is written.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+def split_into_next_page_query_and_continuation_query(
+    schema_graph, statistics, query_ast, parameters, num_pages
+):
+    """Split a query into two equivalent queries, one of which will return roughly a page of data.
+
+    First, two parameterized queries are generated that contain filters usable for pagination i.e.
+    filters with which the number of results can be constrained. Parameters for these filters are
+    then generated such that one of the new queries will return roughly a page of results, while the
+    other query will generate the rest of the results. This ensures that the two new queries' result
+    data is equivalent to the original query's result data.
```
bojanserafimov: This is too much internal methodology for a function in `__init__.py`. You can put it in a different module, and make a function in `__init__.py` that just directly calls it.

graphql_compiler/query_pagination/__init__.py
ReaLNeroM: How do you generate the second page query? By calling this function twice. At that point `first_page_query` should actually be called `second_page_query`, unless the user somehow knows that that `first` in `first_page_query` is relative to.

graphql_compiler/query_pagination/__init__.py
ReaLNeroM: Handing in the AST is preferred, as pagination may be called 100 times, and we'd have to pay the price of printing and parsing 100 times.
This is the argument @obi1kenobi put forth. Ultimately, this is an ease-of-use versus performance trade-off that you should probably speak to @obi1kenobi about.

graphql_compiler/query_pagination/__init__.py
bojanserafimov: You can unpack as `first_query, next_query = function_call(...`. I don't disagree. You can make it work if you document it well. I'm just giving a suggestion.

graphql_compiler/query_pagination/__init__.py
bojanserafimov: Call it beginning then.

The problem is that `next` and `continuation` literally mean the same thing. Which one is nexter than the other?

graphql_compiler/query_pagination/__init__.py
obi1kenobi: My suggestion would be to have a public function like `paginate_query_ast` that takes an AST and allows power users to reuse a pre-parsed AST, and separately have another public helper function called `paginate_query` that takes a query string, parses it, and then calls the other function. We've used this pattern successfully elsewhere in the compiler, and it seems to give us the best of both worlds at relatively minor cost.

graphql_compiler/query_pagination/__init__.py
ReaLNeroM: Your unpacking idea is the best solution thus far, I'll use that one! Thanks.

graphql_compiler/query_pagination/__init__.py
ReaLNeroM: `paginate_query` takes care of finding out how many pages to divide the query into, and to see if the query actually needs to be split. `_split_...` does the meat of the work by adding filters and generating parameters. It doesn't, however, find an estimate for the number of pages letting `paginate_query` do the job.
I've made `split...` into a private function.

graphql_compiler/query_pagination/__init__.py
ReaLNeroM: How about `next_page_query` and `remainder_query`?

graphql_compiler/query_pagination/parameter_generator.py
ReaLNeroM: My line of thinking was that I could easily call this function `generate_parameters` and be done with it. However, what we're doing here is we're assigning values to the parameterized queries i.e. creating parameter values for filters which do not have one. 
Replaced `hydrate` with `generate`.

bojanserafimov: the word hydrate still appears throughout the PR

ReaLNeroM: Fixed.

PR #499, opened 2019-08-20
Implemented starts_with and ends_with as filter ops

jecdohmann:
added starts_with and ends_with. I am not sure about the expected_cypher in test_compiler.test_ends_with_op_filter/ test_Compiler.test_starts_with_op_filter.

Please let me know what you think and I will add more test cases

Also I tried removing revisions to TROUBLESHOOTING.md that I made on a previously accepted pull request per pmantica's request, but they keep showing up. Any suggestions?


scripts/generate_test_sql/events.py
```diff
@@ -3,10 +3,10 @@
...
 
 FEEDING_EVENT_NAMES_LIST = (
-    "Breakfast",
-    "Brunch",
-    "Lunch",
-    "Dinner",
+    'Breakfast',
+    'Brunch',
+    'Lunch',
+    'Dinner',
```
chewselene: why the switch to single quotes here and in species.py?

TROUBLESHOOTING.md
```diff
@@ -1,17 +1,19 @@
 # Troubleshooting Guide
 
-## Issues starting MySQL or PostgreSQL with docker-compose
```
obi1kenobi: I'm definitely open to updating the troubleshooting guide, but not in an otherwise-unrelated PR. Would you mind removing these changes from this PR and opening a separate PR just for them?

graphql_compiler/compiler/expressions.py
```diff
@@ -976,6 +978,8 @@ def to_gremlin(self):
             u'not_contains': (u'contains', negated_dotted_operator_format),
             u'intersects': (u'intersect', intersects_operator_format),
             u'has_substring': (u'contains', dotted_operator_format),
+            u'starts_with': (u'==~', immediate_operator_format),
```
obi1kenobi: I think going for regex here is overkill. Since OrientDB Gremlin is just a Groovy interpreter with extra functions, let's just use the existing `startsWith/endsWith` functions on strings if possible?
https://blogs.oracle.com/fadevrel/useful-groovy-string-functions

This will also help us avoid adding to the set of `legal_strings_with_special_chars` in the file below, which is a nice bonus.

graphql_compiler/tests/test_compiler.py
```diff
@@ -2842,6 +2842,88 @@ def test_not_contains_op_filter_with_optional_tag(self):
...
+                    class: Animal,
+                    where: ((name LIKE ({wanted} + '%'))),
+                    as: Animal___1
+                }}
+                RETURN $matches
+            )
+        '''
+        expected_gremlin = '''
+            g.V('@class', 'Animal')
+            .filter{it, m -> (it.name ==~ ('/^' + ($wanted + '.*/')))}
```
chewselene: is this `==~` operation the best way of doing starts and ends with? I can't find much documentation online about it... Just wondering what other possibilities you looked at.

scripts/generate_test_sql/events.py
```diff
@@ -3,10 +3,10 @@
 
 
 FEEDING_EVENT_NAMES_LIST = (
-    "Breakfast",
-    "Brunch",
-    "Lunch",
-    "Dinner",
+    'Breakfast',
```
obi1kenobi: Why change this unrelated file? Was the linter complaining, or was it some other reason?

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -140,6 +140,84 @@ def visitor_fn(expression):
     return new_ir_blocks
 
 
+def lower_ends_with_binary_compositions(ir_blocks):
+    """Lower Filter blocks that use the "ends_with" operation into MATCH-representable form."""
+    def visitor_fn(expression):
+        """Rewrite BinaryComposition expressions with "ends_with" into representable form."""
+        # The implementation of "starts_with" must use the LIKE operator in MATCH, and must
```
chewselene: TYPO: `starts_with` -> `ends_with`

TROUBLESHOOTING.md
jecdohmann: I should've already had these changes merged as part of a previous PR I created - I am going to sync my fork, merge, and then resubmit this PR to try and get rid of these edits

scripts/generate_test_sql/events.py
jecdohmann: Yeah this was the linter's doing - should I revert the change?

graphql_compiler/compiler/expressions.py
jecdohmann: cool! I didn't know Gremlin had these functions. I will switch to that implementation.

scripts/generate_test_sql/events.py
jecdohmann: I think the linter asked for this (specifically flake8, I think)

graphql_compiler/tests/test_compiler.py
jecdohmann: Predrag noted that Gremlin has a native startsWith/endsWith function  https://blogs.oracle.com/fadevrel/useful-groovy-string-functions

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
jecdohmann: thanks!

docs/source/conf.py
```diff
@@ -43,8 +43,8 @@
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-# This theme is the most popular one and the one used by the 
-# Read the Docs website https://docs.readthedocs.io/en/stable/index.html 
+# This theme is the most popular one and the one used by the
+# Read the Docs website https://docs.readthedocs.io/en/stable/index.html
```
chewselene: was this also a cause by a linting error? might be better to put linting fixes in a different PR

jecdohmann: yes - is there a way to disable linting errors for this PR so I can get this through without fixing the linting errors in this PR?

obi1kenobi: @pmantica1 IIRC I think you checked in this code. Could you:
- check your local environment to make sure it is configured to strip whitespace at the end of lines, and
- investigate with @jecdohmann why the linter complained for him but allowed us to merge the original change in?

pmantica1: Will do.

pmantica1:
```# Get all python files or directories that need to be linted.
lintable_locations="graphql_compiler/"
# pylint doesn't support linting directories that aren't packages:
# https://github.com/PyCQA/pylint/issues/352
# Use **/*.py to supply all python files for individual linting.
pylint_lintable_locations="graphql_compiler/**/*.py"
if [ "$diff_only" -eq 1 ] ; then
    # Quotes don't need to be escaped because they nest with $( ).
    lintable_locations="$(git diff --name-only master... | grep ".*\.py$")"
    pylint_lintable_locations="$lintable_locations"
fi```

The default value of `lintable_locations` is the `graphql_compiler/` directory. However, if you run `lint.sh` with the `--diff` parameter you can include files that are not in the `graphql_compiler` directory, such as those in readthedocs.

Comment (pmantica1) [reply docs/source/conf.py]: Also I am not sure that the regex `"\.*\.py$"` is correct. Shouldn't it be ".*\.py$"? (Without the first slash).

**Comment by jecdohmann** [comment]
> > Also I tried removing revisions to TROUBLESHOOTING.md that I made on a previously accepted pull request per pmantica's request, but they keep showing up. Any suggestions?
> 
> You should be able to simply check out the `master` copy of that file and commit it. When your PR gets squashed and merged, those changes should disappear entirely.

I think it should be fixed - TROUBLESHOOTING.md is no longer showing up in the files changed fro this PR

File: docs/source/conf.py (hunk 7 above)
Comment (obi1kenobi) [reply docs/source/conf.py]: I think you might be right. Test it out to be sure and open a PR?

Comment (obi1kenobi) [reply docs/source/conf.py]: Let's also make `docs` be a lintable directory if possible.

File: graphql_compiler/compiler/filters.py [hunk 8]
```diff
@@ -543,6 +543,86 @@ def _process_has_substring_filter_directive(filter_operation_info, location, con
     return blocks.Filter(filter_predicate)
 
 
+@scalar_leaf_only(u'ends_with')
+@takes_parameters(1)
+def _process_ends_with_filter_directive(filter_operation_info, location, context, parameters):
+    """Return a Filter basic block that checks if the directive arg is the string suffix of the field.
+    Args:
```
obi1kenobi: Use a newline between sections. The docstring sections should be like this:

```python
"""Headline, always one line

[Optional additional text, as many lines as you'd like, same indentation as the triple-quotes]

Args:
    foo: explanation
    bar: explanation

Returns: [only if the function returns something]
    return value description (no name, since return values aren't named)

Raises: [only if the function raises something non-obvious]
    description of errors raised
"""
```

This is missing newlines before `Args` and `Returns`.

graphql_compiler/compiler/filters.py
```diff
@@ -543,6 +543,86 @@ def _process_has_substring_filter_directive(filter_operation_info, location, con
...
+        filter_operation_info: FilterOperationInfo object, containing the directive and field info
+                               of the field where the filter is to be applied.
+        location: Location where this filter is used.
+        context: dict, various per-compilation data (e.g. declared tags, whether the current block
+                 is optional, etc.). May be mutated in-place in this function!
+        parameters: list of 1 element, specifying the collection in which the value must exist;
```
obi1kenobi: Whoops, seems like we checked in some bad copy-pasting in the past (e.g. in `has_substring`), and now it has propagated. Mind fixing it both here and elsewhere?

The issue is that `has_substring` as well as these two new filters don't actually take _collection_ values as parameters, and instead take string values. That means the `specifying the collection` bit is incorrect, as is the description of what happens if the parameter is tagged in an optional scope and missing.

graphql_compiler/compiler/ir_lowering_match/__init__.py
```diff
@@ -96,6 +97,8 @@ def lower_ir(ir_blocks, query_metadata_table, type_equivalence_hints=None):
     ir_blocks = rewrite_binary_composition_inside_ternary_conditional(ir_blocks)
     ir_blocks = merge_consecutive_filter_clauses(ir_blocks)
     ir_blocks = lower_has_substring_binary_compositions(ir_blocks)
+    ir_blocks = lower_starts_with_binary_compositions(ir_blocks)
+    ir_blocks = lower_ends_with_binary_compositions(ir_blocks)
```
obi1kenobi: Since the three functions on lines 99-101 have similar behavior and implementations, what do you think about merging them? Would that increase the complexity too much, or might it be okay?

docs/source/conf.py
obi1kenobi: Also `scripts`, it seems like that's not marked as lintable either. Thanks @jecdohmann for finding these issues with our linter setup, and sorry for the frustrations they must have caused.

jecdohmann: No worries - Pedro, let me know if you want to discuss this at all with me. Otherwise, I'll just leave the linting stuff as is for this PR

graphql_compiler/compiler/filters.py
jecdohmann: Could you give an example query snippet that would result in that argument being non-empty? I am not sure I follow what the description here means

jecdohmann: I think I get it. I am changing it to the following:

```parameters: list of 1 element, specifying the variable to which the value must be bound;
                    if the variable is optional and missing, the check will return True.```

Looks like many of the filter directives need their doc changed

File: graphql_compiler/compiler/ir_lowering_match/__init__.py (hunk 10 above)
Comment (jecdohmann) [reply graphql_compiler/compiler/ir_lowering_match/__init__.py]: I think all your suggestions should be resolved in my most recent commit - lmk what you think

File: graphql_compiler/compiler/filters.py (hunk 9 above)
Comment (obi1kenobi) [reply graphql_compiler/compiler/filters.py]: Yeah 😞 I might replace `variable` with `value` just to be a bit more precise, but I think that's the right idea.

File: graphql_compiler/compiler/ir_lowering_match/ir_lowering.py [hunk 11]
```diff
@@ -93,44 +93,98 @@ def visitor_fn(expression):
     return new_ir_blocks
 
 
-def lower_has_substring_binary_compositions(ir_blocks):
-    """Lower Filter blocks that use the "has_substring" operation into MATCH-representable form."""
+def lower_string_operators(ir_blocks):
+    """Lower Filter blocks that use the "has_substring", "starts_with", or
```
obi1kenobi: This is not a compliant docstring, per https://github.com/kensho-technologies/graphql-compiler/pull/499/files#r316730267

The headline must always fit on one line. The rule is that if you need more than one line to describe the function, then it's not that obvious, and you should stick to a one-line summary + a description underneath + write out `Args` and `Returns` sections as well.

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -93,44 +93,98 @@ def visitor_fn(expression):
...
+            #        BinaryComposition(
+            #           u'+',
+            #           Literal("%"),
+            #           Y
+            #        )
+            #    )
+            return BinaryComposition(
+                u'LIKE',
+                expression.left,
+                BinaryComposition(
```
obi1kenobi: There's a lot of repetition here, this is literally just the three functions copy-pasted into one function. I was hoping for a refactor that gets rid of all the repetitions. Consider how left-addition of `%` is repeated twice, right-addition of `%` is repeated twice, and putting all of that in a `LIKE` clause is repeated three times – each time also requiring a lengthy comment describing the transformation.

We should be able to define and document some helper functions that compose nicely together and reduce the complexity and repetitiveness of the code.

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -93,95 +93,47 @@ def visitor_fn(expression):
     return new_ir_blocks
 
 
-def lower_string_operators(ir_blocks):
-    """Lower Filter blocks that use the "has_substring", "starts_with", or
-    "ends_with" operation into MATCH-representable form."""
+def prepend_wildcard(expression):
```
obi1kenobi: Any reason to make these functions public, instead of prefixing them with an underscore and therefore signaling that they are meant to be private?

Also, please add docstrings.

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -93,95 +93,47 @@ def visitor_fn(expression):
...
+    )
 
+
+def lower_string_operators(ir_blocks):
+    """Lower Filters with "has_substring", "starts_with", or "ends_with" operation into MATCH"""
     def visitor_fn(expression):
         if not isinstance(expression, BinaryComposition):
             return expression
         elif expression.operator == u'has_substring':
-            # Rewrite BinaryComposition expressions with "has_substring" into representable form.
```
obi1kenobi: Let's keep one of these detailed comments as an example (I'd keep this one since it's the most detailed), and explain the slight differences compared to the lowering of `starts_with` and `ends_with`. Deleting all three of them seems like an unnecessary loss of documentation.

jecdohmann: sounds good, done

graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
jecdohmann: I didn't know about that convention, thank you

PR #509, opened 2019-08-21
Use the SchemaInfo classes, fixing SQL tests and SQL integration tests

bojanserafimov:
The main problem is fixing the SQL interface and testing setup. The **client-facing SQL interface does not take sufficient information to compile correctly, so it had to be changed**. After changing the interface, I updated the SQL tests and integration tests to use this new interface.

I made refactoring changes (no api redesigns) of other backends, so that all backends can be treated in the same way. The changes mostly consist of propagating the `Backend` and `SchemaInfo` classes where they belong:
- the `_compile_graphql_generic` used to take mostly undocumented parameters, with a questionable `compiler_metadata` wildcard argument. Now it takes a `Backend` object, which can be documented.
- the `compiler_metadata` concept is now gone. It was too vague, and existed solely for the SQL backend. SQL schema information is now instead passed around in a `SQLAlchemySchemaInfo`
- The `schema` and `type_equivalence_hints` are now grouped into a `CommonSchemaInfo` object. They are `the sufficient schema information needed to compile`, so it makes sense to group them.
- The `ir_blocks` and `query_metadata_table` are passed around together in an IR object. Over time we have slowly made the `query_metadata_table` a more important part of the IR than the actual blocks, so I think this is progress towards a better IR.

Making room for future work:
- The integration tests for all backends currently assume no `type_equivalence_hints`
- Better test setup (via the `BackendTester` common interface)
- Better integration test data, consistent across backends

I couldn't find a good place to split this PR up while still passing tests. Let me know if you think it's necessary, or if you have ideas.


bojanserafimov:
@draft_reviewers: Please see if you are ok with the breaking change to the compile SQL interface before I go on actually implementing this

pmantica1:
I agree that this is necessary. Let's just keep in mind that next version bump is going to be a major version bump.

bojanserafimov:
> I agree that this is necessary. Let's just keep in mind that next version bump is going to be a major version bump.

The readme on master currently promises some things about our SQL backend that we don't actually deliver on, like traversals. I might need to fix that too.

bojanserafimov:
> > I agree that this is necessary. Let's just keep in mind that next version bump is going to be a major version bump.
> 
> The readme on master currently promises some things about our SQL backend that we don't actually deliver on, like traversals. I might need to fix that too.

Addressed in latest commit.

obi1kenobi:
It seems like the SQL tests in `test_compiler.py` have moved to MSSQL syntax. Not a problem, just pointing it out.

When we add other backends back in there, we might want to standardize to something more common. Perhaps SQLite (if it supports array-typed columns) so we can use it for a fast(er) sub--set of integration tests? If not that, then maybe Postgres since it's much more commonly used.

graphql_compiler/compiler/common.py
```diff
@@ -146,41 +116,29 @@ def compile_graphql_to_cypher(schema, graphql_string, type_equivalence_hints=Non
...
     """Compile the GraphQL input, lowering and emitting the query using the given functions.
 
     Args:
-        language: string indicating the target language to compile to.
-        lowering_func: Function to lower the compiler IR into a compatible form for the target
-                       language backend.
-        query_emitter_func: Function that emits a query in the target language from the lowered IR.
-        schema: GraphQL schema object describing the schema of the graph to be queried.
+        target_backend: Backend used to compile the query
+        schema_info: target_backend.schemaInfoClass containing all necessary schema information.
```
pmantica1: Shouldn't this be SchemaInfoClass? (In uppercase).

graphql_compiler/__init__.py
```diff
@@ -58,31 +58,13 @@ def graphql_to_match(schema, graphql_query, parameters, type_equivalence_hints=N
...
-                                Recursive expansion of type equivalence hints is not performed,
-                                and only type-level correctness of this argument is enforced.
-                                See README.md for more details on everything this parameter does.
-                                *****
-                                Be very careful with this option, as bad input here will
-                                lead to incorrect output queries being generated.
-                                *****
 
     Returns:
         a CompilationResult object, containing:
```
pmantica1: Do you mind fixing the docstring for CompilationResult? Specifically for SQL, compilation_result.query is not a string, but it is some SQLAlchemy object.

PR #510, opened 2019-08-21
Move "compare_*" functions from test_helpers to query_comparison.py

bojanserafimov:
Just organizing code into modules with meaningful names and making it easier to avoid circular imports (we shouldn't be importing `test_helpers` from outside tests).

### DIFF
--- a/graphql_compiler/query_comparison.py
+++ b/graphql_compiler/query_comparison.py
@@ -1,0 +1,77 @@
+from pprint import pformat
+import re
+import six
+
+from .debugging_utils import pretty_print_gremlin, pretty_print_match
+from .query_formatting.graphql_formatting import pretty_print_graphql
+
+
+def _get_mismatch_message(expected_blocks, received_blocks):
+    """Create a well-formated error message indicating that two lists of blocks are mismatched."""
+    pretty_expected = pformat(expected_blocks)
+    pretty_received = pformat(received_blocks)
+    return u'{}\n\n!=\n\n{}'.format(pretty_expected, pretty_received)
+
+
+def strip_whitespace(emitted_output):
+    """Transform emitted_output into a unique representation, regardless of lines / indentation."""
+    # The strings which we will be comparing have newlines and spaces we'd like to get rid of,
+    # so we can compare expected and produced emitted code irrespective of whitespace.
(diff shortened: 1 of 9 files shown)


graphql_compiler/query_comparison.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def _get_mismatch_message(expected_blocks, received_blocks):
+    """Create a well-formated error message indicating that two lists of blocks are mismatched."""
+    pretty_expected = pformat(expected_blocks)
+    pretty_received = pformat(received_blocks)
+    return u'{}\n\n!=\n\n{}'.format(pretty_expected, pretty_received)
+
+
+def strip_whitespace(emitted_output):
```
ReaLNeroM: I see how `emitted_output` makes sense for backends like match and gremlin, but I don't think it fits with `compare_graphql`, as it's not emitted. My logic is that I've seen functions called `emit_sql` and `emit_match` throughout the code, so I would expect an input to this function to have been outputted from an `emit_*` function. This isn't true for graphql queries.
Perhaps rename to `query_string`? Your call.

graphql_compiler/query_comparison.py
```diff
@@ -0,0 +1,79 @@
...
+    """Compare expected and received code, ignoring whitespace, with the given failure message."""
+    test_case.assertEqual(strip_whitespace(expected), strip_whitespace(received), msg=msg)
+
+
+def compare_ir_blocks(test_case, expected_blocks, received_blocks):
+    """Compare the expected and received IR blocks."""
+    mismatch_message = _get_mismatch_message(expected_blocks, received_blocks)
+
+    if len(expected_blocks) != len(received_blocks):
+        test_case.fail(u'Not the same number of blocks:\n\n'
```
ReaLNeroM: This code is no longer in the `tests` directory, but still contains `test_case` inside itself.

obi1kenobi: Agreed, we should not be depending on `TestCase` objects being passed in here. Let's rethink the API we want to expose instead, to make it suitable for use outside of test code.

bojanserafimov:
Good point vlad. I'm coming with a different approach, separate PR.

PR #511, opened 2019-08-21
Add BackendTester class

bojanserafimov:
Splitting up the production-usable and test-only parts of the `Backend` class.

### DIFF
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -1,5 +1,6 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Common test data and helper functions."""
+from collections import namedtuple
 from pprint import pformat
 import re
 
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/test_helpers.py
```diff
@@ -204,6 +205,28 @@
...
+
+    # Given a SchemaInfo and a connection pool to a database, install the given schema into
+    # the database, erasing content if necessary.
+    'setup_schema'
+
+    # Given a SchemaInfo, a dict representation of data fitting the schema, and a connection pool
+    # to a database with the same schema, install the given data into the database, erasing any
+    # existing data.
+    'setup_data'
+))
```
bojanserafimov: I will make the docstrings more rigorous when I implement the first `BackendTester`. I can't predict what's the best precise signature of each term here.

PR #515, opened 2019-08-23
Add linting for docs directory

pmantica1:


obi1kenobi:
Could we set the linters to look at all files in the entire repo, and not just specific directories?

PR #516, opened 2019-08-23
New sql backend

bojanserafimov:
New stuff:
- traversals
- tags
- tests

Restructuring:
- Using common lowering passes.
- Using the `query_metadata_table` instead of recomputing it.
- Removed the `SqlTree`. It might come back in a lighter form, if we need it for `@optional` or `@recurse`.
- Moved the expression `to_sql` methods in the `expressions.py` module for consistency with other backends

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -40,7 +40,6 @@ def _traverse_and_validate_blocks(ir):
         yield block
 
 
-
 class CompilationState(object):
     """Mutable class used to keep track of state while emitting a sql query."""
 


graphql_compiler/tests/test_compiler.py
```diff
@@ -406,7 +406,14 @@ def test_traverse_and_output(self):
...
         '''
-        expected_sql = NotImplementedError
+        expected_sql = '''
+            SELECT
+                [Animal_1].name AS parent_name
+            FROM
+                db_1.schema_1.[Animal] AS [Animal_2]
+                JOIN db_1.schema_1.[Animal] AS [Animal_1]
+                    ON [Animal_2].parent = [Animal_1].uuid
+        '''
```
bojanserafimov: graphql_input = '''{
    Animal {
        out_Animal_ParentOf {
        }
    }
}'''

graphql_compiler/tests/test_compiler.py
```diff
@@ -2973,7 +2980,16 @@ def test_has_substring_op_filter_with_tag(self):
...
+        expected_sql = '''
+            SELECT
+                [Animal_1].name AS animal_name
+            FROM
+                db_1.schema_1.[Animal] AS [Animal_1]
+                JOIN db_1.schema_1.[Animal] AS [Animal_2]
+                    ON [Animal_1].parent = [Animal_2].uuid
+            WHERE
+                ([Animal_2].name LIKE '%' + [Animal_1].name + '%')
+        '''
```
bojanserafimov:
```        
graphql_input = '''{
    Animal {
        Animal {
            name @output(out_name: "animal_name") @tag(tag_name: "root_name")
            name @output(out_name: "animal_name") @tag(tag_name: "root_name")
            out_Animal_ParentOf {
            out_Animal_ParentOf {
                name @filter(op_name: "has_substring", value: ["%root_name"])
                name @filter(op_name: "has_substring", value: ["%root_name"])
            }
            }
        }
        }
    }'''
 }'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4604,7 +4620,20 @@ def test_no_op_coercion_with_eligible_subpath(self):
...
+                db_1.schema_1.[Animal] AS [Animal_2]
+                JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                    ON [Animal_2].parent = [Animal_3].uuid
+                JOIN db_1.schema_1.[Animal] AS [Animal_1]
+                    ON [Animal_3].parent = [Animal_1].uuid
+                JOIN db_1.schema_1.[Entity] AS [Entity_1]
+                    ON [Animal_3].related_entity = [Entity_1].uuid
+            WHERE
+                [Entity_1].name IN ([EXPANDING_entity_names])
+        '''
```
bojanserafimov:
```
graphql_input = '''{
    Animal {
        out_Animal_ParentOf {
            ... on Animal {
                out_Animal_ParentOf {
                }
                out_Entity_Related {
                    ... on Entity {
                        name @filter(op_name: "in_collection", value: ["$entity_names"])
                    }
                }
            }
        }
    }
}'''
```

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1,285 +1,110 @@
...
+        """Move to a different location in the query, updating the _classname and _alias."""
+        self._location = new_location
+        new_location_type = self._ir.query_metadata_table.get_location_info(new_location).type
+        self._classname = new_location_type.name
+        if self._location.query_path in self._aliases:
+            self._alias = self._aliases[self._location.query_path]
+        else:
+            # mark_name, _ = self._location.get_location_name()
+            # XXX Use a consistent mark name with other backends, or stick with the
+            #     one generated by sqlalchemy?
```
bojanserafimov: opinions? Our generated names look like `Animal__out_Animal_OfSpecies___1` (full path + gremlin revisit count), whereas the sqlalchemy-generated names look like `Animal_3` (table name + disambiguation number).

README.rst
```diff
@@ -1758,15 +1758,22 @@ the query were written in the opposite order:
...
 
-Currently we only support filtering and outputting properties of a single vertex/table. Support
-for edge traversals and most directives is in development.
+Relational databases are supported by compiling to SQLAlchemy core as an intermediate
+language, and then relying on SQLAlchemy's compilation of the dialect specific SQL string to query
+the target database.
+
+Our SQL backend supports basic traversals, filters, tags and outputs, but there's still some pieces
+in development:
+- Directives: `@optional`, `@fold`, `@recurse`
```
pmantica1: This will not render correctly in rst. All lists must be preceded with a blank line. Additionally, also please change `@optional` to :code:`optional` and similarly for other highlighted code blocks.

README.rst
```diff
@@ -1758,15 +1758,23 @@ the query were written in the opposite order:
 SQL
 ---
 
-Currently we only support filtering and outputting properties of a single vertex/table. Support
-for edge traversals and most directives is in development.
+Relational databases are supported by compiling to SQLAlchemy core as an intermediate
+language, and then relying on SQLAlchemy's compilation of the dialect specific SQL string to query
```
obi1kenobi: nit: I think `dialect specific` maybe should be hyphenated, `dialect-specific`

README.rst
```diff
@@ -1758,15 +1758,23 @@ the query were written in the opposite order:
...
+the target database.
+
+Our SQL backend supports basic traversals, filters, tags and outputs, but there's still some pieces
```
obi1kenobi: nit: `there's` -> `there are`

README.rst
```diff
@@ -1758,15 +1758,23 @@ the query were written in the opposite order:
...
+in development:
+
+- Directives: :code:`@optional`, :code:`@fold`, :code:`@recurse`
+- Filter operators: :code:`has_edge_degree`, :code:`is_null`, :code:`is_not_null`
```
obi1kenobi: Of everything here, I think `has_edge_degree` has by far the lowest priority, feel free to ignore it.

graphql_compiler/compiler/emit_sql.py
obi1kenobi: No preference. Seems like using SQLAlchemy's default names might be less work? If so, let's stick with them.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1,285 +1,110 @@
...
+    def __init__(self, sql_schema_info, ir):
+        """Initialize a CompilationState, setting the current location at the root of the query."""
+        # Metadata
+        self._sql_schema_info = sql_schema_info
+        self._ir = ir
+
+        # Current query location state. Only mutable by calling _relocate.
+        self._location = None  # the current location in the query
+        self._classname = None  # the classname at the current location
+        self._alias = None  # a sqlalchemy table Alias at the current location
```
obi1kenobi: Maybe `_current_alias`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1,285 +1,110 @@
...
-    Returns:
-        Expression, SQLAlchemy expression.
-    """
-    column_name = expression.field_name
-    column = sql_context_helpers.get_column(column_name, node, context)
-    return column
+    state = CompilationState(sql_schema_info, ir)
+    for block in ir.ir_blocks:
+        if isinstance(block, blocks.QueryRoot):
+            pass
```
obi1kenobi: Might there be an assertion worth doing here?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1,285 +1,110 @@
...
+        elif isinstance(block, blocks.Backtrack):
+            state.backtrack(block.location)
+        elif isinstance(block, blocks.Traverse):
+            if block.optional:
+                raise NotImplementedError(u'The SQL backend does not support @optional.')
+            state.traverse(u'{}_{}'.format(block.direction, block.edge_name))
+        elif isinstance(block, blocks.Filter):
+            state.filter(block.predicate)
+        elif isinstance(block, blocks.GlobalOperationsStart):
+            pass
```
obi1kenobi: Maybe prevent further traversals or backtracks in the state at this point?

graphql_compiler/compiler/expressions.py
```diff
@@ -127,6 +130,10 @@ def _to_output_code(self):
     to_match = _to_output_code
     to_cypher = _to_output_code
 
+    def to_sql(self, aliases, current_alias):
+        """Return the value."""
+        return self.value
```
obi1kenobi: For consistency, this probably should have a `self.validate()` inside it.

graphql_compiler/compiler/expressions.py
```diff
@@ -305,6 +319,18 @@ def to_gremlin(self):
...
 
+    def to_sql(self, aliases, current_alias):
+        """Return a sqlalchemy Column picked from the current_alias."""
+        self.validate()
+
+        if isinstance(self.field_type, GraphQLList):
+            raise NotImplementedError(u'We dont support lists yet')
+
+        if '@' in self.field_name:
+            raise NotImplementedError(u'We dont support __typename yet')
```
obi1kenobi: nit: `dont` -> `don't` and let's mention which field and type were the problematic ones.

graphql_compiler/compiler/expressions.py
```diff
@@ -545,6 +592,16 @@ def to_cypher(self):
...
 
+    def to_sql(self, aliases, current_alias):
+        """Return a sqlalchemy Column picked from the appropriate alias."""
+        if isinstance(self.field_type, GraphQLList):
+            raise NotImplementedError(u'The SQL backend does not support lists.')
+
+        if '@' in self.location.field:
+            raise NotImplementedError(u'The sql backend does not support typename.')
+
+        return aliases[self.location.at_vertex().query_path].c[self.location.field]
```
obi1kenobi: Should we maybe have a helper function that takes `aliases` and a location object and returns `aliases[self.location.at_vertex().query_path].c[self.location.field]` after doing all the checks for unsupported cases? Seems like this code exists in a few places.

graphql_compiler/compiler/expressions.py
```diff
@@ -1033,6 +1104,32 @@ def to_cypher(self):
...
+            u'=': python_operator.__eq__,
+            u'!=': python_operator.__ne__,
+            u'<': python_operator.__lt__,
+            u'>': python_operator.__gt__,
+            u'<=': python_operator.__le__,
+            u'>=': python_operator.__ge__,
+            u'&&': sql.expression.and_,
+            u'||': sql.expression.or_,
+            u'has_substring': sql.operators.ColumnOperators.contains,
+            # in_collection and not_in_collection are implemented using contains and not_contains
```
obi1kenobi: doesn't this mean that `in_collection` and `not_in_collection` should still exist in this dict but be mapped to `contains` and `not_contains` instead? I think right now those cases would end up in the `NotImplementedError` on line 1126.

graphql_compiler/compiler/expressions.py
```diff
@@ -1127,3 +1224,11 @@ def to_cypher(self):
             predicate=self.predicate.to_cypher(),
             if_true=self.if_true.to_cypher(),
             if_false=self.if_false.to_cypher())
+
+    def to_sql(self, aliases, current_alias):
+        """Return a sqlalchemy Case representing this TernaryConditional."""
+        self.validate()
+        return sql.expression.case(
```
obi1kenobi: Maybe split this into multiple expressions for the sake of readability?

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+import sqlalchemy
+
+
+def contains_operator(collection, element):
+    """Return a sqlalchemy BinaryExpression representing this operator.
+
+    Args:
+        - collection: sqlalchemy BindParameter, a collection runtime parameter
```
obi1kenobi: nit: remove the `-` before the parameter names, both here and in the other functions

graphql_compiler/compiler/expressions.py
bojanserafimov: It appears twice. If I create a helper, all it would do is force the reader to hop up and down to understand the code (ruins locality in reading). Do you see a problem that the change would solve?

graphql_compiler/compiler/expressions.py
bojanserafimov: What I meant was IR generation converts an `in_collection` to a `contains`, so an `in_collection` in the query is a `contains` in the IR. You can see that even the `MATCH` backend doesn't bother to define `in_collection`. I will clarrify.

PR #517, opened 2019-08-24
SQLAlchemySchemaInfo generation specification

pmantica1:
Please no nits at these point. Just look at the high level structure and see if you agree.


graphql_compiler/schema_generation/sqlalchemy/sqlalchemy_schema_builder.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+SQLEdgeDescriptor = namedtuple(
+    'SQLEdgeDescriptor',
+    (
+        'from_table',  # Identifier of the source table.
```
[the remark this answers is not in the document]
pmantica1: I think that we should not call the field `from_property` but instead `from_column`. Suppose someone uses the schema we expose to create a subset schema, (displaying a subset of the types in the original schema). If that person were to hide the 'from_property' property field in the GraphQL schema then the SQLEdgeDescriptor might refer to a non-existent property.  That person might not even have access to the properties used in the SQLEdgeDescriptor objects, so essentially there is no way for the person to safely hide fields in this case.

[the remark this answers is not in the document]
pmantica1: This goes back to the point that we should be referering to tables and their columns vs vertices and their properties.

I still don't understand what do you find ambiguos about the use of the word identifier when referring to tables.

graphql_compiler/__init__.py
```diff
@@ -218,3 +224,57 @@ def graphql_to_redisgraph_cypher(schema, graphql_query, parameters, type_equival
...
+    graphql_schema = get_graphql_schema_from_schema_graph(
+        schema_graph, class_to_field_type_overrides, hidden_classes)
+
+    join_descriptors = get_restructured_edge_descriptors(sql_edge_descriptors)
+    # Note: We do not have inheritance in SQL so we should have type_equivalence_hints.
+    #       I think it's ok to have type_equivalence_hints to make the tests easier but let's not
+    #       expose those here.
+    type_equivalence_hints = {}
+    return make_sqlalchemy_schema_info(
+        graphql_schema, tables, join_descriptors, type_equivalence_hints, dialect)
```
bojanserafimov: The `tables` in the schema info map GraphQL vertices to tables. Here you're giving a mapping from identifiers to tables, which are not the same.

pmantica1: Since the identifiers will be used for vertex names based on the docstrings, then it will still be a "dict mapping every graphql object type or interface type name in the schema to a sqlalchemy table"

graphql_compiler/__init__.py
```diff
@@ -218,3 +219,52 @@ def graphql_to_redisgraph_cypher(schema, graphql_query, parameters, type_equival
...
         query=insert_arguments_into_query(compilation_result, parameters))
+
+
+def get_sqlalchemy_schema_info(
+    tables, sql_edge_descriptors, junction_tables, dialect, class_to_field_type_overrides=None
+):
+    """Return a SQLAlchemyInfo from the metadata.
+
+    Args:
+        tables: dict, str -> SQLAlchemy Table, mapping every GraphQL object in the schema to a
```
bojanserafimov: Which graphql schema are you talking about in `the schema`?

graphql_compiler/__init__.py
```diff
@@ -218,3 +219,52 @@ def graphql_to_redisgraph_cypher(schema, graphql_query, parameters, type_equival
...
+                SQLAlchemy Table. The columns of each table, (with a supported type), will me mapped
+                to a GraphQL field with the same name as the column in the corresponding GraphQL
+                object.
+        sql_edge_descriptors: dict, str-> SQLEdgeDescriptor, mapping the names of edges in
+                              the schema to namedtuple objects specifying the source and destination
```
bojanserafimov: nobody cares if they're classes or namedtuples. Just use their name.

graphql_compiler/__init__.py
pmantica1: The schema that is about to be constructed.

graphql_compiler/__init__.py
pmantica1: Ok gotcha will do.

graphql_compiler/schema_generation/sqlalchemy/edge_descriptors.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from collections import namedtuple
+
+SQLEdgeDescriptor = namedtuple(
```
bojanserafimov: I'd call this a `ForeignKeyEdgeDescriptor` or `DirectEdgeDescriptor`. Now you're implying the `JunctionTableEdgeDescriptor` is not a sql edge descriptor.

pmantica1: Ok.

graphql_compiler/__init__.py
```diff
@@ -218,3 +222,69 @@ def graphql_to_redisgraph_cypher(schema, graphql_query, parameters, type_equival
...
+
+
+def get_sqlalchemy_schema_info(
+    vertex_name_to_table, direct_edges, junction_table_edges, dialect,
+    class_to_field_type_overrides=None
+):
+    """Return a SQLAlchemyInfo from the metadata.
+
+    Args:
+        vertex_name_to_table: dict, str -> SQLAlchemy Table. This dictionary is used to generate the
```
pmantica1: I use the variable name `vertex_name_to_table` instead of `tables` since there are other `tables`, (junction tables), that are not included in this dict.

bojanserafimov:
close this PR?

obi1kenobi:
Should we delete the branch here? Let's try to not leave unmerged and inactive branches lying around.

pmantica1:
No. This diff includes the potential specification for adding junction tables into the SQLAlchemySchemaInfo, which we might want to do in the future.

> Should we delete the branch here? Let's try to not leave unmerged and inactive branches lying around.

PR #518, opened 2019-08-25
Tests for restrict_schema_types

qqi0O0:
Some behaviors in the tests are probably bad, such as keeping an interface/union but not all the subtypes.


pmantica1:
Thank you so much!

LWprogramming:
closing for same reason as #834 and #828, the bulk of it has been merged in

PR #528, opened 2019-08-28
Sql optionals

bojanserafimov:
Implemented:
- tag in optional scope
- filter in optional scope
- mandatory traverse in optional scope

todo:
- Remove the hack (abusing python features)
- Rewrite `([FeedingEvent_3].uuid IS NOT NULL) = 0` as `[FeedingEvent_3].uuid IS NULL`

### DIFF
--- a/graphql_compiler/compiler/ir_lowering_sql/__init__.py
+++ b/graphql_compiler/compiler/ir_lowering_sql/__init__.py
@@ -1,4 +1,6 @@
 # Copyright 2018-present Kensho Technologies, LLC.
+import six
+
 from .. import blocks, expressions
 from ...compiler.compiler_frontend import IrAndMetadata
 from ..helpers import FoldScopeLocation, get_edge_direction_and_name
@@ -45,17 +47,30 @@ def _find_non_null_columns(schema_info, query_metadata_table):
 
 
 class ContextColumn(expressions.Expression):
-    """A column drawn from the global context."""
+    """A column drawn from the global context.
+
+    It is different than an expressions.ContextField because it does not reference a property
+    type in the GraphQL schema, but a column name in the actual SQL table. Some columns are
+    not even represented in the GraphQL schema as properties. An example is Animals.parent
+    in the test schema.
+    """
 
     def __init__(self, vertex_query_path, column_name):
         """Construct a new ContextColumn."""
-        super(ContextColumn, self).__init__()
+        super(ContextColumn, self).__init__(vertex_query_path, column_name)
         self._vertex_query_path = vertex_query_path
         self._column_name = column_name
+        self.validate()
 
     def validate(self):
         """Validate that the ContextColumn is correctly representable."""
-        pass
+        if not isinstance(self._vertex_query_path, tuple):
+            raise AssertionError(u'vertex_query_path was expected to be a tuple, but was {}: {}'
+                                 .format(type(self._vertex_query_path), self._vertex_query_path))
+
+        if not isinstance(self._column_name, six.string_types):
+            raise AssertionError(u'column_name was expected to be a string, but was {}: {}'
+                                 .format(type(self._column_name), self._column_name))
 
     def to_match(self):
         """Not implemented, should not be used."""
@@ -74,6 +89,7 @@ class ContextColumn(expressions.Expression):
 
     def to_sql(self, aliases, current_alias):
         """Return a sqlalchemy Column picked from the appropriate alias."""
+        self.validate()
         return aliases[self._vertex_query_path].c[self._column_name]
 
 
@@ -83,13 +99,14 @@ def _lower_sql_context_field_existence(schema_info, ir_blocks, query_metadata_ta
 
     def visitor_fn(expression):
         """Convert ContextFieldExistence expressions to TrueLiteral."""
-        if isinstance(expression, expressions.ContextFieldExistence):
-            query_path = expression.location.query_path
-            return expressions.BinaryComposition(
-                u'!=',
-                ContextColumn(query_path, non_null_columns[query_path]),
-                expressions.NullLiteral)
-        return expression
+        if not isinstance(expression, expressions.ContextFieldExistence):
+            return expression
+
+        query_path = expression.location.query_path
+        return expressions.BinaryComposition(
+            u'!=',
+            ContextColumn(query_path, non_null_columns[query_path]),
+            expressions.NullLiteral)
 
     return [block.visit_and_update_expressions(visitor_fn) for block in ir_blocks]
 
--- a/graphql_compiler/tests/test_compiler.py
+++ b/graphql_compiler/tests/test_compiler.py
@@ -5756,7 +5756,6 @@ class CompilerTests(unittest.TestCase):
                     ON [Animal_3].species = [Species_1].uuid
             WHERE
                 [Species_1].uuid IS NOT NULL OR [Animal_3].uuid IS NULL
-
         '''
         expected_cypher = SKIP_TEST
 


obi1kenobi:
Assuming #531 should merge first, and then we work on removing the hacks here?

graphql_compiler/tests/test_compiler.py
```diff
@@ -1345,7 +1356,36 @@ def test_complex_optional_variables(self):
...
+                JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                    ON [Animal_2].uuid = [Animal_3].parent
+                LEFT OUTER JOIN db_2.schema_1.[FeedingEvent] AS [FeedingEvent_3]
+                    ON [Animal_3].fed_at = [FeedingEvent_3].uuid
+                JOIN db_1.schema_1.[Animal] AS [Animal_4]
+                    ON [Animal_1].uuid = [Animal_4].parent
+                JOIN db_2.schema_1.[FeedingEvent] AS [FeedingEvent_2]
+                    ON [Animal_4].fed_at = [FeedingEvent_2].uuid
+            WHERE (
+                ([FeedingEvent_1].uuid IS NOT NULL) = 0 OR
```
obi1kenobi: We should be able to lower `([FeedingEvent_1].uuid IS NOT NULL) = 0` into `[FeedingEvent_1].uuid IS NULL`, right?

graphql_compiler/tests/test_compiler.py
```diff
@@ -5343,7 +5434,20 @@ def test_optional_and_traverse(self):
...
+                [Animal_2].name AS grandchild_name,
+                [Animal_3].name AS name
+            FROM
+                db_1.schema_1.[Animal] AS [Animal_3]
+                LEFT OUTER JOIN db_1.schema_1.[Animal] AS [Animal_1]
+                    ON [Animal_3].uuid = [Animal_1].parent
+                LEFT OUTER JOIN db_1.schema_1.[Animal] AS [Animal_2]
+                    ON [Animal_1].uuid = [Animal_2].parent
+            WHERE
+                [Animal_2].parent IS NOT NULL OR [Animal_1].parent IS NULL
```
obi1kenobi: Let's imagine a situation where someone is running a SQL database without constraint checking. It might be possible that `[Animal_2].parent IS NOT NULL` is satisfied but the join to `Animal_3` still doesn't find anything. For the not-null case, might we be better off checking that the primary key on `Animal_3` is non-null?

Put another way, the foreign key column being `null` guarantees there's no edge there. But it being non-null only guarantees there's an edge there if constraint-checking is on, and deletes run with `CASCADE` set. While it's a good idea to do this, many databases and systems do not.

If you think it's too much work, it's fine either way. But if it's about equal work either way, let's err on the side of supporting more use cases.

graphql_compiler/tests/test_compiler.py
bojanserafimov: The way to do this is to lower `ContextFieldExistence(V)` into `BinaryExpression('!=', V.came_from, None)`. The `came_from` value is not known during lowering. It's possible to compute it, but it's not trivial, and maybe even not worth considering this is purely a cosmetic fix.

graphql_compiler/tests/test_compiler.py
bojanserafimov: I don't get what you mean. We're not joining to `Animal_3`, that's where we start. Talk offline?

obi1kenobi: Resolved offline. Just a misunderstanding.

obi1kenobi:
Takeaways from in-person conversation:
- Let's ignore the lowering of `([FeedingEvent_1].uuid IS NOT NULL) = 0` into `[FeedingEvent_1].uuid IS NULL` for now, and just open an "enhancement" issue describing what needs to be done.
- Let's remove the `came_from` hack and use a dict mapping aliases to `came_from` values. We might need to pass additional info to `to_sql()` methods as a result.

After those two steps, this is ready to merge.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -21,6 +22,78 @@ def visitor_fn(expression):
...
+        return aliases[self._vertex_query_path].c[self._column_name]
+
+
+def _lower_sql_context_field_existence(schema_info, ir_blocks, query_metadata_table):
+    """Lower ContextFieldExistence to BinaryComposition."""
+    non_null_columns = _find_non_null_columns(schema_info, query_metadata_table)
+
+    def visitor_fn(expression):
+        """Convert ContextFieldExistence expressions to TrueLiteral."""
+        if isinstance(expression, expressions.ContextFieldExistence):
```
obi1kenobi: Mild preference for early-return like `if not ...: return expression` instead of this, since here the logic inside the `if` is likely to grow whereas the outside is likely to stay `return expression`.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -21,6 +22,78 @@ def visitor_fn(expression):
...
+
+            # The value of the column used to join to this table is an indicator of whether
+            # the left join was a hit or a miss.
+            non_null_column[child_location.query_path] = edge.to_column
+
+    return non_null_column
+
+
+class ContextColumn(expressions.Expression):
+    """A column drawn from the global context."""
```
obi1kenobi: Can you add some more documentation about what this is, and how it's different than the existing expressions?

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -21,6 +22,78 @@ def visitor_fn(expression):
...
+
+    def __init__(self, vertex_query_path, column_name):
+        """Construct a new ContextColumn."""
+        super(ContextColumn, self).__init__()
+        self._vertex_query_path = vertex_query_path
+        self._column_name = column_name
```
obi1kenobi: Call `self.validate()` at the end here.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -21,6 +22,78 @@ def visitor_fn(expression):
...
+
+    return non_null_column
+
+
+class ContextColumn(expressions.Expression):
+    """A column drawn from the global context."""
+
+    def __init__(self, vertex_query_path, column_name):
+        """Construct a new ContextColumn."""
+        super(ContextColumn, self).__init__()
```
obi1kenobi: Pass the args to `__init__`, otherwise the printing of this expression will be broken.

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -21,6 +22,78 @@ def visitor_fn(expression):
...
+        self._vertex_query_path = vertex_query_path
+        self._column_name = column_name
+
+    def validate(self):
+        """Validate that the ContextColumn is correctly representable."""
+        pass
```
obi1kenobi: Presumably there's at least something you can check about the state of the `ContextColumn`, maybe just the types of the attributes?

graphql_compiler/compiler/ir_lowering_sql/__init__.py
```diff
@@ -21,6 +22,78 @@ def visitor_fn(expression):
...
+                             u'in Gremlin, so this is a bug. This function should not be called.')
+
+    def to_cypher(self):
+        """Not implemented, should not be used."""
+        raise AssertionError(u'ContextColumns are not used during the query emission process '
+                             u'in cypher, so this is a bug. This function should not be called.')
+
+    def to_sql(self, aliases, current_alias):
+        """Return a sqlalchemy Column picked from the appropriate alias."""
+        return aliases[self._vertex_query_path].c[self._column_name]
```
obi1kenobi: We've been calling `validate()` in `to_x()` methods elsewhere, might be good to keep that up even though here it's a bit redundant.

graphql_compiler/tests/test_compiler.py
```diff
@@ -5599,7 +5740,24 @@ def test_traverse_and_optional_and_traverse(self):
...
+                db_1.schema_1.[Animal] AS [Animal_1]
+                JOIN db_1.schema_1.[Animal] AS [Animal_2]
+                    ON [Animal_1].uuid = [Animal_2].parent
+                LEFT OUTER JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                    ON [Animal_2].parent = [Animal_3].uuid
+                LEFT OUTER JOIN db_1.schema_1.[Species] AS [Species_1]
+                    ON [Animal_3].species = [Species_1].uuid
+            WHERE
+                [Species_1].uuid IS NOT NULL OR [Animal_3].uuid IS NULL
+
```
obi1kenobi: nit: delete this blank line

PR #531, opened 2019-08-28
Implement recurse in SQL

bojanserafimov:
Implement `@recurse` in `SQL` using recursive CTEs.

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -113,7 +113,7 @@ class CompilationState(object):
         edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
         self._relocate(self._current_location.navigate_to_subpath(vertex_field))
 
-        # Construct literal columns to be used in the query
+        # Sanitize literal columns to be used in the query
         depth_string = str(depth)
         if not depth_string.isnumeric():
             raise AssertionError(u'Depth must be a number. Received {}'.format(depth_string))
@@ -121,26 +121,25 @@ class CompilationState(object):
         literal_0 = sqlalchemy.literal_column('0')
         literal_1 = sqlalchemy.literal_column('1')
 
-        # TODO(bojanserafimov): double-check interactions with filters
-        # TODO(bojanserafimov): take optional scope into account
-
-        on_clause = previous_alias.c[edge.from_column] == self._current_alias.c[edge.to_column]
-        all_columns = self._current_alias.c
-
         recursive_query = sqlalchemy.select(
-            all_columns + [literal_0.label(CTE_DEPTH_NAME)]
+            self._current_alias.c + [literal_0.label(CTE_DEPTH_NAME)]
         ).cte(recursive=True)
 
         self._current_alias = recursive_query.union_all(sqlalchemy.select(
-            all_columns + [(recursive_query.c[CTE_DEPTH_NAME] + literal_1).label(CTE_DEPTH_NAME)]
+            [col for col in recursive_query.c if col.name != CTE_DEPTH_NAME]
+            + [(recursive_query.c[CTE_DEPTH_NAME] + literal_1).label(CTE_DEPTH_NAME)]
         ).select_from(
-            previous_alias.join(self._current_alias, onclause=on_clause)
+            previous_alias.alias().join(
+                self._current_alias,
+                onclause=previous_alias.c[edge.from_column] == self._current_alias.c[edge.to_column])
         ).where(
             recursive_query.c[CTE_DEPTH_NAME] < literal_depth)
         )
 
         # Join to where we came from
-        self._from_clause = self._from_clause.join(self._current_alias, onclause=on_clause)
+        self._from_clause = self._from_clause.join(
+            self._current_alias,
+            onclause=previous_alias.c[edge.from_column] == self._current_alias.c[edge.to_column])
 
     def start_global_operations(self):
         """Execute a GlobalOperationsStart block."""
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -104,6 +104,23 @@ class CompilationState(object):
             onclause=(previous_alias.c[edge.from_column] == self._current_alias.c[edge.to_column]),
             isouter=optional)
 
+    def recurse(self, vertex_field):
+        previous_alias = self._current_alias
+        edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
+        self._relocate(self._current_location.navigate_to_subpath(vertex_field))
+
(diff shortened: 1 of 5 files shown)


bojanserafimov:
Just noticed a lot of things are wrong. Don't review yet

bojanserafimov:
> If I understood the generated queries correctly (which I very well might not have), I think they might have not been starting at depth 0: https://github.com/kensho-technologies/graphql-compiler#recurse

Yeah, a lot of things were wrong. Try now.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -104,6 +149,61 @@ def traverse(self, vertex_field, optional):
...
 
+    def recurse(self, vertex_field, depth):
+        """Execute a Recurse Block."""
+        previous_alias = self._current_alias
+        edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
+        primary_key = self._sql_schema_info.primary_keys[self._current_classname]
+        self._relocate(self._current_location.navigate_to_subpath(vertex_field))
+
+        if self._is_in_optional_scope():
+            raise NotImplementedError(u'The SQL backend does not implement mandatory '
```
obi1kenobi: This error message is maybe not the best. Recurse is a mandatory traversal, but since it starts at depth 0 (i.e. the original node) it's always satisfiable by at least one path.

Let's reword this for clarity. I think the problem here is that `@recurse` inside `@optional` isn't supported, so let's say that. Keep in mind that when the user writes a big, complicated query that hits this error, this text is the only thing they'll see.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -104,6 +149,61 @@ def traverse(self, vertex_field, optional):
...
+                                      u'traversals inside an @optional scope.')
+
+        # Sanitize literal columns to be used in the query
+        depth_string = str(depth)
+        if any(not char.isdigit() for char in depth_string):
```
obi1kenobi: `depth` here is supposed to be a positive integer, right? Any reason to sanitize its `str` representation instead of checking the `int` value directly?

graphql_compiler/schema/schema_info.py
```diff
@@ -84,6 +84,11 @@
     # a sqlalchemy table. Column types that do not exist for this dialect are not allowed.
     'tables',
 
+    # dict mapping every graphql object type or interface type name in the schema to
+    # a string name of its primary key column. The column cannot contain nulls and must
+    # be unique.
+    'primary_keys',
```
obi1kenobi: How much work would it be to design this to allow composite (multi-column) primary keys from day 1? Maybe if we have `Dict[str, Collection[str]]` as the type?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -104,6 +149,61 @@ def traverse(self, vertex_field, optional):
             onclause=(previous_alias.c[edge.from_column] == self._current_alias.c[edge.to_column]),
             isouter=optional)
 
+    def recurse(self, vertex_field, depth):
+        """Execute a Recurse Block."""
+        previous_alias = self._current_alias
+        edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
+        primary_key = self._sql_schema_info.primary_keys[self._current_classname]
```
obi1kenobi: If we go with the option for supporting (at configuration level only) the composite primary keys situation, I'd just put a `NotImplementedError` for the case where the primary key spans more than one column.

graphql_compiler/schema/schema_info.py
```diff
@@ -118,6 +123,9 @@ def make_sqlalchemy_schema_info(schema, type_equivalence_hints, dialect, tables,
         dialect: sqlalchemy.engine.interfaces.Dialect
         tables: dict mapping every graphql object type or interface type name in the schema to
                 a sqlalchemy table
+        primary_keys: dict mapping every graphql object type or interface type name in the schema
```
obi1kenobi: Does the SQLAlchemy table object not have a notion of what its primary key is? I find that a bit surprising.

If that is indeed the case, and you decide to go with this approach, then let's validate that the columns specified here actually exist in the tables given in `tables`. If you also choose to allow composite primary keys, then the list of columns in such a composite key must be non-empty.

graphql_compiler/compiler/emit_sql.py
bojanserafimov: Just being paranoid. I can subclass `int` and override the `to_string` method to return some sql injection string.

graphql_compiler/schema/schema_info.py
bojanserafimov: I wouldn't do it in this PR, since it's already day 2. If we support multi-column primary keys, then the `join_descriptors` should also assume multi-column foreign keys.

graphql_compiler/schema/schema_info.py
bojanserafimov: It does. I went for an explicit external definition because we already ignore the `Table` object's information on foreign keys. If you think there's no reason to have implicit primary keys without formally declaring them as primary keys in your database, I'll switch to using the `Table`-defined primary keys.

graphql_compiler/compiler/emit_sql.py
obi1kenobi: I think it might be unnecessary. The `depth` parameter here is not user-provided, and would come from the AST of the `@recurse` directive, where that value is a raw Python `int` object. If you can define subclasses of `int` and interpose on function calls between compiler internals, then you have already gained so many capabilities that I doubt this will be even so much as a speed bump.

I'd just do a type check for `int`, assert it's a positive number, and call it a day :)

graphql_compiler/schema/schema_info.py
bojanserafimov: Reverted to using the primary key info from the sqlalchemy `Table`s, and documented that we expect tables to have primary keys.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -94,32 +169,64 @@ def traverse(self, vertex_field, optional):
...
+    def recurse(self, vertex_field, depth):
+        """Execute a Recurse Block."""
+        previous_alias = self._current_alias
+        edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
+        if not self._current_alias.primary_key:
+            raise AssertionError(u'The table for vertex {} has no primary key specified. This '
+                                 u'information is required to emit a @recurse directive.'
+                                 .format(self._current_classname))
+        if len(self._current_alias.primary_key) > 1:
+            raise AssertionError(u'The table for vertex {} has a composite primary key {}. '
```
obi1kenobi: `NotImplementedError` maybe?

PR #537, opened 2019-09-04
Add fold in example

bojanserafimov:
The example is mentioning a `@fold` directive that's not in the query.

While we're talking about the `@fold` documentation, there's something else we should clarify: If I filter on `name` inside a fold scope, and there are no matches, do i get an empty list, or does the row get removed? The documentation leads me to believe the row is removed, but I'm not sure that makes sense.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -551,9 +551,9 @@ whose names contain the substring :code:`substr`:
 .. code::
 
     {
-        Animal @fold {
+        Animal {
             name @output(out_name: "animal_name")
-            out_Animal_ParentOf {
+            out_Animal_ParentOf @fold {
                 _x_count @filter(op_name: ">=", value: ["$count"])
                 name @filter(op_name: "has_substring", value: ["$substr"])
             }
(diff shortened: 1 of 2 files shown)


obi1kenobi:
> While we're talking about the @fold documentation, there's something else we should clarify: If I filter on name inside a fold scope, and there are no matches, do i get an empty list, or does the row get removed? The documentation leads me to believe the row is removed, but I'm not sure that makes sense.

Can you open a `documentation` issue for this?

README.rst
```diff
@@ -551,7 +551,7 @@ whose names contain the substring :code:`substr`:
 .. code::
 
     {
-        Animal {
+        Animal @fold {
```
obi1kenobi: The `@fold` should actually be one level deeper, on `out_Animal_ParentOf`. Putting it here should fail to compile since it violates:

```
-  All @output fields within a @fold traversal must be present
   at the innermost scope. It is invalid to expand vertex fields within
   a @fold after encountering an @output directive.
```

PR #538, opened 2019-09-05
Make parameters optional in get_graphql_schema_from_schema_graph

pmantica1:
In this PR, I make `hidden_classes` and `class_to_field_type_override` optional in `get_graphql_schema_from_schema_graph`. This PR is among the first refactors that I need to land to land the `SQLAlchemySchemaInfo` generation. 

I looked at everywhere this function is called in the repo and changed those function calls to use keyword arguments.

### DIFF
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -339,15 +339,7 @@ def estimate_query_result_cardinality(
         float, expected query result cardinality. Equal to the number of root vertices multiplied by
         the expected number of result sets per full expansion of a root vertex.
     """
-    if class_to_field_type_overrides is None:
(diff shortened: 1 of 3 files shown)


graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -345,7 +345,8 @@ def estimate_query_result_cardinality(
         hidden_classes = set()
 
     graphql_schema, type_equivalence_hints = get_graphql_schema_from_schema_graph(
-        schema_graph, class_to_field_type_overrides, hidden_classes
+        schema_graph, class_to_field_type_overrides=class_to_field_type_overrides,
+        hidden_classes=hidden_classes
```
pmantica1: Even though this function call would still work as is, it is best practice to call keyword arguments using the keywords.

PR #539, opened 2019-09-05
Change table parameter name to vertex_name_to_table

pmantica1:
This PR changes the `tables` field in SQLAlchemySchemaInfo to `vertex_name_to_table`. The `tables` field is misleading since it seems to imply that all SQLAlchemyTable objects are going to mapped to vertices. However,  junction tables may just be represented as edges. This PR is another refactor that needs to happen before landing SQLAlchemySchemaInfo generation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1877,8 +1877,7 @@ for configuring and running SQLAlchemy in a production system.
     engine = create_engine('<connection string>')
 
     # Wrap the schema information into a SQLAlchemySchemaInfo object
-    sql_schema_info = make_sqlalchemy_schema_info(schema, {}, engine.dialect, vertex_name_to_table,
-     {})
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -1839,51 +1839,51 @@ for configuring and running SQLAlchemy in a production system.
     from graphql_compiler.compiler.ir_lowering_sql.metadata import SqlMetadata
     from graphql_compiler.schema.schema_info import make_sqlalchemy_schema_info
     from graphql_compiler import graphql_to_sql
-    
+
```
pmantica1: My editor deletes trailing spaces automatically. I don't think that we should not have trailing spaces in the README.rst, but I'd rather not make a pull request just to delete the trailing spaces so please allow this change in this PR.

bojanserafimov: try using a better editor :)

PR #542, opened 2019-09-05
Add SQLAlchemySchemaInfo generation

pmantica1:
In this PR I add SQLAlchemySchemaInfo generation. For a discussion of the API specification please see https://github.com/kensho-technologies/graphql-compiler/pull/517. Junction tables are not specified here, but the plan of how to add them is specified in the referenced PR.

I will add documentation to the README.rst in a latter PR.


graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2019-present Kensho Technologies, LLC.
```
bojanserafimov: I've seen and reviewed this file in `add-sqlalchemy-schema-generation` feature branch. Assuming no changes.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2019-present Kensho Technologies, LLC.
```
bojanserafimov: I've seen and reviewed this file in add-sqlalchemy-schema-generation feature branch. Assuming no changes.

graphql_compiler/schema_generation/sqlalchemy/exceptions.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+from ..exceptions import SchemaError
+
+
+class InvalidSQLEdgeReferenceError(SchemaError):
+    """Raised when a SQL edge references a non-existent vertex or column."""
```
bojanserafimov: 1. Since you decided to make a separate file for sqlalchemy exceptions, why not do the same for orientdb? `..exceptions` contains some orientdb-specific ones.

graphql_compiler/tests/schema_generation_tests/test_sqlalchemy_schema_generation.py
```diff
@@ -0,0 +1,159 @@
...
+    def test_reference_to_non_existent_source_vertex(self):
+        direct_edges = {
+            'invalid_source_vertex': DirectEdgeDescriptor(
+                'InvalidVertexName',
+                'source_column',
+                'ArbitraryObjectName',
+                'destination_column'
+            )
+        }
+        with self.assertRaises(InvalidSQLEdgeReferenceError):
```
bojanserafimov: Shouldn't this be an `InvalidClassError`? That would remove the need for `InvalidSQLEdgeReferenceError`, which is very obscure.

graphql_compiler/schema_generation/sqlalchemy/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        vertex_name_to_table: dict, str -> SQLAlchemy Table. This dictionary is used to generate the
+                              GraphQL objects in the schema in the SQLAlchemySchemaInfo. Each
+                              SQLAlchemyTable will be represented as a GraphQL object. The GraphQL
+                              object names are the dictionary keys. The fields of the GraphQL
+                              objects will be inferred from the columns of the underlying tables.
+                              The fields will have the same name as the underlying columns and
+                              columns with unsupported types, (SQL types with no matching GraphQL
+                              type), will be ignored.
+        direct_edges: dict, str-> DirectEdgeDescriptor. Direct edges are edges that do not
+                      use a junction table, (see junction_table_edges). The traversal of a direct
```
bojanserafimov: since you removed `junction_table_edges`, you can't talk about them. You can add them back as an argument, and immediately raise `NotImplementedError` if they are used, or update the docs not to reference them. Doing the former means fewer api changes, the latter means the api is less misleading in the meantime.

graphql_compiler/schema_generation/sqlalchemy/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Args:
+        vertex_name_to_table: dict, str -> SQLAlchemy Table. This dictionary is used to generate the
+                              GraphQL objects in the schema in the SQLAlchemySchemaInfo. Each
+                              SQLAlchemyTable will be represented as a GraphQL object. The GraphQL
+                              object names are the dictionary keys. The fields of the GraphQL
+                              objects will be inferred from the columns of the underlying tables.
+                              The fields will have the same name as the underlying columns and
+                              columns with unsupported types, (SQL types with no matching GraphQL
+                              type), will be ignored.
```
bojanserafimov: This might be hard to construct for users not familiar with `sqlalchemy`. Maybe suggest that the `reflect()` method is one way to generate the data?

bojanserafimov: Can you specify that we expect tables to have primary keys, and their `schema` field to be defined? Not sure if `engine.reflect()` automatically fills in the `schema` field, or just fills in the `metadata.schema`.

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
pmantica1: Correct. No changes.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
pmantica1: There are changes. The input is now different and I've removed default value parsing since we don't seem to be interested in it at the time and I would prefer to have no to maintain it.

graphql_compiler/schema_generation/sqlalchemy/__init__.py
pmantica1: I'll remove the reference.

graphql_compiler/schema_generation/sqlalchemy/__init__.py
pmantica1: Sounds like a good idea.

pmantica1: I'll check and add it.

graphql_compiler/schema_generation/sqlalchemy/exceptions.py
pmantica1: I'll remove this file and change the exception to InvalidClassError. It makes sense.

graphql_compiler/schema_generation/sqlalchemy/edge_descriptors.py
```diff
@@ -0,0 +1,57 @@
...
+            DirectJoinDescriptor(from_column, to_column)
+        )
+        join_descriptors[direct_edge_descriptor.to_vertex][in_edge_name] = (
+            DirectJoinDescriptor(to_column, from_column)
+        )
+    return join_descriptors
+
+
+def validate_edge_descriptors(vertex_name_to_table, direct_edges):
+    """Validate that the edge descriptors do not reference non-existent vertices or columns."""
```
bojanserafimov: Additional validation can be done. For example see that the column types are comparable, or maybe that there's a unique index on those columns. I'd leave a TODO just not to mislead that the validation has everything you'd expect.

graphql_compiler/schema_generation/sqlalchemy/schema_graph_builder.py
```diff
@@ -0,0 +1,66 @@
...
+                              unsupported types, (SQL types with no matching GraphQL type), will be
+                              ignored.
+        direct_edges: dict, str -> DirectEdgeDescriptor. This dictionary will be used to generate
+                      EdgeType objects. The name of the EdgeType objects will be dictionary keys and
+                      the connections will be deduced from the DirectEdgeDescriptor objects.
+
+    Returns:
+        SchemaGraph reflecting the specified metadata.
+    """
+    # TODO(pmantica1): Use merge_non_overlapping_dicts when macro_system is merged into master.
```
bojanserafimov: The time has come. Dun dun dun ..

PR #544, opened 2019-09-06
MSSQL Statistics collection

bojanserafimov:


graphql_compiler/cost_estimation/statistics_collection.py
```diff
@@ -0,0 +1,104 @@
...
+        LocalStatistics object that can be used for query cost estimation.
+    """
+    return LocalStatistics(
+        dict(_stream_class_counts(sql_schema_info, engine)),
+        dict(_stream_vertex_edge_vertex_counts(sql_schema_info, engine)),
+        dict(_stream_distinct_field_values_counts(sql_schema_info, engine)),
+    )
+
+
+def write_mssql_statistics_to_csv_file(filename, sql_schema_info, engine):
```
obi1kenobi: Any specific reason to make this CSV as opposed to JSON or something else that's a bit more easily machine-readable?

lodrion: Maybe split this into two functions: one that generates the stats and another that writes them... Or leave writing as an exercise to the reader

bojanserafimov:
This approach is too slow in practice. I will open-source the better approach once it's working.

PR #550, opened 2019-09-10
Fix DateTime value parsing

pmantica1:
This pull request is to fix the CI issues for 
https://github.com/kensho-technologies/graphql-compiler/pull/548

If you look at the logs, you'll notice that it is failing for a reason completely unrelated to the changes in the PR. It seems that during CI a later version of arrow gets installed in the pipenv enviroment and with it our test fails. My best guess is that the later version of arrow fixed a bug that we relied on for datetime parsing. Anyways, I found the correct way to correct timezone tokens in the Supported Tokens section of  https://arrow.readthedocs.io/en/latest/ and am submitting the fix in this PR.

### DIFF
--- a/graphql_compiler/schema/__init__.py
+++ b/graphql_compiler/schema/__init__.py
@@ -242,7 +242,6 @@ def _parse_datetime_value(value):
         # Strip the "Z" and add an explicit time zone instead.
         value = value[:-1] + '+00:00'
 
-
     return arrow.get(value, 'YYYY-MM-DDTHH:mm:ssZZ').datetime
 
 
--- a/graphql_compiler/schema/__init__.py
+++ b/graphql_compiler/schema/__init__.py
@@ -242,8 +242,7 @@ def _parse_datetime_value(value):
         # Strip the "Z" and add an explicit time zone instead.
         value = value[:-1] + '+00:00'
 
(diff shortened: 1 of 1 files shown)


graphql_compiler/schema/__init__.py
```diff
@@ -242,8 +242,8 @@ def _parse_datetime_value(value):
         # Strip the "Z" and add an explicit time zone instead.
         value = value[:-1] + '+00:00'
 
-    return arrow.get(value, 'YYYY-MM-DDTHH:mm:ssZ').datetime
 
+    return arrow.get(value, 'YYYY-MM-DDTHH:mm:ssZZ').datetime
```
obi1kenobi: If this is conditional upon a specific arrow version to keep working, then let's set that version as a minimum in our `setup.py`.

pmantica1: This just seems to have been the right way always. 

I ran the datetime parsing tests using this format with arrow versions 0.7.0, (the earliest arrow version specified allowed by setup.py) and 0.15.0, (the current latest version). 

On a separate note, the package requirements in the Pipfile file seem to be out of sync with the setup.py file. Do you know why that may be?

pmantica1:
To be clear what I meant to say when I dismissed ob1kenobi's review is that I addressed his concern in the comment.

PR #551, opened 2019-09-10
Draft implementation for cross-db query plan execution.

obi1kenobi:
Not ready to merge. Please look at it for big-picture design issues and not small nits.

There's a bunch of debugging code currently in there as well, since it has no unit tests and we have no cross-db integration tests, and I've been testing it by hand instead.

### DIFF
--- a/graphql_compiler/schema_transformation/query_plan.py
+++ b/graphql_compiler/schema_transformation/query_plan.py
@@ -11,10 +11,12 @@ from graphql.language.ast import (
 from ..ast_manipulation import get_only_query_definition
 from ..exceptions import GraphQLValidationError
 from ..schema import FilterDirective, OutputDirective
+from .utils import get_query_runtime_arguments
 
 
 SubQueryPlan = namedtuple(
     'SubQueryPlan', (
+        'plan_id',  # int, unique identifier for this sub-plan
         'query_ast',  # Document, representing a piece of the overall query with directives added
         'schema_id',  # str, identifying the schema that this query piece targets
         'parent_query_plan',  # SubQueryPlan, the query that the current query depends on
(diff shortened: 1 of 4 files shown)


graphql_compiler/schema_transformation/query_plan.py
```diff
@@ -276,3 +296,164 @@ def _get_plan_and_depth_in_dfs_order_helper(query_plan, depth):
...
+
+    stitching_output_names_by_parent_plan_id = dict()
+    for join_descriptor in query_plan_descriptor.output_join_descriptors:
+        parent_plan_id = join_descriptor.child_query_plan.parent_query_plan.plan_id
+        stitching_output_names_by_parent_plan_id.setdefault(parent_plan_id, []).append(
+            join_descriptor.output_names)
+
+    full_query_args = dict(query_args)
+
+    plan_and_depth = _get_plan_and_depth_in_dfs_order(query_plan_descriptor.root_sub_query_plan)
```
bojanserafimov: Does the code below depend on the order being a dfs order? In some very trivial cases we will have to start execution from the leafs. Just making sure we don't lock ourselves from that execution plan with the code structure.

obi1kenobi: One thing we should document better than we currently have is the distinction between the output of `split_query()` and the output of `make_query_plan()` -- I think that would clarify the situation here. `split_query()` tells us where we cross schemas, but does not specify an execution order. `make_query_plan()` specifies an order between the subqueries, and is free to specify any valid order -- and may choose an optimized order if it has access to statistics etc.

Given that, once a query plan is made, it's always executed from the root onward. We may in the future allow different styles of executors (e.g. DFS, BFS, async + parallel across children, etc.), but this executor function is just a simple sync DFS.

PR #560, opened 2019-09-19
Remove costly schema generation from cost estimation

bojanserafimov:
Currently, cost estimation takes among other things a `schema_graph` and `statistics`, and the first thing it does is generate a graphql schema using `graphql_schema, type_equivalence_hints = get_graphql_schema_from_schema_graph(schema_graph)`.

It does not make sense to generate schema inside cost estimation because:
1. Cost estimation should not rely on schema generation
2. It can be costly to run

Cost estimation needs the `graphql_schema` in order to compile the query to IR, and use its `query_metadata_table`. So there's two ways to solve this problem: Either provide a `graphql_schema` as an argument, or provide the IR of the query as an argument. This PR takes the first approach because it is more flexible. It allows us to remove the schema graph as a dependency if we decide to do so (maybe to remove the `orientdb_client` as a test dependency), and access the query in all stages: text, AST and IR.

In this PR, I add the `schema_graph` and `type_equivalence_hints` as additional arguments to all cost estimation and auto pagination functions. To keep the number of arguments small, I also group all schema information arguments (everything that does not depend on the particular query) into a `QueryPlanningSchemaInfo`.

### DIFF
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -84,8 +84,8 @@ def _get_base_class_names_of_parent_and_child_from_edge(schema_graph, current_lo
     return parent_base_class_name, child_base_class_name
 
 
-def _query_statistics_for_vertex_edge_vertex_count(
-    statistics, query_metadata, parent_location, child_location):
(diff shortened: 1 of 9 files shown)


obi1kenobi:
Seems reasonable overall, no concerns here.

PR #561, opened 2019-09-20
Add test cases that demonstrate incomplete macro edge validation.

obi1kenobi:
The `@macro_edge_target` must not be at or within an `@optional` scope, same as it must not be at or within a `@fold` scope. This is to avoid non-obvious behavior where the macro edge use is not optional, but due to the optional-ness in the definition, ends up having optional semantics applied anyway.

We added the `@fold` bit to the spec and tested for it, but we forgot to add and test for the `@optional`.

Two of the three test cases added here fail, since they pass validation but shouldn't.


obi1kenobi:
Added the fixes, this is ready for review!

PR #562, opened 2019-09-20
Remove ordered field from index definition

pmantica1:
We do not have a use for the ordered field and so I believe it to best if we stop maintaining it.


bojanserafimov:
It's relevant for query planning (cost estimation, paging, cross-db plans, etc.). Just add an `UNKNOWN_INDEX_TYPE` or something similar.

pmantica1:
> It's relevant for query planning (cost estimation, paging, cross-db plans, etc.). Just add an `UNKNOWN_INDEX_TYPE` or something similar.

I still do not think that maintaining this field is worth it. I think that the odds that we end up using it extremely low. Also, SQLAlchemy does not support the concept of ordered indexes.

obi1kenobi:
I agree with @bojanserafimov, it's essentially impossible to do a good job of estimating query costs without knowing whether index lookups are O(log N) or O(1).

Even if SQLAlchemy doesn't give us the magic reflection secret sauce we'd like, it would be worth implementing it from scratch for ourselves and perhaps even contributing it upstream to SQLAlchemy.

PR #563, opened 2019-09-20
Add pagination keys to query planning info

bojanserafimov:
Adding the option to paginate on a field that's not `uuid`. This information could come from the schema graph if it was included, but including it is a whole ordeal that meddles with schema generation, and the test schema itself.

### DIFF
--- a/graphql_compiler/schema/schema_info.py
+++ b/graphql_compiler/schema/schema_info.py
@@ -197,4 +197,9 @@ QueryPlanningSchemaInfo = namedtuple('QueryPlanningSchemaInfo', (
 
     # A Statistics object giving statistical information about all objects in the schema.
     'statistics',
+
+    # Dict mapping all vertex names in the schema to the Int or ID type property name
+    # to be used for pagination on that vertex. This property should be non-null and
+    # unique for all rows. An easy choice in most situations is the primary key.
+    'pagination_keys',
 ))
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema/schema_info.py
```diff
@@ -197,4 +197,9 @@ def make_sqlalchemy_schema_info(schema, type_equivalence_hints, dialect, vertex_
 
     # A Statistics object giving statistical information about all objects in the schema.
     'statistics',
+
+    # Dict mapping all vertex names in the schema to the Int or ID type property name
```
pmantica1: I think that you want to make a stronger pre-requirement no? Does it not need to be a 16 character uuid if it is a GraphQLID? 

Also, remember that the uuid field is a GraphQLString in the test SchemaGraph, but gets displayed to as a GraphQLID through the `class_to_field_type_overrides` parameter so you might want to rethink your wording here.

pmantica1: Also, I am not sure that it's reasonable to expect that all vertices have a pageable key.

bojanserafimov: > I think that you want to make a stronger pre-requirement no? Does it not need to be a 16 character uuid if it is a GraphQLID?
> 
> Also, remember that the uuid field is a GraphQLString in the test SchemaGraph, but gets displayed to as a GraphQLID through the `class_to_field_type_overrides` parameter so you might want to rethink your wording here.

Good point. I forgot type overrides exist. I'll use the `graphql_schema` as the source of truth for field types.

bojanserafimov: > Also, I am not sure that it's reasonable to expect that all vertices have a pageable key.

Also good point, I'll make it optional. We do require that all vertices have primary keys, but it's good to be able to specify that a certain vertex is not a good vertex to paginate on, by not providing a pagination key.

PR #564, opened 2019-09-20
Add histogram stats

bojanserafimov:
This information is necessary for auto-pagination on non-uuid fields, and collaterally could make cost estimation more accurate.

When naming these functions, i considered other names that didn't work out as well:
- histogram: This is not a statistical term, but a visualization method.
- percentile: More widely known than quantile, but implies we use percents.

### DIFF
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -83,21 +83,6 @@ class Statistics(object):
         """
         return None
 
-    def get_field_quantile(self, vertex_name, field_name, fraction):
-        """Return the field value that's larger than a given fraction of all values.
-
-        Args:
-            vertex_name: str, name of a vertex defined in the GraphQL schema.
-            field_name: str, name of a vertex field.
-            fraction: float from 0 to 1, the fraction of values smaller than the result
-
-        Returns:
-            None or a value of the given field with the appropriate type that is larger than
-            a given fraction of all current values for that field. None is returned when the
-            result is unknown.
-        """
-        return None
-
     def get_field_quantiles(self, vertex_name, field_name):
         """Return a list dividing the field values in equally-sized groups.
 
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -121,11 +121,9 @@ class LocalStatistics(Statistics):
                              and property field name to a list of N quantiles, a sorted list of
                              values separating the values of the field into N-1 groups of almost
                              equal size. The first element of the list is the smallest known value,
-                             and the last element is the largest known value. Generally, the i-th
-                             element is a value greater than around i/N of all present values.
-                             The number N can be different for each entry. When a table has very
-                             few values, or no data is available, it is better not to provide
-                             information than to provide interpolated information.
+                             and the last element is the largest known value. The i-th
+                             element is a value greater than or equal to aroune i/N of all present
+                             values. The number N can be different for each entry.
         """
         if vertex_edge_vertex_counts is None:
             vertex_edge_vertex_counts = dict()
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -122,7 +122,7 @@ class LocalStatistics(Statistics):
                              values separating the values of the field into N-1 groups of almost
                              equal size. The first element of the list is the smallest known value,
                              and the last element is the largest known value. The i-th
-                             element is a value greater than or equal to aroune i/N of all present
+                             element is a value greater than or equal to i/N of all present
                              values. The number N can be different for each entry.
         """
         if vertex_edge_vertex_counts is None:
--- a/graphql_compiler/cost_estimation/statistics.py
+++ b/graphql_compiler/cost_estimation/statistics.py
@@ -92,7 +92,7 @@ class Statistics(object):
 
         Returns:
             None or a sorted list of N quantiles dividing the values of the field
-            into N-1 groups of equal size. The first element of the list is the smallest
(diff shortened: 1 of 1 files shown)


graphql_compiler/cost_estimation/statistics.py
```diff
@@ -82,6 +82,35 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
...
+    def get_field_quantile(self, vertex_name, field_name, fraction):
+        """Return the field value that's larger than a given fraction of all values.
+
+        Args:
+            vertex_name: str, name of a vertex defined in the GraphQL schema.
+            field_name: str, name of a vertex field.
+            fraction: float from 0 to 1, the fraction of values smaller than the result
+
+        Returns:
+            None or a value of the given field with the appropriate type that is larger than
```
pmantica1: Can you make this documentation consistent with the rest of the statistics class?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -82,6 +82,35 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
...
+            result is unknown.
+        """
+        return None
+
+    def get_quantile_rank_of_value(self, vertex_name, field_name, field_value):
+        """Return a [0-1] float representing the fraction of values that are smaller.
+
+        Args:
+            vertex_name: str, name of a vertex defined in the GraphQL schema.
+            field_name: str, name of a vertex field.
```
pmantica1: same as above.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -82,6 +82,35 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
...
+            field_value: An appropriate value for the given field
+
+        Returns:
+            None or the fraction of values for the given field that are smaller than the
```
pmantica1: same as above.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -82,6 +82,35 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
         """
         return None
 
+    def get_field_quantile(self, vertex_name, field_name, fraction):
+        """Return the field value that's larger than a given fraction of all values.
```
pmantica1: nit: Return the smallest field value that's larger than a given fraction of all values.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -82,6 +82,35 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
...
+
+        Args:
+            vertex_name: str, name of a vertex defined in the GraphQL schema.
+            field_name: str, name of a vertex field.
+            fraction: float from 0 to 1, the fraction of values smaller than the result
```
pmantica1: nit: remove, "the fraction of values smaller than the result". It does not add much information and the result may be None so it is technically not true.

pmantica1:
Looks good mostly nits.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -102,15 +117,24 @@ class name) to count of edge instances of given class
             distinct_field_values_counts: optional dict, (str, str) -> int, mapping vertex class
                                           name and property field name to the count of distinct
                                           values of that vertex class's property field.
+            field_quantiles: optional dict, (str, str) -> list, mapping vertex class name
+                             and property field name to a list of N quantiles, a sorted list of
+                             values seperating the values of the field into N-1 groups of equal
+                             size. The first element of the list is the smallest known value,
+                             and the last element is the largest known value. The number N can be
+                             different for each entry.
```
pmantica1: Do the number of values have to be divisible by N-1 in order for the groups to be of equal size? What interpolation method should I use to compute quantiles?

(See the parameters section in the link below so that we have a common language to talk about interpolations).
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.quantile.html

bojanserafimov: 1. Exact equality is not important, the statistics are stale anyway.
2. Don't interpolate.

pmantica1: Can you define the ith-quantile here then as the smallest field value that is larger of the fraction i/N of other field values? Just to make it clear that no interpolation is used, (especially since libraries often compute some sort of interpolation to evaluate quantiles).

Also, instead of writing:
```into N-1 groups of equal size```
Write:
```into N-1 groups of (possibly) equal size```

File: graphql_compiler/cost_estimation/statistics.py [hunk 7]
```diff
@@ -82,13 +82,28 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
         """
         return None
 
+    def get_field_quantiles(self, vertex_name, field_name):
+        """Return a list dividing the field values in equally-sized groups.
+
+        Args:
+            vertex_name: str, name of a vertex defined in the GraphQL schema.
+            field_name: str, name of a vertex field. This field has to have a type on which
```
obi1kenobi: nit: `name of a property field on that vertex type` perhaps, to avoid ambiguity relative to the existing `vertex field` definition in the spec?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -82,13 +82,28 @@ def get_distinct_field_values_count(self, vertex_name, field_name):
         """
         return None
 
+    def get_field_quantiles(self, vertex_name, field_name):
+        """Return a list dividing the field values in equally-sized groups.
+
+        Args:
+            vertex_name: str, name of a vertex defined in the GraphQL schema.
```
obi1kenobi: nit: consider `name of a vertex type` for clarity

PR #566, opened 2019-09-23
Allow unary filter directives to omit value argument

### LINKED ISSUE #546: Allow the "value" argument to "@filter" to be optional and missing
Some filter operators, such as `is_null` and `is_not_null`, are unary and do not require passing additional values. Currently, because the `@filter` directive specifies the `value` argument as non-null (`[String!]!`), that means that the syntax for those filters is the following:
```graphql
@filter(op_name: "is_null", value: [])
```

If we changed the definition to make the list nullable and keep only the contents non-nullable (`[String!]`), we could instead have the possibly more ergonomic syntax:
```graphql
@filter(op_name: "is_null")
```

Do you think this is an improvement in ergonomics and worth considering?

jecdohmann:
When processing filter directives, if the operation is in a list of registered unary directives and the argument has been omitted, insert an empty list argument. Now dynamically enforces that non-unary directives have an argument.

Closes #546.


pmantica1:
I think you forgot to modify the `FilterDirective` in  `graphql_compiler/schema/__init__.py`.

graphql_compiler/compiler/filters.py
```diff
@@ -829,6 +829,11 @@ def _get_filter_op_name_and_values(directive):
     if 'op_name' not in args:
         raise AssertionError(u'op_name not found in filter directive arguments!'
                              u'Validation should have caught this: {}'.format(directive))
+    if args['op_name'].value.value in UNARY_FILTERS and 'value' not in args:
+        args['value'] = Argument(Name('value'), ListValue([], loc=args['op_name'].loc))
```
obi1kenobi: This seems rather circuitous, and potentially dangerous. It's not clear whether `args` contains references to objects provided by the input parameter, and so it's unclear whether we are mutating function inputs or not.

Why are we taking this risk instead of refactoring the code to not require this kind of mutation?

graphql_compiler/compiler/filters.py
```diff
@@ -829,6 +829,11 @@ def _get_filter_op_name_and_values(directive):
...
+    if args['op_name'].value.value not in UNARY_FILTERS and 'value' not in args:
+        raise GraphQLValidationError(u'Directive value omitted for'
```
obi1kenobi: nit: we've adopted the convention of leaving the whitespace before the line break, rather than after it

graphql_compiler/compiler/filters.py
```diff
@@ -829,6 +829,11 @@ def _get_filter_op_name_and_values(directive):
...
+                                     u' non-unary filter: {}'.format(directive))
```
obi1kenobi: `directive` here is an AST object, which is not user-friendly when printed in this error message. Let's try to make this message be more explicit (explain what exactly is wrong, and why that is unacceptable) as well as actionable for the non-expert user (explain how to fix it).

graphql_compiler/tests/test_input_data.py
```diff
@@ -1399,6 +1399,27 @@ def starts_with_op_filter():  # noqa: D103
         type_equivalence_hints=None)
 
 
+def starts_with_op_filter_no_value():  # noqa: D103
+    graphql_input = '''{
+        Animal {
+            name @filter(op_name: "starts_with")
```
obi1kenobi: Is this a legal input? It doesn't seem like it to me.

graphql_compiler/tests/test_input_data.py
```diff
@@ -1565,6 +1586,27 @@ def has_edge_degree_op_filter_with_fold():  # noqa: D103
...
 
 
+def is_null_op_filter_optional_value():  # noqa: D103
+    graphql_input = '''{
+        Animal {
+            name @output(out_name: "name")
+            net_worth @filter(op_name: "is_null")
+        }
+    }'''
+
```
obi1kenobi: nit: follow the same whitespace pattern as the other functions unless there's good reason to break from it, this blank line and the one on line 1600 should be removed for consistency

graphql_compiler/tests/test_input_data.py
```diff
@@ -1565,6 +1586,27 @@ def has_edge_degree_op_filter_with_fold():  # noqa: D103
         type_equivalence_hints=None)
 
 
+def is_null_op_filter_optional_value():  # noqa: D103
```
obi1kenobi: nit: use `missing` rather than `optional` here, since `optional` is an overloaded that also means something else (`@optional`)

graphql_compiler/tests/test_input_data.py
jecdohmann: Yes - this test cases is meant to create an error during compilation

graphql_compiler/compiler/filters.py
jecdohmann: Let me know what you think of the new code

graphql_compiler/tests/test_input_data.py
obi1kenobi: Sorry, legal = compiles without errors. All inputs in this file are expected to compile correctly. `test_ir_generation_errors.py` is for inputs that are expected to raise errors.

graphql_compiler/compiler/filters.py
```diff
@@ -830,12 +832,18 @@ def _get_filter_op_name_and_values(directive):
         raise AssertionError(u'op_name not found in filter directive arguments!'
                              u'Validation should have caught this: {}'.format(directive))
 
+    if args['op_name'].value.value not in UNARY_FILTERS and 'value' not in args:
+        raise GraphQLValidationError(u'Filter directive value argument omitted for non-unary '
+                                     u'filter operation: {} \nPlease provide a value argument'
```
obi1kenobi: Avoid newline characters in error messages -- they tend to cause issues, for example in logging systems that might assume that newline characters are the separator between different log lines.

graphql_compiler/compiler/filters.py
```diff
@@ -949,7 +961,9 @@ def process_filter_directive(filter_operation_info, location, context):
 
     context['metadata'].record_filter_info(
         location,
-        FilterInfo(fields=fields, op_name=op_name, args=tuple(operator_params))
+        FilterInfo(fields=fields,
+                   op_name=op_name,
+                   args=tuple(operator_params) if operator_params else None)
```
obi1kenobi: It seems like setting this to `None` is causing a lot of edge case issues in other code. Can we instead make this be an empty tuple if `operator_params` is empty or `None`? I think that we avoid almost all of the `len(parameters) if parameters else 0` and similar logic.

graphql_compiler/compiler/filters.py
```diff
@@ -860,6 +868,10 @@ def _get_filter_op_name_and_values(directive):
     u'is_null',
     u'is_not_null',
 })
+UNARY_FILTERS = frozenset({
```
obi1kenobi: I can't make up my mind on whether this is confusing or not. On one hand, the filter _is_ unary because `is_null` operates on the field. On the other, it's actually a nullary operation because it doesn't take any arguments. @bojanserafimov what do you think?

graphql_compiler/compiler/filters.py
```diff
@@ -830,13 +830,21 @@ def _get_filter_op_name_and_values(directive):
...
-    #                https://github.com/graphql-python/graphql-core/issues/97
-    if not isinstance(args['value'].value, ListValue):
-        raise GraphQLValidationError(u'Filter directive value was not a list: {}'.format(directive))
-
     op_name = args['op_name'].value.value
-    operator_params = [x.value for x in args['value'].value.values]
+    if 'value' not in args and args['op_name'].value.value not in UNARY_FILTERS:
+        raise GraphQLValidationError(u'Filter directive value argument omitted for non-unary '
+                                     u'filter operation: {}. Please provide a value argument'
+                                     u' for all filters not in the following list: {}'
```
obi1kenobi: nit: our convention has been to leave the spaces on the previous line, to make all the text be nicely aligned on the left

graphql_compiler/compiler/filters.py
jecdohmann: That's a good point, would we traditionally consider something like `has_substring` to be binary?

bojanserafimov: I think unary makes sense. It's even pythonic, with `self` being a function argument in class methods.

PR #567, opened 2019-09-23
Fix SQL integration data

bojanserafimov:
Match the SQL integration data with the one we use for all other databases, while at the same time moving towards a backend-agnostic integration data representation.

### DIFF
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -571,11 +571,12 @@ def get_sqlalchemy_schema_info():
         'Species': sqlalchemy.Table(
             'Species',
             sqlalchemy_metadata_2,
-            sqlalchemy.Column('description', sqlalchemy.String(40), nullable=False),
+            sqlalchemy.Column('description', sqlalchemy.String(40), nullable=True),
             sqlalchemy.Column('uuid', uuid_type, primary_key=True),
             sqlalchemy.Column('name', sqlalchemy.String(40), nullable=False),
             sqlalchemy.Column('eats', uuid_type, nullable=True),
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/test_helpers.py
```diff
@@ -571,11 +571,12 @@ def get_sqlalchemy_schema_info():
         'Species': sqlalchemy.Table(
             'Species',
             sqlalchemy_metadata_2,
-            sqlalchemy.Column('description', sqlalchemy.String(40), nullable=False),
+            sqlalchemy.Column('description', sqlalchemy.String(40), nullable=True),
             sqlalchemy.Column('uuid', uuid_type, primary_key=True),
             sqlalchemy.Column('name', sqlalchemy.String(40), nullable=False),
             sqlalchemy.Column('eats', uuid_type, nullable=True),
-            sqlalchemy.Column('limbs', sqlalchemy.Integer, nullable=False),
+            sqlalchemy.Column('limbs', sqlalchemy.Integer, nullable=True),
```
obi1kenobi: Any reason not to populate this field, while we're at it?

bojanserafimov: Yes, we should. But it's not the only problem with the test schema. I'll get to it.

bojanserafimov: If I fix it now, i'll be required to provide a value, therefore changing what the standard test data is, and this PR will explode. I'll have to change the test data in other backends to match this one.

obi1kenobi: Good call. Let's not change it now then.

PR #568, opened 2019-09-24
Implement backend agnostic integration data

bojanserafimov:
Removed backend-specific integration data, and implemented integration setup for all backends from a common data object. Now we can write cross-backend integration and snapshot tests without having to manually set up the data in each backend separately.

The code does not handle all edge cases, but this is testing code, and we can make it more sophisticated over time as needed. For instance, if the backend data uses a datetime object, a `NotImplementedError` will be raised.


graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -15,58 +17,211 @@
 from ..test_helpers import get_sqlalchemy_schema_info
 
 
+def get_integration_data():
```
bojanserafimov: This is not green code, just moved it from a different place.

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -15,58 +17,211 @@
...
+
+def _write_orient_equality(field_name, field_value):
+    """Write a '{} = {}' statement to be used in queries, validating against SQL injection."""
+    _validate_name(field_name)
+    if isinstance(field_value, six.string_types):
+        allowed_characters = string.ascii_uppercase + string.ascii_lowercase + string.digits + '-_ '
+        if any(c not in allowed_characters for c in str(field_value)):
+            raise AssertionError(u'String value {} contains disallowed characters'
+                                 .format(field_value))
+    elif isinstance(field_value, (six.integer_types, float, datetime.date)):
```
obi1kenobi: I think `six.integer_types` is already a tuple, and nested tuples don't work correctly in `isinstance` -- they aren't flattened.

```
$ python
Python 2.7.16 (default, Sep  2 2019, 11:59:44) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import six
>>> six.integer_types
(<type 'int'>, <type 'long'>)
>>> exit()

$ python3
Python 3.7.4 (default, Sep  7 2019, 18:27:02) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import six
>>> six.integer_types
(<class 'int'>,)
```

Use `six.integer_types + (float, datetime.date)` instead.

bojanserafimov: https://docs.python.org/3/library/functions.html#isinstance `isinstance` takes nested tuples.

obi1kenobi: I thought it didn't work in python 2, but I stand corrected -- the py2 version of that page says the same thing. My bad.

PR #569, opened 2019-09-24
Add ReadTheDocs Home Page

pmantica1:
In this PR I specify the home page for readthedocs, outlining the documentation for the entire project. 

To see how it looks like when rendered follow this link:
https://graphql-compiler.readthedocs.io/en/latest/ 

Please, no nits at this point.


docs/source/index.rst
```diff
@@ -1,20 +1,202 @@
...
+
+   OrientDB <databases/orientdb>
+   SQL <databases/sql>
+
+Advanced Features
+~~~~~~~~~~~~~~~~~
+
+To learn more about the advanced features in the GraphQL compiler see:
+
+    - :doc:`Macro System <advanced_features/macro_system>`, to learn how to capture complex
```
obi1kenobi: I don't think this description is accurate.

docs/source/index.rst
```diff
@@ -1,20 +1,202 @@
...
+   :caption: Core Specification
+   :hidden:
+
+   Definitions <core_specification/definitions>
+   Schema Types <core_specification/schema_types>
+   Query Directives <core_specification/query_directives>
+
+Databases
+~~~~~~~~~
+
```
JulianGoetz: Intro this "to learn more" sentence with a description of why I should care about this Databases section.

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+        name: String
+        in_AnimalLivesIn: [Animal]
+    }
+
+In the :code:`schema` above:
+
+-   :code:`Animal` represents a non-abstract vertex. For relational databases, we think of
+    tables as the non-abstract vertices.
+-   :code:`name` is a **property field** and :code:`out_Animal_LivesIn` is a **vertex field**,
+    which represent an outbound edge to a vertex in the graph. For graph databases,
```
obi1kenobi: `represent` -> `represents`

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+    type Continent {
+        name: String
+        in_AnimalLivesIn: [Animal]
+    }
+
+In the :code:`schema` above:
+
+-   :code:`Animal` represents a non-abstract vertex. For relational databases, we think of
+    tables as the non-abstract vertices.
+-   :code:`name` is a **property field** and :code:`out_Animal_LivesIn` is a **vertex field**,
```
obi1kenobi: Remove the trailing comma here. The `which` refers to `vertex field`, and the comma is just confusing in that case. Bonus points if you split this into two bullet points, one explaining `property field` and one explaining `vertex field`.

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+                name @filter(op_name: "=", value: ["$continent"])
+            }
+        }
+    }
+    """
+    parameters = {'continent': 'Africa'}
+
+There are a couple of things to notice about queries:
 
+- All queries start with a vertex and branch to other vertices using **vertex fields**.
```
obi1kenobi: I'd replace `branch` with `expand` here, for clarity.

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+- **Directives**, such as :code:`@output` and :code:`@filter`, specify the semantics of a query.
+  :code:`@output` indicates which properties to emit and :code:`@filter` specifies a filter
```
obi1kenobi: `emit` is not the right word here. Code is emitted, values of property fields are returned. Please rewrite for clarity. Try not to unnecessarily cram unrelated concepts into one sentence -- output and filter can be in separate sentences here.

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+  operation.
 
+With the query at hand, we can compile it and get the corresponding results from OrientDB.
```
obi1kenobi: Not just the query, but also its arguments.

docs/source/index.rst
```diff
@@ -86,10 +86,12 @@ In the :code:`schema` above:
 
 -   :code:`Animal` represents a non-abstract vertex. For relational databases, we think of
     tables as the non-abstract vertices.
--   :code:`name` is a **property field** and :code:`out_Animal_LivesIn` is a **vertex field**,
-    which represent an outbound edge to a vertex in the graph. For graph databases,
-    edges can be reflected from the database schema. However, for relational databases edges have
-    to be manually specified. See :doc:`SQL <databases/sql>` for more information.
+-   :code:`name` is a **property field** which represents a property of the :code:`Animal` vertex.
+    Think of **property fields** as the conceptual equivalent to table columns.
```
obi1kenobi: Columns can be foreign keys and therefore not properties in the compiler's sense of the word. This is not a good analogy.

docs/source/index.rst
```diff
@@ -86,10 +86,12 @@ In the :code:`schema` above:
...
+-   :code:`out_Animal_LivesIn` is a **vertex field** which represents an outbound edge to a vertex
+    in the graph. For graph databases, edges can be reflected from the database schema. However,
```
obi1kenobi: `reflected` is not good word choice here. Consider `generated automatically` or something to that effect. Also consider adding a `currently` in `for relational databases edges have to be manually specified`:
```
for relational databases, edges currently have to be manually specified and cannot be automatically generated from the schema
```

docs/source/index.rst
```diff
@@ -113,12 +115,12 @@ that live in Africa:
 
 There are a couple of things to notice about queries:
 
-- All queries start with a vertex and branch to other vertices using **vertex fields**.
-- **Directives**, such as :code:`@output` and :code:`@filter`, specify the semantics of a query.
-  :code:`@output` indicates which properties to emit and :code:`@filter` specifies a filter
-  operation.
+- All queries start with a vertex and expand to other vertices using **vertex fields**.
+- **Directives** specify the semantics of a query. :code:`@output` indicates which properties to
+  emit. :code:`@filter` specifies a filter operation.
```
obi1kenobi: `emit` is still here. How about `indicates the properties whose values should be returned`?

pmantica1: Sorry I realized earlier that I had missed this comment and was about to submit a commit to address it. That sounds good.

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+   :description: Turn complex GraphQL queries into optimized database queries.
+   :keywords: graphql compiler, database, orientdb, sql
+
+The GraphQL Compiler is a library that simplifies database querying and exploration by exposing one
+common query language for multiple database backends.  The query language is:
+
+.. EDUCATIONAL: The pattern below is what you would call a definition list in restructuredtext.
+   The "terms" get special rendering in the readthedocs html file.
+
+Written in valid GraphQL syntax
```
MichaelaShtilmanMinkin: I'm not sure about the formatting here. I think this is more typically used for function definitions. e.g. https://pika.readthedocs.io/en/stable/modules/connection.html

I would consider making this a bulleted list or a few small headers.

MichaelaShtilmanMinkin: Note how pika also points to the source code in the docs - this would be an awesome feature to have in our docs.

docs/source/advanced_features/schema_stitching.rst
```diff
@@ -0,0 +1,2 @@
+Schema Stiching
```
MichaelaShtilmanMinkin: /s/Stiching/Stitching

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+        get_graphql_schema_from_orientdb_schema_data
+    )
+    from graphql_compiler.schema_generation.orientdb.utils import ORIENTDB_SCHEMA_RECORDS_QUERY
+
+    client = your_function_that_returns_an_orientdb_client()
+    schema_records = client.command(ORIENTDB_SCHEMA_RECORDS_QUERY)
+    schema_data = [record.oRecordData for record in schema_records]
+    schema, type_equivalence_hints = get_graphql_schema_from_orientdb_schema_data(schema_data)
+
+At the core of generated schema info is the GraphQL :code:`schema`. The database might be
```
MichaelaShtilmanMinkin: I'm not sure it makes sense to decorate words as `code` if it's not actually code / an object name.

docs/source/index.rst
MichaelaShtilmanMinkin: This would actually be a good place to reference the code from the query, that way a user can associate `Animal` as being the vertex, etc

docs/source/index.rst
```diff
@@ -1,20 +1,204 @@
...
+    from graphql_compiler import graphql_to_match
+
+    compilation_result = graphql_to_match(
+        schema, graphql_query, parameters, type_equivalence_hints)
+
+    # Execute query assuming a pyorient client. Other clients may have a different interface.
+    print([result.oRecordData for result in client.query(query)])
+    # [{'animal_name': 'Elephant'}, {'animal_name': 'Lion'}, ...]
+
+What's Next?
```
MichaelaShtilmanMinkin: I'm not sure if this is the best title for this section, since it doesn't exist any specific follow up actions by the user nor us. We could probably get rid of it or replace it with something more accurate.

docs/source/index.rst
pmantica1: schema is part of the code above.

docs/source/index.rst
pmantica1: In restructuredtext you cannot skip section headers.  
https://stackoverflow.com/questions/33149169/skipping-heading-levels-in-restructuredtext
So even if this header doesn't say much it helps as an intermediate header.

docs/source/index.rst
obi1kenobi: I agree with @MichaelaShtilmanMinkin, I think the overall paragraph could use some clarification:
- Saying the schema is at the core of the schema is ... kind of circular. Some clarification could help.
- The wording is also somewhat misleading, and the reader could believe that `schema` is a string with the contents starting on line 75, which is wrong in two ways -- it's not a string, and if it were, those would not be the (full) contents of it.
- `type_equivalence_hints` is assigned and named in the code, but then ignored altogether. At least mention where the curious reader can go to learn more about it.

Also, "GraphQL `schema`" just looks awkward from a purely stylistic angle, in my opinion.

docs/source/index.rst
pmantica1: I was going for the way this looks https://docs.readthedocs.io/en/stable/index.html
This is also a definition list though for some reason it is rendered differently. I made a github issues asking why.  

As to sphinx autodoc, I think it is a good idea, but it's not relevant here.

docs/source/index.rst
obi1kenobi: I think you are confusing content problems and technical limitations. If the title were more appropriate, the fact that you cannot skip heading levels would be irrelevant – because the heading would be actually useful.

docs/source/index.rst
pmantica1: I changed the definition list terms for bolded text instead. I will try to fix this in the future to look as to how I signaled it above. For now, I would prefer to move on.

docs/source/index.rst
pmantica1: Any recommendations for a better title? I can't think of a substantially better one.

docs/source/index.rst
pmantica1: Actually, for now, I will just add a TODO. I will fix this once they get back to me.

docs/source/index.rst
MichaelaShtilmanMinkin: We just synced about this. A few things:
- For now it's fine, but we want to make sure the Home Page doesn't grow out of scope / get too lengthy. It would be nice to have to have a nav for Home either way like here: https://mkdocs.readthedocs.io/en/stable/ From our chat it sounded like there might be some bugs, so let's leave some comments about next steps.
- As for a better title, it would be something along the lines of `Features`.

docs/source/index.rst
pmantica1: Ok they responded and pointed out that the sphinx read the docs theme is not ready for Sphinx 2.0. I bumped down the version of Sphinx to 1.8 which is the version they are using.

docs/source/index.rst
pmantica1: I addressed the comments. I didn't want to talk about schema printing here. I didn't want to expose the entire schema either. I left a TODO in the `Schema Types` file to talk about schema printing there.

docs/source/index.rst
pmantica1: I added a TODO above to the restructuring and the nav bar at the top of the page. 

I renamed the section to `Features`. 

I left `Advanced Features` as `Advanced Features` instead of `Advanced` since  `Advanced Features` also appears in the nav bar where the word `Features` is not mentioned.

PR #571, opened 2019-09-25
Recurse integration test

bojanserafimov:
Add more integration test data for MSSQL, use this data to write a test for recurse in SQL. After this, the SQL integration data will be ahead of the integration data of other backends. The issue will be fixed with https://github.com/kensho-technologies/graphql-compiler/pull/568

### DIFF
--- a/graphql_compiler/tests/test_data_tools/data_tool.py
+++ b/graphql_compiler/tests/test_data_tools/data_tool.py
@@ -1,6 +1,5 @@
 # Copyright 2018-present Kensho Technologies, LLC.
 import datetime
-from decimal import Decimal
 from glob import glob
 from os import path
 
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -206,17 +225,29 @@ def generate_sql_integration_data(sql_test_backends):
...
+                                              .format(edge_name, edge_value['to_uuid']))
+                existing_foreign_key_values[join_descriptor.to_column] = edge_value['from_uuid']
+            elif is_to_uuid:
+                existing_foreign_key_values = uuid_to_foreign_key_values.setdefault(
+                    edge_value['from_uuid'], {})
+                if join_descriptor.from_column in existing_foreign_key_values:
+                    raise NotImplementedError(u'The SQL backend does not support many-to-many edges.'
+                                              u'Found multiple edges of class {} from vertex {}'
+                                              .format(edge_name, edge_value['from_uuid']))
+                existing_foreign_key_values[join_descriptor.from_column] = edge_value['to_uuid']
```
bojanserafimov: Without this change, we get `NotImplementedError` when using the `Animal_ParentOf` edge, since it's `to_column` is not `uuid`.

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -206,17 +224,31 @@ def generate_sql_integration_data(sql_test_backends):
...
+
+            is_from_uuid = join_descriptor.from_column == 'uuid'
+            is_to_uuid = join_descriptor.to_column == 'uuid'
+            if is_from_uuid == is_to_uuid:
+                raise NotImplementedError(u'Exactly one of the join columns was expected to'
+                                          u'be uuid. found {}'.format(join_descriptor))
+
+            if is_from_uuid:
+                existing_foreign_key_values = uuid_to_foreign_key_values.setdefault(
+                    edge_value['to_uuid'], {})
```
pmantica1: Please do not do this. Do not select a subdict of a nested dict that is meant to be mutated later. It makes it much more difficult to see what is going on. Just write `uuid_to_foreign_key_values['to_uuid'][join_descriptor.to_column]` later.

PR #572, opened 2019-09-25
Find orientdb recurse inconsistency

bojanserafimov:
This test will fail, pointing out that our Match and MSSQL backends have different recurse semantics.

The semantic question is, if a node can be reached in multiple different ways, should it be included in the result multiple times? Match says no, MSSQL says yes.

The docs don't precisely specify the desired semantics, so please comment if you have opinions.

### DIFF
--- a/graphql_compiler/tests/integration_tests/test_backends_integration.py
+++ b/graphql_compiler/tests/integration_tests/test_backends_integration.py
@@ -363,7 +363,7 @@ class IntegrationTests(TestCase):
         # TODO(bojanserafimov): Only tested in MSSQL. To test in other backends, they should read in
         #                       the standard integration test schema used by the SQL backend.
         for graphql_query, expected_results in queries:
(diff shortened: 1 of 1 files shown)


bojanserafimov:
Actually, the definition of result set as `an assignment of vertices in the graph to scopes (locations) in the query` implies results should be deduplicated. Do we want to explicitly mention this in the `@recurse` constraints and rules, or do we want to NOT deduplicate, and redefine result sets?

bojanserafimov:
Update: Neo4j does something entirely different with `@recurse`. It does not completely deduplicate like OrientDB, and it doesn't find all paths of with `length < recurse_depth` like our SQL backend. It's not explicitly stated in the docs, but after experimenting with different inputs, it looks like Neo4j will find all different paths that use each edge at most once. Two paths are different if the list (not set) of edges traversed is different (so, following two different loops at a given vertex in a different order makes two different paths). Worth noting that Neo4j also lets you distinguish all the different matches by outputting the paths used to get to them.

PR #573, opened 2019-09-26
WIP SQL fold

chewselene:
First pass at implementing basic folds (no filtering, _x_count, traversals, etc) in SQL. 

There a couple TODOs left to complete this (also noted in the code):
- update the keys in the `self._alias` dictionary (currently keys are query paths, but they need to be tuples of (query path, fold path)). This breaks one of the tests
- the SQL queries reuse alias names (ex `[Animal_1]`) need to validate that this is okay or update so that alias names are not reused
- intermediate names of fold outputs and fold subqueries might benefit from more descriptive names
- I updated the tests, but would like some extra eyes on verifying that those are the expected outputs

Thoughts @obi1kenobi @bojanserafimov @jecdohmann?


graphql_compiler/tests/test_compiler.py
```diff
@@ -4117,7 +4117,23 @@ def test_fold_on_output_variable(self):
...
+                SELECT
+                    array_agg([Animal_1].name) AS fold_output_1,
+                    [Animal_1].parent AS parent
+                FROM
+                    db_1.schema_1.[Animal] AS [Animal_1]
+                GROUP BY [Animal_1].parent
+            ) AS folded_subquery_1
+            ON [Animal_1].uuid = folded_subquery_1.parent
+        '''
+
```
bojanserafimov:
```
graphql_input = '''{
    animal {
        out_animal_parentof @fold {
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4170,7 +4186,21 @@ def test_fold_after_traverse(self):
...
+        JOIN db_1.schema_1.[Animal] AS [Animal_2]
+        ON [Animal_1].uuid = [Animal_2].parent
+        LEFT OUTER JOIN (
+            SELECT
+                array_agg([Animal_2].name) AS fold_output_1,
+                [Animal_2].parent AS parent
+            FROM db_1.schema_1.[Animal] AS [Animal_2]
+            GROUP BY [Animal_2].parent
+        ) AS folded_subquery_1
+        ON [Animal_2].uuid = folded_subquery_1.parent'''
```
bojanserafimov:
```
graphql_input = '''{
    Animal {
        in_Animal_ParentOf {
            out_Animal_ParentOf @fold {
            }
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4835,7 +4866,19 @@ def test_no_op_coercion_inside_fold(self):
...
+            coalesce(folded_subquery_1.fold_output_1, ARRAY[]::VARCHAR[]) AS related_entities
+        FROM db_1.schema_1.[Animal] AS [Animal_1]
+        LEFT OUTER JOIN (
+            SELECT
+                array_agg([Animal_1].name) AS fold_output_1,
+                [Animal_1].related_entity AS related_entity
+            FROM db_1.schema_1.[Animal] AS [Animal_1]
+            GROUP BY [Animal_1].related_entity
+        ) AS folded_subquery_1
+        ON [Animal_1].uuid = folded_subquery_1.related_entity'''
```
bojanserafimov:
```
graphql_input = '''{
    Animal {
        out_Entity_Related @fold {
            ... on Entity {
                name @output(out_name: "related_entities")
            }
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -6975,7 +7018,23 @@ def test_optional_and_fold(self):
...
+            db_1.schema_1.[Animal] AS [Animal_2]
+        ON [Animal_1].uuid = [Animal_2].parent
+        LEFT OUTER JOIN (
+            SELECT
+                array_agg([Animal_1].name) AS fold_output_1,
+                [Animal_1].parent AS parent
+            FROM db_1.schema_1.[Animal] AS [Animal_1]
+            GROUP BY [Animal_1].parent
+        ) AS folded_subquery_1
+        ON [Animal_1].uuid = folded_subquery_1.parent'''
```
bojanserafimov:
```
graphql_input = '''{
    Animal {
        }
        out_Animal_ParentOf @fold {
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -7050,7 +7109,23 @@ def test_fold_and_optional(self):
...
+            SELECT
+                array_agg([Animal_1].name) AS fold_output_1,
+                [Animal_1].parent AS parent
+            FROM db_1.schema_1.[Animal] AS [Animal_1]
+            GROUP BY [Animal_1].parent
+        ) AS folded_subquery_1
+        ON [Animal_1].uuid = folded_subquery_1.parent
+        LEFT OUTER JOIN
+            db_1.schema_1.[Animal] AS [Animal_2]
+        ON [Animal_1].uuid = [Animal_2].parent'''
```
bojanserafimov:
```
graphql_input = '''{
    Animal {
        out_Animal_ParentOf @fold {
        }
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
obi1kenobi: I don't think the order in the predicate matters, so long as we're consistent they should both work. I don't have a strong preference for which order we pick, so maybe let's keep it as-is unless that's somehow more work?

jecdohmann: I don't mean the order should be changed, the columns should be taken from different tables. Namely, uuid belongs to the subquery -

I think the query should read:

```

        SELECT
            [Animal_1].name AS animal_name,
            coalesce(folded_subquery_1.fold_output_1, ARRAY[]::VARCHAR[]) AS related_entities
        FROM db_1.schema_1.[Animal] AS [Animal_1]
        LEFT OUTER JOIN (
            SELECT
                array_agg([Entity_1].name) AS fold_output_1,
                [Entity_1].uuid AS uuid
            FROM db_1.schema_1.[Entity] AS [Entity_1]
            GROUP BY [Entity_1].related_entity
        ) AS folded_subquery_1
        ON folded_subquery_1.uuid = [Animal_1].related_entity
```

obi1kenobi:
In general, @bojanserafimov is a lot more familiar with this code than me, so I'll let him do the final review though I'm happy to poke at it if you need more feedback. I'll just have to read up more on all this since it's not my code.

graphql_compiler/tests/test_compiler.py
bojanserafimov: Yes, the table used should be `Entity`, not `Animal`. My impression that this error only happens only when using fold. The test `no_op_coercion_with_eligible_subpath` covers a similar case, without fold, and in that case the `Entity` table is used.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -167,6 +202,8 @@ def backtrack(self, previous_location):
 
     def traverse(self, vertex_field, optional):
         """Execute a Traverse Block."""
+        if self._in_fold:
+            raise NotImplementedError("Traversals inside a fold are not implemented yet.")
```
bojanserafimov: I'm worried implementing this will cause a major redesign of the state we store and update for fold. Do you have an idea of what is involved in implementing it?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -93,11 +110,14 @@ def __init__(self, sql_schema_info, ir):
         # Metadata
         self._sql_schema_info = sql_schema_info
         self._ir = ir
-        self._used_columns = _find_columns_used(sql_schema_info, ir)
+        self._used_columns = _find_columns_used_outside_folds(sql_schema_info, ir)
+        self._all_folded_outputs = _find_folded_outputs(ir)
+        self._in_fold = False
```
bojanserafimov: This is not immutable metadata, but very mutable state that depends on the current progress along the ir. I'd move it lower down, and maybe rename `# Metadata` to `# Immutable metadata` to be more explicit.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -107,6 +127,21 @@ def __init__(self, sql_schema_info, ir):
...
+
+        # Information needed to LEFT JOIN fold subquery to main Selectable (reset for each fold)
+        self._pre_fold_location = None  # the current Location prior to fold
+        self._fold_location = None  # the fold's FoldScopeLocation
+        self._fold_join_vertex_field = None  # field to perform LEFT JOIN on
+
+        # Information to keep track of fold aliases
+        self._folded_output_aliases = {}  # mapping FoldScopeLocations to their intermediate alias
+        self._folded_output_count = 1
+        self._fold_count = 1
```
bojanserafimov: There's 9 mutable variables needed just for fold. This is fine in this draft PR, to demonstrate that `@fold` works. However, this is where most of the code complexity comes from, and can lead to bugs. Mutable state is the hardest to document, so let's have as little of it as possible. I think simplifying the state is a worthwhile investment, and it looks to me like it would be necessary to implement traversals inside fold. Some techniques you can try:
- Try to adopt the existing fields instead of defining new ones when possible. Is it necessary to store both `_current_location` and `_fold_location`? Do you have to update both of them when you're inside fold? Is it necessary to have both `_aliases` and `_folded_output_aliases`?
- Group some fields into a namedtuple, reducing visual clutter. Examples:
    - The query components like `_fold_from_clause`, `_fold_outputs` and `_fold_group_by` are all part of one query. You could create a `QueryBuilder` class or namedtuple that contains all parts of the query, which will simplify even the non-fold code that generates the final result.
    - The `_fold_count` and `_folded_output_count` are only used to generate unique names, but take up a central position in the complicated state we need to manage. You can pull out this functionality out of this class into a `UniqueAliasGenerator`, making these things invisible.
- Derive some of the fields from others, converting them into pure functions. See `self._current_classname` as an example.

There might be other ways that I'm not seeing. Feel free to play around and see what works.

graphql_compiler/tests/test_compiler.py
jecdohmann: I think line 7128 should read `ON [Animal_1].parent = [Animal_2].uuid`

bojanserafimov: This might be a merge conflict issue. I pushed a change to master correcting the direction of the `Animal_ParentOf` edge. I'm guessing the tests are incorrect (because of the merge conflict), not your code. You can see that the tests are failing.

PR #575, opened 2019-09-26
Implement basic integration test for optional

bojanserafimov:
Added basic `@optional` test. Orientdb and redisgraph (or maybe our test framework for them) have some problems I'm working on.


bojanserafimov:
Turns out our Match backend omits keys in the output dicts instead of including `None` values on `@optional` miss.

obi1kenobi:
> Turns out our Match backend omits keys in the output dicts instead of including `None` values on `@optional` miss.

OrientDB's representation of `None` is to drop that key from the row dict, there's no way to represent an explicit `None` as far as I know, at least in OrientDB 2.x

bojanserafimov:
> > Turns out our Match backend omits keys in the output dicts instead of including `None` values on `@optional` miss.
> 
> OrientDB's representation of `None` is to drop that key from the row dict, there's no way to represent an explicit `None` as far as I know, at least in OrientDB 2.x

On second thought, this is not a problem. Technically the output representation of every database is completely different (they all might have custom binary formats). The fact that pyorient interprets the result this way is not an inconsistency, as long as we provide a way to normalize it client side the way it should be.

obi1kenobi:
> On second thought, this is not a problem. Technically the output representation of every database is completely different (they all might have custom binary formats). The fact that pyorient interprets the result this way is not an inconsistency, as long as we provide a way to normalize it client side the way it should be. In my latest commit I normalized the pyorient output in the tests.

I agree. We have the output metadata produced by the compiler, so we should even be able to coerce datetime strings back into datetime objects (if needed) etc.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -45,8 +45,19 @@ def compile_and_run_match_query(schema, graphql_query, parameters, orientdb_clie
...
+    # Get results, adding None for optional columns with no matches
     query = compilation_result.query
-    results = [row.oRecordData for row in orientdb_client.command(query)]
+    results = []
+    for row in orientdb_client.command(query):
+        row_dict = row.oRecordData
+        for output_name, output_info in six.iteritems(compilation_result.output_metadata):
+            if output_name not in row_dict:
+                if not output_info.optional:
+                    raise AssertionError(
```
obi1kenobi: This can get hit if the database has `null` as the value in some column. Is that something we guarantee the test data does not have?

bojanserafimov: Oh OrientDB returns nothing even for null values? In that case I'll take this assertion back.

PR #581, opened 2019-09-27
Add SchemaGraph section

pmantica1:
This is copy-paste.

### DIFF
--- a/docs/source/advanced_features/schema_graph.rst
+++ b/docs/source/advanced_features/schema_graph.rst
@@ -1,2 +1,62 @@
 Schema Graph
 ============
+
+When building a GraphQL schema from the database metadata, we first
+build a :code:`SchemaGraph` from the metadata and then, from the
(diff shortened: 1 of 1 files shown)


docs/source/advanced_features/schema_graph.rst
```diff
@@ -1,2 +1,62 @@
...
+
+    # Get the superclass of all classes allowed as the destination vertex of an edge class.
+    print(schema_graph.get_edge_schema_element_or_raise('Animal_Eats').base_out_connection)
+    # Food
+
+    # Get the unique indexes defined on a class.
+    print(schema_graph.get_unique_indexes_for_class('Animal'))
+    # [IndexDefinition(name='uuid', 'base_classname'='Animal', fields={'uuid'}, unique=True, ordered=False, ignore_nulls=False)]
+
+In the future, we plan to add :code:`SchemaGraph` generation from SQLAlchemy
```
bojanserafimov: Either fix or label this as incorrect info. I think somewhere else in the docs we already state we have schema generation from sqlalchemy.

PR #587, opened 2019-10-02
Add fast sqlalchemy reflection

pmantica1:


pmantica1:
I was not able to use the database name as a parameter of the query. To prevent SQL injection, I decided to instead raise an assertion error if the input did not match a database regex. Even SQLAlchemy allows for SQL Injection the same. https://github.com/sqlalchemy/sqlalchemy/issues/4883

graphql_compiler/schema_generation/sqlalchemy/sqlalchemy_reflector.py
```diff
@@ -0,0 +1,142 @@
...
+                to reflect into the metadata.
+        primary_key_selector: optional function that takes in a table name and list of dicts
+                              specifying column metadata and returns a set of column names to use
+                              as the primary key for the corresponding SQLAlchemy Table object.
+                              The compiler requires each SQLAlchemy Table to have a primary key.
+                              The primary keys do not need to be the primary keys in the underlying
+                              tables. They just need to be an unique and non-null identifier of
+                              each row. This parameter should be used to amend SQLAlchemy Table
+                              objects with such non-explicitly enforced primary keys.
+    """
```
bojanserafimov: Can you return a list of ignored types, a list of primary keys that were inferred, etc. to make the process more transparent? If you run a production system and realize days after generating the schema that it was incorrect, you can't reproduce the old state to find what went wrong. You can only rely on logs and dumps.

obi1kenobi:
> I was not able to use the database name as a parameter of the query. To prevent SQL injection, I decided to instead raise an assertion error if the input did not match a database regex. Even SQLAlchemy allows for SQL Injection the same. [sqlalchemy/sqlalchemy#4883](https://github.com/sqlalchemy/sqlalchemy/issues/4883)

That's disappointing, from a defense-in-depth perspective of course. May I suggest something simpler than database regex? I think we can probably safely limit database names to matching `[A-Za-z0-9\-_]+`, and hopefully simplify our code as a result.

PR #590, opened 2019-10-02
Improve filter selectivity computation

### LINKED ISSUE #589: Filters with multiple fields are handled incorrectly in cost estimation.
A field can apply to multiple fields. For example, the `name_or_alias` field applies to both `name`, or `alias`, returning true if at least one of those fields has the desired value.

A unique index can also apply to multiple fields, but that means something else. It means that the pair of those two values are unique.

In `filter_selectivity_utils.py`, with `_are_filters_uniquely_indexed`, we check if an index applies to the fields of an equality filter, and if so, assume the filter will retain at most one vertex. However, it is possible that there's a unique index on `(name, alias)`, and `name_or_alias` still find more than one match.

bojanserafimov:
Resolving some well-known problems with cost-estimation to make it more usable in production:
1. We currently compute the selectivity of each filter separately, and multiply those selectivities at the end. Because of this, the filter `x > median and x < median` will return a selectivity of `1/4`. This becomes a problem when using cost estimation for auto pagination. In this PR I solve this problem with a refactor, grouping together all filters relating to the same field.
2. We currently assume uuid4-valued fields are called `uuid`. This is true in the test schema, but not generally. I remove this assumption by providing information about uuid4-valued fields in the `QueryPlanningSchemaInfo`
3. This pr resolves https://github.com/kensho-technologies/graphql-compiler/issues/589 by not computing selectivities of `name_or_alias` filters (instead of computing them incorrectly).

The next step (coming in next the PR) is to implement inequality filter selectivities for integer fields using the quantile statistics.


graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -50,6 +52,8 @@ def _has_any_absolute(selectivities):
     return False
 
 
+# TODO(bojanserafimov): The class name should be checked against the class name of the index.
+# TODO(bojanserafimov): This is not correct for len(filter_fields) > 1.
```
bojanserafimov: Will address this in a future PR. In this PR I just make sure not to call it with `len(filter_fields) > 1`

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -507,6 +336,112 @@ def _combine_filter_selectivities(selectivities):
...
+    Args:
+        schema_info: QueryPlanningSchemaInfo
+        filter_infos: list of FilterInfos, filters on the location being filtered
+        parameters: dict, parameters with which query will be executed
+        location_name: string, type of the location being filtered
+        counts: float, result count that we're adjusting for filters
+
+    Returns:
+        Selectivity object
+    """
```
bojanserafimov: I know this function can be split up into smaller ones, but it's hard to tell where before integrating inequality computations for non-uuid field filters. That's coming up next.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -1,8 +1,12 @@
 # Copyright 2019-present Kensho Technologies, LLC.
+from __future__ import division
```
obi1kenobi: This is probably not necessary, I think all our divisions explicitly cast to float prior to dividing.

obi1kenobi (approved):

graphql_compiler/cost_estimation/filter_selectivity_utils.py
bojanserafimov: It's the recommended py2 compatible way of doing it in `six`, and our linters complain about it too. Not sure if it's necessary either.

obi1kenobi: Hmm okay. I thought that bug in `pylint` got fixed, but maybe we just haven't updated to the new version or something.

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

PR #591, opened 2019-10-03
Implement int filter inequality selectivity

bojanserafimov:
Use the quantile statistics to compute how selective an inequality filter on an int field is.

The formula used to compute the selectivity could be improved, but it will do well if the resolution of the quantile statistics is high enough. The formula makes sure not to return 0 or negative values for selectivity.

When reviewing this, please focus on whether you think the formula has the desired properties, and not so much on code structure. Addressing both issues in one PR will make it harder to review and merge. Once we decide on the formula, I will split up the whole function into manageable chunks.

### DIFF
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -373,6 +373,7 @@ def _get_selectivity_fraction_of_interval(interval, quantiles):
         lower_bound_quantile = bisect.bisect_left(quantiles, interval.lower_bound)
         upper_bound_quantile = bisect.bisect_left(quantiles, interval.upper_bound)
         if lower_bound_quantile == upper_bound_quantile:
+            # https://math.stackexchange.com/questions/195245/
             interval_size = 1.0 / 3
         else:
             interval_size = upper_bound_quantile - lower_bound_quantile
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -1,10 +1,12 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from __future__ import division
 
+import bisect
 from collections import namedtuple
 import sys
 from uuid import UUID
 
+from graphql import GraphQLInt
 import six
 
 from ..compiler.helpers import get_parameter_name
@@ -337,6 +339,11 @@ def _combine_filter_selectivities(selectivities):
     return Selectivity(kind=combined_selectivity_kind, value=combined_selectivity_value)
 
 
+def _is_int_field_type(schema_info, location_name, field_name):
+    field_type = schema_info.schema.get_type(location_name).fields[field_name].type
+    return field_type == GraphQLInt
+
+
 def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters, location_name):
     """Get the combined selectivity of all filters at the vertex.
 
@@ -368,7 +375,9 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters, 
 
         # Process inequality filters
         is_uuid4_field = field_name in schema_info.uuid4_fields.get(location_name, {})
-        if is_uuid4_field:
+        is_int_field = _is_int_field_type(schema_info, location_name, field_name)
+        is_inequality_filter_estimation_supported = is_uuid4_field or is_int_field
+        if is_inequality_filter_estimation_supported:
             for filter_info in filters_on_field:
                 if filter_info.op_name in INEQUALITY_OPERATORS:
                     parameter_values = [
@@ -377,9 +386,10 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters, 
                     ]
 
                     # Map uuid4 values to their corresponding integer values
-                    parameter_values = [
-                        _convert_uuid_string_to_int(value) for value in parameter_values
-                    ]
+                    if is_uuid4_field:
+                        parameter_values = [
+                            _convert_uuid_string_to_int(value) for value in parameter_values
+                        ]
 
                     filter_interval = _get_query_interval_of_integer_inequality_filter(
                         parameter_values, filter_info.op_name)
@@ -387,7 +397,7 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters, 
 
             if interval is None:
                 selectivity_at_field = Selectivity(kind=ABSOLUTE_SELECTIVITY, value=0.0)
-            else:
+            elif is_uuid4_field:
                 domain_interval = _create_integer_interval(MIN_UUID_INT, MAX_UUID_INT)
                 domain_interval_size = domain_interval.upper_bound - domain_interval.lower_bound + 1
                 interval = _get_intersection_of_intervals(interval, domain_interval)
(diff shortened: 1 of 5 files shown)


ReaLNeroM:
A few notes so that my internship thoughts don't get lost:
There's two ways to use quantiles:
0-1 interpolation: for a given bucket, see if all the values are within the query interval. If so, include the complete bucket. If a bucket is partially inside the query interval, do not add anything. 
Linear interpolation: Assume that values within a given quantile group are uniformly distributed. So if the query interval cuts a quantile bucket in half, you assume that half of the values in the bucket get included.

Benefits of 0-1 interpolation: You are guaranteed that the number you get is a lower bound on the actual number of results you'll get (as good as a lower bound as the statistics can get you). You can use 0-1 interpolation for strings (you can't linearly interpolate strings).
Benefits of linear interpolation: If a field is uniformly distributed, you can very easily implement support for it: create a quantile with the minimum and maximum values. This would make the UUID code branch and quantile code branch identical.

Hope this comes in useful.

pmantica1:
I am ok with the formula as is. One thing though is that you could return the worst case instead of expected case estimations. It would probably raise it one just a small constant so not a huge deal.

pmantica1:
I think that a bigger question is: What are the desired properties of the query cost estimator? (For instance, should it always overestimate vs underestimate). You don't need to answer them in this diff, but I think that this design principle should be listed somewhere.

bojanserafimov:
What's important: When you paginate a query with page size N, the cost estimator should return N as the estimated number of results for any page. This is important to enable nested pagination (paginating a query returned from the paginator). I don't care about any other details for now. It's easy to improve the cost estimator without breaking backwards compatibility.

The question of interpolating is important but not now. Interpolation can always be added later to improve the results if its necessary.

The same goes for the question of average vs max. Instead of writing perfect code to deal with it, I can get away with just generating more quantiles.

@pmantica1 If this makes sense I'll complete the current prototype implementation so that tests pass (there's stuff commented out).

pmantica1:
> What's important: When you paginate a query with page size N, the cost estimator should return N as the estimated number of results for any page. This is important to enable nested pagination (paginating a query returned from the paginator). I don't care about any other details for now. It's easy to improve the cost estimator without breaking backwards compatibility.
> 
> The question of interpolating is important but not now. Interpolation can always be added later to improve the results if its necessary.
> 
> The same goes for the question of average vs max. Instead of writing perfect code to deal with it, I can get away with just generating more quantiles.
> 
> @pmantica1 If this makes sense I'll complete the current prototype implementation so that tests pass (there's stuff commented out).

Makes sense to me.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -337,6 +339,47 @@ def _combine_filter_selectivities(selectivities):
...
+    """Get the fraction of values contained in an interval.
+
+    We ignore the interval endpoint values, and only consider the quantile they
+    are in. We treat the endpoints as uniform random variables within their quantile
+    and compute the expected value of the length of the interval divided by the size
+    of the whole domain.
+
+    Args:
+        interval: IntegerInterval defining the range of values
+        quantiles: Sorted list of N quantiles dividing the values into N+1 groups of
```
pmantica1: nit: sorted

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -397,6 +445,19 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
                     kind=FRACTIONAL_SELECTIVITY, value=fraction_of_domain_queried)
                 selectivity_at_field = _combine_filter_selectivities(
                     [selectivity_at_field, selectivity])
+            elif is_int_field:
+                # int fields are not uniformly distributed but we have quantile information.
+                quantiles = schema_info.statistics.get_field_quantiles(location_name, field_name)
+                quantiles = quantiles[1:-1]  # Discard the min and max value
```
pmantica1: I understand this is happening, but can you explain in your comment why?

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1194,6 +1194,82 @@ def test_inequality_filters_on_uuid(self):
...
+            statistics=statistics,
+            pagination_keys=pagination_keys,
+            uuid4_fields=uuid4_fields)
+
+        # Test <= filter in the middle
+        filter_info_list = [FilterInfo(fields=('limbs',), op_name='<=', args=('$limbs_upper',))]
+        params = {'limbs_upper': 8}
+        result_counts = adjust_counts_for_filters(
+            schema_info, filter_info_list, params, 'Species', 32.0)
+        # The value 8 is in the middle of the third quantile out of six.
```
pmantica1: I would say "bucket" here instead of "quantile". Can you also make this change in all the comments below?

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1194,6 +1194,82 @@ def test_inequality_filters_on_uuid(self):
...
+        self.assertAlmostEqual(expected_counts, result_counts)
+
+        # Test weak <= filter
+        filter_info_list = [FilterInfo(fields=('limbs',), op_name='<=', args=('$limbs_upper',))]
+        params = {'limbs_upper': 90}
+        result_counts = adjust_counts_for_filters(
+            schema_info, filter_info_list, params, 'Species', 32.0)
+        # The value 90 is in the middle of the last quantile.
+        expected_counts = 32.0 * (5.5 / 6.0)
+        self.assertAlmostEqual(expected_counts, result_counts)
```
pmantica1: Can you include at least one `>=` tests? Doesn't seem like much effort to add this?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -337,6 +339,47 @@ def _combine_filter_selectivities(selectivities):
...
+        interval_size = 0.5 + upper_bound_quantile
+    elif interval.upper_bound is None:
+        lower_bound_quantile = bisect.bisect_left(quantiles, interval.lower_bound)
+        interval_size = len(quantiles) - lower_bound_quantile
+    else:
+        lower_bound_quantile = bisect.bisect_left(quantiles, interval.lower_bound)
+        upper_bound_quantile = bisect.bisect_left(quantiles, interval.upper_bound)
+        if lower_bound_quantile == upper_bound_quantile:
+            # https://math.stackexchange.com/questions/195245/
+            interval_size = 1.0 / 3
```
pmantica1: Can you make this a constant and put that link above? That way we don't need to include that comment in multiple places.

bojanserafimov: I think that's less readable. I don't have to create a whole concept about this, give it a name, and a special place. It's not an important piece of this code and it does not deserve more attention.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
bojanserafimov: Lol actually had a bug here.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -397,6 +446,25 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
...
+                if quantiles is not None:
+                    selectivity = Selectivity(
+                        kind=FRACTIONAL_SELECTIVITY,
+                        value=_get_selectivity_fraction_of_interval(
+                            interval,
+                            # Since we can't be sure the minimum observed value is the
+                            # actual minimum value, we treat values less than it as part
+                            # of the first quantile. That's why we drop the minimum and
+                            # maximum observed values from the quantile list.
+                            quantiles[1:-1],
```
pmantica1: Why not move this inside `_get_selectivity_fraction_of_interval`?  The function does not what it promises right now.

pmantica1: Also, can you add a test when the number of quantiles is less than 2?

PR #592, opened 2019-10-03
Backend refactor, adding run_func

bojanserafimov:
We've implemented custom `run_query` functions for each of our backends, based on some python client for the respective database. These run functions show the intended way to run the queries generated for the backend, and resolve any minor backend differences, like enforcing a unifying date format, or standardizing optional output behavior (return `None` if not present).

So far we only use those functions in tests, and we import them from `integration_test_helpers` whenever needed. By standardizing and exposing them we benefit in the following ways:
- The codebase is more structured.
- The work necessary to implement a backend is more clearly defined.
- We avoid plumbing whenever we define a new backend.
- We explicitly document postprocessing steps on the result returned from the database for each backend.
- Make it easier to define a provider and implement cross-db queries


bojanserafimov:
This should be done in `test_helpers.BackendTester` instead of the `Backend` class.

PR #593, opened 2019-10-04
Add xfail tests for cost estimation rotation invariance

bojanserafimov:
If you rewrite the exact same query in a different way, you get a different result. That's bad.

I've added a failing test case. Looking for solutions. The bug has to do with absolute filter selectivities.


bojanserafimov:
This is the problem:
An absolute filter selectivity with value `1` means "in the entire query, there's only one match for that vertex". However, due to difficulties with the recursive structure of the computation, we treat it as "Treat the many-to-many edge going into this scope as a many-to-one edge", which is much weaker.

bojanserafimov:
The second test case I added demonstrates the bug appears even with fractional filter selectivities. The problem is that in the recursive algorithm, the selectivity is only propagated to the children, but not up towards the root.

bojanserafimov:
One of the tests was broken. Turns out the only problem is that absolute selectivity is interpreted locally.

obi1kenobi (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

Issue #535, opened 2019-09-03
`@fold` support for SQL


chewselene:
Here's the initial plan for how to handle `@fold` in SQL. May need additional considerations for interactions with other directives.

https://www.db-fiddle.com/f/cD9aczVx5B8gRiAXavWBhB/3

Let me know what you think @obi1kenobi @bojanserafimov @pmantica1

obi1kenobi:
I think this is a step in the right direction, thank you for putting it together. A few observations about opportunities for further iteration:

- `FULL JOIN` seems like it might not be the right thing there. I think all of those joins should be outer only on one side -- the side that isn't being folded on.

- I'm also confused about the `DISTINCT` in the `array_agg` calls. The comment says something about eliminating duplicates, but I'm not sure what duplicates there are, and why they are being eliminated. Essentially, I'm unable to match up this `DISTINCT` to any part of the compiler spec for `@fold`.

- Also worth considering is how filters get applied, and how `_x_count` outputting and filtering would work.

chewselene:
In the end `DISTINCT` was unnecessary (query 5) since each array aggregation will occur in a sub-select statement. It was join an intermediate thing I tried since the first query produces dupes.

Will look into what type of `JOIN` to use and filtering.

chewselene:
Here's the plan for how to handle filters and `_x_count`:

https://www.db-fiddle.com/f/4EXkttLVoWTdbMZQBsNDDB/1

The basic idea is to have `@fold` and `_x_count` be a part of a sub-`SELECT` statement that is joined with a `LEFT JOIN`. These fields are then `COALESCE`-ed to `[]` or `0`, respectively, if a `null` was returned. Any `@filter` operations will part of a `HAVING` clause, which will require that the query have a `GROUP BY` clause on all fields not used in a `@filter`.

Next I will be looking at traversals within `@fold`.

obi1kenobi:
I'm a bit worried I might have misunderstood the last bit around the `HAVING` clause and which fields go in the `GROUP BY` – not sure why the fields in the `GROUP BY` change on the basis of what we're filtering on. Imagine we define a filter that is always satisfied – the result should be the same whether or not that filter was there, but I am concerned that by changing which fields go in the `GROUP BY` we might get different results.

Also, does the above cause any issues if we try to both filter and output a given field?

chewselene:
Yeah, that's my bad in the explanation; this part 

[...]

PR #595, opened 2019-10-04
Implement `@fold` for PostgreSQL

jecdohmann:
This is a reorganization of PR #573 which is in turn based on issue #535 

I have modified Selene's implementation of standard folds (no traversals within the fold) to be a special case of her implementation of folds followed by a traversal. So far only standard folds are implemented in this PR. Similarly, _x_count hasn't been added yet.

In particular the original spec called for regular folds to be implemented as

1. 
```
SELECT
    Cities.name AS city,
    COALESCE(agg_neighborhoods, ARRAY[]::VARCHAR[]) AS neighborhoods,
FROM Cities
LEFT JOIN (
  SELECT 
  array_agg(Neighborhoods.name) AS agg_neighborhoods,
  Neighborhoods.city,
  FROM Neighborhoods
  GROUP BY Neighborhoods.city
) AS temp_table 
ON Cities.name = temp_table.city
```

With folds followed by a traversal implemented as:

2.

```
SELECT
	Cities.name AS city,
        COALESCE(temp_table.streets, ARRAY[]::VARCHAR[]) AS streets
FROM Cities
LEFT JOIN (
  SELECT
    Cities.name AS name,
    array_agg(Streets.name) AS streets
  FROM Cities
  INNER JOIN Neighborhoods ON Cities.name = Neighborhoods.city
  INNER JOIN Streets ON Streets.neighborhood = Neighborhoods.name
  GROUP BY Cities.name
) AS temp_table
ON Cities.name = temp_table.name;
```

Note that in the first case the subquery selects only elements from the folded vertex and then joins to the outer vertex using info you'd need to get from a join descriptor.

On the other hand, in the second example the query begins from the outer vertex and performs all the necessary joins in sequence until it reaches the final vertex table. The resulting query is then joined to the outermost table via a join on the equality of the unique identifier used in both tables for each row (in particular, Cities.name).

I've rewritten standard folds (1.) so that they are a general case of the structure give in (2.), see below:

3.
```
SELECT
    Cities.name AS city,
    COALESCE(agg_neighborhoods, ARRAY[]::VARCHAR[]) AS neighborhoods,
FROM Cities
LEFT JOIN (
  SELECT 
  array_agg(Neighborhoods.name) AS agg_neighborhoods,
  Citites.name,
  FROM Citites
  JOIN Neighborhoods <- Only difference between 2 and 3 is # of joins
  ON Citites.name = Neighborhoods.city
  GROUP BY Citites.name
) AS temp_table 
ON Cities.name = temp_table.name
```

This makes it so that the only difference between a fold with no additional traversals and a fold with multiple traversals within it is the number of joins in the subquery.

If a fold occurs AFTER a number of traversals, then the joins corresponding to those traversals occur outside the subquery, and the rightmost table in the sequence of joins plays the role of the leftmost table within the subquery.

An  example is below:
```
{
 Animal {
   out_Animal_LivesIn {
     in_Animal_LivesIn @fold {
       name @output(out_name: "neighbor_and_self_names_list")
     }
    }
  }
}
```
with corresponding sql:
4.
```
SELECT
  [Animal_1].name AS animal_name,
  coalesce(folded_subquery_1.fold_output_1, ARRAY[]::VARCHAR[]) AS neighbor_and_self_names_list
FROM db_1.schema_1.[Animal] AS [Animal_1]
JOIN db_1.schema_1.[Location] AS [Location_1] <- rightmost table outside subquery
ON [Animal_1].lives_in = [Location_1].uuid
LEFT OUTER JOIN (
  SELECT
    array_agg([Animal_2].name) AS fold_output_1,
    [Location_2].uuid AS uuid
 FROM db_1.schema_1.[Location] AS [Location_2] <- Location appears again as leftmost table inside subquery 
 JOIN db_1.schema_1.[Animal] AS [Animal_2]
 ON [Location_2].uuid = [Animal_2].lives_in
  GROUP BY [Location_2].uuid
) AS folded_subquery_1
ON [Location_1].uuid = folded_subquery_1.uuid
```

This change should make it substantially easier to implement folds with embedded traversals. @bojanserafimov @obi1kenobi @pmantica1 @chewselene  let me know what you think. I will start writing folds with embedded traversals and push to this as it's finished


graphql_compiler/tests/test_compiler.py
```diff
@@ -4117,7 +4119,23 @@ def test_fold_on_output_variable(self):
...
         '''
-        expected_sql = NotImplementedError
+        expected_sql = '''
+            SELECT
+                [Animal_1].name AS animal_name,
+                coalesce(folded_subquery_1.fold_output_1, ARRAY[]::VARCHAR[]) AS child_names_list
+            FROM 
+                db_1.schema_1.[Animal] AS [Animal_1]
+            LEFT OUTER JOIN (
+                SELECT
```
jecdohmann: graphql_input = '''{
    animal {
        out_animal_parentof @fold {
        }
    }
}'''

graphql_compiler/tests/test_compiler.py
```diff
@@ -4117,7 +4119,23 @@ def test_fold_on_output_variable(self):
...
+                    array_agg([Animal_2].name) AS fold_output_1,
+                    [Animal_3].uuid AS uuid
+                FROM db_1.schema_1.[Animal] AS [Animal_3]
+                JOIN db_1.schema_1.[Animal] AS [Animal_2] ON [Animal_3].uuid = [Animal_2].parent
+                GROUP BY 
+                    [Animal_3].uuid
+            ) AS folded_subquery_1
+            ON [Animal_1].uuid = folded_subquery_1.uuid
+        '''
```
jecdohmann:
```
graphql_input = '''{
    animal {
        out_animal_parentof @fold {
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4170,7 +4187,23 @@ def test_fold_after_traverse(self):
...
+            LEFT OUTER JOIN (
+                SELECT
+                    array_agg([Animal_3].name) AS fold_output_1,
+                    [Animal_4].uuid AS uuid
+                FROM db_1.schema_1.[Animal] AS [Animal_4]
+                JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                ON [Animal_4].uuid = [Animal_3].parent
+                GROUP BY [Animal_4].uuid
+            ) AS folded_subquery_1
+            ON [Animal_2].uuid = folded_subquery_1.uuid'''
```
jecdohmann:
```
graphql_input = '''{
    Animal {
        in_Animal_ParentOf {
            out_Animal_ParentOf @fold {
            }
        }
    }
}'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4191,6 +4224,62 @@ def test_fold_after_traverse(self):
...
+                GROUP BY [Location_2].uuid
+            ) AS folded_subquery_1
+            ON [Location_1].uuid = folded_subquery_1.uuid'''
+
+        result = compile_graphql_to_sql(self.sql_schema_info, test_data.graphql_input)
+        string_result = str(result.query.compile(dialect=self.sql_schema_info.dialect))
+        compare_sql(self, expected_sql, string_result)
+        self.assertEqual(test_data.expected_output_metadata, result.output_metadata)
+        compare_input_metadata(self, test_data.expected_input_metadata,
+                               result.input_metadata)
```
jecdohmann:
```
{
            Animal {
                out_Animal_LivesIn {
                    in_Animal_LivesIn @fold {
                        name @output(out_name: "neighbor_and_self_names_list")
                    }
                }
            }
        }
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4191,6 +4224,62 @@ def test_fold_after_traverse(self):
...
```
jecdohmann:
```
{
                Animal {
                    out_Animal_LivesIn {
                        name @output(out_name: "location_name")
                        in_Animal_LivesIn @fold {
                            name @output(out_name: "neighbor_and_self_names_list")
                        }
                    }
                }
            }
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4230,7 +4319,23 @@ def test_fold_and_traverse(self):
...
+                SELECT
+                    [Animal_2].uuid,
+                    array_agg([Animal_4].name) as sibling_and_self_names_list
+                FROM db_1.schema_1.[Animal] AS [Animal_2]
+                INNER JOIN db_1.schema_1.[Animal] AS [Animal_3] ON [Animal_2].parent = [Animal_3].uuid
+                INNER JOIN db_1.schema_1.[Animal] AS [Animal_4] ON [Animal_3].uuid = [Animal_4].parent
+                GROUP BY [Animal_2].uuid
+            ) AS folded_subquery_1
+            ON [Animal_1].uuid = folded_subquery_1.uuid
+        '''
```
jecdohmann:
```
{
        Animal {
            in_Animal_ParentOf @fold {
                out_Animal_ParentOf {
                }
            }
        }
    }
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -4742,7 +4847,33 @@ def test_fold_date_and_datetime_fields(self):
...
+                    LEFT OUTER JOIN (
+                        SELECT
+                            array_agg([FeedingEvent_1].event_date) AS fold_output_2,
+                            [FeedingEvent_1].uuid AS uuid
+                        FROM
+                            db_2.schema_1.[FeedingEvent] AS [FeedingEvent_1]
+                        GROUP BY [FeedingEvent_1].uuid
+                    ) AS folded_subquery_2
+                    ON [Animal_1].fed_at = folded_subquery_1.uuid
+                '''
```
jecdohmann:
```
{
        Animal {
            out_Animal_FedAt @fold {
                event_date @output(out_name: "fed_at_datetimes_list")
            }
            out_Animal_ParentOf @fold {
                birthday @output(out_name: "child_birthdays_list")
            }
        }
    }
```

DATETIME objects aren't implemented in sql fold.

graphql_compiler/tests/test_compiler.py
```diff
@@ -4835,7 +4966,22 @@ def test_no_op_coercion_inside_fold(self):
...
+        LEFT OUTER JOIN (
+            SELECT
+                array_agg([Entity_1].name) AS fold_output_1,
+                [Animal_2].uuid AS uuid
+            FROM db_1.schema_1.[Animal] AS [Animal_2] 
+            JOIN db_1.schema_1.[Entity] AS [Entity_1]
+            ON [Animal_2].uuid = [Entity_1].related_entity
+            GROUP BY [Animal_2].uuid
+        ) AS folded_subquery_1
+        ON [Animal_1].uuid = folded_subquery_1.uuid'''
```
jecdohmann:
```
{
        Animal {
            out_Entity_Related @fold {
                ... on Entity {
                    name @output(out_name: "related_entities")
                }
            }
        }
    }
```

This fails because hyper edges aren't yet implemented completely so the join descriptor returns the edges in the wrong from/to order for the  Animal -> Entity join

graphql_compiler/tests/test_compiler.py
```diff
@@ -6975,7 +7121,25 @@ def test_optional_and_fold(self):
...
+        LEFT OUTER JOIN (
+            SELECT
+                array_agg([Animal_3].name) AS fold_output_1,
+                [Animal_4].uuid AS uuid
+            FROM db_1.schema_1.[Animal] AS [Animal_4]
+            JOIN db_1.schema_1.[Animal] AS [Animal_3]
+            ON [Animal_4].uuid = [Animal_3].parent
+            GROUP BY [Animal_4].uuid
+        ) AS folded_subquery_1
+        ON [Animal_1].uuid = folded_subquery_1.uuid'''
```
jecdohmann:
```
{
        Animal {
            }
            out_Animal_ParentOf @fold {
            }
        }
    }
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -7050,7 +7214,25 @@ def test_fold_and_optional(self):
...
+                [Animal_4].uuid AS uuid
+            FROM db_1.schema_1.[Animal] AS [Animal_4]
+            JOIN db_1.schema_1.[Animal] AS [Animal_3]
+            ON [Animal_4].uuid = [Animal_3].parent
+            GROUP BY [Animal_4].uuid
+        ) AS folded_subquery_1
+        ON [Animal_1].uuid = folded_subquery_1.uuid
+        LEFT OUTER JOIN
+            db_1.schema_1.[Animal] AS [Animal_2]
+        ON [Animal_1].parent = [Animal_2].uuid'''
```
jecdohmann:
```
{
        Animal {
            out_Animal_ParentOf @fold {
            }
            }
        }
    }
```

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+        ].alias()
+        outer_vertex_table = self._current_alias.alias()
+        inner_vertex_type = self._ir.query_metadata_table \
+            .get_location_info(fold_scope_location) \
+            .type \
+            .name
+        outer_vertex_type = self._current_classname
+
+        self._aliases[
+            (self._current_location.query_path, self._fold_inner_vertex_location)
```
bojanserafimov: Not clear to me why we're using the `_current_location.query_path` here. We're assigning the `inner_vertex_table` as an alias for the outer vertex location?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+            raise AssertionError(u'Reached an invalid fold state.')
+
+        # begin the fold
+        self._in_fold = True
+        # outer vertex is whatever vertex occurs immediately before the fold
+        self._fold_outer_vertex_location = self._current_location
+        self._fold_inner_vertex_location = fold_scope_location
+
+        # basic info about the outer vertex and inner vertex
+        inner_vertex_table = self._sql_schema_info.vertex_name_to_table[
```
bojanserafimov: Let's call these aliases for consistency.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+
+        if self._in_fold:
+            raise AssertionError(u'Reached an invalid fold state.')
+
+        # begin the fold
+        self._in_fold = True
+        # outer vertex is whatever vertex occurs immediately before the fold
+        self._fold_outer_vertex_location = self._current_location
+        self._fold_inner_vertex_location = fold_scope_location
+
```
bojanserafimov: Shouldn't  `self._current_location` be updated to be the `fold_scope_location`? That's what the spec for that field implies should happen.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+            .name
+        outer_vertex_type = self._current_classname
+
+        self._aliases[
+            (self._current_location.query_path, self._fold_inner_vertex_location)
+        ] = inner_vertex_table
+
+        # collect edge information to join the fold subquery to the main selectable
+        edge_direction = fold_scope_location.fold_path[0][0]
+        edge_name = fold_scope_location.fold_path[0][1]
```
bojanserafimov: For better readability and less indexing, you can unpack this as `edge_direction, edge_name = fold_scope_location.fold_path[0]`

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,30 +77,113 @@ def _find_columns_used(sql_schema_info, ir):
...
+    folded_outputs = {}
+    # Find outputs used for each fold path
+    for _, output_info in ir.query_metadata_table.outputs:
+        if isinstance(output_info.location, FoldScopeLocation):
+            fold_path = output_info.location.fold_path
+            folded_outputs.setdefault(fold_path, set()).add(output_info.location)
+    return folded_outputs
+
+
+class SQLFoldObject(object):
```
bojanserafimov: Please document what this is and why it exists. Is it a mutable class? I see it being constructed and then mutated by directly going into its private fields. Is that the intended usage?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,30 +77,113 @@ def _find_columns_used(sql_schema_info, ir):
...
+    def __init__(self, inner_vertex_table,
+                 outer_vertex_table,
+                 inner_vertex_type,
+                 outer_vertex_type,
+                 join_info):
```
bojanserafimov: These arguments are underdocumented, even taking the getter docstrings into account. `join_info` is never fully specified.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+                     inner_vertex_type,
+                     self._current_alias)
+
+        self._fold = SQLFoldObject(inner_vertex_table,
+                                   outer_vertex_table,
+                                   inner_vertex_type,
+                                   outer_vertex_type,
+                                   join_info)
+        # find outputs for this fold in self._all_folded_outputs and add to self._fold_outputs
+        if fold_scope_location.fold_path in self._all_folded_outputs:
```
bojanserafimov: isn't it an assertion error if you don't find this? That's what the spec of the function that computes `self._all_folded_outputs` implies.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -46,21 +47,23 @@ def _traverse_and_validate_blocks(ir):
         yield block
 
 
-def _find_columns_used(sql_schema_info, ir):
+def _find_columns_used_outside_folds(sql_schema_info, ir):
```
bojanserafimov: Update the docstring too.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -107,11 +193,21 @@ def __init__(self, sql_schema_info, ir):
         self._outputs = []  # sqlalchemy Columns labelled correctly for output
         self._filters = []  # sqlalchemy Expressions to be used in the where clause
 
+        self._fold = None
+        self._in_fold = False
+        self._fold_inner_vertex_location = None
+        self._fold_outer_vertex_location = None
+
+        # Information to keep track of fold aliases
+        self._fold_output_column_aliases = {}
```
bojanserafimov: Can you document what this one means, and how it's formatted?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,30 +77,113 @@ def _find_columns_used(sql_schema_info, ir):
...
+
+
 class CompilationState(object):
     """Mutable class used to keep track of state while emitting a sql query."""
 
     def __init__(self, sql_schema_info, ir):
         """Initialize a CompilationState, setting the current location at the root of the query."""
-        # Metadata
+        # Immutable metadata
+        self._fold_location_info = None
```
bojanserafimov: Can you document what this one means? Is it even used?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+                                   outer_vertex_type,
+                                   join_info)
+        # find outputs for this fold in self._all_folded_outputs and add to self._fold_outputs
+        if fold_scope_location.fold_path in self._all_folded_outputs:
+            for fold_output in self._all_folded_outputs[fold_scope_location.fold_path]:
+                if fold_output.field == '_x_count':
+                    raise NotImplementedError(u'_x_count not implemented in SQL')
+
+                intermediate_fold_output_name = self._alias_generator.gen_output_column()
+                self._fold_output_column_aliases[fold_output] = intermediate_fold_output_name
```
bojanserafimov: This is suspicious for two reasons (maybe I don't understand it):
1. I'm guessing we generate `intermeiate_fold_output_name` because `fold_output`s can collide from different outputs on different scopes inside the fold, but on fields with the same name. If that's the case, can you try to find an example query where that happens? I'm pretty sure the rules and constraints on `@fold` in the README rule out that case.
2. What's the point of generating unique names if we're using `fold_output` as a unique key in a dictionary? If `fold_output` is not unique, we will get collisions in this dictionary. Furthermore this `self._fold_output_column_aliases` is shared between all folds we process, and we can get into some really nasty bugs if information about one fold overrides information about another fold in the query.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+        self._fold.outputs.append(self._fold.outer_vertex_table.c[edge.from_column])
+        self._fold.group_by.append(self._fold.outer_vertex_table.c[edge.from_column])
+
+    def unfold(self):
+        """Complete the execution of a Fold Block."""
+        # Create the folded select statement, updating the fold count and adding the subquery
+        # to self._aliases
+        # left join to the self._from_clause on self._fold_join_vertex_field
+        edge = self._fold.join_info[0]
+        j = self._fold.outer_vertex_table
```
bojanserafimov: Can you pick a better name than `j`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+
+        # the group by field is the unique identifier for the outer vertex
+        self._fold.outputs.append(self._fold.outer_vertex_table.c[edge.from_column])
+        self._fold.group_by.append(self._fold.outer_vertex_table.c[edge.from_column])
+
+    def unfold(self):
+        """Complete the execution of a Fold Block."""
+        # Create the folded select statement, updating the fold count and adding the subquery
+        # to self._aliases
+        # left join to the self._from_clause on self._fold_join_vertex_field
```
bojanserafimov: According to the comments, there's 3 things going on here, and it's not clear what lines of code these comments are talking about. The `_alias` updating is in the next code paragraph. Can you try to organize it into a readable narrative? As I see it, the one thing that happens here is we join the subquery to the main query. Everything else is secondary, as it is maintaining the correctness of our state.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+        outer_vertex_table = self._current_alias.alias()
+        inner_vertex_type = self._ir.query_metadata_table \
+            .get_location_info(fold_scope_location) \
+            .type \
+            .name
+        outer_vertex_type = self._current_classname
+
+        self._aliases[
+            (self._current_location.query_path, self._fold_inner_vertex_location)
+        ] = inner_vertex_table
```
bojanserafimov: When is this alias ever going to be used? As I understand this, we also add to `self._aliases[(something, None)]` when we unfold, which is the value we use when emitting a `FoldedContextField`.

graphql_compiler/tests/test_compiler.py
```diff
@@ -4191,6 +4224,64 @@ def test_fold_after_traverse(self):
         check_test_data(self, test_data, expected_match, expected_gremlin, expected_sql,
                         expected_cypher)
 
+    def test_fold_after_traverse_different_types(self):
```
chewselene: what is different types referring to here? Location vs. Animal? Should other languages use this test?

graphql_compiler/tests/test_compiler.py
```diff
@@ -4191,6 +4224,64 @@ def test_fold_after_traverse(self):
...
+                    ON [Location_1].uuid = folded_subquery_1.uuid'''
+
+        result = compile_graphql_to_sql(self.sql_schema_info, test_data.graphql_input)
+        string_result = str(result.query.compile(dialect=self.sql_schema_info.dialect))
+        compare_sql(self, expected_sql, string_result)
+        self.assertEqual(test_data.expected_output_metadata, result.output_metadata)
+        compare_input_metadata(self, test_data.expected_input_metadata,
+                               result.input_metadata)
+
+    def test_fold_after_traverse_no_output_on_root(self):
```
chewselene: Same question as above: should other languages use this test?

graphql_compiler/tests/test_compiler.py
```diff
@@ -4835,7 +4981,23 @@ def test_no_op_coercion_inside_fold(self):
                 )
             ])}
         '''
-        expected_sql = NotImplementedError
+        # TODO Not working b/c join descriptors for hyper edges not working properly
```
chewselene: What's the solution for this? Is there a plan for how to make hyper edges work?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+        # the group by field is the unique identifier for the outer vertex
+        self._fold.outputs.append(self._fold.outer_vertex_table.c[edge.from_column])
+        self._fold.group_by.append(self._fold.outer_vertex_table.c[edge.from_column])
+
+    def unfold(self):
+        """Complete the execution of a Fold Block."""
+        # Create the folded select statement, updating the fold count and adding the subquery
+        # to self._aliases
+        # left join to the self._from_clause on self._fold_join_vertex_field
+        edge = self._fold.join_info[0]
```
chewselene: could join_info be a named tuple? this is a little confusing to follow

graphql_compiler/compiler/emit_sql.py
chewselene: I was thinking the keys to `self._aliases` should be `(query_path, fold_path)` (as in the TODO on 186). 

For folds, the `query_path` comes from it's `base_location`. For non-folds no `fold_path` is needed so the key becomes `(query_path, None)`. This change was needed because the `fold_path` alone is not a unique key.

Looks like we're using the entire `FoldScopeLocation` as the second part of the tuple though, which seems unnecessary to me.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+        )
+        folded_subquery = sqlalchemy.select(
+            self._fold.outputs
+        ).select_from(
+            j
+        ).group_by(
+            *self._fold.group_by
+        ).alias(self._alias_generator.gen_subquery())
+
+        self._aliases[(self._fold_inner_vertex_location.fold_path, None)] = folded_subquery
```
chewselene: I don't think this is necessarily a unique key. Isn't it possible to take the same `fold_path` starting from different base locations?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
 
     def filter(self, predicate):
         """Execute a Filter Block."""
-        sql_expression = predicate.to_sql(self._aliases, self._current_alias)
+        if self._in_fold or isinstance(predicate.left, FoldedContextField) \
+                or isinstance(predicate.right, FoldedContextField):
```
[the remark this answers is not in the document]
jecdohmann: Yes, I'm not sure its possible to produce such an IR because we only permit filters that take variables or values as arguments, not other predicates. A complete solution to this would probably involve traversing the left and right predicates until you get to a leaf value or variable. Implementing filters for folds will be one of the next things to do so it might be worth it to just leave the fully complete check until then.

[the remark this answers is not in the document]
obi1kenobi: One other thing to keep in mind here. Use implicit line joining and avoid backticks (`\`) whenever possible, per the Google Style Guide (note our line length is 100 chars, otherwise the same rules apply): https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,30 +77,113 @@ def _find_columns_used(sql_schema_info, ir):
...
+        return self._join_info
+
+
+class UniqueAliasGenerator(object):
+    def __init__(self):
+        """Create unique subquery and column aliases by tracking counters."""
+        self._folded_output_count = 1
+        self._fold_count = 1
+
+    def gen_subquery(self):
```
obi1kenobi: Avoid abbreviations. Optimize for readability, not brevity / typing speed.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+        # outer vertex is whatever vertex occurs immediately before the fold
+        self._fold_outer_vertex_location = self._current_location
+        self._fold_inner_vertex_location = fold_scope_location
+
+        # basic info about the outer vertex and inner vertex
+        inner_vertex_table = self._sql_schema_info.vertex_name_to_table[
+            self._ir.query_metadata_table.get_location_info(fold_scope_location).type.name
+        ].alias()
+        outer_vertex_table = self._current_alias.alias()
+        inner_vertex_type = self._ir.query_metadata_table \
```
obi1kenobi: More backticks here, all of these need to go. Consult the link in my other comment for advice on how to do that.

graphql_compiler/compiler/emit_sql.py
jecdohmann: I actually think storing this table in the aliases isn't necessary because it is stored inside the SQLFoldObject. The aliases dictionary doesn't ever actually get queried with any tuples for which the second element is non-null.

I am going to remove this redundant book-keeping in this PR. We can discuss adding it again if it is necessary for traversals inside fold (though I think the SQLFoldObject state should be sufficient in that case anyway).

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +340,152 @@ def start_global_operations(self):
...
+                # add array aggregated intermediate output to self._fold_outputs
+                self._fold.outputs.append(
+                    sqlalchemy.func.array_agg(
+                        self._fold.inner_vertex_table.c[fold_output.field]
+                    ).label(intermediate_fold_output_name)
+                )
+
+        # the group by field is the unique identifier for the outer vertex
+        self._fold.outputs.append(self._fold.outer_vertex_table.c[edge.from_column])
+        self._fold.group_by.append(self._fold.outer_vertex_table.c[edge.from_column])
```
obi1kenobi: This is a [code smell](https://en.wikipedia.org/wiki/Code_smell). Just because some state got encapsulated into `self._fold` of type `SQLFoldObject`, it doesn't mean that this is composable or understandable code. And yes, you may have avoided the linter's complaint about mutating object internal state by making it public, but that is treating the symptoms and not the underlying problem.

Here's the underlying problem: `def fold()` is almost a 100 lines long, and is full of extremely complex code that I'm finding quite difficult to understand already. Code only becomes less understandable over time, and that means trouble.

Instead, try to describe in a few words what this function needs to do. You'll notice that you'll need to define some terminology to keep the description concise — great! Make helper functions for those bits of terminology, and go from there.

graphql_compiler/compiler/emit_sql.py
jecdohmann: I don't think so because we don't actually traverse into the folded location. If we did, then I think we'd need to backtrack after the Unfold.

Imagine we have 
```
{
        Animal {
            out_Animal_ParentOf @fold {
            }
            }
        }
    }
```

If we changed current location inside the Fold block, there would be no relocate before the Traverse of in_Animal_ParentOf. Basically, I don't think the IR treats folds as changes of location like it does a regular traversal, because there is no backtrack or traverse call.

graphql_compiler/compiler/emit_sql.py
chewselene: yep, should be changed to an assertion error instead

graphql_compiler/compiler/emit_sql.py
jecdohmann: The primary motivation is that if you don't provide some name then the resulting output column:

```
    sqlalchemy.func.array_agg(
         self._fold.inner_vertex_alias.c[fold_output.field]
     )
```

the resulting column is given an `anon_label` instead of a `label`. The `anon_label` isn't actually generated until the column is selected (to my knowledge). This would make it difficult to associate a fold scope location with stable alias - but we need to know the stable alias because at line 740 in expressions.to_sql() we index directly into the fold subquery table by the name of the column being selected. I don't really see a way to get this to work without explicitly providing _some_ label. That said, providing globally unique names, whatever those names are makes sure that we can always index into the table containing this column and get precisely that column back.

I don't know if collisions is actually an issue because fold_output should be unique to a specific fold and a specific output column? I'm not sure though, could you elaborate on that reasoning?

graphql_compiler/compiler/emit_sql.py
jecdohmann: You're right - I will remove this line

graphql_compiler/compiler/emit_sql.py
jecdohmann: Yeah this is tricky because _fold_inner_vertex_location is too specific, but the fold path is both too vague and not specific enough, because the subquery can be defined by multiple outputs inside the fold, so it doesn't make sense to identify it with only one. Furthermore, so is self._fold_inner_vertex_location.base_location is also too vague. 

How do you think I should resolve this?

chewselene: It seems like the key for `self._aliases` is ill defined. Prior to fold, it was simply a `query_path`, but we also need to accommodate `fold_path`s, which are not necessarily unique. 

In the TODO I left about this, I had proposed the use of a tuple of `(query_path, fold_path)`, where the `query_path` comes from the `base_location` for `FoldScopeLocation` and where `fold_path` is `None` for non folded aliases. Does that give you the specificity/uniqueness you need?

graphql_compiler/compiler/emit_sql.py
bojanserafimov: Great, `fold_output` is unique. Can we make `intermediate_fold_output_name = fold_output`? I think that addresses the `anon_label` issue because we explicitly name our columns. It also removes the need for `self._fold_output_column_aliases` entirely. Let me know if I'm missing something.

graphql_compiler/compiler/emit_sql.py
jecdohmann: Hi, you're right. I was mistaken about the relationship between (query_path, fold_path) tuples and folded output columns. I thought each such tuple uniquely identified just one column in a fold, and therefore would've been too specific in the case that there are multiple outputs. FoldScopeLocations uniquely identify output columns (b/c thet contain base_location, fold_path, AND field) so I must've been confused. 

You're right, that system works.

graphql_compiler/compiler/emit_sql.py
jecdohmann: fold_output looks like, e.g. `FoldScopeLocation(Location(('Animal',), None, 1), (('out', 'Animal_ParentOf'),), field=name)` I think I could write `intermediate_fold_output_name = fold_output.field` without getting any errors as long as no table in the schema has duplicate fields and as long as the same field isn't output twice in the fold. Those two cases would be validation errors right?

graphql_compiler/compiler/emit_sql.py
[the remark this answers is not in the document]
jecdohmann: I don't know if it's possible to use `with ... as ...` blocks with `isinstance` - it returns `AttributeError: __enter__`

How is just using regular variable assignment to refer to the results of the `isinstance` calls?

graphql_compiler/compiler/emit_sql.py
bojanserafimov: Yes, the `compiler_frontend` should complain if any of those two cases happens.

graphql_compiler/tests/test_compiler.py
jecdohmann: Yes, this test was to make debugging easier b/c you can see when you cross the edge that you've also changed types. I don't know if its necessary for any of my tests to be written into the other languages but if possible I'd like to put that into a different PR b/c I don't really understand Gremlin at all.

graphql_compiler/tests/test_compiler.py
jecdohmann: This isn't an issue with folds, it is a problem with what join descriptor is returned for a given edge. In particular, I think the from and to columns are reversed for this test case. I think @bojanserafimov  might know more about this

graphql_compiler/compiler/emit_sql.py
jecdohmann: I agree - I think the most recent PR addresses most of these issues. It seems like we would prefer the SQLFoldObject to be built to completion all at once, as opposed to built incrementally. I think that is possible for folds with no inner traversals, but won't generalize because different pieces of information will be stored in different parts of the IR, e.g. each traversal within a fold adds another join descriptor and table to our SQLFoldObject. Hopefully the current design represents a good middle ground. I don't think it will need substantial changes to handle the general, multi traversal fold case, but it looks as build-to-completion as it can.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+            raise NotImplementedError('Filters inside a fold are not implemented yet.')
+
+        sql_expression = predicate.to_sql(self._aliases,
+                                          self._current_alias)
         if self._is_in_optional_scope():
             sql_expression = sqlalchemy.or_(sql_expression,
                                             self._came_from[self._current_alias].is_(None))
         self._filters.append(sql_expression)
 
+    def get_join_descriptor(self, fold_scope_location):
```
chewselene: I'd put the word fold in the function name so it is clear that this function is only for folds since the current name makes it sound pretty general.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+        # basic info about the outer vertex and inner vertex
+        inner_vertex_alias = self._sql_schema_info.vertex_name_to_table[
+            self._ir.query_metadata_table.get_location_info(fold_scope_location).type.name
+        ].alias()
+        outer_vertex_alias = self._current_alias.alias()
+        join_descriptor = self.get_join_descriptor(fold_scope_location)
+
+        return inner_vertex_alias, outer_vertex_alias, join_descriptor
+
+    def get_fold_outputs(self, fold_scope_location, join_descriptor):
```
chewselene: Seems like this and other fold helper functions are intended for internal use only so should start with `_`

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+        # find outputs for this fold in self._all_folded_outputs and add to self._fold_outputs
+        if fold_scope_location.fold_path in self._all_folded_outputs:
+            for fold_output in self._all_folded_outputs[fold_scope_location.fold_path]:
+                # this is for folds with the same fold path but different query paths
+                if (fold_output.base_location, fold_output.fold_path) == (
+                        fold_scope_location.base_location, fold_scope_location.fold_path):
+                    if fold_output.field == '_x_count':
+                        raise NotImplementedError(u'_x_count not implemented in SQL')
+
+                    # making explicit label forces column to have label as opposed to anon_label
```
chewselene: nit: different tense used here

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+
+        return inner_vertex_alias, outer_vertex_alias, join_descriptor
+
+    def get_fold_outputs(self, fold_scope_location, join_descriptor):
+        """Generate output columns for innermost fold scope and add them to active SQLFoldObject."""
+        outputs = []
+        # find outputs for this fold in self._all_folded_outputs and add to self._fold_outputs
+        if fold_scope_location.fold_path in self._all_folded_outputs:
+            for fold_output in self._all_folded_outputs[fold_scope_location.fold_path]:
+                # this is for folds with the same fold path but different query paths
```
chewselene: nit: the grammatical structure feels off

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+                    outputs.append(
+                        sqlalchemy.func.array_agg(
+                            self._fold.inner_vertex_alias.c[fold_output.field]
+                        ).label(intermediate_fold_output_name)
+                    )
+
+        # the group by field is the unique identifier for the outer vertex
+        outputs.append(self._fold.outer_vertex_alias.c[join_descriptor.from_column])
+
+        return sorted(outputs, key=lambda x: x.name, reverse=True)
```
chewselene: It is pretty self explanatory here, but could be even more readable with a more descriptive name than x

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,26 +77,111 @@ def _find_columns_used(sql_schema_info, ir):
...
+        """Get a tuple containing edge and table info for the joins within the subquery."""
+        return self._join_info
+
+
+class UniqueAliasGenerator(object):
+    """Immutable class used to generate unique aliases for subqueries and columns."""
+
+    def __init__(self):
+        """Create unique subquery and column aliases by tracking counters."""
+        self._folded_output_count = 1
```
chewselene: Is this getting used?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+        # the group by field is the unique identifier for the outer vertex
+        outputs.append(self._fold.outer_vertex_alias.c[join_descriptor.from_column])
+
+        return sorted(outputs, key=lambda x: x.name, reverse=True)
+
+    def fold(self, fold_scope_location):
+        """Begin execution of a Fold Block."""
+        # This method creates the SELECT clause and GROUP BY for the fold subquery.
+        #
+        # SELECT will contain OuterVertex.uuid, a unique identifier for the
```
chewselene: Isn't the unique identifier dependent on the data? This seems to imply that it will always be uuid.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+        # to the inner vertex where the fold outputs are.
+        #
+        # NOTE: right now inner vertex denotes both the vertex which is folded on AND the vertex at
+        # which the outputs are placed. This is only accurate when there are no traversals inside of
+        # the fold. Once functionality for traversals inside folds is introduced the folded vertex
+        # will no longer necessarily be the one that has the outputs.
+        #
+        # SELECT will also contain an ARRAY_AGG for each columns labeled for output inside the fold
+        #
+        # SELECT will also contain a COUNT(*) if _x_count is referred to somewhere in the query
```
chewselene: Can you make it clear that this is a TODO since we don't have COUNT implemented yet?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+        # This method creates the SELECT clause and GROUP BY for the fold subquery.
+        #
+        # SELECT will contain OuterVertex.uuid, a unique identifier for the
+        # OuterVertex determined by the edge descriptor from the vertex immediately outside the fold
+        # to the inner vertex where the fold outputs are.
+        #
+        # NOTE: right now inner vertex denotes both the vertex which is folded on AND the vertex at
+        # which the outputs are placed. This is only accurate when there are no traversals inside of
+        # the fold. Once functionality for traversals inside folds is introduced the folded vertex
+        # will no longer necessarily be the one that has the outputs.
```
chewselene: Not sure if inner vertex is the best name since it won't make as much sense once traversals are added. Maybe output vertex instead of inner vertex? This could be something to change when we add traversals.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+
+        # 2. initialize fold object
+        self._fold = SQLFoldObject(inner_alias, outer_alias, join_info)
+
+        # 3. add output columns to fold object
+        outputs = self.get_fold_outputs(fold_scope_location, join_descriptor)
+
+        # 4. add group by clause to fold object
+        group_by = [self._fold.outer_vertex_alias.c[join_descriptor.from_column]]
+        self._fold.set_outputs(outputs)
```
chewselene: move this up to step 3

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+    def unfold(self):
+        """Complete the execution of a Fold Block."""
+        edge, outer_vertex = self._fold.join_info
+        subquery_from_clause = self.construct_fold_joins(edge)
+        fold_subquery = self.construct_fold_subquery(subquery_from_clause)
+
+        subquery_key = (self._fold_inner_vertex_location.base_location.query_path,
+                        self._fold_inner_vertex_location.fold_path)
+        # replaces the table first placed in the dict during the MarkLocation
+        # following the open of the fold scope
```
chewselene: Can you elaborate on why we need to replace the table? Or maybe just make a note in the MarkLocation function that the table is temporary and will need to be updated

graphql_compiler/compiler/emit_sql.py
pmantica1: The unique identifier should be the primary key of the SQAlchemy Table object. Even though the primary key of the SQLAlchemy Table may not be the primary key in the underlying table, according to schema generation, it will be a non-null and unique identifier of each row.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +338,182 @@ def start_global_operations(self):
...
+            self._from_clause,
+            fold_subquery,
+            onclause=(outer_vertex.c[edge.from_column] == fold_subquery.c[edge.from_column]),
+            isouter=True
+        )
+        # this line is necessary because the column referenced in the outputs belongs to the
+        # innermost occurrence of the outer vertex's table but we have now aliased and referenced
+        # that table again, so we need  to replace later substitutions of that column with ones
+        # taken from this new alias
+        self._aliases[(self._current_location.at_vertex().query_path, None)] = outer_vertex
```
bojanserafimov: Didn't we conclude that the `, None` part is unnecessary, as it always is `None`?

graphql_compiler/compiler/emit_sql.py
jecdohmann: Yes, I will make clear that the identifier is not necessarily the uuid column

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,26 +77,111 @@ def _find_columns_used(sql_schema_info, ir):
...
+    # Find outputs used for each fold path
+    for _, output_info in ir.query_metadata_table.outputs:
+        if isinstance(output_info.location, FoldScopeLocation):
+            fold_path = output_info.location.fold_path
+            folded_outputs.setdefault(fold_path, set()).add(output_info.location)
+    return folded_outputs
+
+
+class SQLFoldObject(object):
+    """Immutable object used to collect info for folds in order to ensure correct code emission."""
```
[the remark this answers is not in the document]
jecdohmann: I think #2 is a good idea and is similar to a design I had in mind for when we introduce multiple traversals. I was thinking that sort of conceptually belonged in the PR which handles the general case, but I think there's a good argument for doing that right now. If you think such a refactoring is better done now I will implement that now.

graphql_compiler/compiler/emit_sql.py
jecdohmann: @bojanserafimov see this thread on why the keys in the aliases dict are tuples. I was incorrect when I said we could get away without the second part of the tuple.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +460,118 @@ def start_global_operations(self):
...
 
+    def _get_fold_join_descriptor_from_current_classname(self, fold_scope_location):
+        """Convert a fold location and preceding classname to a join descriptor from schema."""
+        # collect edge information to join the fold subquery to the main selectable
+        edge_direction, edge_name = fold_scope_location.fold_path[0]
+        full_edge_name = '{}_{}'.format(edge_direction, edge_name)
+        # only works if fold scope location is the immediate child of self._current_classname
+        return self._sql_schema_info.join_descriptors[self._current_classname][full_edge_name]
+
+    def _get_fold_join_info(self, fold_scope_location):
```
chewselene: What's the reason to not encapsulate this in the SQLFoldObject?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +460,118 @@ def start_global_operations(self):
...
+        self._fold.visit_output_vertex(fold_vertex_alias, fold_scope_location, join_descriptor,
+                                       self._all_folded_outputs)
+
+    def unfold(self):
+        """Complete the execution of a Fold Block."""
+        fold_subquery, from_clause, outer_vertex = self._fold.end_fold(self._alias_generator,
+                                                                       self._from_clause,
+                                                                       self._current_alias)
+        subquery_key = (self._fold_vertex_location.base_location.query_path,
+                        self._fold_vertex_location.fold_path)
```
chewselene: key to what? Perhaps rename to `subquery_alias_key` or just add a comment like `generate a key for self._aliases that maps to the fold subquery's alias`.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +460,118 @@ def start_global_operations(self):
...
+
+        # Replaces the table first placed in the dict during the MarkLocation
+        # following the open of the fold scope. It needs to be replaced because while
+        # inside the fold scope, locations referring to subquery_key access columns
+        # from the output vertex table, while after the Unfold, those locations refer to the
+        # subquery constructed for the fold.
```
chewselene: nit: `Replaces` -> `Replace`, `open` -> `opening` or `start`

I think the second sentence is a bit wordy/confusing here. It might be clearer if you break it up into more than 1 sentence.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +460,118 @@ def start_global_operations(self):
...
+        self._aliases[subquery_key] = fold_subquery
+        self._from_clause = from_clause
+
+        # this line is necessary because the column referenced in the outputs belongs to the
+        # innermost occurrence of the outer vertex's table but we have now aliased and referenced
+        # that table again, so we need  to replace later substitutions of that column with ones
+        # taken from this new alias
```
chewselene: This comment is also wordy/confusing.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,19 +460,118 @@ def start_global_operations(self):
...
+            if self._in_fold else (self._current_location.query_path, None)
+        ] = self._current_alias
 
     def construct_result(self, output_name, field):
         """Execute a ConstructResult Block."""
-        self._outputs.append(field.to_sql(self._aliases, self._current_alias).label(output_name))
+        self._outputs.append(field.to_sql(
+            self._aliases,
+            self._current_alias
+        ).label(output_name))
```
chewselene: nit: remove unnecessary reformatting.

graphql_compiler/compiler/emit_sql.py
jecdohmann: It uses a lot of the folds from the CompilationState, including the schema, the query metadata table, and the current class name. On the other hand, it doesn't use any of the fields of the SQLFoldObject.  Seems like it might be more appropriate for that to be generated from CompilationState

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,244 @@ def _find_columns_used(sql_schema_info, ir):
...
+        # find outputs for this fold in all_folded_outputs and add to self._outputs
+        if fold_scope_location.fold_path in all_folded_outputs:
+            for fold_output in all_folded_outputs[fold_scope_location.fold_path]:
+                # distinguish between folds with the same fold path but different query paths
+                if (fold_output.base_location, fold_output.fold_path) == (
+                        fold_scope_location.base_location, fold_scope_location.fold_path):
+                    if fold_output.field == '_x_count':
+                        raise NotImplementedError(u'_x_count not implemented in SQL')
+
+                    # explicit label forces column to have label as opposed to anon_label
```
chewselene: nit: make this comment imperative.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,244 @@ def _find_columns_used(sql_schema_info, ir):
...
+                    intermediate_fold_output_name = 'fold_output_' + fold_output.field
+                    # add array aggregated output column to self._outputs
+                    self._outputs.append(
+                        sqlalchemy.func.array_agg(
+                            self.output_vertex_alias.c[fold_output.field]
+                        ).label(intermediate_fold_output_name)
+                    )
+
+        # this is the unique identifier for the outer vertex used to join to the outer table
```
chewselene: nit: make this comment imperative.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,244 @@ def _find_columns_used(sql_schema_info, ir):
...
+        self._outputs.append(self.outer_vertex_alias.c[join_descriptor.from_column])
+
+        return sorted(self._outputs, key=lambda column: column.name, reverse=True)
+
+    def visit_output_vertex(self,
+                            output_alias,
+                            fold_scope_location,
+                            join_descriptor,
+                            all_folded_outputs):
+        """Update output and group by columns when visiting the vertex containing output columns."""
```
chewselene: Comment might be off here. I don't see the group by getting updated.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,244 @@ def _find_columns_used(sql_schema_info, ir):
...
+                                               all_folded_outputs)
+
+    def visit_traversed_vertex(self, join_descriptor, from_table, to_table):
+        """Add a new join descriptor for every vertex traversed in the fold."""
+        self._join_info.append((join_descriptor, from_table, to_table))
+
+    def end_fold(self, alias_generator, from_clause, outer_from_table):
+        """Produce the final subquery and join it onto the rest of the query."""
+        # for now we only handle folds containing one traversal (i.e. join)
+        edge, from_alias, to_alias = self.join_info.pop()
```
chewselene: Should we raise a `NotImplementedError` if `self.join_info` has more than 1 join?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,244 @@ def _find_columns_used(sql_schema_info, ir):
...
+        self._used_columns = _find_columns_used_outside_folds(sql_schema_info, ir)
+        self._all_folded_outputs = _find_folded_outputs(ir)
 
         # Current query location state. Only mutable by calling _relocate.
         self._current_location = None  # the current location in the query. None means global.
         self._current_alias = None  # a sqlalchemy table Alias at the current location
-        self._aliases = {}  # mapping marked query paths to table _Aliases representing them
+        # mapping (query_path, fold_path) tuples to corresponding table _Aliases
+        # query_path and fold_path (if in a fold, otherwise None) both come from the location
+        # of the IR block associated with the current alias at the time that location is visited.
```
chewselene: nit: this is more than 1 sentence.

graphql_compiler/compiler/emit_sql.py
chewselene: I still think it might be better to fully encapsulate all of the fold logic in SQLFoldObject. What do you think @bojanserafimov ?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -239,15 +465,112 @@ def start_global_operations(self):
...
+        # inside the fold scope, columns accessed at the fold scope location refer
+        # to the folded vertex table, while after the Unfold, those column accesses refer to
+        # the subquery constructed for the fold.
+        self._aliases[subquery_alias_key] = fold_subquery
+        self._from_clause = from_clause
+
+        # This line is necessary because otherwise references to the outer vertex table
+        # following the Unfold, would select columns from the copy of that table
+        # found in the fold subquery, when in actuality they should refer to a totally new
+        # copy of the outer vertex table.
```
chewselene: nit: make this imperative.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,244 @@ def _find_columns_used(sql_schema_info, ir):
...
+        # table for vertex immediately outside fold
+        self._outer_vertex_alias = outer_vertex_table
+
+        # group by column for fold subquery
+        self._group_by = [self._outer_vertex_alias.c[join_descriptor.from_column]]
+
+        # List of 3-tuples:
+        #  edge: join descriptor for the columns used to join outer and output vertex
+        #  from_table: the table on the left side of the join
+        #  to_table: the table on the right side of the join
```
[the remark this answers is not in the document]
jecdohmann: All of your points are accurate, but the last two are only true in the case of no addl traversals inside the fold. If there are n traversals inside the fold there will be n+1 elements of the list, each of which constitutes its own join that must be inserted into the query. In the current implementation, only the last element (also the first, b/c there is only ever one element) of the list is used for unfold and that join object has from and to set to outer and output, resp.

When there are multiple traversals the first element of the list will be a join from the outer vertex to some other vertex, and the last element of the list will be a join from some other vertex to the output vertex, with a number of other joins in between traversing from outer to output.

Does that make sense?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,249 @@ def _find_columns_used(sql_schema_info, ir):
...
+        return sorted(self._outputs, key=lambda column: column.name, reverse=True)
+
+    def visit_output_vertex(self,
+                            output_alias,
+                            fold_scope_location,
+                            join_descriptor,
+                            all_folded_outputs):
+        """Update output columns when visiting the vertex containing output directives."""
+        self._output_vertex_alias = output_alias
+        self._outputs = self._get_fold_outputs(fold_scope_location,
```
chewselene: we should probably raise an error here if a user tries to call `visit_output_vertex` more than once.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,249 @@ def _find_columns_used(sql_schema_info, ir):
...
+                                               join_descriptor,
+                                               all_folded_outputs)
+
+    def visit_traversed_vertex(self, join_descriptor, from_table, to_table):
+        """Add a new join descriptor for every vertex traversed in the fold."""
+        self._join_info.append((join_descriptor, from_table, to_table))
+
+    def end_fold(self, alias_generator, from_clause, outer_from_table):
```
chewselene: in the doc for SQLFoldObject it says that "2.) at least one traversal, 3.) visiting the output vertex to collect the output columns" must occur before ending the fold. Let's assert that those things have happened at the start of `end_fold`.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -74,31 +77,249 @@ def _find_columns_used(sql_schema_info, ir):
...
+        self._all_folded_outputs = _find_folded_outputs(ir)
 
         # Current query location state. Only mutable by calling _relocate.
         self._current_location = None  # the current location in the query. None means global.
         self._current_alias = None  # a sqlalchemy table Alias at the current location
-        self._aliases = {}  # mapping marked query paths to table _Aliases representing them
+
+        # Dict mapping (some_location.query_path, some_location.fold_path) tuples to
+        # corresponding table _Aliases. some_location is either self._current_location
+        # or an open fold scope location.
```
chewselene: i still think this comment is a bit off. a fold scope location doesn't have a `query_path`. `some_location` is the `base_location` of an open fold scope location

graphql_compiler/compiler/emit_sql.py
```diff
@@ -107,11 +328,17 @@ def __init__(self, sql_schema_info, ir):
         self._outputs = []  # sqlalchemy Columns labelled correctly for output
         self._filters = []  # sqlalchemy Expressions to be used in the where clause
 
+        # SQLFoldObject used to collect fold information and guide output query
+        self._current_fold = None
```
chewselene: nit: can this comment go on the same line instead of above like the others?

PR #596, opened 2019-10-04
Add meta fields section to ReadTheDocs

pmantica1:
In this PR, I add meta fields to readthedocs.

### DIFF
--- a/docs/source/language_specification/schema_types.rst
+++ b/docs/source/language_specification/schema_types.rst
@@ -316,7 +316,7 @@ Meta fields
 Meta fields are fields that do not represent a property/column in the underlying vertex type.
 They are also an advanced compiler feature. To use them, one needs to understand how certain
 compiler directives work. Before reading this section please go over the :doc:`query directives
-<query-directives>`.
+<query_directives>`.
 
 \_\_typename
 ~~~~~~~~~~~~
--- a/docs/source/language_specification/schema_types.rst
+++ b/docs/source/language_specification/schema_types.rst
@@ -46,6 +46,7 @@ since we will immediately proceed to dissect the schema.
     scalar Decimal
 
     type Animal implements Entity {
+        _x_count: Int
         name: String
         alias: [String]
         birthday: Date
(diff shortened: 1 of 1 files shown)


docs/source/language_specification/schema_types.rst
```diff
@@ -298,3 +308,210 @@ and discarded.
 In this query, the :code:`out_Entity_Related` is of :code:`Entity` type.
 However, the query only wants to return results where the related entity
 is a :code:`Species`, which :code:`... on Species` ensures is the case.
+
+
+Meta fields
+-----------
+
+Meta fields are fields that do not represent a property/column in the underlying vertex type.
+They are also an advanced compiler feature. To use them, one needs to understand how certain
```
obi1kenobi: You've described what they aren't, but what are they? :)

Also, "certain compiler directives" is pretty vague, any reason for that? Why not just say "Before continuing, readers should familiarize themselves with the various query directives supported by the compiler." and throw in a link?

pmantica1: Well I was thinking that you need to understand @fold to understand how to use the _x_count meta field, (technically also @output), but the way you phrase it makes sense.

I cannot think of a better way to describe meta fields. I looked at the GraphQL spec and the GraphQL website for inspiration and couldn't find a definition of what a "meta fied" in GraphQL is.

obi1kenobi: Please be mindful that @-syntax is how GitHub allows you to tag users. So unless you surround your directives with backticks (e.g. `@fold`), you are tagging random GitHub users and probably causing notifications on their end.

For the best description, maybe we can sit down with @JulianGoetz and figure out how to describe this idea precisely yet succinctly.

pmantica1: Do you mind approving this PR though beforehand? This PR is for sure an improvement over existing documentation. I can open an issue for the lack of a precise definition of meta fields.

pmantica1:
Here is the issue: https://github.com/kensho-technologies/graphql-compiler/issues/604

Issue #605, opened 2019-10-23
Lack of documentation for @output_source


pmantica1:
It is not clear what the documentation

@output\_source
~~~~~~~~~~~~~~~

See the `Completeness of returned
results <#completeness-of-returned-results>`__ section for a description
of the directive and examples.

Constraints and Rules
^^^^^^^^^^^^^^^^^^^^^

-  May exist at most once in any given GraphQL query.
-  Can exist only on a vertex field, and only on the last vertex field
   used in the query.
-  Cannot be used within a scope marked :code:`@optional` or :code:`@fold`.

PR #606, opened 2019-10-23
Add query directive section in readthedocs

pmantica1:
The vast majority of this pull request is copy-paste. 

I do a couple of controversial things though: 

Remove hyperlinks and mentions of: 
- "Expanding `@optional` vertex fields" . I believe that this section should not be in readthedocs. It is an implementation detail not something that should be documented in readthedocs nor the README.rst. 
- "Performance Penalty", This is specific to OrientDB so we can mention the performance penalty of using `@optional` there.   

I also remove the following paragraph since it is again an implementation detail: 

"If the compiler is able to prove that a type coercion in the @fold` scope is actually a no-op, it may optimize it away. See the `Optional type_equivalence_hints compilation parameter` for more details." 

Also, the `Optional type_equivalence_hints compilation parameter`  section did not contain any more details.

I also mention that the documentation for  `@output_source` section is in development. As explained in this issue https://github.com/kensho-technologies/graphql-compiler/issues/605, the existing documentation for this directive is quite poor.  Also since most of the documentation for `@output_source`, was in a different section, e.g. `Completeness of returned results`, it was a bit hard to include it in this diff.  We can include it's documentation in a later pull request. 

Finally, I make some changes in the exposed schema so that the queries in this page only reference existing GraphQL types.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -57,8 +57,7 @@ html_theme = 'sphinx_rtd_theme'
 html_static_path = ['_static']
 
 # This specifies how many header levels below the title should we include in the table of contents
-# navigation bar. With a navigation depth of 2 we only include the title and the top header of
-# each page.
+# navigation bar.
 html_theme_options = {
-    'navigation_depth': 2,
+    'navigation_depth': 3,
 }
--- a/docs/source/language_specification/schema_types.rst
+++ b/docs/source/language_specification/schema_types.rst
@@ -47,8 +47,10 @@ since we will immediately proceed to dissect the schema.
 
     type Animal implements Entity {
         _x_count: Int
+        uuid: ID
         name: String
         alias: [String]
+        color: String
         birthday: Date
         net_worth: Decimal
         in_Animal_ParentOf: [Animal]
@@ -61,6 +63,7 @@ since we will immediately proceed to dissect the schema.
 
     type Food implements Entity {
         _x_count: Int
+        uuid: ID
         name: String
         alias: [String]
         in_Entity_Related: [Entity]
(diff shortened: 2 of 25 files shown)


docs/source/conf.py
```diff
@@ -57,8 +57,7 @@
 html_static_path = ['_static']
 
 # This specifies how many header levels below the title should we include in the table of contents
-# navigation bar. With a navigation depth of 2 we only include the title and the top header of
-# each page.
+# navigation bar.
 html_theme_options = {
-    'navigation_depth': 2,
+    'navigation_depth': 3,
```
pmantica1: I think the documentation simply looks better with this naviation_depth simply looks better.

docs/source/language_specification/schema_types.rst
```diff
@@ -70,9 +73,9 @@ since we will immediately proceed to dissect the schema.
 
     type Species implements Entity {
         _x_count: Int
+        uuid: ID
         name: String
         alias: [String]
-        description: String
```
pmantica1: This field is not used anywhere.

pmantica1:
Here is what the documentation looks like:
https://graphql-compiler.readthedocs.io/en/latest/language_specification/query_directives.html

bojanserafimov:
"Expanding @optional vertex fields" belongs in the readme because the user should know why queries are slow, and know that nested optionals are trouble. I agree that it doesn't belong here, but in a section describing OrientDB-specific exceptions.

bojanserafimov:
"Expanding @optional vertex fields" belongs in the "Performance Penalty" section. The user should know where to expect performance penalties, and why they're necessary. Might also be a good idea to call it "OrientDB Performance Penalties".

Do you plan to do this in this diff, or the one immediately following this one?

pmantica1:
> "Expanding @optional vertex fields" belongs in the "Performance Penalty" section. The user should know where to expect performance penalties, and why they're necessary. Might also be a good idea to call it "OrientDB Performance Penalties".
> 
> Do you plan to do this in this diff, or the one immediately following this one?

The one immediately following one.

PR #609, opened 2019-10-24
Sql query printing

bojanserafimov:
Implement query printing such that the result can be pasted into a database console and ran.

### DIFF
--- a/graphql_compiler/compiler/sqlalchemy_extensions.py
+++ b/graphql_compiler/compiler/sqlalchemy_extensions.py
@@ -52,7 +52,7 @@ def print_sqlalchemy_query_string(query, dialect):
     Returns:
         string that can be ran using sqlalchemy.sql.text(result)
     """
-    class BindparamCompiler(sqlalchemy.dialects.mssql.base.MSSQLCompiler):
+    class BindparamCompiler(dialect.statement_compiler):
         def visit_bindparam(self, bindparam, **kwargs):
             # A bound parameter with name param is represented as ":param". However,
             # if the parameter is expanding (list-valued) it is represented as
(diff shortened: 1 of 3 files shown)


graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -40,3 +40,26 @@ def not_contains_operator(collection, element):
...
+    """Return a string form of the parameterized query.
+
+    Args:
+        query: sqlalchemy.sql.selectable.Select
+        dialect: sqlalchemy.engine.interfaces.Dialect
+
+    Returns:
+        string that can be ran using sqlalchemy.sql.text(result)
+    """
+    class LiteralCompiler(sqlalchemy.dialects.mssql.base.MSSQLCompiler):
```
pmantica1: Can you either make sure this works for dialects other than mssql, raise an assertion if this is not a mssql dialect,  or change the docstring of this function?

graphql_compiler/tests/test_sqlalchemy_extensions.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_print_query_mssql_list_argument(self):
+        animal = self.sql_schema_info.vertex_name_to_table['Animal'].alias()
+        query = sqlalchemy.select([animal.c.name]).where(
+            animal.c.name.in_(sqlalchemy.bindparam('names', expanding=True))
+        )
+        text = print_sqlalchemy_query_string(query, mssql.dialect())
+        expected_text = '''
+             SELECT [Animal_1].name
+             FROM db_1.schema_1.[Animal] AS [Animal_1]
+             WHERE [Animal_1].name IN :names
```
pmantica1: Have you tried seeing if you could fill in the parameter of this query by using pymssql or pydobc and actually running it?

bojanserafimov: yes

graphql_compiler/compiler/sqlalchemy_extensions.py
bojanserafimov: Made it dialect-agnostic

Issue #555, opened 2019-09-11
`@fold` support in MSSQL


jecdohmann:
@chewselene @obi1kenobi  @pmantica1 let me know your thoughts

MSSQL doesn't have an array aggregation operator so folds need to be done using alternative methods (mostly manipulating the string results of the `FOR XML PATH` operator).

**Edited to use less common delimiters and to escape nulls unambiguously**
**Edited to defer conversion of auto-escaped characters back to their original form until a post processing step SEE BELOW**

The generic structure is:

```
SELECT
    OuterVertex.property AS alias_1,
    COALESCE(STUFF(
        (SELECT 
           '~|*' + REPLACE(tbl_inner.name, '|', '||')
        FROM FoldedVertex tbl_inner
        WHERE (tbl_inner.outEdge = tbl_outer.key)
        FOR XML PATH ('')),
        1,
        3,
        ''), '!|#null!|#') AS alias_2
FROM OuterVertex AS tbl_outer
```

The XML PATH is needed to iterate over each FoldedVertex.

_x_count is captured by adding a nested subquery and the joining:

```
SELECT
    OuterVertex.property AS alias_1,
   COALESCE(STUFF(
        (SELECT 
           '~|*' + REPLACE(tbl_inner.name, '|', '||')
        FROM FoldedVertex tbl_inner
        WHERE (tbl_inner.outEdge = tbl_outer.key)
        FOR XML PATH ('')),
        1,
        3,
        ''), '!|#null!|#') AS alias_2,
    COALESCE(inner_1._x_count, 0) AS _x_count <-- here
FROM OuterVertex AS tbl_outer
LEFT JOIN ( <-- here
  SELECT 
    tbl_inner.outEdge AS group_id,
    COUNT(tbl_inner.property) AS  _x_count
  FROM FoldedVertex tbl_inner
  GROUP BY tbl_inner.group_id
) AS inner_0
ON tbl_outer.key = inner_0.outEdge
```

Applying a filter to the x_count is accomplished by putting a WHERE clause at the very outside of the query (this ensures that it is logically imposed last)

```
SELECT
    OuterVertex.property AS alias_1,
    COALESCE(STUFF(
        (SELECT 
            '~|*' + REPLACE(tbl_inner.name, '|', '||')
        FROM FoldedVertex tbl_inner
        WHERE (tbl_inner.outEdge = tbl_outer.key)
        FOR XML PATH ('')),
        1,
        3,
[...]

PR #613, opened 2019-10-24
Implement `@fold` for MSSQL 2014

jecdohmann:
Implemented based on the design for MSSQL folds given in #555 slightly modified to accord better with PR #595. 

The generic structure has been changed to

```
SELECT
    OuterVertex.property AS alias_1,
    folded_subquery_1.fold_output_column AS alias_2
FROM OuterVertex AS tbl_outer_0
LEFT OUTER JOIN (
    SELECT
       OuterVertex.primary_key AS primary_key,
       COALESCE(
           (SELECT
               '|' + COALESCE(
                 REPLACE(
                   REPLACE(
                     REPLACE(inner_table.agg_column, '^', '^e'), 
                   '~', '^n'),     
                 '|', '^d'),
            '~')
            FROM Table inner_table
            WHERE (inner_table.foreign_key = tbl_outer_1.primary_key) 
            FOR XML PATH ('')),
       '') AS fold_output_column,
    FROM OuterVertex AS tbl_outer_1
) AS folded_subquery_1
ON folded_subquery_1.primary_key = OuterVertex.primary_key
```

When the downstream client receives the output of this query it will need to:

1. Split the `|` delimited list into an array
2. Replace all `~` with the appropriate null-type
3. Undo all of the encodings by replacing `^d` with `|`, `^n` with `~` and finally `^e` with `^`.

The query above is analogous to the equivalent generic query from PostgreSQL fold #595, wherein the folded subquery is identified by the outer vertex's primary key. 

```
SELECT
    OuterVertex.primary_key AS alias_1,
    COALESCE(folded_subquery_1.fold_output_column, ARRAY[]::VARCHAR[]) AS alias_2,
FROM OuterVertex
LEFT JOIN (
  SELECT 
      array_agg(Table.agg_column) AS fold_output_column,
      OuterVertex.primary_key,
  FROM OuterVertex
  JOIN Table 
  ON OuterVertex.primary_key = Table.foreign_key
  GROUP BY OuterVertex.primary_key
) AS folded_subquery_1 
ON OuterVertex.primary_key = folded_subquery_1.primary_key
```

The key difference is that the final join from the penultimate vertex to the output vertex in the Postgres version is missing in the  MSSQL version. Additionally, the GROUP BY on the outer vertex primary key is missing from the MSSQL version. Those two features are missing from the MSSQL implementation because the logic of grouping by the outer vertex primary key and aggregating is accomplished by calling XML PATH to collect the columns into an XML string and then filter the columns collected using a WHERE clause.

The code in SQLFoldObject.end_fold(...) reflects this difference by not incorporating the last join in the join_info list.

This PR also refactors the SQLFoldObject somewhat to make it more general and more amenable to implementing traversals within fold.

NOTE: Ran `test_backends_integration.test_fold_basic` and it correctly returns `child_names: {'child_names': 'Animal 1~|*Animal 2~|*Animal 3'}]` for Query 2. Full integration testing will be performed in a separate PR.
NOTE: The output format above is outdated ^, delimiters now use `~`. Full integration testing will be performed separately.


graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +252,55 @@ def _set_group_by(self, group_by):
...
+                join_clause,
+                to_alias,
+                onclause=(from_alias.c[edge.from_column] == to_alias.c[edge.to_column])
+            )
         return join_clause
 
     def _construct_fold_subquery(self, subquery_from_clause):
         """Combine all parts of the fold object to produce the complete fold subquery."""
-        return sqlalchemy.select(
+        select_stmt = sqlalchemy.select(
```
chewselene: don't use abbreviations since it makes it harder for the user to easily understand what you mean.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -266,7 +348,7 @@ def visit_traversed_vertex(self, join_descriptor, from_table, to_table):
         """Add a new join descriptor for every vertex traversed in the fold."""
         self._join_info.append((join_descriptor, from_table, to_table))
 
-    def end_fold(self, alias_generator, from_clause, outer_from_table):
+    def end_fold(self, alias_generator, from_clause, outer_from_tbl):
```
chewselene: don't undo this change

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,49 @@ def _find_folded_outputs(ir):
...
+        )
+    return compiler.visit_label(element, **kw)
+
+
+def agg_table(output_column, where):
+    """Produce an MSSQL 2014-style XML PATH-based aggregation subquery.
+
+    Args:
+        output_column: Column values being aggregated with XML PATH
+        where: Predicate used to filter
```
chewselene: what are the types of these arguments?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,49 @@ def _find_folded_outputs(ir):
...
+    Returns:
+        An SQLAlchemy Selectable corresponding to an XML PATH subquery
+        which aggregates the values of the output_column from the output vertex.
+        Joining from the preceding vertex to the correct output vertex
+        is performed with the WHERE clause.
+    """
+    col = (u'~|*' + func.REPLACE(output_column, u'|', u'||'))
+    col = XMLPathBinaryExpression(col.left, col.right, col.operator)
```
chewselene: don't use abbreviations

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,49 @@ def _find_folded_outputs(ir):
...
+    """Suppress labeling when compiling XML PATH subqueries, otherwise compile as usual."""
+    if type(element.element) == XMLPathBinaryExpression:
+        # this indicates that no label should be inserted for this element
+        kw.update(
+            within_columns_clause=False
+        )
+    return compiler.visit_label(element, **kw)
+
+
+def agg_table(output_column, where):
```
chewselene: this function could use some more documentation. it is hard to tell why all these operations are being applied.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +252,55 @@ def _set_group_by(self, group_by):
...
-    def _construct_fold_joins(self, edge, from_alias, to_alias):
+    def _construct_fold_joins(self):
         """Use the edge descriptors to create the join clause between the tables in the fold."""
-        join_clause = sqlalchemy.join(
-            from_alias,
-            to_alias,
-            onclause=(from_alias.c[edge.from_column] == to_alias.c[edge.to_column])
-        )
+        _, from_alias, _ = self.join_info[-1]
+        join_clause = from_alias
```
chewselene: why not unpack this as `join_clause`?

graphql_compiler/tests/test_compiler.py
```diff
@@ -4181,6 +4182,93 @@ def test_fold_on_two_output_variables(self):
         compare_input_metadata(self, test_data.expected_input_metadata,
                                result.input_metadata)
 
+    def test_fold_on_two_output_variables_mssql(self):
```
chewselene: I think either the MSSQL tests should be broken out into their own `unittest.TestCase` or have their check incorporated into `check_test_data`. Seems odd to have separated tests for the same input within `CompilerTests` when all the other backends are incorporated into a single test per input.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -169,9 +215,10 @@ def __init__(self, outer_vertex_table, join_descriptor):
         #  from_table: the table on the left side of the join
         #  to_table: the table on the right side of the join
         self._join_info = []
-
         self._outputs = []  # output columns for fold
 
+        self._is_mssql2014 = is_mssql2014  # indicate results should be compatible with MSSQL2014
```
chewselene: This function has a bunch of unnecessary formatting changes, which should be reverted. `is_mssql2014` needs an entry in the args section.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +252,55 @@ def _set_group_by(self, group_by):
...
+            # is moved into the WHERE clause of the SELECT ... FOR XML PATH('')
+            self.join_info.pop()
+        while len(self.join_info) > 0:
+            # more recent joins are at the end of the list
+            edge, from_alias, to_alias = self.join_info.pop()
+            join_clause = sqlalchemy.join(
+                join_clause,
+                to_alias,
+                onclause=(from_alias.c[edge.from_column] == to_alias.c[edge.to_column])
+            )
```
chewselene: can you add a comment for what this does?

graphql_compiler/schema_generation/sqlalchemy/__init__.py
```diff
@@ -7,7 +7,9 @@
 
 
 def get_sqlalchemy_schema_info_from_specified_metadata(
-    vertex_name_to_table, direct_edges, dialect, class_to_field_type_overrides=None
+    vertex_name_to_table, direct_edges,
+    dialect, class_to_field_type_overrides=None,
+    is_mssql_2014=False
```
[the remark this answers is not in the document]
jecdohmann: I agree about factoring the flag into a separate PR, good idea.

I don't think it's necessarily the best idea to make the flag be as general as `supports_array_agg` because this code does very specific stuff that is only supported in MSSQL, namely XML PATH. In theory we could name the flag `is_mssql`, but we kind of have broken the correspondence between dialect and output due to the whole issue with the dialect field having been set to mssql the whole time. Not sure what the best solution is here.

jecdohmann:
> We probably want a more general flag than `is_mssql2014`. Maybe something like `supports_array_agg` instead?
> 
> This won't pass the integration tests as is since post processing has to occur to turn the results back into a list, but would be a good idea to manually check that the results are as you expect.

I replied under Bojan's comment re: naming the flag. W.r.t. integration tests, see my note in the PR description. Do you want me to formally add integration tests to this PR?

graphql_compiler/schema_generation/sqlalchemy/__init__.py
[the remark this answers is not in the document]
jecdohmann: EDIT: I see that SQL SERVER also has XML PATH syntax. I will need to check whether my queries correctly execute on SQL Server 1.0 before making the flag more general.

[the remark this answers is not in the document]
bojanserafimov: Yet another idea is passing a `dialect_specific_flags` dict argument, and validating that the arguments apply to the dialect specified.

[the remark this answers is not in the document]
obi1kenobi: Broadly, I think we want two things here — h/t @lodrion for the fruitful discussion where these ideas crystallized:
- We want the flexibility of feature flags, so we can easily toggle between different implementations (e.g. specifically for `@fold`, we can choose to emit `array_agg` vs JSON-based tricks vs `XML PATH`).
- Simultaneously, we want users to be able to use the compiler with minimal configuration and getting-started cost. This includes not worrying about what values to use for all the flags.

I think in the long run, our architecture might need to look something like this:
- For a given query, we compute the set of things it requires the backend to support (since not all backends support all operations). This is stuff like `@fold` or `@recurse` or `@optional + nested traversals`.
- Each backend has a set of feature flags that could be simple booleans (`recurse_supported == True`) or a choice among many options (`fold_strategy == None` if not supported, or `"array_agg" / "xml_path" ...` otherwise). For a given query, we use this info to check that the backend in question has the ability to support the requested operations.
- We have feature flag presets for various database versions, which we allow users to use / amend / completely ignore and build their own. E.g. for Postgres 11, `fold_strategy` is set to `"array_agg"`, and for MSSQL 2014 it's set to `"xml_path"`. If users set a flag to an invalid value, the backend errors out — definitely advanced, use-with-care functionality that way.
- In the long term, we offer functionality to autodetect the database version and select the dialect automatically when possible (e.g. across different versions of SQL databases).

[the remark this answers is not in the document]
jecdohmann: Hi, the most recent push settles for the short-term solution of simply having postgres and mssql side by side with mssql being default and the desired target determined by the dialect field of the sql schema info. 

I like the idea of, in the long-run, for each choice point in the code, having a flag that toggles the behavior, and bundling those flags into specific dialects/versions on the basis of how those dialects set the flags.

Would you like to see something like that integrated into this PR, or deferred until we have a need for several more meaningful flags?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +259,60 @@ def _set_group_by(self, group_by):
...
         ).select_from(
             subquery_from_clause
-        ).group_by(
-            *self.group_by
+        )
+
+        if self._dialect == 'mssql':
+            # mssql 2014 doesn't rely on a group by
+            return select_statement
+        else:
```
obi1kenobi: Surely this is an `elif` for some dialects, and `raise NotImplementedError()` for everything that we don't recognize? Right now, this code claims that every SQL database out there except MSSQL would use a `GROUP BY` here, which seems like a stretch.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +259,60 @@ def _set_group_by(self, group_by):
...
+            return select_statement.group_by(
+                *self.group_by
+            )
+
+    def _append_default_column(self, intermediate_fold_output_name, fold_output):
+        """Select an array_agg of the fold output field, labeled as requested."""
+        self._outputs.append(
```
obi1kenobi: Rather than state-mutating functions like this, why not restructure this into a stateless function that does the `sqlalchemy.func.array_agg().label()` magic, and then take its output and append it to `self._outputs` in the caller function?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +259,60 @@ def _set_group_by(self, group_by):
...
+            sqlalchemy.func.array_agg(
+                self.output_vertex_alias.c[fold_output.field]
+            ).label(intermediate_fold_output_name)
+        )
+
+    def _append_mssql_column(self, intermediate_fold_output_name, fold_output, last_join):
+        """Select the MSSQL2014 XML PATH agg of the fold output field, labeled as requested."""
+        edge, from_alias, to_alias = last_join
+        self._outputs.append(
```
obi1kenobi: Same question as above on making this a stateless function.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +259,60 @@ def _set_group_by(self, group_by):
...
+        """Select an array_agg of the fold output field, labeled as requested."""
+        self._outputs.append(
+            sqlalchemy.func.array_agg(
+                self.output_vertex_alias.c[fold_output.field]
+            ).label(intermediate_fold_output_name)
+        )
+
+    def _append_mssql_column(self, intermediate_fold_output_name, fold_output, last_join):
+        """Select the MSSQL2014 XML PATH agg of the fold output field, labeled as requested."""
+        edge, from_alias, to_alias = last_join
```
obi1kenobi: It's extremely non-obvious what `last_join` is supposed to be, both as a type (other than a 3-tuple) as well as semantically.

graphql_compiler/compiler/expressions.py
```diff
@@ -719,7 +719,7 @@ def to_cypher(self):
 
         return template.format(mark_name=mark_name, field_name=field_name)
 
-    def to_sql(self, aliases, current_alias):
+    def to_sql(self, aliases, current_alias, dialect):
```
obi1kenobi: I'd move `dialect` to the first position in all of these.

graphql_compiler/schema/schema_info.py
```diff
@@ -80,6 +80,9 @@
     # (e.g. sqlalchemy.dialects.mssql.dialect()).
     'dialect',
 
+    # boolean indicating whether we are compiling for MSSQL-2014
+    'mssql2014',
```
obi1kenobi: Is this still used?

graphql_compiler/schema_generation/sqlalchemy/__init__.py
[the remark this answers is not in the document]
obi1kenobi: Probably deferred is fine and we can stick with the dialect for the sake of this PR.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,53 @@ def _find_folded_outputs(ir):
...
+        which aggregates the values of the output_column from the output vertex.
+        Joining from the preceding vertex to the correct output vertex
+        is performed with the WHERE clause.
+    """
+    xml_column = (u'~|*' + func.REPLACE(output_column, u'|', u'||'))
+
+    # as a custom type, you can't directly construct an XMLPathBinaryExpression from plain text
+    xml_column = XMLPathBinaryExpression(xml_column.left, xml_column.right, xml_column.operator)
+
+    # REPLACEs to undo autoconversion of HTML special characters to entity reference shorthand
```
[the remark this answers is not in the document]
jecdohmann: I forgot to carry over this comment from #555 (in part because I forgot the details - this goes to show you how scary it really is!). I'm not sure if we ever debated the costs/benefits of this design trade-off.

```
I have edited the plan to defer all replacement of auto-escaped chars to their original form until python post processing. The reason why is that in addition to convering <,>,& to &lt;, etc. there is also a whole host of special characters are converted to entity numbers, e.g. &#0xHEX; and it would be best to capture those with one single regex - easier done in python than in mssql. A bit more info is here https://www.w3schools.com/html/html_entities.asp

An important design alternative here is that it is possible to avoid the auto-escaping if we use XML PATH(''), TYPE).value('.', 'VARCHAR(max)') - the only downside is that this form isn't robust to inputs containing certain unicode characters and would need to be both preprocessed to pass through the TYPE function and postprocessed to undo that preprocessing. b/c it is robust to all character inputs (to my knowledge), the method using just XML PATH('') only needs postprocessing.
```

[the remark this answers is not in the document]
jecdohmann: If we defer all of the substitution of the hex characters to post processing, it might be smart to do so for the other characters as well. Then I can just make a list on this method of all the reverse mappings that need to be done and what order to do them in.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,83 @@ def _find_folded_outputs(ir):
...
+class XMLPathBinaryExpression(BinaryExpression):
+    """Special override of BinaryExpression used to trigger `compile_xmlpath` during compile.
+
+    This type of binary expression is used to describe the Selectable selected inside the
+    XML PATH subquery. Using an XMLPathBinaryExpression forces the compiler to produce that
+    Selectable without aliasing (aka labeling) it. This prevents the string resulting from
+    the XML PATH from having extraneous XML tags based on the alias given to the column.
+    """
+
+    pass
```
obi1kenobi: Delete this `pass`, it's unnecessary. It only has to get added when there is nothing at a deeper indentation level after a `:`. In this case, the docstring is at the deeper indentation level, so the `pass` doesn't do anything.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,83 @@ def _find_folded_outputs(ir):
...
+        where: SQLAlchemy BinaryExpression, predicate used to filter
+
+    Returns:
+        An SQLAlchemy Selectable corresponding to an XML PATH subquery
+        which aggregates the values of the output_column from the output vertex.
+        Empty arrays are represented as '', null values are represented as '~', array
+        entries are delimited using '|'.
+
+        All occurrences of '^', '~', and '|' in the original string values are
+        replaced with '^e', '^n', and '^d', resp.
```
obi1kenobi: Give some intuition about the choice of `e/n/d` here -- it's a fine choice but I was originally expecting `caret/tilde/pipe` -> `c/t/p`. Assuming it's for `escape/null/delimiter`, just write that down because we won't remember.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,83 @@ def _find_folded_outputs(ir):
...
+        Post-processing must split on '|', convert '~' to None, and both the encoding above
+        and the XML reference entity encoding auto performed by XML PATH. In particular, undo
+        '^d' -> '|', '^e' -> '^', '^n' -> '~', '&amp;' -> '&', '&gt;' -> '>', '&lt;' -> '<',
+        '&0xHEX;' -> u'0xHEX'
+
+        Joining from the preceding vertex to the correct output vertex
+        is performed with the WHERE clause.
+    """
+    # delimit elements in the array using '|'
+    xml_column = (u'|' + func.COALESCE(  # denote null values with '~'
```
obi1kenobi: This is very difficult to read. Use intermediate variables (maybe even the same variable each time) and assign each `func.REPLACE()` and `func.COALESCE()` to it in steps.

graphql_compiler/compiler/emit_sql.py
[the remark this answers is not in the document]
obi1kenobi: Btw I think https://docs.python.org/3/library/html.html#html.unescape should do the right conversions from XML entity encoding into proper text.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,25 +289,66 @@ def _set_group_by(self, group_by):
...
-    def _construct_fold_joins(self, edge, from_alias, to_alias):
+    def _construct_fold_joins(self):
         """Use the edge descriptors to create the join clause between the tables in the fold."""
-        join_clause = sqlalchemy.join(
-            from_alias,
-            to_alias,
-            onclause=(from_alias.c[edge.from_column] == to_alias.c[edge.to_column])
-        )
+        # Start the join clause with the from_table of the first join descriptor
+        _, join_clause, _ = self.join_info[0]
```
obi1kenobi: How do we know there is at least one thing in `join_info`? What happens if there is more than one?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -144,10 +226,11 @@ class SQLFoldObject(object):
     # ON ...
     # GROUP BY OuterVertex.SOME_COLUMN
 
-    def __init__(self, outer_vertex_table, join_descriptor):
+    def __init__(self, dialect, outer_vertex_table, join_descriptor):
         """Create an SQLFoldObject with table, type, and join information supplied by the IR.
 
         Args:
+            dialect: string passed in from the schema indicating the target dialect
```
obi1kenobi: Are we sure this should be a string and not a SQLAlchemy dialect object? What are the pros/cons of each choice?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +259,60 @@ def _set_group_by(self, group_by):
...
+        _, join_clause, _ = self.join_info[-1]
+
+        # Starting at the first from_table, join traversed vertices until the output vertex is
+        # reached.
+        #
+        # MSSQL doesn't use the last JOIN so we pop it off before construction the JOIN clause
+        if self._dialect == 'mssql':
+            # Move the final join predicate (from the vertex preceding the output to
+            # the output vertex) into the WHERE clause of the SELECT ... FOR XML PATH('')
+            self.join_info.pop()
```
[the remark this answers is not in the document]
obi1kenobi: I'm confused here. We grab stuff from `self.join_info[0]` on line 295, but here we say that we don't use that join in MSSQL even though we're clearly using the `join_clause` value below. This seems like a bug.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,25 +289,66 @@ def _set_group_by(self, group_by):
...
+        # MSSQL doesn't use the JOIN for the very last traversal,
+        # so we pop it off before construction the JOIN clause
+        if self._dialect == 'mssql':
+            # Move the final join predicate (from the vertex preceding the output to
+            # the output vertex) into the WHERE clause of the SELECT ... FOR XML PATH('')
+            self.join_info.pop()
+        while len(self.join_info) > 0:
+            # joins from earlier in the chain of traversals are at the end of the list
+            # b/c joins are appended in the order they are traversed
+            edge, from_alias, to_alias = self.join_info.pop(0)
```
obi1kenobi: Here we seem to be popping from the front of the list, which makes this loop unnecessarily be `O(n^2)`. Can't we just iterate over the list in order to construct the `join_clause` object and then just set the reference to an empty list? Isn't that `O(n)`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,25 +289,66 @@ def _set_group_by(self, group_by):
         """Set output columns for the fold object."""
         self._group_by = group_by
 
-    def _construct_fold_joins(self, edge, from_alias, to_alias):
+    def _construct_fold_joins(self):
```
obi1kenobi: Are we supposed to be mutating `self.join_info` here? Can we guarantee that this is safe to do? Do we gain anything by mutating it?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,25 +289,66 @@ def _set_group_by(self, group_by):
...
+                *self.group_by
+            )
+        else:
+            raise NotImplementedError(
+                u'Fold only supported for MSSQL and '
+                u'PostgreSQL, dialect was set to {}'.format(self._dialect)
+            )
+
+    def _get_default_column(self, intermediate_fold_output_name, fold_output):
+        """Select an array_agg of the fold output field, labeled as requested."""
```
obi1kenobi: This function name and description don't seem like they match each other, and don't really describe what their purpose is. Based on those two data points, I have no idea when I'd ever want to call this function.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,25 +289,66 @@ def _set_group_by(self, group_by):
...
+        return sqlalchemy.func.array_agg(
+            self.output_vertex_alias.c[fold_output.field]
+        ).label(intermediate_fold_output_name)
+
+    def _get_mssql_column(self, intermediate_fold_output_name, fold_output, last_join):
```
obi1kenobi: Please choose a better name here as well — and do a pass over the rest of the PR and consider if there are any other names you could improve.

graphql_compiler/compiler/emit_sql.py
obi1kenobi: This is still a concern. We need a more detailed docstring in addition to a better function name.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -238,11 +363,18 @@ def _get_fold_outputs(self, fold_scope_location, join_descriptor, all_folded_out
...
-                    )
+                    if self._dialect == 'mssql':
+                        self._outputs.append(
+                            self._get_mssql_column(
+                                intermediate_fold_output_name,
+                                fold_output,
+                                self.join_info[-1]  # output vertex was most recently traversed
+                            )
+                        )
+                    else:
```
obi1kenobi: Is this really an `else` or is it more of an `elif:` followed by `else: raise NotImplemented / AssertionError`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -238,11 +363,18 @@ def _get_fold_outputs(self, fold_scope_location, join_descriptor, all_folded_out
...
-                        sqlalchemy.func.array_agg(
-                            self.output_vertex_alias.c[fold_output.field]
-                        ).label(intermediate_fold_output_name)
-                    )
+                    if self._dialect == 'mssql':
+                        self._outputs.append(
+                            self._get_mssql_column(
+                                intermediate_fold_output_name,
+                                fold_output,
+                                self.join_info[-1]  # output vertex was most recently traversed
```
obi1kenobi: How do we know `self.join_info` has anything in it? For example, I remember that `_construct_fold_joins()` pops everything off of that field — how do we know that function hasn't already been called?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -276,10 +408,11 @@ def end_fold(self, alias_generator, from_clause, outer_from_table):
         # for now we only handle folds containing one traversal (i.e. join)
         if len(self.join_info) > 1:
```
obi1kenobi: Why does this line use `self.join_info` whereas just a few lines above we use `self._join_info`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -570,15 +708,27 @@ def mark_location(self):
...
             self._from_clause).where(sqlalchemy.and_(*self._filters))
 
+    @property
+    def sql_schema_info(self):
+        """Get the SQLALchemySchemaInfo for the current query."""
+        return self._sql_schema_info
+
+    @property
+    def dialect(self):
+        """Get whether the current query is written for MSSQL."""
```
obi1kenobi: This is not an accurate description of this property.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -521,7 +659,7 @@ def fold(self, fold_scope_location):
         ][full_edge_name]
 
         # 3. initialize fold object
-        self._current_fold = SQLFoldObject(outer_alias, join_descriptor)
+        self._current_fold = SQLFoldObject(self.dialect, outer_alias, join_descriptor)
```
obi1kenobi: Why do we sometimes use `self.dialect` and sometimes use `self._dialect`? What is the purpose of making `self.dialect` a property?

graphql_compiler/tests/test_compiler.py
```diff
@@ -4237,12 +4318,71 @@ def test_fold_same_edge_type_in_different_locations(self):
                   schema_1."Animal" AS "Animal_5"
                 JOIN schema_1."Animal" AS "Animal_6" ON "Animal_5".uuid = "Animal_6".parent
                 GROUP BY
-                  "Animal_5".uuid
+                    "Animal_5".uuid
```
obi1kenobi: Why the whitespace-only change? This is not what the other `GROUP BY` clauses look like. Overall, the entire query uses a mix of 2-space and 4-space indentation, and changing just this one location doesn't make sense. I'd either fix the whole thing or not change anything in it.

graphql_compiler/tests/test_compiler.py
```diff
@@ -4237,12 +4318,71 @@ def test_fold_same_edge_type_in_different_locations(self):
...
+              LEFT OUTER JOIN(
+                SELECT
+                  [Animal_2].uuid AS uuid,
+                  coalesce((
+                    SELECT
+                      :coalesce_1 + coalesce(
+                        REPLACE(
+                          REPLACE(
+                            REPLACE([Animal_3].name, :REPLACE_1, :REPLACE_2),
+                            :REPLACE_3,
```
obi1kenobi: Is this right? Do we expect these to be query parameters? I don't think they get added to the query parameter list during execution, I would have expected them to be literal characters instead.

graphql_compiler/compiler/emit_sql.py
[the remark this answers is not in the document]
jecdohmann: I'll put this link in a comment on how post-processing should work.

graphql_compiler/compiler/emit_sql.py
jecdohmann: Yeah so we could imagine changing `dialect` to the actual compiler used, then doing some sort of check like `dialect == sqlalchemy.dialects.mssql.dialect()`, but for some reason different calls to dialect() aren't equal to one another :( 

We could also try something like `isinstance(dialect,MSDialect_pyodbc)`. A pro of this is its immutable, its a constant provided by SQLAlchemy, and it is not liable to change much between different releases of SQLAlchemy.

The pros of the string way of doing things is that it is readable, and that we are going to need to use some constant field of the SQLALchemy compiler object since different instances of the same compiler dialect aren't equal. A downside might be that strings are mutable, and SQLAlchemy might change how they represent different dialects some day so that they no longer use string names, but some other identifier. 

I will change this to use the type check instead of passing strings around. Let me know if you object

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,83 @@ def _find_folded_outputs(ir):
...
+    """
+    # delimit elements in the array using '|'
+    xml_column = (u'|' + func.COALESCE(  # denote null values with '~'
+        func.REPLACE(  # replace all occurrences of '|' in the original with '|d'
+            func.REPLACE(  # replace all occurrences of '~' in the original with '|n'
+                func.REPLACE(  # replace all occurrences of '^' in the original with '|e'
+                    output_column, '^', '^e'
+                ), '~', '^n'
+            ), '|', '^d'
+        ), '~'
```
bojanserafimov: Can you see if using literals here instead of strings produces literals in the query?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,81 +354,199 @@ def _set_group_by(self, group_by):
...
+                                 u'No traversed vertices visited.'.format(self))
 
         # for now we only handle folds containing one traversal (i.e. join)
-        if len(self.join_info) > 1:
-            raise NotImplementedError('Folds containing multiple traversals are not implemented.')
-        edge, from_alias, to_alias = self.join_info.pop()
+        if len(self._traversal_descriptors) > 1:
+            raise NotImplementedError(u'Not implemented error encountered for fold: {}'
+                                      u'Folds containing multiple traversals are not '
+                                      u'implemented.'.format(self))
```
pmantica1: Do not start a `NotImplementedError`  with "Not implemented error".  

For `AssertionError`, I think that it is ok that you include as much information as you'd like. They indicate a bug and we need as much information as possible to debug. 

However, `NotImplementedError` are errors that are likely going to be seen multiple times by the users so please make sure that they make sense to someone with little to no understanding of how fold internally works.

pmantica1: The same applies to your other `NotImplementedError`.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +112,109 @@ def _find_folded_outputs(ir):
...
+# 3-tuple describing the join information for each traversal in a fold.
+#
+# Contains join descriptor naming the columns used in the join predicate,
+# the source/from table, and the destination/to table
+SQLFoldTraversalDescriptor = namedtuple('SQLFoldJoinInfo', (
+    # SQLAlchemySchemaInfo join descriptor giving columns used to join from_table/to_table
+    'join_descriptor',
+
+    # Table on the left side of the join, corresponding to the outside vertex of the traversal
+    'from_table',
```
pmantica1: Change this to 'from_vertex' and 'to_vertex' for consistency with the schema generation API.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,23 +259,60 @@ def _set_group_by(self, group_by):
...
-            onclause=(from_alias.c[edge.from_column] == to_alias.c[edge.to_column])
-        )
+        # Start the join clause with the from_table of the first join descriptor
+        _, join_clause, _ = self.join_info[-1]
+
+        # Starting at the first from_table, join traversed vertices until the output vertex is
+        # reached.
+        #
+        # MSSQL doesn't use the last JOIN so we pop it off before construction the JOIN clause
+        if self._dialect == 'mssql':
```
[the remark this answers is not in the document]
bojanserafimov: This exposes a bigger problem I didn't notice when I approved your previous diff. Here https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/tests/test_helpers.py#L491 you set dialect to a string, but it should be a `sqlalchemy.engine.interfaces.Dialect` as documented here https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema/schema_info.py#L81

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +113,112 @@ def _find_folded_outputs(ir):
...
+        encoded_column, expression.literal_column('\'~\'')
+    ))  # allow unambiguously distinguishing (nullable) array elements using scheme above
+
+    # use constructor: you can't directly construct an XMLPathBinaryExpression from plain text
+    xml_column = XMLPathBinaryExpression(xml_column.left, xml_column.right, xml_column.operator)
+
+    return func.COALESCE(  # coalesce to represent empty arrays as ''
+        select([xml_column]).where(where).suffix_with("FOR XML PATH ('')").as_scalar(),
+        expression.literal_column('\'\'')
+    )
```
bojanserafimov: We absolutely need integration tests to see if this logic works as expected. You're free to add a new district of vertices in the integration testing graph filled with string valuies with weird characters. Doesn't have to be in this PR.

My advice actually is to split this PR into 2 PRs:
- xml_aggregation: This has nothing to do with `@fold` and can even be its own module. It's just some utils that help you mimic `array_agg` using `xml`. Then you can create some very simple sql queries and check their correctness with integration tests.
- `@fold` for old mssql versions: This would be this PR minus all the xml stuff. Hoever, now the `emit_sql` module is much smaller and easier to understand, and also reviewers already trust that the xml logic works and they can focus their doubt on the other parts of the PR.

In general, try to split your work into small PRs whenever possible. You will have fewer rounds of code reviews and things will go much faster. My method is to get everything working on my own secret WIP branch, make a merge plan and progressively roll out in multiple PRs. For example if you look at my WIP PR https://github.com/kensho-technologies/graphql-compiler/pull/638 you'll see that it's ugly code (that passes tests) with a merge plan that I will execute in 5 PRs. You don't have to follow my method but that's been working well for me.

graphql_compiler/compiler/expressions.py
```diff
@@ -732,15 +734,27 @@ def to_sql(self, aliases, current_alias):
...
+            empty_array = 'ARRAY[]::{}[]'.format(sql_array_type)
+            return sqlalchemy.func.coalesce(
+                fold_output_column,
+                sqlalchemy.literal_column(empty_array)
+            )
+        else:
+            raise NotImplementedError(
+                u'Fold only supported for MSSQL and '
+                u'PostgreSQL, dialect was set to {}'.format(dialect.name)
+            )
```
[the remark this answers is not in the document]
jecdohmann: I'm not sure what you mean by this. Can you give an example code snippet? `sqlalchemy.dialects.mssql` is a callable function that generates instances of the MSSQL compiler, right?

graphql_compiler/compiler/emit_sql.py
jecdohmann: I have factored out the pure fold changes to #644. Please review it when you get the chance.

The MSSQL stuff relies on those changes so I am going to keep this PR open for now.

graphql_compiler/compiler/expressions.py
```diff
@@ -732,15 +734,27 @@ def to_sql(self, aliases, current_alias):
...
-                self.fold_scope_location.fold_path
-            ].c['fold_output_' + self.fold_scope_location.field],
-            sqlalchemy.literal_column(empty_array)
-        )
+        fold_output_column = aliases[
+            self.fold_scope_location.base_location.query_path,
+            self.fold_scope_location.fold_path
+        ].c['fold_output_' + self.fold_scope_location.field]
+
+        if isinstance(dialect, MSDialect_pyodbc):
```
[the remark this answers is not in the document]
jecdohmann: MSDialect_pyodbc is the type of the SQLAlchemySchemaInfo compiler. In theory it would be great if we could do something like `dialect == sqlalchemy.dialects.mssql()` but different calls to the compiler generator don't return equal compilers.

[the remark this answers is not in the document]
pmantica1: I think what you want to do here is:
```
from sqlalchemy.dialects.mssql.base import MSDialect
if isinstance(dialect, MSDialect)
```
`MSDialect_pyodbc` is a subclass of `MSDialect`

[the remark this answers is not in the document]
jecdohmann: Good idea

graphql_compiler/compiler/expressions.py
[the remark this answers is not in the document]
jecdohmann: Fixed, with `from sqlalchemy.dialects.mssql.base.MSDialect`

graphql_compiler/compiler/emit_sql.py
```diff
@@ -126,6 +134,94 @@ def _find_folded_fields(ir):
...
+    replaced with '^e' (escape), '^n' (null), and '^d' (delimiter), resp.
+
+    Undoing the encoding above as well as the XML reference entity encoding performed
+    by the XML PATH statement is deferred to post-processing when the list is retrieved
+    from the string representation produced by the subquery.
+
+    Post-processing must split on '|', convert '~' to None, and undo both the encoding above
+    and the XML reference entity encoding auto performed by XML PATH. In particular, undo
+    '^d' -> '|', '^e' -> '^', '^n' -> '~', '&amp;' -> '&', '&gt;' -> '>', '&lt;' -> '<',
+    '&0xHEX;' -> u'0xHEX'. XML post processing can be done with
```
obi1kenobi: For the `&0xHEX` cases, do you know what characters trigger that case and what the hex value is referring to? I suspect there's an issue with the suggested `u'0xHEX'` transformation since I think hex codes don't have any meaning in Unicode since Unicode deals in code points, not bytes.

Separately, it might be good to include an example, perhaps in parens, of a specific character that got escaped into `&0xHEX` form and what it should be unescaped into. This would help implementers of this post-processing verify the correctness of their algorithm by providing a ready-made test case.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -126,6 +134,94 @@ def _find_folded_fields(ir):
...
+        A SQLAlchemy Selectable corresponding to the XML PATH subquery
+    """
+    encoded_column = func.REPLACE(  # replace all occurrences of '^' in the original with '^e'
+        output_column, expression.literal_column('\'^\''), expression.literal_column('\'^e\'')
+    )
+    encoded_column = func.REPLACE(  # replace all occurrences of '~' in the original with '^n'
+        encoded_column, expression.literal_column('\'~\''), expression.literal_column('\'^n\'')
+    )
+    encoded_column = func.REPLACE(  # replace all occurrences of '|' in the original with '^d'
+        encoded_column, expression.literal_column('\'|\''), expression.literal_column('\'^d\'')
```
obi1kenobi: Does the linter allow double quotes so you can avoid the constant escaping of `'`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -126,6 +134,94 @@ def _find_folded_fields(ir):
...
+    )
+
+    # delimit elements in the array using '|'and mark nulls with `~`
+    xml_column = (expression.literal_column('\'|\'') + func.COALESCE(  # denote null values with '~'
```
obi1kenobi: Would assigning `expression.literal_column('\'|\'')` to a local (and potentially the same for other literals like it) be allowed? It might make the code easier to read.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -198,11 +294,12 @@ class SQLFoldObject(object):
...
     # INNER JOIN VertexPrecedingOutput
     # ON ...
-
-    def __init__(self, outer_vertex_table, primary_key):
+    def __init__(self, sqlalchemy_compiler, outer_vertex_table, primary_key):
         """Create an SQLFoldObject with table, type, and join information supplied by the IR.
 
         Args:
+            sqlalchemy_compiler: SQLAlchemy compiler object passed in from the schema, indicating
+            the target dialect of SQL
```
obi1kenobi: please align this line to match the visual indentation of the start of the explanation on the line above (i.e. under `SQLAlchemy`)

graphql_compiler/compiler/emit_sql.py
```diff
@@ -198,11 +294,12 @@ class SQLFoldObject(object):
     #          ...
     # INNER JOIN VertexPrecedingOutput
     # ON ...
-
-    def __init__(self, outer_vertex_table, primary_key):
+    def __init__(self, sqlalchemy_compiler, outer_vertex_table, primary_key):
         """Create an SQLFoldObject with table, type, and join information supplied by the IR.
 
         Args:
+            sqlalchemy_compiler: SQLAlchemy compiler object passed in from the schema, indicating
```
obi1kenobi: Is this the dialect value? If so, let's try to name it consistent things across the various places where it is used.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -315,17 +424,69 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+            return select_statement
+        elif isinstance(self._sqlalchemy_compiler, PGDialect):
+            return select_statement.group_by(
+                *self.group_by
+            )
+        else:
+            raise NotImplementedError(
+                u'Fold only supported for MSSQL and '
+                u'PostgreSQL, dialect was set to {}'.format(self._sqlalchemy_compiler.name)
+            )
```
obi1kenobi: neat!

graphql_compiler/compiler/emit_sql.py
```diff
@@ -315,17 +424,69 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+        direction of the edge connecting VertexPrecedingOutput to OutputVertex.
+
+        Args:
+            intermediate_fold_output_name: String label to give to the resulting XML PATH
+            subquery built
+            fold_output_field: String name of the column requested from the output vertex.
+            last_traversal: SQLFoldTraversalDescriptor describing tables/WHERE predicate
+            used in subquery.
+
+        Returns: Selectable for XML PATH aggregation subquery
```
obi1kenobi: `Selectable` needs to go on a new line

graphql_compiler/compiler/emit_sql.py
```diff
@@ -315,17 +424,69 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+        chain of traversals beginning at the vertex immediately outside the fold.
+        VertexPrecedingOutput is the `from_table` of the last traversal descriptor tuple
+        added to the fold's `traversal_descriptors` list.
+
+        - The join predicate may have primary_key and foreign_key reversed depending on the
+        direction of the edge connecting VertexPrecedingOutput to OutputVertex.
+
+        Args:
+            intermediate_fold_output_name: String label to give to the resulting XML PATH
+            subquery built
```
obi1kenobi: Let's make sure the indentation here is consistent with the rest of the repo's docstring style. This should be indented to below `string`.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -315,17 +424,69 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+        - VertexPrecedingOutput is the vertex immediately preceding the output vertex in the
+        chain of traversals beginning at the vertex immediately outside the fold.
+        VertexPrecedingOutput is the `from_table` of the last traversal descriptor tuple
+        added to the fold's `traversal_descriptors` list.
+
+        - The join predicate may have primary_key and foreign_key reversed depending on the
+        direction of the edge connecting VertexPrecedingOutput to OutputVertex.
+
+        Args:
+            intermediate_fold_output_name: String label to give to the resulting XML PATH
```
obi1kenobi: nit: lowercase `String` here and elsewhere since the Python type you are referring to is also in lowercase (`str`).

graphql_compiler/compiler/emit_sql.py
```diff
@@ -348,10 +509,28 @@ def _get_fold_outputs(self, fold_scope_location, all_folded_outputs):
...
                         # add array aggregated output column to self._outputs
-                        self._outputs.append(
-                            self._get_array_agg_column(intermediate_fold_output_name,
-                                                       fold_output.field)
-                        )
+                        # add aggregated output column to self._outputs
+                        if isinstance(self._sqlalchemy_compiler, MSDialect):
+                            # MSSQL uses XML PATH aggregation
+                            self._outputs.append(
+                                self._get_mssql_xml_path_column(
```
obi1kenobi: It seems that all non-error paths here end up calling `append()`, and the question is what to append to `self._outputs`. Consider pulling the `self._outputs.append()` call outside of the `if-elif` and instead use the `if-elif` to just construct the expression that gets appended.

graphql_compiler/tests/test_compiler.py
```diff
@@ -4341,6 +4340,38 @@ def test_fold_on_output_variable(self):
               Animal___1.name AS `animal_name`,
               [x IN collected_Animal__out_Animal_ParentOf___1 | x.name] AS `child_names_list`
         '''
+        expected_mssql = '''
+            SELECT
+              [Animal_1].name AS animal_name,
+              folded_subquery_1.fold_output_name AS child_names_list
+            FROM
+              db_1.schema_1.[Animal] AS [Animal_1]
+              JOIN(
```
obi1kenobi: I think this `JOIN` is a clause at the same level as the top-level `SELECT` so it should be at the same indentation level as the `SELECT`. That probably means that the rest of the query is over-indented by one stop.

Also, please use 4-space indentation here instead of 2-space, otherwise some editor programs might get confused and think that the file is 2-space indented.

graphql_compiler/compiler/emit_sql.py
jecdohmann: Technically this is a compiler object, which _is_ the dialect. E.g. a compiler is either an MSSQL compiler, PostgreSQL compiler, etc.

I refer to it as a compiler everywhere in this file, let me know if you think that isn't clear. Another point of confusion might come from the fact that I pass this compiler into the `to_sql` functions where I then refer to it as `dialect`. I think this might be a good practice because from the perspective of `expressions.py` it doesn't matter as much what the complex relationship is between compilers and dialects in SQLAlchemy so it's not worth going through the trouble to explain it. Let me know what you think.

graphql_compiler/compiler/emit_sql.py
jecdohmann: Let me know if this explanation is what you had in mind

graphql_compiler/compiler/emit_sql.py
```diff
@@ -126,6 +134,101 @@ def _find_folded_fields(ir):
...
+    """
+    if isinstance(element.element, XMLPathBinaryExpression):
+        # this indicates that no label should be inserted for this element
+        kw.update(
+            within_columns_clause=False  # label always gets added if within_columns_clause is True
+        )
+    return compiler.visit_label(element, **kw)
+
+
+def get_xml_path_clause(output_column, where):
```
obi1kenobi: Should this function's name start with an underscore? It seems to only be used within this file, and does not appear to be intended to be exposed outside of it.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -126,6 +134,101 @@ def _find_folded_fields(ir):
...
+    https://docs.python.org/3/library/html.html#html.unescape
+
+    Joining from the preceding vertex to the correct output vertex
+    is performed with the WHERE clause. This is contrasted with PostgreSQL subqueries,
+    in which the join from penultimate vertex to output vertex is done within
+    the FROM/JOIN clause of the aggregation subquery.
+
+    Args:
+        output_column: SQLAlchemy Column, value being aggregated with XML PATH
+        where: SQLAlchemy BinaryExpression, predicate used to filter
```
obi1kenobi: How about calling this `predicate_expression`? That way it's both more expressive and also avoids the `.where(where)` bit of code on line 227.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -471,6 +655,9 @@ def __init__(self, sql_schema_info, ir):
 
         self._alias_generator = UniqueAliasGenerator()  # generates aliases for the fold subqueries
 
+        # indicates the sqlalchemy compiler determining the query's dialect
+        self._sqlalchemy_compiler = self.sql_schema_info.dialect
```
obi1kenobi: I still think this should be called `dialect` throughout. The fact that all of this file calls it `sqlalchemy_compiler` doesn't help that much, when clearly it's both read from a `.dialect` field here, and also passed as the `dialect` parameter in `to_sql()` calls that go outside of this file. The benefits to be gained from a consistent yet technically somewhat fuzzy definition greatly outweigh the drawbacks of that fuzzy definition, in my opinion.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -336,22 +505,37 @@ def _get_fold_outputs(self, fold_scope_location, all_folded_outputs):
                         fold_scope_location.base_location, fold_scope_location.fold_path):
 
                     if fold_output.field == COUNT_META_FIELD_NAME:
```
obi1kenobi: Consider breaking out this entire `if-else` block into a `_get_fold_output_column_clause()` function such that you can just do `column_clause = _get_fold_output_column_clause(...)` here. Then that function can just return the appropriate thing across all branches.

This both helps readability and also makes it much more difficult for a future refactor to accidentally introduce a bug that causes `self._outputs.append(column_clause)` to be reached with `column_clause` undefined.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -315,17 +431,70 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+            fold_output_field: string name of the column requested from the output vertex.
+            last_traversal: SQLFoldTraversalDescriptor describing tables/WHERE predicate
+                            used in subquery.
+
+        Returns:
+            Selectable for XML PATH aggregation subquery
+        """
+        # use join info tuple for most recent traversal to set WHERE clause for XML PATH subquery
+        edge, from_alias, to_alias = last_traversal
+        return get_xml_path_clause(self.output_vertex_alias.c[fold_output_field],
```
obi1kenobi: nit: I'd start the function arguments on a new line here, instead of right next to the name of the function being called.

PR #621, opened 2019-10-30
Implement SQL _x_count

chewselene:
Adding `_x_count` to the SQL backend in the form of a `COUNT(*)` in the folded subquery. Also adds the ability to filter based on `_x_count` (taken from the global context) with a user provided parameter or a tagged value (from a `GlobalContextField`).


graphql_compiler/tests/test_compiler.py
```diff
@@ -5631,7 +5703,34 @@ def test_multiple_filters_on_count(self):
...
+                FROM db_1.schema_1.[Animal] AS [Animal_4]
+                JOIN db_1.schema_1.[Entity] AS [Entity_1]
+                ON [Animal_4].related_entity = [Entity_1].uuid
+                GROUP BY [Animal_4].related_entity
+            ) AS folded_subquery_2
+            ON [Animal_1].related_entity = folded_subquery_2.related_entity
+            WHERE
+                coalesce(folded_subquery_1.fold_output_x_count, 0) >= :min_children AND
+                coalesce(folded_subquery_2.fold_output_x_count, 0) >= :min_related
+    '''
```
chewselene:
```
 graphql_input = '''{
        Animal {
            out_Animal_ParentOf @fold {
                _x_count @filter(op_name: ">=", value: ["$min_children"])
            }
            out_Entity_Related @fold {
                _x_count @filter(op_name: ">=", value: ["$min_related"])
            }
        }
    }'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -5572,7 +5625,26 @@ def test_filter_count_with_tagged_parameter_in_fold_scope(self):
...
+                    count(*) AS fold_output_x_count,
+                    array_agg([Animal_3].name) AS fold_output_name
+                FROM db_1.schema_1.[Animal] AS [Animal_2]
+                JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                ON [Animal_2].uuid = [Animal_3].parent
+                GROUP BY [Animal_2].uuid
+            ) AS folded_subquery_1
+            ON [Animal_1].uuid = folded_subquery_1.uuid
+            WHERE coalesce(folded_subquery_1.fold_output_x_count, 0) >= [Species_1].limbs
+'''
```
chewselene:
```
graphql_input = '''{
        Animal {
            out_Animal_OfSpecies {
                limbs @tag(tag_name: "limbs")
            }
            out_Animal_ParentOf @fold {
                _x_count @filter(op_name: ">=", value: ["%limbs"])
                name @output(out_name: "child_names")
            }
        }
    }'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -5514,8 +5533,24 @@ def test_output_count_in_fold_scope(self):
...
+                SELECT
+                    [Animal_2].uuid AS uuid,
+                    count(*) AS fold_output_x_count,
+                    array_agg([Animal_3].name) AS fold_output_name
+                FROM db_1.schema_1.[Animal] AS [Animal_2]
+                JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                ON [Animal_2].uuid = [Animal_3].parent
+                GROUP BY [Animal_2].uuid
+              ) AS folded_subquery_1
+              ON [Animal_1].uuid = folded_subquery_1.uuid'''
```
chewselene:
```
graphql_input = '''{
        Animal {
            out_Animal_ParentOf @fold {
                _x_count @output(out_name: "number_of_children")
                name @output(out_name: "child_names")
            }
        }
    }'''
```

graphql_compiler/tests/test_compiler.py
```diff
@@ -885,7 +885,26 @@ def test_filter_in_optional_and_count(self):
...
+                FROM db_1.schema_1.[Species] AS [Species_2]
+                JOIN db_1.schema_1.[Species] AS [Species_3]
+                ON [Species_2].uuid = [Species_3].eats
+                GROUP BY [Species_2].uuid
+              ) AS folded_subquery_1
+            ON [Species_1].uuid = folded_subquery_1.uuid
+            WHERE
+                ([Animal_1].name = :animal_name OR [Animal_1].species IS NULL) AND
+                coalesce(folded_subquery_1.fold_output_x_count, 0) >= :predators
+        '''
```
chewselene:
```
graphql_input = '''{
        Species {
            name @output(out_name: "species_name")

            in_Animal_OfSpecies @optional {
                name @filter(op_name: "=", value: ["$animal_name"])
            }

            in_Species_Eats @fold {
                _x_count @filter(op_name: ">=", value: ["$predators"])
            }
        }
    }'''
```

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +103,25 @@ def _find_folded_outputs(ir):
...
+        if isinstance(block, blocks.GlobalOperationsStart):
+            global_operation_start = True
+        # Check the left and right side of Filter predicates to see if _x_count is involved.
+        if global_operation_start and isinstance(block, blocks.Filter):
+            if (isinstance(block.predicate.left, FoldedContextField) and
+                    block.predicate.left.fold_scope_location.field == '_x_count'):
+                _x_count_filters.append(block.predicate.left)
+            if (isinstance(block.predicate.right, FoldedContextField) and
+                    block.predicate.right.fold_scope_location.field == '_x_count'):
+                _x_count_filters.append(block.predicate.right)
```
obi1kenobi: A different equivalent option here, your call on which one to use:

```
for subexpression in (block.predicate.left, block.predicate.right):
    if isinstance(subexpression, FoldedContextField) and subexpression.fold_scope_location.field == '_x_count':
        _x_count_filters.append(subexpression)
```

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +103,25 @@ def _find_folded_outputs(ir):
...
+    _x_count_filters = []
+    global_operation_start = False
+    for block in ir.ir_blocks:
+        # _x_count filters only occur after encountering a GlobalOperationStart block.
+        if isinstance(block, blocks.GlobalOperationsStart):
+            global_operation_start = True
+        # Check the left and right side of Filter predicates to see if _x_count is involved.
+        if global_operation_start and isinstance(block, blocks.Filter):
+            if (isinstance(block.predicate.left, FoldedContextField) and
+                    block.predicate.left.fold_scope_location.field == '_x_count'):
```
obi1kenobi: Do we have a constant we could refer to for `_x_count`? If not, can we make one?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -224,43 +243,72 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+                        fold_scope_location.at_vertex()):
+                    self._outputs.append(
+                        sqlalchemy.func.coalesce(
+                            sqlalchemy.func.count(),
+                            sqlalchemy.literal_column('0')
+                        ).label('fold_output_x_count')
+                    )
+                    break
+
         return sorted(self._outputs, key=lambda column: column.name, reverse=True)
```
obi1kenobi: Interesting, I'm very curious about this `reverse=True` :) I know it's not the subject of the PR though, so we can discuss it another time if that's easier.

graphql_compiler/compiler/expressions.py
```diff
@@ -394,10 +394,19 @@ def to_cypher(self):
...
-                             u'should not be called.')
+        """Return a sqlalchemy Column picked from the appropriate alias."""
+        self.validate()
+        if isinstance(self.field_type, GraphQLList):
+            raise NotImplementedError(u'The SQL backend does not support lists. Cannot '
+                                      u'process field {}.'.format(self.location.field))
+
+        if self.location.field is not None:
+            if '@' in self.location.field:
+                raise NotImplementedError(u'The SQL backend does not support __typename.')
```
obi1kenobi: We don't conclusively know that this is `__typename` here, all we know is that the field is a special field of some sort (there could be more than one in the future).

That said, this is because of an OrientDB-ism (`@class`), so maybe open an issue about using `__typename` as the field name in IR locations and refer to it from a comment here?

graphql_compiler/compiler/expressions.py
```diff
@@ -394,10 +394,19 @@ def to_cypher(self):
...
+            return aliases[(self.location.at_vertex().query_path, None)].c[self.location.field]
+        else:
+            raise AssertionError(u'This is a bug. The SQL backend does not use '
+                                 u'global context fields to point to vertices.')
```
obi1kenobi: Print some more context. If we ever see this message, all we'll know is that we have a bug -- we won't know any contextual info or even what field caused it.

graphql_compiler/tests/test_compiler.py
```diff
@@ -5542,7 +5578,25 @@ def test_filter_count_with_runtime_parameter_in_fold_scope(self):
...
+            LEFT OUTER JOIN (
+                SELECT
+                    [Animal_2].uuid AS uuid,
+                    coalesce(count(*), 0) AS fold_output_x_count,
+                    array_agg([Animal_3].name) AS fold_output_name
+                FROM db_1.schema_1.[Animal] AS [Animal_2]
+                JOIN db_1.schema_1.[Animal] AS [Animal_3]
+                ON [Animal_2].uuid = [Animal_3].parent
+                GROUP BY [Animal_2].uuid
+            ) AS folded_subquery_1
```
obi1kenobi: Nice work, these queries already look like something I wouldn't want to write by hand :)

graphql_compiler/compiler/emit_sql.py
```diff
@@ -224,43 +242,72 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
+        # _x_count must be added if _x_count is used in filtering, but is not an output
+        if not has_x_count:
+            for x_count_filter in x_count_filters:
+                if (x_count_filter.fold_scope_location.at_vertex() ==
+                        fold_scope_location.at_vertex()):
+                    self._outputs.append(
+                        sqlalchemy.func.coalesce(
+                            sqlalchemy.func.count(),
+                            sqlalchemy.literal_column('0')
+                        ).label('fold_output_x_count')
```
bojanserafimov: Does this mean `fold_output_x_count` is disallowed as an `out_name` of a folded output?

graphql_compiler/compiler/expressions.py
chewselene: Addressing this in [#628 ](https://github.com/kensho-technologies/graphql-compiler/pull/628)

graphql_compiler/compiler/emit_sql.py
chewselene: This would actually be a problem with `x_count` since `fold_output_` is prepended to the `out_name` to get the intermediate fold output name. I will change this label to `fold_output__x_count` so that it follows the convention of the others.

bojanserafimov: Aah I see. Can you explan this in a comment next to the definition of the `fold_output__x_count` constant?

chewselene: Switched to constructing this string with constants. Hopefully it is a little clearer now!

graphql_compiler/compiler/emit_sql.py
chewselene: @jecdohmann any insight here?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +108,23 @@ def _find_folded_outputs(ir):
     return folded_outputs
 
 
+def _find_x_count_filters(ir):
+    """For each fold path, find all filters involving _x_count."""
```
bojanserafimov: Isn't it simpler to find just the set of fold scope locations that use `_x_count` in any way (filter or output it)? It seems to me like that's how this result is being used, in a convoluted way. Also it seems like `_get_fold_outputs` will be simplified. Correct me if I'm missing something.

graphql_compiler/compiler/expressions.py
```diff
@@ -785,7 +799,18 @@ def to_cypher(self):
 
     def to_sql(self, aliases, current_alias):
         """Return a sqlalchemy Column picked from the appropriate alias."""
-        # get the type of the folded field
+        # _x_count is a special case that has already been coalesced to 0.
+        # _x_count's intermediate output name is always fold_output__x_count
```
bojanserafimov: Maybe document this expectation in the definition of `CompilationState._aliases`. I think that's the easiest way to make sure:
- it actually gets implemented (or the implementation does not get removed)
- the usage of it is not confusing (it is backed by the spec of the field)

PR #622, opened 2019-10-30
Add argument deserialization helper

bojanserafimov:
This is needed for serving a jsonrpc GraphQL endpoint, or using this library with an editor.

### DIFF
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -1,6 +1,5 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Safely insert runtime arguments into compiled GraphQL queries."""
-import builtins
 import datetime
 import decimal
 
@@ -67,7 +66,12 @@ def _deserialize_anonymous_json_argument(expected_type, value):
         if isinstance(value, six.integer_types):
             return value
         elif isinstance(value, six.string_types):
-            return builtins.int(value)
+            if six.PY2:
+                return long(value)
+            elif six.PY3:
+                return int(value)
+            else:
+                raise AssertionError(u'Unreachable code. Neither py2 nor py3.')
         else:
             raiseValueError(u'Unexpected type {}. Expected one of {}.'
                             .format(type(value), (six.integer_types, six.string_types)))
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
                                                          type(value).__name__))
 
 
+def _deserialize_anonymous_json_argument(expected_type, value):
+    """Deserialize argument. See docstring of deserialize_json_argument."""
+    allowed_types_for_graphql_type = {
+        GraphQLDate: (six.string_types,),
+        GraphQLDateTime: (six.string_types,),
+        GraphQLFloat: (six.string_types, float, six.integer_types),
+        GraphQLDecimal: (six.string_types, float, six.integer_types),
+        GraphQLInt: (six.integer_types,),
+        GraphQLString: (six.string_types,),
+        GraphQLBoolean: (bool,)
+        GraphQLID: (six.string_types,),
+    }
+
+    # Check if the type of the value is correct
+    correct_type = True
+    expected_python_types = allowed_types_for_graphql_type[expected_type]
+    if isinstance(value, bool) and not GraphQLBoolean.is_same_type(expected_type):
+        correct_type = False  # We explicitly disallow passing boolean values for non-boolean types
+    if not isinstance(value, expected_python_types):
+        correct_type = False
+    if not correct_type:
+        raise ValueError(u'Unexpected type {}. Expected one of {}.'
+                         .format(type(value), expected_python_types))
+
+    # Use the default GraphQL parser to parse the value
+    return expected_type.parse_value(value)
+
+
+def deserialize_json_argument(name, expected_type, value):
+    """Deserialize a GraphQL argument parsed from a json file.
+
+    Passing arguments via jsonrpc, or via the GUI of standard GraphQL editors is tricky because
+    json does not support certain types like Date, Datetime, Decimal, and also confuses floats
+    for integers if there are no decimals. This function takes in a value received from a json,
+    and converts it to a standard python representation that can be used in the query. Below are
+    examples of accepted json encodings of all the types:
+        GraphQLDate: "2018-02-01T05:11:54Z"
+        GraphQLDateTime: "2018-02-01"
+        GraphQLFloat: 4.3, "5.0", 5
+        GraphQLDecimal: "5.00000000000000000000000000001"
+        GraphQLInt: 4
+        GraphQLString: "Hello"
+        GraphQLBoolean: True
+        GraphQLID: "13d72846-1777-6c3a-5743-5d9ced3032ed"
+
+    Args:
+        name: string, the name of the argument. It will be used to provide a more descriptive error
+              message if an error is raised.
+        expected_type: GraphQLType we expect. All GraphQLNonNull type wrappers are stripped.
+        value: object that can be interpreted as being of that type
+    """
+    try:
+        return _deserialize_anonymous_json_argument(strip_non_null_from_type(expected_type), value)
+    except ValueError as e:
+        raise GraphQLInvalidArgumentError('Error parsing argument {}: {}'.format(name, e))
+
+
 def validate_argument_type(name, expected_type, value):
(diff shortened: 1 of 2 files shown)


graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
 
 
+def _deserialize_anonymous_json_argument(expected_type, value):
+    """Deserialize argument. See docstring of deserialize_json_argument."""
+    allowed_types_for_graphql_type = {
+        GraphQLDate: (six.string_types,),
+        GraphQLDateTime: (six.string_types,),
+        GraphQLFloat: (six.string_types, float, six.integer_types),
+        GraphQLDecimal: (six.string_types, float, six.integer_types),
+        GraphQLInt: (six.integer_types,),
```
obi1kenobi: Should we allow int-like strings here as well? I believe JSON as a format doesn't actually support the full int64 range, since Javascript i.e. JSON only knows about doubles, and any int above `2**53` would lose precision as a result. The easy-ish fix here is to allow strings that parse into ints to work.

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
+def deserialize_json_argument(name, expected_type, value):
+    """Deserialize a GraphQL argument parsed from a json file.
+
+    Passing arguments via jsonrpc, or via the GUI of standard GraphQL editors is tricky because
+    json does not support certain types like Date, Datetime, Decimal, and also confuses floats
+    for integers if there are no decimals. This function takes in a value received from a json,
+    and converts it to a standard python representation that can be used in the query. Below are
+    examples of accepted json encodings of all the types:
+        GraphQLDate: "2018-02-01"
+        GraphQLDateTime: "2018-02-01T05:11:54Z"
```
obi1kenobi: Let's add an example and test case to prove that explicit timezone info is parsed correctly as well: e.g., `2018-02-01T05:11:54Z` and `2018-02-01T05:11:54+00:00` are the same thing, per the standard.

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
+        GraphQLFloat: 4.3, "5.0", 5
+        GraphQLDecimal: "5.00000000000000000000000000001"
+        GraphQLInt: 4
+        GraphQLString: "Hello"
+        GraphQLBoolean: True
+        GraphQLID: "13d72846-1777-6c3a-5743-5d9ced3032ed"
```
obi1kenobi: It might be worth mentioning here that ID types can be any int or string, per the GraphQL spec. In particular, they are allowed but not required to be UUID-like.

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
+
+    Args:
+        name: string, the name of the argument. It will be used to provide a more descriptive error
+              message if an error is raised.
+        expected_type: GraphQLType we expect. All GraphQLNonNull type wrappers are stripped.
+        value: object that can be interpreted as being of that type
```
obi1kenobi: Need a `Returns:` section here.

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
+    """
+    try:
+        return _deserialize_anonymous_json_argument(strip_non_null_from_type(expected_type), value)
+    except ValueError as e:
+        raise GraphQLInvalidArgumentError('Error parsing argument {}: {}'.format(name, e))
```
pmantica1: This will be a really common error make it as good as possible. This is not sufficient.

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
+
+    # Check if the type of the value is correct
+    correct_type = True
+    expected_python_types = allowed_types_for_graphql_type[expected_type]
+    if isinstance(value, bool) and not GraphQLBoolean.is_same_type(expected_type):
+        correct_type = False  # We explicitly disallow passing boolean values for non-boolean types
+    if not isinstance(value, expected_python_types):
+        correct_type = False
+    if not correct_type:
+        raise ValueError(u'Unexpected type {}. Expected one of {}.'
```
obi1kenobi: nit: Might be worth mentioning why this raises `ValueError` as opposed to the `GraphQLInvalidArgumentError` that this is repackaged into in the below function. I'm assuming this is because `parse_value()` also raises `ValueError` when parsing fails, e.g. when calling `int()` on a string-typed value.

graphql_compiler/tests/test_end_to_end.py
```diff
@@ -217,3 +217,57 @@ def test_non_null_types_pass_validation(self):
...
+            deserialize_json_argument('amount', GraphQLFloat, 'sdg')
+
+        # Bool
+        with self.assertRaises(GraphQLInvalidArgumentError):
+            deserialize_json_argument('amount', GraphQLFloat, True)
+
+        # Valid string
+        self.assertEqual(float(5), deserialize_json_argument('amount', GraphQLFloat, '5'))
+
+        # Valid string with comma
```
obi1kenobi: nit: Technically a period, not a comma, at least in this localization. Maybe say "decimal portion" or something less locale-sensitive? :)

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
                                                          type(value).__name__))
 
 
+def _deserialize_anonymous_json_argument(expected_type, value):
+    """Deserialize argument. See docstring of deserialize_json_argument."""
+    allowed_types_for_graphql_type = {
+        GraphQLDate: (six.string_types,),
+        GraphQLDateTime: (six.string_types,),
```
pmantica1: In this diff, can you make the errors for datetime objects a bit more understandable? You'll get the following error if you enter the value incorrectly.

`arrow.parser.ParserMatchError: Failed to match 'YYYY-MM-DDTHH:mm:ssZZ'`

Can you make it something that does not require a deep understanding of how datetime strings work?

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
                                                          type(value).__name__))
 
 
+def _deserialize_anonymous_json_argument(expected_type, value):
+    """Deserialize argument. See docstring of deserialize_json_argument."""
```
obi1kenobi: Can you add `Args/Returns/Raises` sections? Mostly interested in the `Raises` here so we know what errors to catch and rewrap in the caller function.

graphql_compiler/query_formatting/common.py
```diff
@@ -28,6 +28,63 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
...
+    if isinstance(value, bool) and not GraphQLBoolean.is_same_type(expected_type):
+        correct_type = False  # We explicitly disallow passing boolean values for non-boolean types
+    if not isinstance(value, expected_python_types):
+        correct_type = False
+    if not correct_type:
+        raise ValueError(u'Unexpected type {}. Expected one of {}.'
+                         .format(type(value), expected_python_types))
+
+    # Use the default GraphQL parser to parse the value
+    return expected_type.parse_value(value)
```
obi1kenobi: Per #623, it seems like this line could throw an `arrow` error which I believe might not be a subclass of `ValueError`, so this code might have unexpected behavior.

graphql_compiler/query_formatting/common.py
```diff
@@ -47,8 +47,8 @@ def _deserialize_anonymous_json_argument(expected_type, value):
             GraphQLID: six.string_types
 
     Raises:
-        ValueError or arrow.parser.ParserMatchError if the value is not appropriate for the type.
-        These are the exceptions that the GraphQL parsers raise for invalid values.
+        ValueError if the value is not appropriate for the type. Value error is chosen because
```
obi1kenobi: nit: `Value error` -> `ValueError`?

obi1kenobi: Also I'm guessing this is true because the parser error raised by `arrow` is already a subclass of `ValueError`?

PR #625, opened 2019-10-31
Remove Cypher query parameters documentation

pmantica1:
We never introduce the cypher compilation target anywhere in the documentation, but this documentation block seems to jump "right in". It confuses the reader and leaves him looking for the rest of the cypher documentation, which does not exist.  

Once the cypher compilation target and its backends are well documented we can re-introduce this documentation. For now, it is a net negative.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,7 +90,6 @@ Table of contents
    -  `Optional type_equivalence_hints compilation
       parameter <#optional-type_equivalence_hints-parameter>`__
    -  `SchemaGraph <#schemagraph>`__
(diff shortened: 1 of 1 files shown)


obi1kenobi:
We're already moving to ReadTheDocs, so why not leave this in a "WIP" kind of section? It's perfectly good info that's just missing some context. When the context is added, I'd prefer not to have to write the whole section all over again. Especially that link to RedisGraph is both valuable and would be difficult to find again.

I don't think this is a move we want to make.

PR #626, opened 2019-11-03
Compare directives by name instead of python object references

pmantica1:
`get_schema_for_macro_definition` currently does not work correctly. If you serialize and parse a schema and then run `get_schema_for_macro_definition`, you'll end up with a schema with only the required macro directives. The subtle reason this happens is that we currently compare directives by their python object references instead of their names.

### DIFF
--- a/graphql_compiler/macros/__init__.py
+++ b/graphql_compiler/macros/__init__.py
@@ -16,11 +16,10 @@ from ..ast_manipulation import safe_parse_graphql
 from ..compiler.subclass import compute_subclass_sets
 from ..compiler.validation import validate_schema_and_query_ast
 from ..exceptions import GraphQLValidationError
-from ..schema import check_for_nondefault_directive_names
 from .macro_edge import make_macro_edge_descriptor
 from .macro_edge.directives import (
     DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION, DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION,
-    MacroEdgeDirective
+    MacroEdgeDirective, get_schema_for_macro_edge_definitions
 )
 from .macro_expansion import expand_macros_in_query_ast
 from .validation import (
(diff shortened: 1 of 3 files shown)


graphql_compiler/macros/__init__.py
```diff
@@ -212,18 +207,7 @@ def get_schema_for_macro_definition(schema):
...
-    macro_definition_schema_directives += DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION
-    # Remove disallowed directives from directives list
-    macro_definition_schema_directives = list(set(macro_definition_schema_directives) &
-                                              set(DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION))
-
-    # pylint: disable=protected-access
-    macro_definition_schema._directives = macro_definition_schema_directives
-    # pylint: enable=protected-access
-    return macro_definition_schema
+    return get_schema_for_macro_edge_definitions(schema)
```
pmantica1: I do not remove `get_schema_for_macro_edge_definitions` and  only keep `get_schema_for_macro_definitions` because it's not obvious that we should. These functions just happen to have the same functionality right now. However, in the future, we might have macro properties and might want to use separate schemas for validating macro properties and macro definitions. In either case, removing `get_schema_for_macro_edge_definitions` should be done in it's own PR.

graphql_compiler/macros/macro_edge/directives.py
```diff
@@ -68,8 +69,19 @@
 
 def get_schema_for_macro_edge_definitions(querying_schema):
     """Given a schema object used for querying, create a schema used for macro edge definitions."""
-    new_directives = list(chain(
-        querying_schema.get_directives(), DIRECTIVES_REQUIRED_IN_MACRO_EDGE_DEFINITION))
+    original_directives = querying_schema.get_directives()
+    check_for_nondefault_directive_names(original_directives)
+
+    names_of_allowed_directives = {
+        directive.name for directive in
```
obi1kenobi: nit: move the `for` to the next line for consistency with the list comprehension on lines 80-84.

graphql_compiler/macros/__init__.py
obi1kenobi: Why would we remove `get_schema_for_macro_edge_definitions`? Based on your point that `[...] in the future, we might have macro properties and might want to use separate schemas for validating macro properties and macro definitions` shouldn't we remove `get_schema_for_macro_definition` instead and keep functions specific to each particular kind of macro?

pmantica1: I am not sure that we should remove `get_schema_for_macro_definition`. You might want a schema with all macro definition directives included when building a UI for registering macro definitions. 

The point I am trying to make is that it's not obvious what to do here, so whatever we end up doing should be done in a separate PR.

PR #628, opened 2019-11-05
Move __typename -> @class conversion after IR generation

chewselene:
As @obi1kenobi pointed out in [PR #621](https://github.com/kensho-technologies/graphql-compiler/pull/621), `@class` is an orientDB-ism. Currently, `__typename` is converted to `@class` in the IR generation step. This PR moves the conversion later since the IR must also be used by SQL and any future languages that do not use `@class`.

### DIFF
--- a/graphql_compiler/compiler/expressions.py
+++ b/graphql_compiler/compiler/expressions.py
@@ -8,7 +8,7 @@ from sqlalchemy import bindparam, sql
 
 from . import cypher_helpers, sqlalchemy_extensions
 from ..exceptions import GraphQLCompilationError
-from ..schema import COUNT_META_FIELD_NAME, GraphQLDate, GraphQLDateTime
+from ..schema import COUNT_META_FIELD_NAME, TYPENAME_META_FIELD_NAME, GraphQLDate, GraphQLDateTime
 from .compiler_entities import Expression
 from .helpers import (
     STANDARD_DATE_FORMAT, STANDARD_DATETIME_FORMAT, FoldScopeLocation, Location,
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/expressions.py
```diff
@@ -489,8 +505,9 @@ def to_sql(self, aliases, current_alias):
                                       u'process field {}.'.format(self.location.field))
 
         if self.location.field is not None:
-            if '@' in self.location.field:
-                raise NotImplementedError(u'The SQL backend does not support __typename.')
+            if self.location.field == TYPENAME_META_FIELD_NAME:
+                raise NotImplementedError(u'The SQL backend does not support {}.'.format(
+                    TYPENAME_META_FIELD_NAME))
```
bojanserafimov: It might be better to have the following structure:
```
if self.location.field == '__some_field_that_is_supported_on_this_backend':
    implementation_of_field...
elif is_meta_field(self.location.field)
    raise NotImplementedError(u'...')
```

That way new meta fields will be unsupported by default until they are actually implemented. Also there wouldn't be a need to scatter this logic in multiple files: https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/emit_sql.py#L236

chewselene: Updated to the new structure, but I think the `_x_count` check in `emit_sql` may still necessary. We are appending outputs by adding an array aggregation of `self.output_vertex_alias.c[fold_output.field]` to the list of folded outputs. However, this does not work `fold_output.field == _x_count` since a column called `_x_count` does not exist. All of this is done without accessing an expression's `to_sql` method. Does that make sense?

obi1kenobi: Agreed with both of you. We'd probably still need the logic in more than one place, but it's nice to design for "unsupported by default until explicitly supported" whenever it's possible to do that. 🎉 for merging this!

PR #638, opened 2019-11-08
WIP Pagination parameter generator

bojanserafimov:
Implement `generate_parameters_for_vertex_partition`, returning a generator of values that split the vertex into the desired number of splits. Usually only the first number in the generator is needed, but it's good to have all just in case. The function can be used and tested without access to a parameterized query.

### DIFF
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -68,9 +68,15 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
     return False
 
 
-def _convert_uuid_string_to_int(uuid_string):
+def convert_uuid_string_to_int(uuid_string):
     """Return the integer representation of a UUID string."""
     return UUID(uuid_string).int
+
+
+def convert_int_to_uuid_string(int_value):
+    """Return the integer representation of a UUID string."""
+    # XXX needs more care
+    return str(UUID(int=int(int_value)))
 
 
 def _create_integer_interval(lower_bound, upper_bound):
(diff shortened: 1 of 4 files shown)


graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -68,11 +68,17 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
...
 
 
-def _convert_uuid_string_to_int(uuid_string):
+def convert_uuid_string_to_int(uuid_string):
     """Return the integer representation of a UUID string."""
     return UUID(uuid_string).int
 
 
+def convert_int_to_uuid_string(int_value):
+    """Return the integer representation of a UUID string."""
```
obi1kenobi: I think this is backwards, maybe copy-pasta :)

graphql_compiler/query_pagination/parameter_generator.py
```diff
@@ -1,24 +1,53 @@
...
+# XXX merge plan:
+# - copy _convert_int_to_field_value and _convert_field_value_to_int from query-pagination branch
+# - implement _get_query_path_endpoint_type properly
+# - land to this PR
+# - find existing filters on the pagination_key
+# - Implement int field paging
+
+
+def _get_query_path_endpoint_type(schema, query_path):
+    current_type = schema.get_type('RootSchemaQuery')
```
obi1kenobi: This might end up being a problem in the long run, since technically I'm not sure anything guarantees this choice of name for the root query type. Consider `schema.get_query_type()` instead, which I believe should be the same thing?

bojanserafimov:
splitting up

PR #641, opened 2019-11-11
Add mapping of Postgresql scalar datatypes to GraphQL types

pmantica1:


graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
```diff
@@ -94,8 +96,50 @@
...
+    postgrestypes.TIMESTAMP: GraphQLDateTime,
+    postgrestypes.ENUM: GraphQLString,
+}
+
+UNSUPPORTED_POSTGRES_TYPES = frozenset({
+    # We shouldn't map the Postgresql bit type to the GraphQLBoolean type. The Postgresql bit type
+    # is used to represent a bit string of variable length.
+    # https://www.postgresql.org/docs/8.1/datatype-bit.html
+    postgrestypes.TIME,
+    postgrestypes.BIT,
```
obi1kenobi: nit: Consider moving the `BIT` type such that it's right under the comment that specifically references it

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
```diff
@@ -94,8 +96,50 @@
...
+    postgrestypes.INET,
+    postgrestypes.CIDR,
+    postgrestypes.MACADDR,
+    postgrestypes.MONEY,
+    postgrestypes.OID,
+    postgrestypes.REGCLASS,
+    postgrestypes.BYTEA,
+    postgrestypes.INTERVAL,
+    postgrestypes.ARRAY,
```
obi1kenobi: We don't support array types? What would be required for us to be able to turn them into GraphQL list types?

pmantica1: We don't have the functionality to support SQL ARRAYS. We could add it. It would probably be around 30-50 lines of code with tests included. Would you prefer that I did this in this PR or in a separate PR?

bojanserafimov: I prefer separate because I want `DOUBLE_PRECISION` to merge into master sooner

PR #642, opened 2019-11-11
Compare directives by name in deserialize_json_argument

pmantica1:
See the test case to see why we want to make this change.


graphql_compiler/tests/test_end_to_end.py
```diff
@@ -301,3 +301,14 @@ def test_int_deserialization(self):
             # Long string
             self.assertEqual(50000000000000000000000000000000000000000, deserialize_json_argument(
                 'amount', GraphQLInt, '50000000000000000000000000000000000000000'))
+
+    def test_invalid_directive_comparison(self):
```
pmantica1: I check that this test failed before I changed the code above.

PR #644, opened 2019-11-11
Update SQL Fold Logic to Make MSSQL Implementation Easier

jecdohmann:
PR #613 was becoming very complex, containing a combination of changes to the underlying fold logic and additional utils permitting MSSQL folds. I have factored out all of the non-MSSQL specific changes from that PR into this one.

This PR contains the following changes:

1. Replace unnamed tuples for joins with namedtuple SQLFoldTraversalDescriptor to enhance readability of the code which builds joins in the SQL Fold object
2. Fix the join from the outer query to the folded subquery so that the join constraint is on the primary key of the vertex outside the fold, as opposed to choosing the from column of the most recent edge.
3. Made `_construct_joins` depend on a loop, creating a better generalization of the original functionality which will help with MSSQL fold and with adding traversals inside of fold later on
4. Added various small abstractions to make integrating with MSSQL functionality easier (e.g. moving the array agg into its own method in `_get_fold_outputs`)
5. Added more assertion errors and better error logging. Added a field in SQL Fold to determine whether the fold has been ended and to prevent further methods from being called on an ended fold.
6. Added `test_fold_on_foreign_key` to see whether change 2 permits us to now properly fold on an edge which exits a vertex via that vertex's foreign key.


bojanserafimov:
>     6\. to know properly fold on an edge which exits a vertex via that vertex's foreign key.

Not sure what you mean here. I'm trying to understand what the bug fixed by change 2 is, but I don't understand the test case that the old implementation is supposed to fail.

jecdohmann:
> > ```
> > 6\. to know properly fold on an edge which exits a vertex via that vertex's foreign key.
> > ```
> 
> Not sure what you mean here. I'm trying to understand what the bug fixed by change 2 is, but I don't understand the test case that the old implementation is supposed to fail.

In `test_fold_on_foreign_key` we traverse an edge where the `from_column` for the join descriptor at the folded vertex is a foreign key. The previous implementation would've created a subquery that groups by that foreign key, selects that foreign key, and then joins back to the outside table on that foreign key. But we really want is to select the primary key, group by primary key, etc.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -143,35 +165,71 @@ class SQLFoldObject(object):
...
+        if self._outer_vertex_alias is None:
+            return u'Invalid fold: no vertex preceding fold.'
+        elif self._output_vertex_alias is None:
+            return u'Vertex outside fold: {}. Output vertex for fold: None'.format(
+                self._outer_vertex_alias.original
+            )
+        else:
+            return u'Vertex outside fold: {}. Output vertex for fold: {}'.format(
+                self._outer_vertex_alias.original, self._output_vertex_alias.original
+            )
```
chewselene: I don't know if these messages will read clearly in the error messages that use these methods. Ex: these will be put in the middle of sentences, but are capitalized and won't really complete the sentences.

graphql_compiler/tests/test_compiler.py
```diff
@@ -4171,6 +4170,36 @@ def test_fold_on_output_variable(self):
...
+                GROUP BY
+                    "Animal_2".uuid
+            ) AS folded_subquery_1
+            ON "Animal_1".uuid = folded_subquery_1.uuid
+        '''
+
+        expected_mssql = SKIP_TEST
+        expected_match = SKIP_TEST
+        expected_gremlin = SKIP_TEST
+        expected_cypher = SKIP_TEST
```
chewselene: Should these be implemented eventually? If so, put a TODO. If not, put a comment about why not.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,81 +258,115 @@ def _set_group_by(self, group_by):
...
-                                               join_descriptor,
                                                all_folded_outputs)
 
-    def visit_traversed_vertex(self, join_descriptor, from_table, to_table):
-        """Add a new join descriptor for every vertex traversed in the fold."""
-        self._join_info.append((join_descriptor, from_table, to_table))
+    def visit_traversed_vertex(self, join_descriptor, from_vertex, to_vertex):
+        """Add a new traversal descriptor for every vertex traversed in the fold."""
+        if self._ended:
+            raise AssertionError(u'Cannot visit traversed vertices after end_fold has been called.')
```
chewselene: this one is missing a print out of the fold object.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -143,35 +165,71 @@ class SQLFoldObject(object):
...
+    #
+    # SELECT
+    #   OuterVertex.SOME_COLUMN <- this value is the primary key
+    #   COALESCE((SELECT ... FOR XML PATH(''), '~') AS fold_output
+    # FROM OuterVertex
+    # INNER JOIN ...
+    # ON ...
+    #          ...
+    # INNER JOIN VertexPrecedingOutput
+    # ON ...
```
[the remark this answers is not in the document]
jecdohmann: I believe that having the x_count inside the same query as the aggregation won't work for cases where there are multiple traversals. So GROUP BYs won't be necessary in MSSQL

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +105,24 @@ def _find_folded_outputs(ir):
...
+#
+# Contains join descriptor naming the columns used in the join predicate,
+# the source/from table, and the destination/to table
+SQLFoldTraversalDescriptor = namedtuple('SQLFoldJoinInfo', (
+    # SQLAlchemySchemaInfo join descriptor giving columns used to join from_vertex/to_vertex
+    'join_descriptor',
+
+    # SQLAlchemy table corresponding to corresponding to the outside vertex of the traversal,
+    # appears on the left side of the join.
+    'from_vertex',
```
bojanserafimov: It's not intuitive that `from_vertex` is a table. Call it `from_table`?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,6 +105,24 @@ def _find_folded_outputs(ir):
     return folded_outputs
 
 
+# 3-tuple describing the join information for each traversal in a fold.
+#
+# Contains join descriptor naming the columns used in the join predicate,
+# the source/from table, and the destination/to table
+SQLFoldTraversalDescriptor = namedtuple('SQLFoldJoinInfo', (
+    # SQLAlchemySchemaInfo join descriptor giving columns used to join from_vertex/to_vertex
```
bojanserafimov: `DirectJoinDescriptor`

graphql_compiler/compiler/emit_sql.py
pmantica1: This is the terminology that we use for `DirectEdgeDescriptor` objects. I think that we should have a consistent naming convention. If we make this change here then we should also probably do it in schema generation.

pmantica1: Sorry I did not notice this was referring to SQLAlchemy Table object. I thought that it was referring to the vertex name of the SQLAlchemy Table. `from_table` seems like a good name.

jecdohmann:
> > > > ```
> > > > 6\. to know properly fold on an edge which exits a vertex via that vertex's foreign key.
> > > > ```
> > > 
> > > 
> > > Not sure what you mean here. I'm trying to understand what the bug fixed by change 2 is, but I don't understand the test case that the old implementation is supposed to fail.
> > 
> > 
> > In `test_fold_on_foreign_key` we traverse an edge where the `from_column` for the join descriptor at the folded vertex is a foreign key (`lives_in`). The previous implementation would've created a subquery that groups by that foreign key, selects that foreign key, and then joins back to the outside table on that foreign key (e.g. we would get a subquery looking something like `SELECT [Animal_2].lives_in ... GROUP BY [Animal_2].lives_in`. But what we really want is to select the primary key, group by primary key, etc. e.g. `SELECT [Animal_2].uuid ... GROUP BY [Animal_2].uuid`
> 
> Why is that change necessary? Is the result of running the query any different? Or is it just for convenience?
> 
> Off topic: In many-to-one examples I'm not sure that the `GROUP BY` is even necessary. But let's leave that as a future optimization.

The change is necessary because if you've grouped by and identified each row with the foreign key, what do you join on when you try and stitch the subquery back into the rest of the query?

You can't say:

```
...
FROM Animal AS [Animal_1]
JOIN
(SELECT [Animal_2].lives_in ... FROM Animal AS [Animal_2] ... GROUP BY [Animal_2].lives_in)
ON [Animal_1].lives_in = [Animal_2].lives_in
```

because that gives spurious results.

It only makes sense to join the subquery on the basis of a column which uniquely identifies the folded vertex, e.g.

```
...
FROM Animal AS [Animal_1]
JOIN
(SELECT [Animal_2].uuid = ... FROM Animal AS [Animal_2] ... GROUP BY [Animal_2].uuid =)
ON [Animal_1].uuid = [Animal_2].uuid =
```

graphql_compiler/tests/test_input_data.py
```diff
@@ -1629,6 +1629,28 @@ def fold_on_output_variable():  # noqa: D103
         type_equivalence_hints=None)
 
 
+def fold_on_foreign_key():  # noqa: D103
```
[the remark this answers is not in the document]
jecdohmann: I can do that. I don't quite understand what we meany by `many_to_one_edge` is that GraphQL-ism to describe edges which traverse using a foreign key?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -205,81 +259,120 @@ def _set_group_by(self, group_by):
...
+            from_table = travel_descriptor.from_table
+            to_table = travel_descriptor.to_table
+            join_descriptor = travel_descriptor.join_descriptor
+            join_clause = sqlalchemy.join(
+                join_clause,
+                to_table,
+                onclause=(from_table.c[join_descriptor.from_column] == to_table.c[
+                    join_descriptor.to_column
+                ])
+            )
```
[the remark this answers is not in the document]
jecdohmann: This is just a generalization of the original logic. It isn't quite a fully functional implementation of multiple traversals, because that requires changing some other stuff in the compilation state.

Can I leave this loop as is but put a NotImplementedError at the beginning of the method if traversal_descriptors has more than 1 element?

graphql_compiler/tests/test_input_data.py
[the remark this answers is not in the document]
bojanserafimov: Yes it is. Tests here are intended for all backends, so I'd rather use GraphQL-isms. The concept of `foreign_key` should not exist in this module.

graphql_compiler/tests/test_compiler.py
```diff
@@ -5907,16 +5939,16 @@ def test_multiple_filters_on_count(self):
                 GROUP BY "Animal_2".uuid
             ) AS folded_subquery_1
             ON "Animal_1".uuid = folded_subquery_1.uuid
-            LEFT OUTER JOIN (
+            JOIN (
                 SELECT
-                    "Animal_4".related_entity AS related_entity,
+                    "Animal_4".uuid AS uuid,
```
jecdohmann: Similar as comment above. This was introduced in #621 but this PR changed to using primary keys for the subquery

graphql_compiler/tests/test_compiler.py
```diff
@@ -5722,7 +5754,7 @@ def test_output_count_in_fold_scope(self):
                 "Animal_1".name AS name,
                 folded_subquery_1.fold_output__x_count AS number_of_children
             FROM schema_1."Animal" AS "Animal_1"
-            LEFT OUTER JOIN (
+            JOIN (
```
jecdohmann: All of these changes are because #666 switched to JOINs instead of LEFT OUTER JOINs but these tests were made in #621

graphql_compiler/compiler/emit_sql.py
```diff
@@ -226,33 +280,61 @@ def _set_group_by(self, group_by):
...
     def _construct_fold_subquery(self, subquery_from_clause):
         """Combine all parts of the fold object to produce the complete fold subquery."""
-        return sqlalchemy.select(
+        select_statement = sqlalchemy.select(
             self.outputs
         ).select_from(
             subquery_from_clause
-        ).group_by(
+        )
+        # Factor our GROUP BY because MSSQL won't use it
```
chewselene: nit: I think this is a typo: our -> out

graphql_compiler/compiler/emit_sql.py
```diff
@@ -435,7 +533,8 @@ def backtrack(self, previous_location):
     def traverse(self, vertex_field, optional):
         """Execute a Traverse Block."""
         if self._current_fold is not None:
-            raise NotImplementedError('Traversals inside a fold are not implemented yet.')
+            raise NotImplementedError('Traversals inside a fold are not implemented '
+                                      'yet {}.'.format(self))
```
chewselene: `self` here is a `CompilationState`, right? I don't see a `__str__` method for it so how does it look when printed? I don't know if this is necessary information to give the user.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -267,47 +349,59 @@ def _get_fold_outputs(self, fold_scope_location, join_descriptor, all_folded_out
...
+            raise AssertionError(u'No traversed vertices visited. '
+                                 u'Invalid state encountered during fold {}'.format(self))
 
         # for now we only handle folds containing one traversal (i.e. join)
-        if len(self.join_info) > 1:
-            raise NotImplementedError('Folds containing multiple traversals are not implemented.')
-        edge, from_alias, to_alias = self.join_info.pop()
+        if len(self._traversal_descriptors) > 1:
+            raise NotImplementedError(u'Folds containing multiple traversals are not '
+                                      u'implemented {}.'.format(self))
```
chewselene: I think this error reads weird. The user will see `Folds containing multiple traversals are not implemented SQLFoldObject(...)` and it is kind of unclear to me what they're supposed to do with the information in `SQLFoldObject(...)`. Here the issue is that we haven't implemented something, not that the `SQLFoldObject` is being used incorrectly. Does having the `SQLFoldObject` information really benefit them more than seeing that there are multiple traversals inside a fold in the query itself?

graphql_compiler/tests/test_compiler.py
```diff
@@ -4341,6 +4341,38 @@ def test_fold_on_output_variable(self):
...
         '''
+
+        check_test_data(self, test_data, expected_match, expected_gremlin, expected_mssql,
+                        expected_cypher, expected_postgresql)
+
+    def test_fold_on_many_to_one_edge(self):
+        test_data = test_input_data.fold_on_many_to_one_edge()
+
+        # Even though this is a many to one edge, primary key should
+        # be the join predicate for the folded subquery
```
chewselene: nit: use the edge name instead of `this`. Sounds a little ambiguous as is. Also, the comment is a full sentence so it should have a period.

PR #646, opened 2019-11-12
Validate that there is only one vertex type name for each table

pmantica1:


graphql_compiler/schema_generation/sqlalchemy/utils.py
```diff
@@ -14,6 +14,19 @@ def validate_that_tables_have_primary_keys(tables):
                                          .format(table.name, table.schema))
 
 
+def validate_that_sqlalchemy_tables_have_a_single_vertex_name(vertex_name_to_table):
+    """Validate that each SQLAlchemy Table has only one corresponding vertex type name."""
+    table_to_vertex_name = {}
```
obi1kenobi: Aren't we already constructing this dict in a way where we actually end up using it in one of the other PRs that are already in flight? I'm all for aggressive validation, but let's not duplicate work -- especially when that work scales proportionally to the size of the entire schema.

PR #647, opened 2019-11-12
Create edge descriptors from foreign keys

pmantica1:
In this diff, I add a function that generates DirectEdgeDescriptor objects from foreign keys.  I will add documentation for it in a latter diff. 

I will include documentation in a latter diff. (We might want to document a function that generates both the DirectEdgeDescriptor objects and the corresponding names instead).


graphql_compiler/schema_generation/sqlalchemy/edge_descriptors.py
```diff
@@ -56,3 +57,54 @@ def validate_edge_descriptors(vertex_name_to_table, direct_edges):
...
+
+    return direct_edge_descriptors
+
+
+def _get_table_to_vertex_name(vertex_name_to_table):
+    """Return a mapping of SQLAlchemy Table objects to their vertex type names."""
+    table_to_vertex_name = {}
+
+    for vertex_name, table in vertex_name_to_table.items():
+        table_to_vertex_name[table] = vertex_name
```
obi1kenobi: Assert that this line doesn't accidentally overwrite something? It shouldn't, but better safe than sorry.

graphql_compiler/schema_generation/sqlalchemy/edge_descriptors.py
```diff
@@ -56,3 +57,54 @@ def validate_edge_descriptors(vertex_name_to_table, direct_edges):
...
+    table_to_vertex_name = _get_table_to_vertex_name(vertex_name_to_table)
+    direct_edge_descriptors = set()
+
+    number_of_composite_foreign_keys = 0
+    for vertex_name, table in vertex_name_to_table.items():
+        for fk_constraint in table.foreign_key_constraints:
+            foreign_key_columns = fk_constraint.columns
+            # The .elements attribute refers to a list of ForeignKey objects.
+            referenced_columns = [element.column for element in fk_constraint.elements]
+            if len(foreign_key_columns) != 1 or len(referenced_columns) != 1:
```
obi1kenobi: This doesn't seem right. If either of these is somehow `0`, we'll still do `number_of_composite_foreign_keys += 1` which is wrong. I think this needs to go from `if-else` to `if-elif-else` with maybe an `AssertionError` in the `else`.

graphql_compiler/tests/schema_generation_tests/test_sqlalchemy_schema_generation.py
```diff
@@ -184,6 +186,80 @@ def test_index_generation_from_unique_constraint(self):
         )
 
 
+@pytest.mark.filterwarnings('ignore: Ignored * edges implied by composite foreign keys*')
```
obi1kenobi: Are you sure this is correct? This seems to be written in glob syntax (`*` as a wildcard) rather than regex (`.*` as a wildcard). I don't remember which was expected by `filterwarnings` but I'd be somewhat surprised if it's glob syntax.

graphql_compiler/tests/schema_generation_tests/test_sqlalchemy_schema_generation.py
```diff
@@ -186,7 +186,7 @@ def test_index_generation_from_unique_constraint(self):
         )
 
 
-@pytest.mark.filterwarnings('ignore: Ignored * edges implied by composite foreign keys*')
+@pytest.mark.filterwarnings('ignore: Ignored .* edges implied by composite foreign keys.*')
```
pmantica1: I tested that it works.

PR #671, opened 2019-11-18
Permit multiple traversals within a fold for Postgres and MSSQL

jcd2020:
This PR leaves the SQLFoldObject unmodified but required substantial modifications to the logic of the CompilationState object. This is blocked on #613 

1. Updated `_relocate` function to use query_path, fold_path tuple as alias key if the current location is within a fold.

2. Update `traverse` to call SQLFoldObject's `visit_traversed_vertex` method to update traversal descriptors.

3. Update `fold` method to no longer call `visit_output_vertex` on folded vertex, because output isn't necessarily the same as folded vertex.

4. Update `fold` method to set the current location and current alias to the fold vertex and fold alias.

5. Update `mark_location` to call `visit_output_vertex` if this location is the last location inside a fold before backtracking or unfolding.

6. Updated `unfold` to make sure the outer vertex alias is taken from the alias dictionary at the base location of the fold we are ending.


graphql_compiler/compiler/emit_sql.py
```diff
@@ -929,11 +963,21 @@ def emit_code_from_ir(sql_schema_info, ir):
...
-            state.mark_location()
+            # If we are in a fold and the next block either backtracks
+            # or unfolds, then we have traversed to the deepest point
+            # of the fold and need to set the output columns
+            backtracks_next = isinstance(ir.ir_blocks[index + 1], blocks.Backtrack)
+            unfolds_next = isinstance(ir.ir_blocks[index + 1], blocks.Unfold)
+            if state.is_in_fold() and (unfolds_next or backtracks_next):
+                is_last_fold_block = True
+            else:
+                is_last_fold_block = False
```
bojanserafimov: This needs to be untangled a bit. Let me walk you through my experience as a reader:
1. We see if we're in the deepest point in the fold
2. We rephrase `deepest` as `last`, adding confusion. The meaning of deepest was documented inside the comment, but what last means is left for me to guess.
3. We pass `is_last_fold_block` into `mark_location`, where this happens:
```
        if is_last_fold_block:
            # Pull out the output columns on the last block inside the fold.
            self._current_fold.visit_output_vertex(...
```
I'm supposed to understand that the output columns need to be pulled on the last block. With no memory of what "pulling columns" menas or what the "last block is", this is a large jump in reasoning, and the only comment explaining this is in a totally different place. Furthermore, the function is not called `pull_output_columns` as this comment leads me to believe, but `visit_output_vertex`, which judging by the name may entail many more things .
4. We look inside the `visit_output_vertex` function documentation and see... `Update output columns when visiting`, finally unraveling the mystery of the plumbing. Now I can finally get into trying to understand the content of this code.

This code seems understandable when you work on it, but give it 2 months and you'll have no idea why this is happening. To tidy up, you have to have a higher respect for function specifications. Just follow these 3 rules:
1. For each function, the name and spec fully define the function
2. The implementation of the function implements the spec and nothing else
3. The usage of the function uses the spec and does not rely on anything else
4. The tests of the function test that the implementation implements the spec and nothing else

graphql_compiler/compiler/emit_sql.py
```diff
@@ -876,25 +902,29 @@ def unfold(self):
...
-        self._aliases[(self._current_location.at_vertex().query_path, None)] = outer_vertex
-
         # clear the fold from the compilation state
         self._current_fold = None
         self._fold_vertex_location = None
         self._current_alias = outer_vertex
+        self._current_location = self._current_location.base_location
 
-    def mark_location(self):
+    def mark_location(self, is_last_fold_block):
```
bojanserafimov: The meaning of `is_last_fold_block` is not obvious at all. Either document it well or compute it inside this function (passing the whole ir_block in as an argument). The fact that the very definition of this concept depends on the restriction that there's only one output scope inside fold makes me think that we should not glorify it as a named concept leaking outside the `CompilationState` class.

graphql_compiler/compiler/emit_sql.py
jecdohmann: Hi, good comments, I agree completely. Let me know what you think of the code I just pushed. It's possible also that changing the name or documentation of `mark_location` is necessary, but I'm not sure if we need to jump to that if the current docs suffice.

graphql_compiler/compiler/emit_sql.py
jecdohmann: Good idea, done.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -876,25 +902,41 @@ def unfold(self):
...
+        else:
+            is_deepest_fold_vertex = False
+
+        # Visit output vertex if the current location is the deepest vertex in the fold
+        if is_deepest_fold_vertex:
+            # If we are at the deepest vertex in the fold, then the current location is the output
+            # vertex of the fold and we must visit it now to select its output columns.
+            self._current_fold.visit_output_vertex(self._current_alias,
+                                                   self._current_location,
+                                                   self._all_folded_fields)
```
bojanserafimov: Since `visit_output_vertex` should be called exactly once, can we assert that? Better yet, why not have a `visit_vertex` instead of `visit_output_vertex` that's expected to be called on every vertex inside the fold. That will allow us to move all of this fold logic (checking for outputs) inside the `SQLFoldObject`, which is the class that we made to contain all the fold logic. Inside that method we can assert we've only found one vertex that actually has outputs. That will also remove the need to pass the unintuitive `next_ir_block` in here.

jecdohmann: I believe we would still need to determine whether this is an output vertex while in the context of the CompilationState because the SQLFoldObject has no way of knowing without any additional information whether the vertex is an output vertex.

At best, we could change the logic to have a visit_vertex method that receives a flag indicating this an output vertex when called from inside this if statement. 

No matter what, I don't think we get away with having some sort of version of this if-logic occurring in the CompilationState.

jecdohmann: Also, note that we do assert that `visit_output_vertex` is called exactly once inside of that function. But technically that check only works if we assume that we never pass in `None` as the `output_alias`. I can change the error checking to catch that case. That way we will always be ensured only one call to `visit_output_vertex` occurs.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -537,38 +539,39 @@ def _get_fold_output_column_clause(self, fold_output_field):
...
         if self._output_vertex_alias is not None:
             raise AssertionError(u'Cannot visit multiple output vertices in one fold. '
                                  u'Invalid state encountered during fold {}'.format(self))
-        self._output_vertex_alias = output_alias
-        self._outputs = self._get_fold_outputs(fold_scope_location,
-                                               all_folded_outputs)
+        if fold_scope_location.at_vertex() in all_folded_outputs:
+            # If this fold scope is in the output dict then it contains an output directive.
+            self._output_vertex_alias = output_alias
+            self._outputs = self._get_fold_outputs(fold_scope_location, all_folded_outputs)
```
bojanserafimov: Can we assert that this branch is taken only once?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -876,25 +906,29 @@ def unfold(self):
...
-        self._aliases[
-            (self._fold_vertex_location.base_location.query_path,
-             self._fold_vertex_location.fold_path)
-            if self._current_fold is not None else (self._current_location.query_path, None)
-        ] = self._current_alias
+        if self._current_fold is not None:
+            # If we are in a fold, then we visit the current vertex
+            self._current_fold.visit_vertex(self._current_alias,
+                                            self._current_location,
+                                            self._all_folded_fields)
```
bojanserafimov: You're passing `all_folded_fields` for `all_folded_outputs`, which smells buggy.

jecdohmann:
@chewselene  I realized I made in error in how multiple traversals should work.

The joins should go inside the XML PATH clause because otherwise you get a different XML PATH string for each path you can take to the innermost vertex.

This fiddle has the right implementation http://sqlfiddle.com/#!18/c8ce22/9

If the query in there were written, how this current PR understands traversals, such that Neighborhoods were JOINed to Cities in the FROM clause of the subquery, then we would end up 3 entries for Boston, 2 for NY, 2 for Ithaca.

That example is here: http://sqlfiddle.com/#!18/c8ce22/10

obi1kenobi:
I'd write down the GraphQL query that each of those fiddles attempts to represent as SQL, otherwise it's very difficult (especially in the incorrect case) to figure out what the correct answer should be.

I'm also still confused about the `LEFT OUTER JOIN` in the top-level `SELECT`. Are you confident that it is correct and shouldn't be an `INNER JOIN` instead?

jecdohmann:
> I'd write down the GraphQL query that each of those fiddles attempts to represent as SQL, otherwise it's very difficult (especially in the incorrect case) to figure out what the correct answer should be.
> 
> I'm also still confused about the `LEFT OUTER JOIN` in the top-level `SELECT`. Are you confident that it is correct and shouldn't be an `INNER JOIN` instead?

Yep. You're right. No LEFT JOIN. Should look like this http://sqlfiddle.com/#!18/c8ce22/11

The GraphQL equivalent would be something like

```
{ 
   Cities {
     out_Neighborhoods_NeighborhoodIn @fold {
       out_Streets_StreetIn {
         name @output(out_name: "streets")
       }
     }
   }
}
```

chewselene:
PostgreSQL folded traversals implemented in #803 
MSSQL folded traversals implemented in #830

PR #673, opened 2019-11-19
Add Python 3.8 to the test matrix and to the supported list.

obi1kenobi:
Also removing Python 3.5 since we weren't testing with it, and we won't support it once we drop 2.7 support anyway.


obi1kenobi:
The Python 3.8 bits succeeded. I think the easiest way to merge this would be to drop Python 2 support first, though, because that is holding back a number of packages (`pylint` being one of them) and this is causing issues in the build pipeline.

PR #675, opened 2019-11-20
Add diff-cover as a tool for automated coverage testing

pmantica1:
This is a test to test travis CI. Please do not review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,8 @@ matrix:
   - name: "Python 3.7 unit tests"
     python: "3.7"
     script:
-      - pipenv run py.test --cov=graphql_compiler graphql_compiler/tests
+      - pipenv run py.test --cov=graphql_compiler --cov-report=xml graphql_compiler/tests
+      - pipenv run diff-cover coverage.xml --fail-under=80
 after_success:
   - docker-compose down
   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,7 @@ matrix:
   - name: "Python 3.7 unit tests"
     python: "3.7"
     script:
-      - pipenv run py.test --cov=graphql_compiler --cov-report=xml graphql_compiler/tests
-      - pipenv run diff-cover coverage.xml --fail-under=80
+      - pipenv run py.test --cov=graphql_compiler graphql_compiler/tests
 after_success:
   - docker-compose down
   - coveralls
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -25,11 +25,13 @@ matrix:
...
     script:
-      - pipenv run py.test --cov=graphql_compiler graphql_compiler/tests
+      - pipenv run py.test --cov=graphql_compiler --cov-report=xml graphql_compiler/tests
+      - pipenv run diff-cover coverage.xml --fail-under=80
   - name: "Python 3.6 unit tests"
     python: "3.6"
     script:
-      - pipenv run py.test --cov=graphql_compiler graphql_compiler/tests
+      - pipenv run py.test --cov=graphql_compiler --cov-report=xml graphql_compiler/tests
+      - pipenv run diff-cover coverage.xml --fail-under=80
```
pmantica1: For python 2.7 and 3.6, I changed the coverage report format to xml since `diff-cover` requires an xml format. For python 3.7, I did not change the coverage report format for python3.7 since coveralls expects the coverage report to be emitted stdout in the standard format. So essentially, I kept one report that we could send to coveralls. 

I am not sure how coveralls merges the coverage report of parallel builds. So maybe we could be underrepresenting our code coverage somehow by only sending coveralls a coverage report of a test run with python3+. However, we are about to discontinue python2 and we don't really have that many tests that behave differently depending on whether the code is python2 or 3.

obi1kenobi: I don't think having different coverage configurations is a good idea. I also don't think we should only partially enforce `diff-cover` rules or only partially send coverage to Coveralls.

Are there any alternatives?

pmantica1: Nevermind I was wrong about this. coveralls relies on the `.coverage` file that is generated every time that `pipenv run py.test --cov=graphql_compiler graphql_compiler/tests` is run. It is also generated even when we use the '--cov-report=xml' option. So we can fully enforce `diff-cover` and we can fully send coverage to coveralls. 

Again, I still don't know completely how the results from `.coverage` get merge, but it seems overall that adding `diff-cover` is a good idea. 

The reason I had mistaken beliefs was that when I made all of the scripts return the coverage report in xml format, coveralls took forever to registered a report. But I guess that this was just bad luck. As you can see in my latest commit, I have all of the pytest scripts give me an xml report and coveralls registered a report.

pmantica1: I just thought of something. This will fail if we have feature branches. I am going to close this PR for this reason.

obi1kenobi: Why will it fail? Does it always compare against `master` or something? And how do you know this?

PR #681, opened 2019-11-27
Add APIs for building query interpreters with our usual GraphQL semantics

obi1kenobi:


bojanserafimov:
I'm still not sure we have the right approach for the separation of interpreter and cross-schema query plan execution. What makes the most sense for me is to only define the interpreter for single-vertex queries (no traversals), and use the cross-schema query plan executor for the rest. It already has intelligence on which part of the query to execute first, and we don't want to replicate that. It also knows to convert edge traversals to filters, which is something you manually had to add as a "hint" to the interpreter.

obi1kenobi:
Agreed that there is a lot of overlap and there's still some design work to be done, hence the `draft PR` status.

My thoughts were that the interpreter shouldn't try to do query planning, and should instead just take the IR it is given and execute it as-is. I think there's an intermediate component to be built here that would perform the query planning step and decide on the order of operations, and that component would have a lot in common (in design and hopefully in code) with the cross-schema execution planner.

> already knows how to format the result
While the cross-schema execution knows how to format the result, it's not designed for the workload the interpreter is handling (nor should it be). In the cross-schema execution setting, we issue a query and get a set of result "dataframes" with their rows in arbitrary order, and we have to join and fold them appropriately -- which we do by assembling join indexes. In the interpreter, there is no arbitrary order -- we know precisely and explicitly which rows go together, and we don't need join indexes. This is going to be significantly more performant and I'd be happy to work with you on some benchmarks if you'd like hard evidence for it.

> It also knows to convert edge traversals to filters, which is something you manually had to add as a "hint" to the interpreter.
I didn't get what you meant here. Would you mind clarifying?

obi1kenobi:
Thoroughly superseded by the `interpreted_mode_v3` branch. That one is still somewhat raw (more code with less documentation), will open a draft PR once it's worth reviewing.

Issue #672, opened 2019-11-19
Upgrade to GraphQL-core v3 as soon as it is available; drop Python 2.x support


obi1kenobi:
This is a blocker for our 2.0 release:
https://github.com/graphql-python/graphql-core-next/issues/53

pmantica1:
GraphQL-core v3 does not allow for multiple directives in the same field in a query. In other words this query is invalid:
```
{
     ObjectType {
           field @directive1 @directive 
     }
}
```
We need to change tests/documentation/other code to adjust for this change.

obi1kenobi:
In your example, did you mean to make the directives be differently-named, `directive1` and `directive2`? If I understand this correctly, you are saying that a field may only have up to one directive?

pmantica1:
I am just saying that this issue https://github.com/kensho-technologies/graphql-compiler/issues/346
is a blocker. While GraphQL-core does not enforce directive uniqueness peer location, GraphQL-core-next does.

obi1kenobi:
Directives of different names are allowed to be in the same location by the GraphQL spec, see the link in #346. GraphQL-core-next additionally supports repeatable directives, which we should make use of to avoid breaking existing queries: https://github.com/graphql-python/graphql-core-next/blob/0832f0a73d60e8954f1defb7a397ff02cc7621e4/tests/validation/test_unique_directives_per_location.py#L16

obi1kenobi:
GraphQL-core v3 is out: https://github.com/graphql-python/graphql-core-next/releases

PR #683, opened 2019-11-27
Remove Python 2 support

obi1kenobi:


CONTRIBUTING.rst
```diff
@@ -127,43 +124,6 @@ that your files start with a line like:
...
-   ./scripts/make_pipenv_lockfiles.sh
-
-Then, re-run
-
-::
-
-   pipenv sync --dev
-
-to install the relevant dependencies.
-
```
pmantica1: Remember to update the readthedocs rst files too. Maybe we should just update readthedocs at this point. We are only missing the query printing and the execution model sections.

obi1kenobi: I think I updated both `CONTRIBUTING.rst` files. Is there another file I should update?

pmantica1: Sorry, I missed the changes in the bottom.

pmantica1: Sorry, I missed the changes in the bottom.

obi1kenobi:
GraphQL-core v3 is out, and merging this PR is a prerequisite for migrating to it: #672 is related.

scripts/make_pipenv_lockfile.sh
```diff
@@ -11,22 +11,16 @@ shopt -s globstar nullglob
...
-pipenv lock --python="$(which python2)" --dev
-mv Pipfile.lock Pipfile.py2.lock
+# Delete the existing lockfile.
+echo 'Deleting old lockfile...'
+rm Pipfile.lock || true  # Don't error if the lockfile wasn't there to start.
 
 echo 'Creating Python 3 lockfile...'
 pipenv --rm
-pipenv lock --python="$(which python3)" --dev
+pipenv lock --python="$(which python)" --dev
```
pmantica1: I don't think we need this file anymore.

obi1kenobi:
It helps make sure you have the correct version of Python before locking.

I don't see the harm in keeping it and encouraging people to use it to
avoid Python version issues. It doesn't stop anyone from doing `pipenv
lock` directly, but it does minimize the odds that locking fails because of
a Python version mismatch.

pmantica1:
I am still not a huge fan of not this `scripts/make_pipenv_lock.sh` file for the following reasons:
1. pipenv already gives you a warning when you are using a version of python other than the one specified by the Pipfile. 
2. If we are going to have a script to enforce this sanity check when running `pipenv lock` why not also have a script to enforce it when running `pipenv install` or `pipenv sync`. It seems weird to me to special case `pipenv lock` since when you are creating your virtual environment with `pipenv install` you also need to make sure you are using the right version of python. 

You mentioned that you don't see the harm in keeping it, but I think that these file contents are confusing for a new user. A new user would:
1. Randomly stumble onto this file because there is currently no documentation about it in the contributing docs.
2. Parse all the semi-complicated bash. 
3. Become confused about what is the purpose of this file, specifically why we have a file like this one for creating a lock file but not a virtual environment. 

At the end of the day, I think it makes more sense just include in the contributing docs the following phrase "be sure you are using the right version of python <link_to_how_to_specify_the_version_of_python_in_pipenv>". However, I don't think that we really need to do so because pipenv already gives you a warning if you are using the wrong python version.

obi1kenobi:
> If we are going to have a script to enforce this sanity check when running pipenv lock why not also have a script to enforce it when running pipenv install or pipenv sync. It seems weird to me to special case pipenv lock since when you are creating your virtual environment with pipenv install you also need to make sure you are using the right version of python.

`pipenv sync` and `pipenv install` affect only your local environment state, whereas `pipenv lock` updates the lockfile which affects everyone else's environment state too.

That said, `pipenv lock` seems to have weird behavior which threw me off. If you already have a pipenv-managed virtualenv, it seems to use that one and ignore whatever `python` resolves to in the current shell — so I thought that it doesn't actually verify the Python version. If you remove the virtualenv, it does check the Python version, so the script isn't necessary.

PR #686, opened 2019-12-02
Add filter with tagged parameter inside a fold tests

chewselene:
Adding a couple of tests about filtering with tagged parameters inside of a fold scope. For PostgreSQL, we are able to filter `_x_count` using tagged parameters, but not regular fields. I would particularly like some eyes on the results match/gremlin/cypher since I'm not very familiar with writing queries in these languages.

These tests also made me realize that we were only searching for `_x_count`s in the first level of the binary expression used in the filter. This PR includes an update to recursively search for `_x_count`s used in filters.


graphql_compiler/compiler/emit_sql.py
```diff
@@ -107,6 +108,37 @@ def _find_columns_used_outside_folds(sql_schema_info, ir):
     return used_columns
 
 
+def _find_tagged_parameters(expression_from_filter):
+    """Return True if the expression contains a ContextColumn (i.e. a tagged parameter)."""
```
obi1kenobi: From the docstring of `ContextColumn` it seems that it is used when referencing concrete columns in a table, rather than fields in the GraphQL type corresponding to that table. However, for tagged parameters, we know the tag must have been on a field in the GraphQL type. Do we use `ContextColumn` here because that field access (i.e. `ContextField` type) has already been lowered away in a previous step to a `ContextColumn`, or is there another reason for it?

Maybe I'm being paranoid here. I'm just worried that here we seem to be implicitly assuming that `ContextColumn` has to mean the presence of a tagged parameter. Whether that is true or false right now isn't obvious to me, and I'm even more worried about whether that will remain true or false in the future.

And maybe the answer is just "this is temporary code until we implement this functionality". If that's the case, I'd like to make sure we have a plan on how we'd actually implement it, so that the code doesn't accidentally become permanent :)

graphql_compiler/compiler/emit_sql.py
```diff
@@ -125,10 +157,8 @@ def _find_folded_fields(ir):
         # Check the left and right side of Filter predicates to see if _x_count is involved.
         if global_operation_start and isinstance(block, blocks.Filter):
             for subexpression in (block.predicate.left, block.predicate.right):
-                if (isinstance(subexpression, FoldedContextField) and
-                        subexpression.fold_scope_location.field == COUNT_META_FIELD_NAME):
-                    fold_scope_location = subexpression.fold_scope_location
-                    fold_path = fold_scope_location.fold_path
+                fold_scope_location, fold_path = _find_x_count_used_in_global_filter(subexpression)
```
obi1kenobi: I don't think I originally reviewed this code, so please pardon my confusion here. I'm just not sure I understand why we seem to be inferring (from the lowered IR) which folded fields are present where, when this information is already available from the query metadata and must have remained unchanged throughout lowering (or else the semantics of the query were altered by the lowering process and we have a bug in the system).

chewselene: good point! i will switch to using the output metadata table from the IR.

graphql_compiler/compiler/emit_sql.py
chewselene: looking into this now...

graphql_compiler/compiler/emit_sql.py
```diff
@@ -113,6 +113,22 @@ def _find_columns_used_outside_folds(sql_schema_info, ir):
     return used_columns
 
 
+def _find_tagged_parameters(expression_from_filter: Expression) -> bool:
+    """Return True if the expression contains a ContextField (i.e. a tagged parameter)."""
+    has_context_fields = False
+
+    def visitor_fn(expression_to_visit: Expression) -> Expression:
+        """Update has_context_fields if a ContextField is found."""
+        nonlocal has_context_fields
```
obi1kenobi: Yay Python 3!

PR #688, opened 2019-12-03
Add the first batch of type hints.

obi1kenobi:
Steps:
- `pip install pytest-annotate` (another Kensho open-source project, a plugin for `pytest` that uses `pyannotate` on each test)
- `pytest --annotate-output=./localdata/annotations.json` to run tests and capture the *actual* types with which functions are called.
- Work around https://github.com/dropbox/pyannotate/issues/92 by opening the file and removing all entries with `<locals>` in them. If any annotations list becomes empty, delete the entire dict.
- Pick a few files to annotate, and run `pyannotate --py3 --type-info ./localdata/annotations.json <your files>`. This will preview all the changes, spot-check them quickly for any major issues. Don't worry about smaller things, you'll touch this up by hand in the next step.
- The last command was a dry-run, append `-w` to actually perform the changes.
- Touch up the types in the file as necessary. You'll probably see a lot of things like `Union[List[str], List[date], List[int]]` that may need to be replaced with `List[Any]` depending on what the function does. Read the function and *engage your brain* to figure out what makes sense :)
- Fix import style with `isort --settings-path=setup.cfg --recursive ./graphql_compiler/`
- Paint it black: `black .`
- Argue with `mypy` until no errors show up in `mypy --ignore-missing-imports <your files>`. Repeat the last few steps as necessary, using `type: ignore` only as a last resort.
- Yay more type hints! Open a PR.

### DIFF
--- a/graphql_compiler/tests/integration_tests/integration_test_helpers.py
+++ b/graphql_compiler/tests/integration_tests/integration_test_helpers.py
@@ -1,13 +1,25 @@
 # Copyright 2018-present Kensho Technologies, LLC.
+from datetime import date
 from decimal import Decimal
+from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
 
+from graphql.type.schema import GraphQLSchema
+from pyorient.orient import OrientDB
+from redisgraph.client import Graph
 import six
+from sqlalchemy.engine.base import Engine
+
+from graphql_compiler.schema.schema_info import SQLAlchemySchemaInfo
+from graphql_compiler.tests.test_data_tools.neo4j_graph import Neo4jClient
 
 from ... import graphql_to_match, graphql_to_redisgraph_cypher, graphql_to_sql
 from ...compiler import compile_graphql_to_cypher
 
 
-def sort_db_results(results):
+T = TypeVar('T')
+
+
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -16,15 +27,15 @@ def sort_db_results(results):
     Returns:
         List[Dict], sorted DB results.
     """
-    sort_order = []
+    sort_order: List[str] = []
     if len(results) > 0:
         sort_order = sorted(six.iterkeys(results[0]))
 
-    def sort_key(result):
+    def sort_key(result: Dict[str, Any]) -> Tuple[Tuple[bool, Any], ...]:
```
MichaelaShtilmanMinkin: I've gone down a bit of a rabbithole learning about `...` and I'm still not entirely sure what it means in this context. Would you mind explaining?

obi1kenobi: This is a special case explained in https://github.com/python/mypy/issues/184

`Tuple[T, ...]` means an arbitrary-length tuple of `T` elements -- the `tuple()` analogue of `List[T]`. This is different than `Tuple[T]` which is a tuple of *one* `T` element.

PR #689, opened 2019-12-03
Add mypy dependency and linter check.

obi1kenobi:
We currently don't have any types, so we can't yet turn on mypy strict mode. We'll need many more PRs like #688 before we have enough types for that to work.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,6 +21,7 @@ flake8 = ">=3.6.0,<4"
 flake8-bugbear = ">=19.8.0"
 flake8-print = ">=3.1.0,<4"
 isort = ">=4.3.4,<5"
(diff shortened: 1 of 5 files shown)


pmantica1:
I took the liberty of fixing your isort linting error.

obi1kenobi:
> Please merge after testing that `scripts/lint.sh --diff` works as expected for black and mypy.

Good call. It seems to work fine:

```
*** Running black... ***

All done! ✨ 🍰 ✨
157 files would be left unchanged.

*** End of black run; exit: 0 ***
<...>
*** Running mypy... ***

Success: no issues found in 1 source file

*** End of mypy run, exit: 0 ***
```

PR #690, opened 2019-12-03
Add a documentation test

pmantica1:
In this diff, I add a sphinx test in travis to prevent common documentation errors such as invalid internal links.


scripts/lint.sh
```diff
@@ -83,6 +83,12 @@ pylint $pylint_lintable_locations
 pylint_exit_code=$?
 echo -e "\n*** End of pylint run, exit: $pylint_exit_code ***\n"
 
+echo -e '\n*** Running sphinx-build to test documentation... ***\n'
+# The dummy builder doesn't produce any output but still needs a build directory parameter.
+sphinx-build -nqW -b dummy docs/source/ docs/build/ --keep-going
```
pmantica1: This reads as "run sphinx-build in a nitpicky and quiet mode. Treat warnings as warnings. Use the dummy builder and don't produce any output. If you encounter a warning keep going until the end, and output all warnings as errors." 

More information here. 
https://www.sphinx-doc.org/en/master/man/sphinx-build.html

`sphinx-build` seems to be command that is very similar to `make html`. For some reason though, `make html` doesn't have that many options.

pmantica1: I was inspired by https://medium.com/akeneo-labs/documentation-on-steroids-with-sphinx-github-travis-ci-and-platform-sh-89218513d136

pmantica1: (I could not find another repo running sphinx tests).

obi1kenobi: Does `sphinx-build` have long-form versions of the single-letter flags? E.g. `--quiet` instead of `-q`? If so, consider using them -- the verbose version of the flag is optimizing for readability.

Meta point: if you feel that you have to explain your code via a comment in the PR, consider either working on further clarifying the code, or adding a comment to the code itself. When someone has questions about the code down the line, they'd much rather look at an inline comment than look at the PR.

pmantica1: It does not. I just added mostly copy-pasted documentation from the sphinx website as comments to this script. (All of it is copy-pasted except for the dummy builder part, which is not documented in the link).

scripts/lint.sh
```diff
@@ -107,6 +123,7 @@ if  [[
     echo -e "pydocstyle on tests exit: $pydocstyle_test_exit_code"
     echo -e "pylint exit: $pylint_exit_code"
     echo -e "bandit exit: $bandit_exit_code"
+    echo -e "sphinx-build exit: $sphinx_build_exit_code"
```
obi1kenobi: Either move this up one or move the sphinx check to after the bandit check. Let's keep these lists of commands parallel.

PR #691, opened 2019-12-04
Add type hints to test_helpers.py.

obi1kenobi:
Same deal as #688, same process, different file.

### DIFF
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -26,7 +26,12 @@ from ..compiler.subclass import compute_subclass_sets
 from ..debugging_utils import pretty_print_gremlin, pretty_print_match
 from ..macros import MacroRegistry, create_macro_registry, register_macro_edge
 from ..query_formatting.graphql_formatting import pretty_print_graphql
-from ..schema import CUSTOM_SCALAR_TYPES, is_vertex_field_name
+from ..schema import (
+    CUSTOM_SCALAR_TYPES,
+    ClassToFieldTypeOverridesType,
+    GraphQLSchemaFieldType,
+    is_vertex_field_name,
+)
 from ..schema.schema_info import (
     CommonSchemaInfo,
     DirectJoinDescriptor,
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/test_helpers.py
```diff
@@ -3,30 +3,56 @@
...
+from .test_compiler import CompilerTests
+from .test_emit_output import EmitCypherTests, EmitGremlinTests, EmitMatchTests
+from .test_end_to_end import QueryFormattingTests
+from .test_ir_generation import IrGenerationTests
+from .test_macro_expansion import MacroExpansionTests
+from .test_sqlalchemy_extensions import CommonIrLoweringTests
 
 
+# The valid types that a field inside an object or interface in the GraphQL schema may be.
+GraphQLSchemaFieldType = Union[GraphQLList, GraphQLNonNull, GraphQLScalarType]
```
pmantica1: I think that we probably want this type in a type defs file. I can see it being used multiple times throughout the codebase.

graphql_compiler/tests/test_helpers.py
```diff
@@ -744,7 +784,11 @@ def generate_schema_graph(orientdb_client):
     return get_orientdb_schema_graph(schema_data, index_query_data)
 
 
-def generate_schema(orientdb_client, class_to_field_type_overrides=None, hidden_classes=None):
+def generate_schema(
+    orientdb_client: OrientDB,
+    class_to_field_type_overrides: Optional[Dict[str, Dict[str, GraphQLSchemaFieldType]]] = None,
```
MichaelaShtilmanMinkin: I think we reference `class_to_field_type_overrides` a few times in the code. Given `Optional[Dict[str, Dict[str, GraphQLSchemaFieldType]]]` is quite lengthy, too, I would consider storing the whole annotation so it can be re-used.

graphql_compiler/tests/test_helpers.py
```diff
@@ -744,7 +779,11 @@ def generate_schema_graph(orientdb_client):
     return get_orientdb_schema_graph(schema_data, index_query_data)
 
 
-def generate_schema(orientdb_client, class_to_field_type_overrides=None, hidden_classes=None):
+def generate_schema(
+    orientdb_client: OrientDB,
+    class_to_field_type_overrides: Optional[ClassToFieldTypeOverridesType] = None,
+    hidden_classes: Optional[Set[str]] = None,
+) -> Tuple[GraphQLSchema, Dict[Union[GraphQLInterfaceType, GraphQLObjectType], GraphQLUnionType]]:
```
pmantica1: I am sorry for not mentioning this earlier, but I think it might be worthwhile to create a `TypeEquivalenceHintsType` too. We are probably going to be redefining it a bunch of times otherwise.

PR #692, opened 2019-12-04
Add query pretty-printing documentation to ReadTheDocs

pmantica1:
I chose to place the query pretty-printing documentation at the top of the Query Directives file for the following reasons:
- I don't think that query pretty-printing is important enough to be on the home page.
- I don't think that query pretty-printing merits a page for itself. (It might be a good idea to place it in a "Tools" page in the future). 
- The Query Directives file is a must-see for someone that is new to the compiler and committed to learning about it.


pmantica1:
I updated the placement.

PR #693, opened 2019-12-05
Add type hints to the compiler's test input data file.

obi1kenobi:
Also auto-generated with the same process as #691.

### DIFF
--- a/graphql_compiler/tests/test_input_data.py
+++ b/graphql_compiler/tests/test_input_data.py
@@ -7,12 +7,7 @@ from typing import Dict
 from graphql import GraphQLID, GraphQLInt, GraphQLList, GraphQLString
 
 from ..compiler.compiler_frontend import OutputMetadata
-from ..schema import (
-    GraphQLDate,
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/test_input_data.py
```diff
@@ -29,7 +30,7 @@ def immediate_output():  # noqa: D103
     expected_output_metadata = {
         "animal_name": OutputMetadata(type=GraphQLString, optional=False),
     }
-    expected_input_metadata = {}
+    expected_input_metadata: Dict[str, GraphQLSchemaFieldType] = {}
```
bojanserafimov: Not sure if this is adding to readability or anything. Was it necessary for type checking to work?

obi1kenobi:
Yes. Mypy isn't yet able to refine type hints based on use, so it requires
that empty dicts get explicit type hints. Otherwise it would have had to
infer `Dict[Any, Any]` here and that would stay the type of this variable
regardless of how it's used.

PR #697, opened 2019-12-09
Upgrade to newer major linter and test framework versions.

obi1kenobi:
Upgrade `pydocstyle` from v3 to v5, `pylint` from v1.9 to v2.4, and `pytest` from v4 to v5.

Also fixed a handful of newly-discovered lint issues.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -42,7 +42,10 @@ graphql-core = ">=2.1,<3"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
-graphql-compiler = {editable = true, path = "."}  # necessary to make some pylint passes work
+
+# The below is necessary to make a few pylint passes work properly, since pylint expects to be able
+# to run "import graphql_compiler" in the environment in which it runs.
+graphql-compiler = {editable = true, path = "."}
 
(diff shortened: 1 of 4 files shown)


obi1kenobi:
This PR got bitten by an issues in a dev pre-release of a dependency of a dependency: https://github.com/sphinx-doc/sphinx/issues/6887

The advice in https://github.com/pypa/pipenv/issues/1760 suggests that `pipenv` allows a specific pre-release version to be installed without allowing any other pre-releases, by explicitly pinning to the pre-release we want and then *not* including `allow_prereleases = true` in the Pipfile. So that's what I did -- I pinned `black==19.10b0` and removed the `allow_prereleases = true`.

Pipfile
```diff
@@ -42,9 +42,7 @@ graphql-core = ">=2.1,<3"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
+graphql-compiler = {editable = true, path = "."}  # necessary to make some pylint passes work
```
pmantica1: Can you explain how it solves it in a comment? This statement seems to have more implications that I what I am seeing if it is able to fix pylint passes.

PR #699, opened 2019-12-11
Fix a couple of type errors and add more types.

obi1kenobi:
A couple of the uses of `use_all_backends` here were accidentally passing `str` instead of `Tuple[str]`. Thank mypy for catching them :)

### DIFF
--- a/graphql_compiler/tests/integration_tests/test_backends_integration.py
+++ b/graphql_compiler/tests/integration_tests/test_backends_integration.py
@@ -1,7 +1,7 @@
 # Copyright 2018-present Kensho Technologies, LLC.
 import datetime
 from decimal import Decimal
-from typing import Any, Dict, List, Tuple
+from typing import Any, Callable, Dict, List, Tuple
 from unittest import TestCase
 
 from graphql.type import GraphQLID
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -421,10 +421,10 @@ def test_recurse(self):
         )
     )
     @integration_fixtures
-    def test_fold_basic(self, backend_name):
+    def test_fold_basic(self, backend_name: str) -> None:
         # (query, args, expected_results) tuples.
         # The queries are ran in the order specified here.
-        queries = [
+        queries: List[Tuple[str, Dict[str, Any], List[Dict[str, Any]]]] = [
```
MichaelaShtilmanMinkin: Could we re-use the `expected_results` type here to make this a bit more readable?

obi1kenobi: I'm leaning toward typing first, refactoring as necessary later. Ideally, we'd get type hints everywhere, then take a look at how to streamline and improve the user-friendliness of the public and user-facing functions (e.g. with named typedefs), and then use those typedefs consistently throughout the codebase. For example, it's not clear to me on the basis of the integration tests file whether `QueryParameters = Dict[str, Any]` as a typedef would help or harm readability of the user-facing code.

PR #702, opened 2019-12-12
Query Planning: Add generic range reasoning method

bojanserafimov:
This allows the cost estimator and the paginator to reason about int, uuid, date and datetime ranges in a generic way. This reasoning entails understanding inequality filters, computing filter selectivity given quantile statistics, and choosing good pagination parameters given quantile statistics, respecting existing filters on the query.

### DIFF
--- a/graphql_compiler/cost_estimation/helpers.py
+++ b/graphql_compiler/cost_estimation/helpers.py
@@ -1,7 +1,7 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from graphql import GraphQLInt
 
-from ..schema import GraphQLDateTime, GraphQLDate
+from ..schema import GraphQLDate, GraphQLDateTime
 
 
 def is_datetime_field_type(schema_info, vertex_name, field_name):
--- a/graphql_compiler/cost_estimation/helpers.py
+++ b/graphql_compiler/cost_estimation/helpers.py
@@ -3,26 +3,33 @@ from graphql import GraphQLInt
 
 from ..global_utils import is_same_type
 from ..schema import GraphQLDate, GraphQLDateTime
+from ..schema.schema_info import QueryPlanningSchemaInfo
 
 
-def is_datetime_field_type(schema_info, vertex_name, field_name):
+def is_datetime_field_type(
+    schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
+) -> bool:
     """Return whether the field is of type GraphQLDateTime."""
     field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
     return is_same_type(GraphQLDateTime, field_type)
 
 
-def is_date_field_type(schema_info, vertex_name, field_name):
+def is_date_field_type(
+    schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
+) -> bool:
     """Return whether the field is of type GraphQLDate."""
     field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
     return is_same_type(GraphQLDate, field_type)
 
 
-def is_int_field_type(schema_info, vertex_name, field_name):
+def is_int_field_type(
+    schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
+) -> bool:
     """Return whether the field is of type GraphQLInt."""
     field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
     return is_same_type(GraphQLInt, field_type)
 
 
-def is_uuid4_type(schema_info, vertex_name, field_name):
+def is_uuid4_type(schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str) -> bool:
     """Return whether the field is a uniformly distributed uuid4 type."""
     return field_name in schema_info.uuid4_fields.get(vertex_name, {})
--- a/graphql_compiler/cost_estimation/helpers.py
+++ b/graphql_compiler/cost_estimation/helpers.py
@@ -1,6 +1,20 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from graphql import GraphQLInt
 
+from ..schema import GraphQLDateTime, GraphQLDate
+
+
+def is_datetime_field_type(schema_info, vertex_name, field_name):
+    """Return whether the field is of type GraphQLDateTime."""
+    field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
+    return GraphQLDateTime.is_same_type(field_type)
+
+
+def is_date_field_type(schema_info, vertex_name, field_name):
+    """Return whether the field is of type GraphQLDate."""
+    field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
+    return GraphQLDate.is_same_type(field_type)
+
 
 def is_int_field_type(schema_info, vertex_name, field_name):
     """Return whether the field is of type GraphQLInt."""
--- a/graphql_compiler/cost_estimation/int_value_conversion.py
+++ b/graphql_compiler/cost_estimation/int_value_conversion.py
@@ -67,15 +67,15 @@ def convert_int_to_field_value(
     """
     if is_int_field_type(schema_info, vertex_class, property_field):
         return int_value
-    if is_datetime_field_type(schema_info, vertex_class, property_field):
+    elif is_datetime_field_type(schema_info, vertex_class, property_field):
         return DATETIME_EPOCH_UTC + datetime.timedelta(microseconds=int_value)
-    if is_date_field_type(schema_info, vertex_class, property_field):
+    elif is_date_field_type(schema_info, vertex_class, property_field):
         return datetime.date.fromordinal(int_value)
     elif is_uuid4_type(schema_info, vertex_class, property_field):
         if not MIN_UUID_INT <= int_value <= MAX_UUID_INT:
             raise AssertionError(
-                u"Integer value {} could not be converted to UUID, as it"
-                u" is not in the range of valid UUIDs {} - {}: {} {}".format(
+                u"Integer value {} could not be converted to UUID, as it "
+                u"is not in the range of valid UUIDs {} - {}: {} {}".format(
                     int_value, MIN_UUID_INT, MAX_UUID_INT, vertex_class, property_field
                 )
             )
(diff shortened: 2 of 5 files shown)


graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,94 @@
...
+# and generating parameters for pagination. Since integers are the easiest type to deal with
+# in this context, when we encounter a different type we represent it as an int, do all the
+# computation in the integer domain, and transfer the computation back into the original domain.
+#
+# In order to be able to reason about value intervals and successor/predecessor values, we
+# make sure these mappings to integers are increasing bijective functions.
+#
+# This kind of mapping is easy to do for int, uuid and datetime types, but not possible for other
+# types, like string. When the need for other types arises, the precise interface for range
+# reasoning can be defined and implemented separately for each type.
```
bojanserafimov: This is what we can do whenever we need more generality from this: https://github.com/kensho-technologies/graphql-compiler/blob/c54e165b32e461c3c6947a9e5390c1697c4f73e1/graphql_compiler/cost_estimation/filter_set.py

bojanserafimov:
If reviewing the draft PR, please only focus on intent: whether this is a good idea in general. This is how our most up-to-date paginator (in `feature_pagination` branch) currently works. I know there's documentation and tests missing, which I will add before marking "Ready for review".

pmantica1:
I am completely in favor of this diff. Though I have an issue with the way we convert datetime to seconds. See the following comment.

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,94 @@
...
+                             u' only uniform uuid4 and int fields are supported.'
+                             .format(int_value, vertex_class, property_field))
+
+
+def convert_field_value_to_int(schema_info, vertex_class, property_field, value):
+    """Return the integer representation of a property field value."""
+    if is_int_field_type(schema_info, vertex_class, property_field):
+        return value
+    if is_datetime_field_type(schema_info, vertex_class, property_field):
+        return int((value - datetime.datetime(1970, 1, 1)).total_seconds())
```
pmantica1: We are losing milisecond precision here for datetimes. How about running?
```
delta = (value - datetime.datetime(1970, 1, 1) 
return ((delta.days * 86400 + delta.seconds) * 10*6 + delta.microseconds) * 10**6
```
The code above is mostly copy pasted from the the `total_seconds` method of `timedelta` objects.
```
return ((self.days * 86400 + self.seconds) * 10**6 +
                self.microseconds) / 10**6
```

pmantica1: Also, maybe should also use the UTC timezone for `datetime.datetime(1970, 1, 1)`?

obi1kenobi: For the precision issue, just follow the [documentation's advice](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds): `td / timedelta(microseconds=1)`

For the timezones, be mindful of the fact that datetime arithmetic requires that both operands are of the same flavor: timezone-aware or timezone-naive. Mixing them will cause errors.

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,94 @@
...
+        is_uuid4_type(schema_info, vertex_class, property_field),
+        is_int_field_type(schema_info, vertex_class, property_field),
+        is_datetime_field_type(schema_info, vertex_class, property_field),
+        is_date_field_type(schema_info, vertex_class, property_field),
+    ))
+
+
+def convert_int_to_field_value(schema_info, vertex_class, property_field, int_value):
+    """Return the given integer's corresponding property field value.
+    Note that the property field values need not be integers. For example, all UUID values can be
```
obi1kenobi: This docstring won't pass lint.

bojanserafimov: It's a draft

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+from .helpers import is_date_field_type, is_datetime_field_type, is_int_field_type, is_uuid4_type
+
+
+# UUIDs are defined in RFC-4122 as a 128-bit identifier. This means that the minimum UUID value
+# (represented as a natural number) is 0, and the maximal value is 2^128-1.
+MIN_UUID_INT = 0
+MAX_UUID_INT = 2**128 - 1
+
+
+DATETIME_1970 = datetime.datetime(1970, 1, 1, tzinfo=pytz.utc)
```
obi1kenobi: Consider calling this `DATETIME_EPOCH_UTC`, 1970-01-01 is canonically the epoch: https://en.wikipedia.org/wiki/Epoch_(computing)

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def field_supports_range_reasoning(schema_info, vertex_class, property_field):
+    """Return whether range reasoning is supported. See module docstring for definition."""
+    return any((
```
obi1kenobi: You probably know this already, but this won't actually break early when it finds the first "True" value. The evaluation order requires the 4-tuple be calculated first (calling all four functions), and then the tuple is passed to `any()`.

These calculations aren't super expensive, but they do have quite a few attribute accesses and dict lookups, and they are on the critical path of query planning, so it might not be a terrible idea to rewrite this with `or` instead.

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    if is_int_field_type(schema_info, vertex_class, property_field):
+        return int_value
+    if is_datetime_field_type(schema_info, vertex_class, property_field):
+        return DATETIME_1970 + datetime.timedelta(microseconds=int_value)
+    if is_date_field_type(schema_info, vertex_class, property_field):
+        return datetime.date.fromordinal(int_value)
+    elif is_uuid4_type(schema_info, vertex_class, property_field):
+        if not MIN_UUID_INT <= int_value <= MAX_UUID_INT:
+            raise ValueError(u'Integer value {} could not be converted to UUID, as it'
```
obi1kenobi: Can this ever actually happen and not be a bug in the compiler? My impression is that `convert_int_to_field_value` is only called with `int_value` values produced by the pagination code, and if that is the case then hitting this would appear to be a bug in the pagination. If so, I'd make this `AssertionError` instead.

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+        raise NotImplementedError(u'Could not represent int {} as {} {}.'
+                                  .format(int_value, vertex_class, property_field))
+
+
+def convert_field_value_to_int(schema_info, vertex_class, property_field, value):
+    """Return the integer representation of a property field value."""
+    if is_int_field_type(schema_info, vertex_class, property_field):
+        return value
+    if is_datetime_field_type(schema_info, vertex_class, property_field):
+        if value.tzinfo != pytz.utc:
```
obi1kenobi: If the datetime is timezone-naive, `tzinfo` will be None and this will get hit. We probably want to handle that case.

graphql_compiler/cost_estimation/helpers.py
```diff
@@ -1,6 +1,20 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from graphql import GraphQLInt
 
+from ..schema import GraphQLDate, GraphQLDateTime
+
+
+def is_datetime_field_type(schema_info, vertex_name, field_name):
+    """Return whether the field is of type GraphQLDateTime."""
+    field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
+    return GraphQLDateTime.is_same_type(field_type)
```
obi1kenobi: This code will have to change slightly when #708 is merged.

graphql_compiler/cost_estimation/int_value_conversion.py
bojanserafimov: Why not enforce one standard across the codebase? I can't guarantee that converting to int and back to datetime will return an equal object if we allow 2 equivalent representations of datetime values.

obi1kenobi: I don't see a way for us to enforce this. Timezone awareness is a property of the data source being queried, and I don't see a way for us to affect that without corrupting the data we're supposed to return. Any ideas?

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,125 @@
...
+        ValueError, if the given int_value is outside the range of valid values for the given
+        property field.
+    """
+    if is_int_field_type(schema_info, vertex_class, property_field):
+        return int_value
+    if is_datetime_field_type(schema_info, vertex_class, property_field):
+        return DATETIME_EPOCH_UTC + datetime.timedelta(microseconds=int_value)
+    if is_date_field_type(schema_info, vertex_class, property_field):
+        return datetime.date.fromordinal(int_value)
+    elif is_uuid4_type(schema_info, vertex_class, property_field):
```
obi1kenobi: This is a bit strange. For consistency, I'd say either make line 70 and 72 also be `elif`, or add a blank line here and make this a regular `if` since all previous ones return.

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -0,0 +1,125 @@
...
+        if not MIN_UUID_INT <= int_value <= MAX_UUID_INT:
+            raise AssertionError(
+                u"Integer value {} could not be converted to UUID, as it"
+                u" is not in the range of valid UUIDs {} - {}: {} {}".format(
```
obi1kenobi: nit: prefer putting this space on the previous line rather than starting the string literal with it here.

graphql_compiler/cost_estimation/int_value_conversion.py
bojanserafimov: The inputs and outputs to the compiler are iso formatted strings, which are always tz-aware. Unless we change that, we have to coerce naive zones to utc. But I can see how if there is a time zone in the database, preserving it in the output might be good.

This specific piece of code is dealing with datetimes from user input, and datetimes from statistics. Since we only support utc and naive, and we're not dealing with database output, it's safe to coerce everything into utc.

obi1kenobi: What's the plan here? Are we going to assume that the caller will have normalized the field value to a UTC datetime (if it's supposed to be a datetime)?

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1398,3 +1406,101 @@ def test_disjoint_intervals(self) -> None:
...
+        )
+
+        datetime_values = [
+            datetime(2000, 1, 1),
+            datetime(3000, 1, 1, tzinfo=None),
+            datetime(1000, 1, 1, tzinfo=pytz.utc),
+            datetime(1, 1, 1, tzinfo=pytz.utc),
+            datetime(2000, 1, 1, 20, 55, 40, 877633, tzinfo=pytz.utc),
+            datetime(2000, 1, 1, 20, 55, 40, 877633, tzinfo=pytz.timezone("GMT")),
+        ]
```
obi1kenobi: How about also adding a non-UTC and non-UTC-equivalent timezone as a test case?
```suggestion
            datetime(2000, 1, 1, 20, 55, 40, 877633, tzinfo=pytz.timezone("America/New_York")),
        ]
```

PR #706, opened 2019-12-15
Use GitHub Actions instead of Travis CI for testing and linting

obi1kenobi:
This cuts the time CI needs to test and lint our code approximately in half, to somewhere around 5-6min and with many components of the test/lint suite finishing significantly earlier than that.

In addition, GitHub has very nice UI integrations that make GitHub Actions more pleasant to use.

The one thing to watch out for is that Codecov may temporarily display incomplete coverage, since we generate coverage from multiple runs which then needs to be merged together before computing the final coverage numbers. I suspect Codecov has some intelligent logic to avoid posting incomplete data in PRs, but thought I'd flag this just in case.


elvin666666:
> Evaluating GitHub Actions-based testing and linting, to see if it's any faster or more convenient than our current setup.

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,114 @@
...
+          NEO4J_AUTH: neo4j/root
+      redisgraph:
+        image: redislabs/redisgraph:1.2.2
+        ports:
+          - "127.0.0.1:6379:6379"
+    strategy:
+      max-parallel: 3
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+    steps:
```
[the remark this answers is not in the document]
obi1kenobi: Unfortunately, it doesn't seem like it's possible to share steps between jobs at the moment, at least not with any reasonable setup...

obi1kenobi:
@billyjanitsch if you have a moment, I'd love it if you could take a look at the actions setup since I don't think anyone working on GraphQL compiler has any experience with GitHub Actions so far.

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,96 @@
...
+        # Use a matrix for the python version here to make the rest of the lint job setup
+        # as similar as possible to the test job below.
+        python-version: [3.8]
+        lint-flags:
+          - "--run-only-fast-linters"
+          - "--run-only-pylint"
+          - "--run-only-mypy"
+          - "--run-only-bandit"
+          - "--run-only-sphinx-build"
+          - "--run-only-typing-copilot-tighten"
```
billyjanitsch, line 26: Is splitting the different types of linting across a matrix for performance reasons, or so that the results are reported separately? If the former, is this really faster when accounting for setup time?

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,96 @@
...
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+      - name: Cache the Python dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('**/Pipfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-${{ matrix.python-version }}-
```
billyjanitsch, line 46: Is this what the GH docs recommend? I'm surprised that there's so much boilerplate, especially to establish the cache directory. The Node cache setup is much terser; totally possible that Python is different but wanted to double check.

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,96 @@
+name: Tests and lint
```
billyjanitsch, line 1: I've usually seen the workflow filename align with the name of the workflow. It doesn't affect anything, just FYI.

billyjanitsch (approved):
A few sanity check questions but otherwise lgtm! 🙂

.github/workflows/pythonpackage.yml
obi1kenobi: `pylint` is the "long pole" in the entire test suite, at 5-6min on its own. `typing_copilot tighten` is around 4min. `bandit` and `mypy` are also 1-2 minutes each. `sphinx-build` is super fast but didn't feel like it made any sense to combine with the "fast linters" which are useful for local development (`flake8, black, isort` etc.).

So not all of them need to be split up — e.g. `mypy` and `bandit` could be combined without a net slowdown of the suite. But `pylint, mypy, typing_copilot, bandit` all do some pretty heavyweight AST munging that doesn't run particularly quickly, and at least to me there didn't seem to be a "conceptually clean" way to combine tools into fewer runs.

.github/workflows/pythonpackage.yml
obi1kenobi: What does the node cache setup look like? This is based on https://github.com/actions/cache/blob/main/examples.md#using-pip-to-get-cache-location

We don't currently do multi-OS testing, but I didn't feel like debugging the Actions setup if we ever do end up doing it.

.github/workflows/pythonpackage.yml
obi1kenobi: I can definitely rename it 👍

.github/workflows/pythonpackage.yml
billyjanitsch: Ah, I use something like [this](https://github.com/actions/cache/blob/d606e039ae32f64a8593bf4a37b0bf205c695237/examples.md#macos-and-ubuntu) but it only works on *nix and the Windows snippet they recommend in the subsequent section looks more like the pip one. Makes sense.

.github/workflows/pythonpackage.yml
billyjanitsch: Oh I missed that this is running `./scripts/lint.sh` which calls out to different services. I thought that the whole thing was one linter that could be configured to run different portions of itself.

obi1kenobi: Yeah, we kind of bend over backwards here to make sure that the bot's lint checks are 100% reproducible locally. If we didn't care about that so much, in principle it would have been a lot easier to just use the GitHub Actions version of each of the linter tools when available, and have a CI-specific script for the rest. There are pros and cons both ways, and in this PR I tried to keep to the spirit of the old CI system as much as possible.

billyjanitsch: So do we. `npm test` 🙃

billyjanitsch (approved):

PR #707, opened 2019-12-16
 Annotate more test files with type hints.

obi1kenobi:


pmantica1:
Please fix the linting issues. Also, please run the linting script before submitting a pull request.

graphql_compiler/tests/test_explain_info.py
```diff
@@ -14,17 +15,23 @@
...
+    def compare_output_info(self, expected: OutputInfo, received: OutputInfo) -> None:
         """Compare two OutputInfo objects, using proper GraphQL type comparison operators."""
         self.assertEqual(expected.location, received.location)
         self.assertTrue((expected.type).is_same_type(received.type))
         self.assertEqual(expected.optional, received.optional)
 
-    def check(self, graphql_test, expected_filters, expected_recurses, expected_outputs):
+    def check(
+        self,
+        graphql_test: Callable[[], test_input_data.CommonTestData],
```
MichaelaShtilmanMinkin: What does `[]` represent here? I'm assuming it has something to do with self but I'm not sure and having trouble googling 😅

obi1kenobi: `Callable` is generic on two arguments, a list of positional input argument types (here, empty), and an output type (here, `CommonTestData`). https://docs.python.org/3/library/typing.html#callable is where I learned this, it has an example specifically for this.

MichaelaShtilmanMinkin: Ah! Makes sense - thanks. This PR looks good to me - I can approve once it passes linting.

PR #685, opened 2019-12-02
Apply and enforce "black" on all Python code in the project

obi1kenobi:
Sibling of #684.

Initially generated with running "black" on all Python files, produced with the configuration in #684. Then, manually applied fixes to make lint pass, e.g. compensating for https://github.com/psf/black/issues/1161 and moving `bandit`'s `nosec` comments around until `bandit` is happy.

All substantive changes are in #684, so please review that first. Any changes here but not in #684 are either autogenerated by `black`, or purely stylistic and with no functional changes.

PR #708, opened 2019-12-17
Update graphql-core version

### LINKED ISSUE #672: Upgrade to GraphQL-core v3 as soon as it is available; drop Python 2.x support

chewselene:
This PR updates to graphql-core3. 

This required a number of changes, which I will summarize below:

- AST node types had their names updated from {NodeDescription} -> {NodeDescription}Node
- `is_same_type` functionality on graphql types was removed from the core library. Since the compiler relies on this functionality it has been moved to `global_utils.py`
- ~~schema getter methods have been replaced with `to_kwargs()`, which returns a dictionary the keys corresponding to the old getter functions.~~ However, `type_map` is not included in `to_kwargs()` so `type_map` is directly accessed with `schema.type_map`. All schema properties are accessed with direct access.
- schema modification and macro expansion no longer done in place due to switch from `list`s to graphql's `FrozenList`s
- in `macros/macro_edge_directives.py` switched from frozen sets of directives to frozen sets of directive names since directives are no longer hashable
- added a special case for parsing strings to floats since the default parser errors on this input
- made the `filter` directive repeatable
- added macro_edge and stitch directives to test schema and allowable compiler directives
- updated visitor functions based on new `QUERY_DOCUMENT_KEYS` (used to be camel case and are now snake case)
- use new AST type `OperationType.QUERY` instead of using the string `query`
- adding some new mypy type hints
- update test schema to match updated version of print_schema (with descriptions)

This list is probably not exhaustive, but hopefully that covers most of the changes.

Resolves #672


obi1kenobi:
That seems like a ton of merge conflicts... Based on inspecting the commit history, it seems like you might not have pulled `master` before making the new branch for this, so #685 (the PR where we ran `black` on all the files) wasn't in there — hence the merge conflicts.

Try grabbing the `pyproject.toml` file from the master branch, installing `black`, and running `black .` from the project root directory — hopefully that will minimize many of the merge conflicts. If that still doesn't do the trick, let's try to work this out in person together.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -342,14 +344,15 @@ def check_ast_schema_is_valid(ast):
     """
     try:
         schema = build_ast_schema(ast)
-    except Exception as e:  # Can't be more specific -- see graphql/utils/build_ast_schema.py
+    except Exception as e:  # Can't be more specific -- see graphql/utilities/build_ast_schema.py
```
obi1kenobi: My impression is that this is resolved in the new version of the GraphQL library, so we can remove the `try-except` entirely and just have a bare `schema = build_ast_schema(ast)`. No need to rewrap the exception now that it isn't of type `Exception`.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -342,14 +344,15 @@ def check_ast_schema_is_valid(ast):
...
+
         raise SchemaStructureError(u"Input is not a valid schema. Message: {}".format(e))
 
-    if schema.get_mutation_type() is not None:
+    if schema.to_kwargs()["mutation"] is not None:
```
obi1kenobi: It might be possible to avoid the `to_kwargs()` here and below on line 355 similarly to previous instances as well.

graphql_compiler/tests/schema_transformation_tests/example_schema.py
```diff
@@ -243,6 +243,8 @@ def _get_type_equivalence_hints(schema_id_to_ast, type_equivalence_hints_names):
   Creature: Creature
 }
 
+directive @stitch(source_field: String!, sink_field: String!) on FIELD_DEFINITION
```
obi1kenobi: Nice catch!

graphql_compiler/tests/schema_transformation_tests/input_schema_strings.py
```diff
@@ -9,6 +9,8 @@ class InputSchemaStrings(object):
           query: SchemaQuery
         }
 
+        directive @stitch(source_field: String!, sink_field: String!) on FIELD_DEFINITION
```
obi1kenobi: Another nice catch!

graphql_compiler/tests/schema_transformation_tests/input_schema_strings.py
```diff
@@ -101,7 +103,7 @@ class InputSchemaStrings(object):
           age: Int
         }
 
-        type Human implements Character, Creature {
+        type Human implements Character & Creature {
```
obi1kenobi: This is super subtle but a massive compatibility improvement, it was such a struggle without this!

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -14,11 +15,12 @@
 class TestRenameSchema(unittest.TestCase):
     def test_rename_visitor_type_coverage(self):
         """Check that all types are covered without overlap."""
-        all_types = set(ast_type.__name__ for ast_type in QUERY_DOCUMENT_KEYS)
+        all_types_snake_case = set(ast_type for ast_type in QUERY_DOCUMENT_KEYS)
```
obi1kenobi: What is the type of `QUERY_DOCUMENT_KEYS`? This should probably be `set(QUERY_DOCUMENT_KEYS)` or maybe even just `QUERY_DOCUMENT_KEYS` used directly.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -14,11 +15,12 @@
...
         type_sets = [
             RenameSchemaTypesVisitor.noop_types,
             RenameSchemaTypesVisitor.rename_types,
         ]
+        all_types = set([snake_to_camel(node_type) + "Node" for node_type in all_types_snake_case])
```
obi1kenobi: Can we use a set comprehension here, instead of wrapping a list comprehension inside `set()`?

graphql_compiler/tests/test_graphql_pretty_print.py
```diff
@@ -37,9 +37,8 @@ def test_graphql_pretty_print_indentation(self) -> None:
 
     def test_filter_directive_order(self) -> None:
         bad_query = """{
-            Animal @filter(value: ["$name"], op_name: "name_or_alias") {
-                uuid @filter(value: ["$max_uuid"], op_name: "<=")
-
```
obi1kenobi: Since these tests cover query pretty-printing, let's try to minimize or entirely eliminate changes to test inputs here. Issues like extraneous whitespace, poor formatting, directive argument order that causes poor readability, etc. are probably intentional and may in fact be precisely what the tests are testing.

graphql_compiler/tests/test_helpers.py
```diff
@@ -59,7 +60,7 @@
         query: RootSchemaQuery
     }
 
-    directive @filter(op_name: String!, value: [String!]) on FIELD | INLINE_FRAGMENT
+    directive @filter(op_name: String!, value: [String!]) repeatable on FIELD | INLINE_FRAGMENT
```
obi1kenobi: Woot! Awesome!

graphql_compiler/tests/test_macro_schema.py
```diff
@@ -46,25 +46,25 @@ def test_get_schema_for_macro_definition_addition(self) -> None:
         original_schema = self.macro_registry.schema_without_macros
         macro_definition_schema = get_schema_for_macro_definition(original_schema)
         macro_schema_directive_names = {
-            directive.name for directive in macro_definition_schema.get_directives()
+            directive.name for directive in macro_definition_schema.to_kwargs()["directives"]
```
obi1kenobi: Let's try to avoid all of these `to_kwargs()` calls if possible and use direct attribute access.

graphql_compiler/tests/test_macro_validation.py
```diff
@@ -133,7 +133,6 @@ def test_macro_with_target_at_fold(self) -> None:
             macro_registry = get_empty_test_macro_registry()
             with self.assertRaises(GraphQLInvalidMacroError):
                 register_macro_edge(macro_registry, macro_edge_definition, args)
-
```
obi1kenobi: This seems unintentional, consider undoing it.

graphql_compiler/schema_transformation/utils.py
obi1kenobi: It's probably fine either way, my gut feeling is to go for simplicity and just let the underlying error bubble up. The alternative would be to do something like `raise SchemaStructureError("Input is not a valid schema.") from e` and that might be okay too — though in that case we should probably tighten down what kinds of exceptions we expect. But that seems like a lot of work for not that much benefit at all.

graphql_compiler/macros/macro_edge/validation.py
```diff
@@ -99,7 +99,7 @@ def _validate_non_required_macro_definition_directives(
         inside_fold_scope: bool, whether the subtree is within a @fold scope
     """
     names_of_allowed_directives = frozenset(
-        {directive.name for directive in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION}
+        {directive for directive in DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION}
```
obi1kenobi: I think `DIRECTIVES_ALLOWED_IN_MACRO_EDGE_DEFINITION` is already a frozenset of the names of those directives, so I think `names_of_allowed_directives` is unnecessary as a whole — the global constant should be fine to use throughout.

graphql_compiler/schema/__init__.py
```diff
@@ -210,6 +211,52 @@
     locations=[DirectiveLocation.FIELD,],
 )
 
+# TODO(selene): comments for the macro directives
```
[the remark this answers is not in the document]
obi1kenobi: Just to be safe, let's go with opening an issue for this (make sure to tag it with the v2.0 milestone) and then if we get to resolving this before the merge — great, and if not then we can still merge and unblock other work, and we won't forget to come back to this.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -222,29 +230,31 @@ def _accumulate_types(
...
     for new_definition in new_definitions:
-        if isinstance(new_definition, ast_types.SchemaDefinition):
+        if isinstance(new_definition, ast_types.SchemaDefinitionNode):
             continue
         elif (
-            isinstance(new_definition, ast_types.ObjectTypeDefinition)
+            isinstance(new_definition, ast_types.ObjectTypeDefinitionNode)
             and new_definition.name.value == current_query_type
         ):  # query type definition
             new_query_type_fields = new_definition.fields  # List[FieldDefinition]
```
obi1kenobi: the comment here might need to be updated to say `List[FieldDefinitionNode]`

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -307,24 +334,34 @@ def _process_directive_definition(directive, existing_directives, merged_schema_
                 )
             )
     # new directive
-    merged_schema_ast.definitions.append(directive)
-    existing_directives[directive_name] = directive
+    new_definitions = list(merged_schema_ast.definitions)
+    new_definitions.append(directive)
+    new_merged_schema_ast = ast_types.DocumentNode(definitions=new_definitions)
+    new_existing_directives = deepcopy(existing_directives)
```
obi1kenobi: This should be able to be safely replaced with just `copy()` (which in this case does the same thing as `dict(existing_directives)` so feel free to use that if that's easier) since we don't make changes within keys or values of the dict — we only add a new key.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -338,8 +375,12 @@ def _process_scalar_definition(scalar, existing_scalars, type_name_to_schema_id,
             )
         )
     # new, valid scalar
-    merged_schema_ast.definitions.append(scalar)
-    existing_scalars.add(scalar_name)
+    new_definitions = list(merged_schema_ast.definitions)
+    new_definitions.append(scalar)
+    new_merged_schema_ast = ast_types.DocumentNode(definitions=new_definitions)
+    new_existing_scalars = deepcopy(existing_scalars)
```
obi1kenobi: Similarly, this should be able to be just a `set()`/`copy()` call instead of a deep copy.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -371,8 +420,13 @@ def _process_generic_type_definition(
                 type_name, schema_id, type_name, type_name_to_schema_id[type_name], type_name
             )
         )
-    merged_schema_ast.definitions.append(generic_type)
-    type_name_to_schema_id[type_name] = schema_id
+
+    new_definitions = list(merged_schema_ast.definitions)
+    new_definitions.append(generic_type)
+    new_merged_schema_ast = ast_types.DocumentNode(definitions=new_definitions)
+    new_type_name_to_schema_id = deepcopy(type_name_to_schema_id)
```
obi1kenobi: Here as well, I think.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -736,30 +812,55 @@ def _add_edge_field(
...
+    new_type_fields.append(new_edge_field_node)
+    if type(source_type_node) == ast_types.ObjectTypeDefinitionNode:
+        new_source_type_node = ast_types.ObjectTypeDefinitionNode(
+            description=source_type_node.description,
+            name=source_type_node.name,
+            directives=source_type_node.directives,
+            fields=new_type_fields,
+            interfaces=source_type_node.interfaces,
+        )
+    elif type(source_type_node) == ast_types.InterfaceTypeDefinitionNode:
```
obi1kenobi: This might be another good candidate for use of `to_kwargs()` — it might help merge the object/interface branches into a single branch.

graphql_compiler/tests/test_schema.py
```diff
@@ -26,11 +27,26 @@ def _get_directives_in_string_form(directives):
...
-            schema_lines = [line.strip() for line in print_schema(fake_schema).split("\n")]
+            # Split schema on double line breaks where the following character is not a space.
+            # It is not possible to simply split on double line breaks because print_schema puts a
+            # double line break between GraphQLArguments. The not space character is retained and
+            # reattached to the rest of the line.
+            split_schema_lines = [
+                line.strip() for line in re.split("\n\n([^ ])", print_schema(fake_schema))
+            ]
+
+            # Reattach the deliminator's character to the rest of the line. The first line does
```
obi1kenobi: nit: I think `delimiter` might be the more commonly-recognized nomenclature here and in the variable name below

graphql_compiler/macros/macro_edge/directives.py
```diff
@@ -93,7 +61,7 @@ def get_schema_for_macro_edge_definitions(querying_schema):
     # altering the object's internals directly.
     macro_edge_schema = copy(querying_schema)
```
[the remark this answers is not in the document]
chewselene: Nice! this is much prettier

obi1kenobi:
If you can make the linters and tests happy, this looks good to me. You might need to generate the lockfile in a Python 3.8 environment to fix the `black` issue.

I'm also not positive on why the query pretty-printer seems to flip the argument order. I think that's something we added some code to address a while back, but it's possible the new version might need some more attention to make that work again.

graphql_compiler/tests/test_graphql_pretty_print.py
obi1kenobi: nit: It probably doesn't matter a ton, but I'd feel better about it if we didn't change anything about these tests, including just removing the blank lines.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -45,10 +46,20 @@ def execute_graphql(
...
+        #     for key, value in six.iteritems(test_data.type_equivalence_hints)
+        # }
+        for key, value in six.iteritems(test_data.type_equivalence_hints):
+            key_type = schema.get_type(key)
+            value_type = schema.get_type(value)
+            if (
+                key_type and value_type
+                and (isinstance(key_type, GraphQLInterfaceType)
+                     or isinstance(key_type, GraphQLObjectType))
+                and isinstance(value_type, GraphQLUnionType)
```
obi1kenobi: Neat! You should be able to assert that this branch is always taken — the `else` here should `raise AssertionError`.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -45,10 +46,20 @@ def execute_graphql(
     if test_data.type_equivalence_hints:
         # For test convenience, we accept the type equivalence hints in string form.
         # Here, we convert them to the required GraphQL types.
-        schema_based_type_equivalence_hints = {
-            schema.get_type(key): schema.get_type(value)
-            for key, value in six.iteritems(test_data.type_equivalence_hints)
-        }
+        schema_based_type_equivalence_hints = {}
+        #     schema.get_type(key): schema.get_type(value)
```
obi1kenobi: Remember to remove this before we merge.

graphql_compiler/tests/snapshot_tests/test_orientdb_match_query.py
```diff
@@ -45,10 +46,20 @@ def execute_graphql(
     if test_data.type_equivalence_hints:
         # For test convenience, we accept the type equivalence hints in string form.
         # Here, we convert them to the required GraphQL types.
-        schema_based_type_equivalence_hints = {
-            schema.get_type(key): schema.get_type(value)
-            for key, value in six.iteritems(test_data.type_equivalence_hints)
-        }
+        schema_based_type_equivalence_hints = {}
```
obi1kenobi: To make `mypy` happy, you might need to declare the type here: `Dict[Union[GraphQLInterfaceType, GraphQLObjectType], GraphQLUnionType]`

graphql_compiler/tests/test_helpers.py
```diff
@@ -554,13 +554,21 @@ def get_schema() -> GraphQLSchema:
...
     """Get the default type_equivalence_hints used for testing."""
     schema = get_schema()
-    return {
-        schema.get_type(key): schema.get_type(value)
-        for key, value in [
-            ("Event", "Union__BirthEvent__Event__FeedingEvent"),
-            ("FoodOrSpecies", "Union__Food__FoodOrSpecies__Species"),
-        ]
-    }
+    type_equivalence_hints = {}
```
obi1kenobi: `mypy` might want the same kind of explicit type hint here as well.

obi1kenobi:
I suspect the `black` error might be solvable by running `pipenv lock` in a Python 3.8-based environment, it's possible some 3.8 dependency isn't getting picked up and installed.

PR #710, opened 2019-12-28
Add type hints to a few files in the compiler submodule.

obi1kenobi:
Before reviewing, please make sure you are familiar with [this part of PEP 484](https://www.python.org/dev/peps/pep-0484/#annotating-instance-and-class-methods), since some of the generic type hints depend on that functionality.

This PR consists of:
- auto-generated type hints where possible;
- some hand-written type hints where more sophistication (e.g. generics) was necessary;
- some minor tweaks to logic and a couple of helper methods to make mypy happy, and
- a fix in the `test_emit_output.py` file's Cypher section which I caught thanks to mypy.

Would appreciate it if someone more familiar with the SQL backend double-checked the type signature on `to_sql()` and looked into whether we can put a tighter signature there, since the current one uses a couple of `Any` types.

### DIFF
--- a/graphql_compiler/compiler/blocks.py
+++ b/graphql_compiler/compiler/blocks.py
@@ -30,9 +30,6 @@ class QueryRoot(BasicBlock):
                          This will generally be a set of length 1, except when using Gremlin
                          with a non-final class, where we have to include all subclasses
                          of the start class. This is done using a Gremlin-only IR lowering step.
-
-        Returns:
-            new QueryRoot object
         """
         super(QueryRoot, self).__init__(start_class)
         self.start_class = start_class
@@ -80,9 +77,6 @@ class CoerceType(BasicBlock):
                           This will generally be a set of length 1, except when using Gremlin
                           with a non-final class, where we have to include all subclasses
                           of the target class. This is done using a Gremlin-only IR lowering step.
-
-        Returns:
-            new CoerceType object
         """
         super(CoerceType, self).__init__(target_class)
         self.target_class = target_class
@@ -122,9 +116,6 @@ class ConstructResult(BasicBlock):
         Args:
             fields: dict, variable name string -> Expression
                     see rules for variable names in validate_safe_string().
-
-        Returns:
-            new ConstructResult object
         """
         self.fields = {ensure_unicode_string(key): value for key, value in six.iteritems(fields)}
 
@@ -227,9 +218,6 @@ class MarkLocation(BasicBlock):
 
         Args:
             location: BaseLocation object, must not be at a property field in the query
-
-        Returns:
-            new MarkLocation object
         """
         super(MarkLocation, self).__init__(location)
         self.location = location
@@ -265,9 +253,8 @@ class Traverse(BasicBlock):
             edge_name: string obeying variable name rules (see validate_safe_string).
             optional: optional bool, specifying whether the traversal to the given location
                       is optional (i.e. non-filtering) or mandatory (filtering).
-
-        Returns:
-            new Traverse object
+            within_optional_scope: optional bool, set to True to indicate that this Traverse
+                                   is located within a scope marked @optional
         """
         super(Traverse, self).__init__(
             direction, edge_name, optional=optional, within_optional_scope=within_optional_scope
@@ -361,9 +348,8 @@ class Recurse(BasicBlock):
             direction: string, 'in' or 'out'.
             edge_name: string obeying variable name rules (see validate_safe_string).
             depth: int, always greater than or equal to 1.
-
-        Returns:
-            new Recurse object
+            within_optional_scope: optional bool, set to True to indicate that this Recurse
+                                   is located within a scope marked @optional
         """
         super(Recurse, self).__init__(
             direction, edge_name, depth, within_optional_scope=within_optional_scope
@@ -430,9 +416,6 @@ class Backtrack(BasicBlock):
             location: BaseLocation object, specifying where to backtrack to
             optional: optional bool, specifying whether the steps between the current location
                       and the location to which Backtrack is returning were optional or not
-
-        Returns:
-            new Backtrack object
         """
         super(Backtrack, self).__init__(location, optional=optional)
         self.location = location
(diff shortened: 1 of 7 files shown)


obi1kenobi:
Seems like `pipenv sync` didn't properly handle the migration to the new `graphql-core` library. As a result, my local `mypy` checks passed but they are failing on Travis at the moment.

obi1kenobi:
I think I can make this be a bit more elegant, but it'll need some more work.

obi1kenobi:
I don't think Python's generics are expressive enough to do the trick I wanted to do.

In the long run, I think we may end up redesigning the `BaseLocation` and `Expression` type hierarchies to better fit within the capabilities of the Python type system. However, I don't think the time for that is now, and I think it might be better to get *some* type-checked release in the near term rather than wait for a perfect solution.

graphql_compiler/compiler/blocks.py
```diff
@@ -145,7 +148,9 @@ def validate(self):
                     u"{} -> {}".format(key, value)
                 )
 
-    def visit_and_update_expressions(self, visitor_fn):
+    def visit_and_update_expressions(
+        self, visitor_fn: Callable[[Expression], Expression]
+    ) -> "ConstructResult":
```
MichaelaShtilmanMinkin: TIL about forward references!
https://www.python.org/dev/peps/pep-0484/#forward-references
I think I've actually run into this issue before and couldn't figure out how to resolve it 😅

graphql_compiler/compiler/compiler_entities.py
```diff
@@ -66,23 +68,26 @@ def __eq__(self, other):
...
 
-    @abstractmethod
-    def to_gremlin(self):
+    def to_gremlin(self) -> str:
         """Return the Gremlin unicode string representation of this object."""
         raise NotImplementedError()
 
 
+AliasType = Union[Alias, CTE]
+AliasesDictType = Dict[Tuple[Tuple[str, ...], Optional[Tuple[Tuple[str, str], ...]],], AliasType]
```
MichaelaShtilmanMinkin: Suggestion: consider adding a brief description of what these are meant to represent at a high level.

graphql_compiler/compiler/blocks.py
```diff
@@ -244,7 +251,13 @@ class Traverse(BasicBlock):
...
     __slots__ = ("direction", "edge_name", "optional", "within_optional_scope")
 
-    def __init__(self, direction, edge_name, optional=False, within_optional_scope=False):
+    def __init__(
+        self,
+        direction: str,
+        edge_name: str,
+        optional: bool = False,
+        within_optional_scope: bool = False,
+    ) -> None:
```
chewselene: Upon first read through I was confused because the documentation for this function has a `Returns` section, but the type hint says nothing is returned. Might just be a special case for `__init__` functions, but is it correct to say that this returns a new traverse object?

Also (not particularly related to the scope of this PR), the `Args` section is missing `within_optional_scope`.

obi1kenobi: Yeah, the `Returns` on `__init__` is something we probably should eliminate everywhere, maybe as a tech debt item. More generally, it is `__new__` that returns a new (uninitialized) object, which then `__init__` initializes by mutating the `self` reference. So while a `Returns` section on `__init__` can reasonably be thought of as describing the *intent*, it is technically incorrect.

I'm happy to remove the `Returns` section here, and document `within_optional_scope`. Nice catch! I'm not sure why `pydocstyle` didn't catch that.

graphql_compiler/compiler/compiler_entities.py
obi1kenobi: I'd love to — but I actually don't know if they are even correct, let alone how to best describe them. @bojanserafimov, help? 😄

graphql_compiler/compiler/blocks.py
obi1kenobi: https://github.com/python/mypy/issues/731 is also good to read about, since it's not currently implemented but is being actively worked on. I ran into this while working on the interpreter module, and it took me a while to diagnose the problem and find that Github issue.

graphql_compiler/compiler/blocks.py
obi1kenobi: I found one more instance where `within_optional_scope` wasn't described, and I removed all the `Returns` sections from `__init__` methods.

graphql_compiler/compiler/compiler_entities.py
obi1kenobi: To avoid playing a game of telephone, this might be a good fit for a separate PR by @bojanserafimov.

MichaelaShtilmanMinkin: Gotcha - maybe add a TODO/issue? Because right now I'm not sure what I'm looking at :)...

graphql_compiler/compiler/expressions.py
```diff
@@ -49,10 +51,17 @@
 )
 
 
-def make_replacement_visitor(find_expression, replace_expression):
+ExpressionT = TypeVar("ExpressionT", bound=Expression)
+ReplacementExpressionT = TypeVar("ReplacementExpressionT", bound=Expression)
+ReplacementT = Union[ExpressionT, ReplacementExpressionT]
```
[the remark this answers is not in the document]
obi1kenobi: I don't know if there's a common convention, and if one exists I'm happy to stick to it. I've tried to use `T` suffixes for generic types (including their derivations like `Union[FooT, BarT]`), and `Type` suffixes for situations where I wanted some shorthand for a complex non-generic type. I think this instance is consistent with that approach, but if I'm happy to fix any situations where I might have deviated from it.

graphql_compiler/compiler/compiler_entities.py
obi1kenobi: Opened #711 for it.

graphql_compiler/compiler/expressions.py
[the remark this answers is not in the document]
MichaelaShtilmanMinkin: Sounds good. I was a little unsure as to which category `Union` fell into in both cases.

obi1kenobi:
In this case it's easy because it's a union of two generics, so I'd say
generic rules apply. If it was a union of a generic and a non-generic type,
we might have to decide on a case-by-case basis.

PR #712, opened 2020-01-03
Add python 3.8 troubleshooting

bojanserafimov:


TROUBLESHOOTING.rst
```diff
@@ -147,4 +147,4 @@ can be resolved on Ubuntu by running:
 .. bash::
 
     # Please modify command to match the version of python you use for development.
-    sudo apt-get install python3.6-dev
+    sudo apt-get install python3.6-dev python 3.8-dev
```
MichaelaShtilmanMinkin: This is not necessarily what we want - we already have `python3.6-dev` and we write `Please modify command to match the version of python you use for development.` right above this line. Maybe we need to make this instruction more obvious? e.g. replace the value with an x? e.g. 3.x?

bojanserafimov: Ok. I'll set the default to 3.8, and add the same comment ("Please modify...") to line 117

pmantica1: Please perform the equivalent modification in `docs/source/about/contributing.rst`. Sorry I know that it is tedious that we have to modify it in both places. I will try to move us to readthedocs soon.

obi1kenobi:
Superseded by #718.

PR #713, opened 2020-01-03
Fold Post-Processing

chewselene:
This PR adds post-processing functionality for XML PATH-based array aggregation performed by compiled MSSQL queries. See #613 for original MSSQL fold implementation.


graphql_compiler/post_processing/sql_post_processing.py
```diff
@@ -0,0 +1,128 @@
...
+    # Get the internal representation of the query.
+    ir_and_metadata = graphql_to_ir(
+        schema_info.schema,
+        graphql_query,
+        type_equivalence_hints=schema_info.type_equivalence_hints,
+    )
+    lowered_ir_blocks = sql_backend.lower_func(schema_info, ir_and_metadata)
+
+    # Loop through ir_blocks; for all ConstructResult for FoldedContextFields, look up string result
+    # in query_results and update to a list.
```
bojanserafimov: The output metadata is intuitively the sensible thing to use because:
- It's made for this purpose. If it's not serving it's purpose, we should improve it instead of avoiding it if it's not too much work.
- It's easier to understand. The IR is hard, and you don't want to assume IR knowledge to write post-processing steps.

The problem with the output metadata for this purpose is that it only knows which outputs are list type, and can't differentiate outputting list typed properties from outputting folded non-list properties. I think it's worth amending the output metadata with this information.

graphql_compiler/tests/test_post_processing.py
```diff
@@ -0,0 +1,221 @@
...
+                }"""
+
+        expected_result = [{"child_names": []}]
+
+        post_process_mssql_folds(self.mssql_schema_info, graphql_query, query_output)
+        self.assertEqual(query_output, expected_result)
+
+    def test_convert_basic(self):
+        """Test basic XML path encoding (only pipe separations) is correctly decoded."""
+        query_output = [{"child_names": "Animal 1|Animal 2|Animal 3|",}]
```
bojanserafimov: Can you put two consecutive empty string elements in the middle? Just protecting against different string splitting methods (I can imagine a sensible one that merges multiple occurances of the delimiter).

graphql_compiler/post_processing/sql_post_processing.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Post-processing steps:
+        1. split on "|",
+        2. convert "~" to None
+        3. convert caret escaped characters (excluding "^" itself)
+            i.e. "^d" (delimiter) to "|" and "^n" (null) to "~"
+        4. with caret escaped characters removed, convert "^e" to "^"
+        5. convert ampersand escaped characters (excluding "&" itself)
+            i.e. "&gt;" to ">", "&lt;" to "<" and "&#xHEX;" to "\xHEX"
+        6. with ampersand escaped characters removed, convert "&amp;" to "&"
```
bojanserafimov: This comment explains why the order of operations matters. I'd note that near the implementation, to avoid regression bugs when other people mess with the code.

graphql_compiler/tests/test_post_processing.py
```diff
@@ -0,0 +1,221 @@
...
+                }
+            }
+        }"""
+
+        expected_result = [{"child_net_worths": [500, 1000, 400, None,],}]
+
+        post_process_mssql_folds(self.mssql_schema_info, graphql_query, query_output)
+        self.assertEqual(query_output, expected_result)
+
+    def test_convert_basic_datetime(self):
```
bojanserafimov: The test says it's testing datetimes, but it's testing dates.

bojanserafimov: I'd actually want to see a test with datetimes. Do we enforce a particular datetime format when encoding the datetimes? Do we include timezone information?

bojanserafimov:
LGTM. Only thing I think is missing is a test with datetimes, with some different timezones (if mssql supports timezones). We're using the default GraphQL type parsers, and the default mssql datetime output format. We'd be very lucky if they just happen to work together nicely.

graphql_compiler/post_processing/sql_post_processing.py
```diff
@@ -0,0 +1,108 @@
...
+        for result in list_result
+    ]
+
+    return list_result_to_return
+
+
+def post_process_mssql_folds(
+    query_results: List[Dict[str, Any]], output_metadata: Dict[str, OutputMetadata]
+) -> None:
+    r"""Convert XML PATH fold results from a string to a list of the appropriate type.
```
MichaelaShtilmanMinkin: Sneaky `r`

chewselene:
> We're using the default GraphQL type parsers, and the default mssql datetime output format. We'd be very lucky if they just happen to work together nicely.

I think we might have an issue here. I'm having trouble telling what the default return format is for MSSQL (and the fiddle I usually test in is not working at the moment). Looks like it is possible to specify the output format within the query https://www.mssqltips.com/sqlservertip/1145/date-and-time-conversions-using-sql-server/. Would like to avoid adding special cases on dates and datetimes in the query if possible, but I'm not sure what the alternative is... Thoughts?

graphql_compiler/post_processing/sql_post_processing.py
chewselene: Needs to be a raw string (the `\xHEX` gives a `SyntaxError` otherwise) 😄

MichaelaShtilmanMinkin: That's kinda hilarious. Not sure if warrants a comment but it definitely sticks out a little.

chewselene:
Got the sql fiddle to cooperate and it looks like the defaults for MSSQL 2017 (at least on sqlfiddle.com) are:
- date: `YYYY-MM-DD`
- datetime: `YYYY-MM-DDTHH:mm:ssZ`
which match with the parsing functions. However, if I added a datetime to my table to milliseconds they would be returned as well.

I suspect that this is not a guarantee of the return format though. It might be possible to configure per database.

bojanserafimov:
What's the downside of using `convert` and specifying the format?

obi1kenobi:
> However, if I added a datetime to my table to milliseconds they would be returned as well.

This should be fine, we should be able to support both having and not having explicit milliseconds in the output. I think it's fine to make that change for the datetime parser logic in general — both of these are allowed by ISO-8601.

chewselene:
Down side of `convert` is that it needs to be added in to the query itself, not the post processing. I'll update the datetime parser to check for millisecond format first.

bojanserafimov:
> Down side of `convert` is that it needs to be added in to the query itself, not the post processing. I'll update the datetime parser to check for millisecond format first.

I doubt that impacts query performance though. The upside is that it gives a version-agnostic way to make sure this works, without hoping the defaults don't change. In a client-server setup of the compiler, the client will get all the post-processing bugs, and fixing client bugs is harder, which is why I'm being careful here.

obi1kenobi:
> > Down side of `convert` is that it needs to be added in to the query itself, not the post processing. I'll update the datetime parser to check for millisecond format first.
> 
> I doubt that impacts query performance though. The upside is that it gives a version-agnostic way to make sure this works, without hoping the defaults don't change. In a client-server setup of the compiler, the client will get all the post-processing bugs, and fixing client bugs is harder, which is why I'm being careful here.

I definitely agree with the sentiment and overall principle. That said, I think it's not unreasonable to expect that if the server currently returns ISO-8601-format datetimes, it will keep doing so in the absence of a new ISO datetime formatting standard :)

I definitely appreciate the abundance of caution, but it seems like a major component is stuck on a low-risk concern. I'm not picky about which direction we pick, but I would like to unblock this sooner rather than later since we have much bigger fish to fry than worrying about Microsoft SQL Server all of a sudden changing their datetime format 😄

chewselene:
I can make an issue for adding `convert` to the query. This post processing PR doesn't actually rely on that being implemented so we can come back to it later if we feel like being cautious :)

Issue #663, opened 2019-11-15
Improve API consistency between SQL and other compilation targets


obi1kenobi:
Here is a comparison between our MATCH and SQL APIs at two levels.

MATCH:

`def graphql_to_match(schema, graphql_query, parameters, type_equivalence_hints=None):`
https://github.com/kensho-technologies/graphql-compiler/blob/9489ac4e41fb46d97d6fe51bd7d62f38f428d6c2/graphql_compiler/__init__.py#L25

`def compile_graphql_to_match(schema, graphql_string, type_equivalence_hints=None):`
https://github.com/kensho-technologies/graphql-compiler/blob/9489ac4e41fb46d97d6fe51bd7d62f38f428d6c2/graphql_compiler/compiler/common.py#L24

SQL:

`def graphql_to_sql(sql_schema_info, graphql_query, parameters):`
https://github.com/kensho-technologies/graphql-compiler/blob/9489ac4e41fb46d97d6fe51bd7d62f38f428d6c2/graphql_compiler/__init__.py#L61

`def compile_graphql_to_sql(sql_schema_info, graphql_string):`
https://github.com/kensho-technologies/graphql-compiler/blob/9489ac4e41fb46d97d6fe51bd7d62f38f428d6c2/graphql_compiler/compiler/common.py#L82

I think we should try to make our APIs be consistent between languages, as much as that is possible (e.g. the different languages might require different types of schema info objects, and this is fine).

At the moment, the transition to schema info objects for SQL happens in the user's code. For the non-SQL APIs, however, the user passes in the schema info components separately (the GraphQL schema object and the type equivalence hints). Then the transition happens internally in the package's code, only toward the private _compile_graphql_generic() function.

For consistency, I think the MATCH and other languages' APIs should also take a `schema_info` object that carries the schema and hints, instead of taking a schema and optional hints separately. I think we should make this update as soon as `graphql-core-next` releases a non-beta version [0], and we move to type-checked Python 3.6+ only code.

[0]: https://github.com/graphql-python/graphql-core-next/issues/53

PR #714, opened 2020-01-03
Schema info refactor for usability

MichaelaShtilmanMinkin:
A step towards addressing Issue #663 

Goal:
- To make it easier for users to go from "I have x database, how do I query it?" to actually querying it. A user should easily be able to answer the following questions:
   - How do I get the schema object? And which one?
   - How do I compile?

The idea for the further refactor is to have import paths such as:
- `from graphql_compiler.match import ...`
- `from graphql_compiler.postgresql import ...`
etc.

That way, a user who is using a specific database has a clear path to access everything they need, including the `graphql_to...` or `compile...` functions.

Decisions we've decided to stick to during this redesign
- Keep the core codebase sectioned by functionality (as it is now).
- Create an obvious import path sectioned by database/flavor for the user to get all functionality they need for their database/flavor of choice. That is, the user-facing component will be sectioned by database/flavor.
- Modify the structure of the schema info object to use composition.

This PR will be only to add the new API but will NOT update/replace anything yet. That will be reserved for a future PR in the interest of easier reviewing and transitioning to the new API.

The future PR may or may not include documentation updates as it depends on the size of the code changes.


Pipfile
```diff
@@ -37,6 +37,7 @@ sphinx = ">=1.8,<2"
 [packages]  # Make sure to keep in sync with setup.py requirements.
 arrow = ">=0.15.0,<1"
 funcy = ">=1.7.3,<2"
+dataclasses = ">=0.6"
```
obi1kenobi: Per [PEP508](https://www.python.org/dev/peps/pep-0508/), I think we may actually want `>=0.7,<1; python_version < 3.7` here, though I'm not sure how that will manifest itself in the lockfile. The real longer-term answer we might want down the road might be to switch to `poetry` which seems a bit better suited to package development.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,116 @@
...
+    # interface in the key is the most-derived common supertype
+    # of every GraphQL type in the "value" GraphQL union.
+    # Recursive expansion of type equivalence hints is not performed,
+    # and only type-level correctness of this argument is enforced.
+    # See README.md for more details on everything this parameter does.
+    # *****
+    # Be very careful with this option, as bad input here will
+    # lead to incorrect output queries being generated.
+    # *****
+    type_equivalence_hints: Dict[str, str]  # Optional
```
obi1kenobi: I'm confused by the `# Optional` here, and the lack of `Option[]` around the type. What's the intended use here?

I don't think this value is truly optional, since I think that the auto-generated `__init__` method will expect a positional argument for this field.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,116 @@
...
+
+    # ####
+    # The fields below are required for SQL databases.
+    # ####
+    # Specifying the dialect for which we are compiling, if a dialect exists
+    # e.g. sqlalchemy.dialects.mssql.dialect()
+    dialect: Dialect
```
obi1kenobi: The `if a dialect exists` part is confusing given that the type here is `Dialect` and not `Option[Dialect]`.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,116 @@
...
+    # and only type-level correctness of this argument is enforced.
+    # See README.md for more details on everything this parameter does.
+    # *****
+    # Be very careful with this option, as bad input here will
+    # lead to incorrect output queries being generated.
+    # *****
+    type_equivalence_hints: Dict[str, str]  # Optional
+
+    # ####
+    # The fields below are required for SQL databases.
```
obi1kenobi: I'm not a huge fan of this layout, where these fields are required in one case (SQL) and disallowed (ignored and must not be set) in another case (non-SQL). I'm worried this could end up causing additional complexity and allow bugs to creep in. For example, we'll have to validate these fields differently depending on the backend, which means that having constructed `SchemaInfo` instance does not necessarily mean that the instance is valid and usable.

Instead, consider a `SchemaInfo[BackendSpecificDataT]` generic dataclass, where the `BackendSpecificDataT` can contain different things for different backends. We may then be able to use `NewType` from [typing](https://docs.python.org/3/library/typing.html#newtype) to create "new" types for each of the few combinations of `SchemaInfo` and backend-specific data. Not sure if this solution is ideal either — there may be something better still. But it seems to have fewer risks than this option, where `SchemaInfo` carries all the backend-specific fields directly, even when they are not allowed to be used.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,116 @@
...
+    # They are inferred from the schema and the tables dictionary.
+    join_descriptors: Dict[str, Dict[str, DirectJoinDescriptor]]
+
+    # TODO: consider extending this to being able to support QueryPlanningSchemaInfo
+
+    # TODO: consider - maybe move this outside of the class for easy import for users?
+    # e.g. it would be nice if folks could do graphql_compiler.match import create_match_schema_info
+    # It would be a bit of a bigger refactor so I'd like to float this by first - please opine :)
+    def create_match_schema_info(
+            self, schema: GraphQLSchema, type_equivalence_hints: Optional[str, str] = None
```
obi1kenobi: nit: this seems over-indented, does `black` do this?

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,116 @@
...
+    # dict mapping every GraphQL object type or interface type name in the schema to
+    # dict mapping every vertex field name at that type to a DirectJoinDescriptor.
+    # The tables the join is to be performed on are not specified.
+    # They are inferred from the schema and the tables dictionary.
+    join_descriptors: Dict[str, Dict[str, DirectJoinDescriptor]]
+
+    # TODO: consider extending this to being able to support QueryPlanningSchemaInfo
+
+    # TODO: consider - maybe move this outside of the class for easy import for users?
+    # e.g. it would be nice if folks could do graphql_compiler.match import create_match_schema_info
```
obi1kenobi: Yes, I think they should be outside of here and part of the respective backend.

One nit: while the backend is `MATCH` to differentiate it from `Gremlin`, it would probably be a good idea to make the import be `graphql_compiler.orientdb` based on the vendor to make it easier to find.

Pipfile
MichaelaShtilmanMinkin: Unfortunately, it looks like pipenv explicitly doesn't support multiple versions so that syntax fails:
https://github.com/pypa/pipenv/issues/1050

I believe the workaround is to not specify that Python version in the Pipfile. I got around this for now by specifying the version as 0.6, which is compatible with Python 3.8, though I agree 0.7 would be preference.

graphql_compiler/schema/schema_info.py
bojanserafimov: Very much agree.

Btw, the schema info as it was is precisely the `BackendSpecificDataT` you're talking about. Not sure what the right thing to do is, as I don't understand what the goal is very clearly.

MichaelaShtilmanMinkin:
> This is a very breaking change, and I don't see how it's justified. I also don't see what we gain from this. Is it a common base class for all schema infos? How does that help the user?

There is definitely some refactoring involved and I'm going to switch up how this is structured when I update the PR. After talking to Predrag offline I will give composition a go, where each specific backend is composed of things like `GenericSchemaInfo`.

This was mostly to kick off a brainstorming convo, but the idea is to increase usability by making it obvious from where one needs to import anything and how to construct all required objects. Would love to hear some more thoughts/suggestions about what we can do to make the user experience better.

bojanserafimov:
Initialization for different backends cannot look the same because the schema info initialization for each backend has a different function signature. And we don't want it to look the same. In SQL you have to specify edges manually, whereas there's nothing like that in graph databases. The user needs to be aware of this to make an informed choice. We could make the quick-start setup same across backends though. It would just take a database connection and create a best-effort schema info of the appropriate type. For SQL, it will not create any edges.

I agree that at least having a single import path per backend makes things easier. To bring backend functionality in one place for testing uniformity, i created `backend.py`. It's unusable externally because it's not very user-friendly. It obscures all function docstrings with the indirection. That can be fixed by upgrading `Backend` to an abstract base class, and making `MatchBackend` and the others extend it. Then we'll have a place for nice public docstrings. This is still not perfect, and we can get rid of it if you think of a better solution. There are other approaches for bringing backend functionality in one place:
1. Make a module for each backend that imports all the things relevant to it inside the module. This way the original function docstrings will be preserved, but no uniformity across backends will be guaranteed (that all backends have an emit function for example).
2. Currently the codebase is factorized by functionality first, and backend second (first it's subdivided into `compiler`, `schema`, etc, and then subdivided by backend). We can refactor it to be backend first, functionality second. I think this is too much work for a refactor.

MichaelaShtilmanMinkin:
Per our discussion yesterday, here is the next jab at the refactor. The core ideas for the refactor going forward are as such:
1) Keep the core codebase sectioned by functionality (as it is now).
2) Create an obvious import path for the user to get all functionality they need for their database/flavor of choice.
3) Modify the structure of the schema info object to use composition instead.
4) This PR will be only to add the new API but will NOT update/replace anything yet. That will be reserved for a future PR in the interest of easier reviewing and transitioning to the new API.

I just updated the PR to the latest WIP to make sure we have a tight feedback loop. There are some TODOs / RFC bits in there. There are some incomplete imports, too, as I'm digging around the codebase.

No tests yet.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,166 @@
...
+    # TODO: make sure we are treating empty dict the same as None
+    type_equivalence_hints: Optional[Dict[str, str]]  # Optional
+
+
+@dataclass
+class MatchSchemaInfo:
+    generic_schema_info: GenericSchemaInfo
+
+
+def create_match_schema_info(
```
pmantica1: Discussed offline, but restating it here. 
I don't think that we want to expose any of these `create_*_schema_info` functions as API functions. It is too difficult to come up with the input. Instead we probably want to use `get_sqalchemy_schema_info` and `get_graphql_schema_from_orientdb_schema_data`, though those functions could probably use some refactoring themselves.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,166 @@
...
+    "DirectJoinDescriptor",
+    (
+        "from_column",  # The column in the source table we intend to join on.
+        "to_column",  # The column in the destination table we intend to join on.
+    ),
+)
+
+
+@dataclass
+class GenericSchemaInfo:
```
pmantica1: I don't think that using composition substantially improves usability. You know how to write `schema_info.generic_schema_info.schema` to get the schema and you definitely want to access the schema to see what it looks like.

graphql_compiler/schema/schema_info.py
obi1kenobi: Why not expose them? I agree that most users would just want to auto-infer their database's schema — that means we should expose the functions for that, and I don't think it means we shouldn't have these helper functions as well. I don't see any harm in exposing these, and I can think of use cases where we'd want them.

graphql_compiler/schema/schema_info.py
obi1kenobi: I think we can define a common base class for all the backend-specific schema info types that has properties for each of the `GenericSchemaInfo` fields, and hides the fact that the data actually lives one nesting level deeper.

graphql_compiler/schema/schema_info.py
MichaelaShtilmanMinkin: I was hoping to expose those additional helper functions to help the user form the necessary input. They can then use those results with these helper functions.

graphql_compiler/schema/schema_info.py
MichaelaShtilmanMinkin: @pmantica1 makes a good point though I don't think I quite understand what you're proposing @obi1kenobi - would you mind providing an example?

graphql_compiler/schema/schema_info.py
bojanserafimov: Easy interfaces are good for quick-start, modular interfaces are good for serious projects (immediately after quick-start). The only down side of exposing something is that we lock ourselves into keeping it stable, which should be the case for the schema info classes.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,166 @@
...
+    join_descriptors: Dict[str, Dict[str, DirectJoinDescriptor]]
+
+    # TODO: consider extending this to being able to support QueryPlanningSchemaInfo?
+
+
+@dataclass
+class SQLSchemaInfo:
+    # TODO: consider making a class per flavor? Please opine!
+    #       Pros: every flavor has its dedicated suite of classes / functions
+    #       Cons: redundant code + maintenance
```
bojanserafimov: There's some features where our approach could be very different for each dialect. For example inheritance, or list values. So it might be useful to have separate ones in the long run. However, we should have a clear design of these features first, so we can at least know where the difference in compilation would be. Do they all have the same emit function, but different lowering passes?

I think it's too early to tell, and I don't think we'll need such a redesign until version 3.0. I'd leave it for now.

MichaelaShtilmanMinkin: Sounds reasonable! I wasn't sure how different they could end up being so this is helpful. I'll adjust the comment.

graphql_compiler/api/sql/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2019-present Kensho Technologies, LLC.
+# TODO: consider: import functions that help with setting up edges?
```
obi1kenobi: Probably `from <file_in_this_directory> import *`, if someone really wants to actually import this at all?

graphql_compiler/schema/schema_info.py
obi1kenobi:
```python
@dataclass
class BackendSpecificSchemaInfo:
    generic_schema_info: GenericSchemaInfo

    @property
    def schema(self):
        return self.generic_schema_info.schema
    # etc for all properties of the generic schema 

@dataclass
class SQLSpecificSchemaInfo(BackendSpecificSchemaInfo):
    # SQL-specific fields here
    dialect: Dialect
    # etc for all other properties
```

Then, if `s = SQLSpecificSchemaInfo(...)`, the GraphQL schema is accessible as `s.schema` even though it's in reality stored as `s.generic_schema_info.schema`. This seems like the best of both worlds to me, though I'd love to hear if you have any concerns that I might have missed.

pmantica1: If you run `get_sqlachemy_schema_info` you should get a `SQLSchemaInfo` not a `SQLSpecificSchemaInfo`  so you would still have to type: `s.sql_specific_info.schema`.

graphql_compiler/api/sql/__init__.py
MichaelaShtilmanMinkin: Not sure what this refers to - are you saying the imports shouldn't live here?

graphql_compiler/schema/schema_info.py
```diff
@@ -103,8 +103,30 @@ def create_cypher_schema_info(
...
+    This helps hide that the data actually lives one nesting level deeper.
+    """
+    generic_schema_info: GenericSchemaInfo
+
+    @property
+    def schema(self):
+        return self.generic_schema_info.schema
+
+    @property
+    def type_equivalence_hints(self):
```
obi1kenobi: These need type hints.

graphql_compiler/schema/schema_info.py
obi1kenobi: @pmantica1 I'm not 100% sure I understand why this would be, but even if that is indeed the case, doesn't the same trick apply one more time? Add more property getter methods as necessary.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,174 @@
...
+    # of every GraphQL type in the "value" GraphQL union.
+    # Recursive expansion of type equivalence hints is not performed,
+    # and only type-level correctness of this argument is enforced.
+    # See README.md for more details on everything this parameter does.
+    # *****
+    # Be very careful with this option, as bad input here will
+    # lead to incorrect output queries being generated.
+    # *****
+    # TODO: make sure we are treating empty dict the same as None
+    type_equivalence_hints: Optional[Dict[str, str]]  # Optional
```
obi1kenobi: nit: Probably a good idea to remove the `# Optional` here, unless it's intended to communicate something I'm not seeing.

graphql_compiler/schema/schema_info.py
pmantica1: I still don't think we should be exposing `create_mysql_schema_info`, `create_postgresql_schema_info`, etc. These functions take both `schema`and `vertex_name_to_table` as input. These two arguments both describe the schema of the underlying database and are therefore extremely correlated. Having two highly pieces of correlated data as arguments to a function is something that is prone to introduce bugs.

pmantica1: `get_sqlalchemy_schema_info` does not have this issue since it generates the schema from `vertex_name_to_table` and `direct_edges`

graphql_compiler/schema/schema_info.py
pmantica1: Yes this trick sounds a good idea to me. What we have right now looks good in that sense. (I was trying to clarify that we should have `SQLSchemaInfo(BackendSpecificSchemaInfo)` which is what we have below. Also I want to make clear here that before `SQLSpecificSchemaInfo` was a field of `SQLSchemaInfo` so this was not only a naming issue)

graphql_compiler/schema/schema_info.py
pmantica1: Side note: We now have `get_sqlalchemy_schema_info`, `make_sqlalchemy_schema_info` and `create_sqlalchemy_schema_info`.

obi1kenobi: Agreed on the naming issues. As for whether these functions are exposed, I think they should be.

These functions are not intended as the "default" approach — the default is to create the schema info from an existing database. Calling these functions directly is effectively the power-user "I know what I'm doing" switch, and the documentation should indicate that misuse of these functions can produce a very broken compiler state.

We should do our best to validate that the inputs make sense (e.g. that vertex names in `vertex_name_to_table` match the schema types). But I don't think we should take away power from our power users just because that power could be misused to cause bugs.

graphql_compiler/schema/schema_info.py
MichaelaShtilmanMinkin: (Not in the scope of this diff) I'm starting to think that instead of a `schema info` object we might want to create an even more generic object per language that contains everything that language needs (including methods for compilation, etc). Thoughts?

graphql_compiler/schema/schema_info.py
MichaelaShtilmanMinkin: That's a good point - those names are pretty confusing.

Looking at it now, we shouldn't be exposing such similarly named functions (especially since they have very similar signatures). How about we prioritize the most obvious one to use? Would love to hear your thoughts about which one it is as I haven't tried using either (though I should!).

graphql_compiler/schema/schema_info.py
obi1kenobi: I think that would probably be unnecessary and net-negative. I'd like to distinguish between data (this schema info object) and computation (compilation functions etc.). With that split, there is no hidden state kept by the compilation functions other than what is accepted as input and what is produced as output.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,192 @@
...
+    # interface in the key is the most-derived common supertype
+    # of every GraphQL type in the "value" GraphQL union.
+    # Recursive expansion of type equivalence hints is not performed,
+    # and only type-level correctness of this argument is enforced.
+    # See README.md for more details on everything this parameter does.
+    # *****
+    # Be very careful with this option, as bad input here will
+    # lead to incorrect output queries being generated.
+    # *****
+    # TODO: make sure we are treating empty dict the same as None
```
obi1kenobi: Did you want to leave this `TODO` in? I think this is currently already the case in the code, so I'm not sure what it's referring to.

graphql_compiler/schema/schema_info.py
```diff
@@ -1,13 +1,192 @@
...
+    type_equivalence_hints: Optional[Dict[str, str]]
+
+
+@dataclass
+class GenericSchemaInfoOnly:
```
obi1kenobi: My impression was that this was going to be an abstract base class, but it doesn't seem to set `metaclass=ABCMeta`.

If making it an ABC, then I'd name it something like `BackendSpecificSchemaInfo` instead, since its current name is somewhat confusing.

graphql_compiler/schema/schema_info.py
MichaelaShtilmanMinkin: It was per your request somewhere earlier in the comment or mentioned in-person. I wasn't sure and it was out of the scope of this step in the process so I chose to punt it for later. I can remove if you're confident it's fine.

Issue #665, opened 2019-11-15
Amending custom scalar types in the `input_metadata` dict.


pmantica1:
During compilation, we rely on the `.serialize` attributes  of `GraphQLScalarType` objects. However, when a GraphQLSchema is serialized and parsed, the `.parse_value`, `.parse_literal` and `.serialize` function definitions of custom scalar types are lost because there is no way to serialize them. That is why we have the `amend_custom_scalar_types` function.  

However, it is both tedious to document what this function does and it is easy for someone to forget to call the `amend_custom_scalar_types` when serializing and parsing the schema. 

Therefore, I think that it would be a good idea to stop relying on the `.serialize` attributes of the input schema to be correctly set. We could simply "fix" them during compilation. Doing so would actually be pretty simple. We would just have to "amend" the `input_metadata` dict in a similar fashion.

obi1kenobi:
> We could simply "fix" them during compilation. Doing so would actually be pretty simple. We would just have to "amend" the input_metadata dict in a similar fashion.

Could you explain your proposed approach in a bit more detail? I'm not sure what you have in mind when you say "fix during compilation".

pmantica1:
So the only place that we use `.parse_value` attributes is in the `insert_arguments_into_query` function. `insert_arguments_into_query` takes in a `CompilationResult` and uses its `input_metadata` attribute to infer the types of the arguments. The `input_metadata` dict uses the `GraphQLScalarType` objects of the schema that was used during compilation. Currently, if a `GraphQLScalarType` object in the schema has an invalid `.parse_value` function then `insert_arguments_into_query` will not work correctly since the `GraphQLScalarType` objects used by `insert_arguments_into_query` are ultimately the ones in the `GraphQLSchema` used for compilation.  

What I am saying is that instead of "amending" the schema, we amend the `input_metadata` dict that is returned by  `_compile_graphql_generic` function.

bojanserafimov:
Why not just hard code the parsing functions into the library, and don't pass them around as a part of some object, but just import them?

pmantica1:
Ok yeah that makes sense. Let's do that.

> Why not just hard code the parsing functions into the library, and don't pass them around as a part of some object, but just import them?

Issue #610, opened 2019-10-24
Lack of proper schema serialization function


pmantica1:
We currently have a function called `amend_custom_scalar_types` that amends the definition and descriptions of custom scalar types.  We include documentation for it and expose it as public api. However, this function only amends part of the information loss that happens when serializing a GraphQL schema, (we also lose descriptions for everything else). 

We should have two functions that serialize and parse the GraphQL schema info with minimal information loss and expose that as public api.

obi1kenobi:
The part of the issue related to descriptions, directives, and other AST elements lost in the print-parse transformation might be solved by the release of the new GraphQL core library: https://github.com/graphql-python/graphql-core-next/issues/53

PR #717, opened 2020-01-06
Remove the need for and delete `amend_custom_scalar_types` function

pmantica1:
When a schema is serialized, the definitions of the serialization and parsing functions of custom scalar types are lost. To deal with this issue, we created the `amend_custom_scalar_types` function. However, this solution was not great since it is easy for someone to forget to use this function when serializing and parsing a schema. Therefore, it is better to stop relying on the `parse_value`, `parse_literal` and `serialize` functions of custom  `GraphQLScalarType` that possibly came from a parsed schema, which is what this PR achieves.

Resolves https://github.com/kensho-technologies/graphql-compiler/issues/610 
Resolves https://github.com/kensho-technologies/graphql-compiler/issues/665


graphql_compiler/schema/__init__.py
```diff
@@ -291,9 +295,13 @@ def _parse_date_value(value):
 
 def _serialize_datetime(value):
     """Serialize a DateTime object to its proper ISO-8601 representation."""
-    if not isinstance(value, (datetime, arrow.Arrow)):
+    # Python datetime.datetime is a subclass of datetime.date, but in this case, the two are not
+    # interchangeable. Rather than using isinstance, we will therefore check for exact type
+    # equality.
+    if type(value) not in (datetime, arrow.Arrow):
```
obi1kenobi: nit: why use a tuple with `in` instead of a set?

README.rst
```diff
@@ -2629,22 +2629,6 @@ Neo4j Python client called :code:`neo4j_client`:
     with neo4j_client.driver.session() as session:
         result = session.run(compilation_result.query, parameters)
 
-Amending Parsed Custom Scalar Types
```
obi1kenobi: Do you need to update the readthedocs version of this as well?

pmantica1: Nope. There is no equivalent section in readthedocs.

PR #718, opened 2020-01-06
Fix contributing docs

pmantica1:
The contributing docs in readthedocs and the old contributing and troubleshoting docs got out of sync. Each set of docs contained improvements/additions that the other did not contain. In this diff, I add these missing improvements/additions.

I do not suspect that other docs are suffering from the same issue since we have only recently updated the contributing docs and the schema info API, which was adjusted correctly in both the README and in readthedocs. I am trying to move us to readthedocs soon to prevent these sort of errors from happening again.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -69,12 +69,11 @@ Once the dev environment is prepared, you can run the tests, from the root repos
    py.test graphql_compiler/tests
 
 Some snapshot and integration tests take longer to setup, run, and
-teardown. These can be optionally skipped during development by running
-the tests with the :code:`--skip-slow` flag:
+teardown. These can be optionally skipped during development by running:
 
 .. code:: bash
 
-   py.test graphql_compiler/tests --skip-slow
+   py.test -m 'no-slow'
 
 If you run into any issues, please consult the TROUBLESHOOTING.md file.
 If you encounter and resolve an issue that is not already part of the
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -66,14 +66,14 @@ Once the dev environment is prepared, you can run the tests, from the root repos
    pipenv sync --dev
    pipenv shell
 
-   py.test graphql_compiler/tests
+   pytest graphql_compiler/tests
 
 Some snapshot and integration tests take longer to setup, run, and
 teardown. These can be optionally skipped during development by running:
 
 .. code:: bash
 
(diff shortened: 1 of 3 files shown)


TROUBLESHOOTING.rst
```diff
@@ -135,8 +135,6 @@ because that only installs the interface to MySQL.
 Issues with pyodbc
 ------------------
 
-.. TODO: Add link to ODBC driver installation instructions once we move to Read The Docs.
-
```
pmantica1: This is a tangential fix, but I deleted this comment because I didn't think it was needed in readthedocs since all the contributing info is on one page and you can simply run CTRL+F ODBC.

TROUBLESHOOTING.rst
```diff
@@ -147,4 +145,4 @@ can be resolved on Ubuntu by running:
 .. bash::
 
     # Please modify command to match the version of python you use for development.
-    sudo apt-get install python3.6-dev
+    sudo apt-get install python3.8-dev
```
pmantica1: This change is unrelated, but I make it here in order to prevent conflicts with:
https://github.com/kensho-technologies/graphql-compiler/pull/712

pmantica1: (This is exactly what Bojan said he would do in that PR).

MichaelaShtilmanMinkin: #712 seems like low hanging fruit - let's try to merge it in now or keep the fix here? cc @bojanserafimov

CONTRIBUTING.rst
```diff
@@ -69,12 +69,11 @@ Once the dev environment is prepared, you can run the tests, from the root repos
...
 
 Some snapshot and integration tests take longer to setup, run, and
-teardown. These can be optionally skipped during development by running
-the tests with the :code:`--skip-slow` flag:
+teardown. These can be optionally skipped during development by running:
 
 .. code:: bash
 
-   py.test graphql_compiler/tests --skip-slow
+   py.test -m 'no-slow'
```
obi1kenobi: `py.test` is deprecated in favor of `pytest`: https://github.com/pytest-dev/pytest/issues/1629
```suggestion
   pytest -m 'no-slow'
```

TROUBLESHOOTING.rst
obi1kenobi: Let's make this one supersede #712, they both do essentially the same thing and this PR is more comprehensive.

Issue #611, opened 2019-10-24
Lack of documentation for gremlin


pmantica1:
We currently do not document compilation to gremlin in readthedocs. 

The readthedocs documentation is structured is centered about talking about backends, e.g. OrientDB, not about the compilation language, e.g. MATCH. This makes sense since when you the compiler is meant to serve as a "front end" for someone to be able to target a backend without having to worry about the underlying compiled language.

However, currently, we do not have any backend where it currently makes sense to compile to gremlin. Therefore, we never gremlin anywhere except for the documentation in `@output_source`, which explains how we use this directive to deal with known issues when compiling to gremlin. 

We should consider adding a Compilation Languages section,

PR #719, opened 2020-01-06
Add Execution Model section

pmantica1:
This is copy-paste, except for the bit I added in index.rst

### DIFF
--- a/docs/source/about/execution_model.rst
+++ b/docs/source/about/execution_model.rst
@@ -1,11 +1,10 @@
 Execution model
 ===============
 
-Since the GraphQL compiler can target multiple different query
-languages, each with its own behaviors and limitations, the execution
-model must also be defined as a function of the compilation target
-language. While we strive to minimize the differences between
-compilation targets, some differences are unavoidable.
+Since the GraphQL compiler can target multiple different query languages, each with its own
+behaviors and limitations, the execution model must also be defined as a function of the
+compilation target language. While we strive to minimize the differences between compilation
+targets, some differences are unavoidable.
 
 The compiler abides by the following principles:
 
--- a/docs/source/about/execution_model.rst
+++ b/docs/source/about/execution_model.rst
@@ -19,7 +19,8 @@ The compiler abides by the following principles:
     each result may appear in the nested tree format used by the standard
     GraphQL specification.
 - Each such result must satisfy all directives and types in its corresponding GraphQL query.
-- The returned list of results is **not** guaranteed to be complete!
+- The returned list of results is **not** guaranteed to be complete! (This currently only applies
+  to Gremlin - please follow this :ref:`link <output_source>` for more information on the issue).
 
   - In other words, there may have been additional result sets that satisfy all directives and
     types in the corresponding GraphQL query, but were not returned by the database.
--- a/docs/source/about/execution_model.rst
+++ b/docs/source/about/execution_model.rst
@@ -1,0 +1,27 @@
+Execution model
+===============
+
+Since the GraphQL compiler can target multiple different query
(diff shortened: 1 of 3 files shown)


docs/source/about/execution_model.rst
```diff
@@ -0,0 +1,27 @@
...
+  form of a list of results.
+- The precise format of each such result is defined by each compilation target separately.
+
+  - :code:`gremlin`, :code:`MATCH` and :code:`SQL` return data in a tabular format, where each
+    result is a row of the table, and fields marked for output are columns.
+  - However, future compilation targets may have a different format. For example,
+    each result may appear in the nested tree format used by the standard
+    GraphQL specification.
+- Each such result must satisfy all directives and types in its corresponding GraphQL query.
+- The returned list of results is **not** guaranteed to be complete!
```
MichaelaShtilmanMinkin: As a user, a few questions immediately come to mind:
- Will I know when this happens?
- How will I know?
- What can I do about?

While we don't need to go into the details, a short sentence explaining how this might be handled could be useful.

pmantica1: I see your point.

"The returned list of results is **not** guaranteed to be complete!" only applies to gremlin. We kind of explain that this  happens for gremlin in the `@output_directive` section, (pretty much the only reason this directive exists is to deal with this gremlin issue). However, there is no currently almost no reason why one should be compiling to gremlin. It can only be used to target OrientDB since we don't support other gremlin databases and when targetting OrientDB, compiling to MATCH is strictly the better choice in this case. 
See https://github.com/kensho-technologies/graphql-compiler/issues/611

So I think that it might be a good idea to remove this sentence entirely and maybe add a warning in the OrientDB page not to use gremlin. (Or maybe even take more drastic measures and remove all mention of gremlin from the documentation altogether). 

What do you think? 

Note: I would prefer to keep this as a minimal PR since I want to move to readthedocs as soon as possible in order to avoid issues like the ones in https://github.com/kensho-technologies/graphql-compiler/pulls

obi1kenobi: I think all these questions are on point — thank you @MichaelaShtilmanMinkin for flagging.

In the interest of completing the move to `readthedocs` and then not having duplicate docs to maintain, I'd somewhat prefer we open an issue and work with @JulianGoetz to handle it in the next doc-a-thon. That would mean that for the time being we'd rely on a copy-paste of the slightly outdated old docs. Thoughts?

MichaelaShtilmanMinkin: I think that for the most part, that is fine, but a quick note of `(this only applies to Gremlin - please see section x for details)` would alleviate a lot of concern. Would definitely flag it for more thorough re-phrasing during the doc-a-thon, though.

MichaelaShtilmanMinkin: I think that for the most part, that is fine, but a quick note of `(this only applies to Gremlin - please see section x for details)` would alleviate a lot of concern. Would definitely flag it for more thorough re-phrasing during the doc-a-thon, though.

docs/source/about/execution_model.rst
```diff
@@ -0,0 +1,28 @@
+Execution model
+===============
+
+Since the GraphQL compiler can target multiple different query
+languages, each with its own behaviors and limitations, the execution
+model must also be defined as a function of the compilation target
+language. While we strive to minimize the differences between
+compilation targets, some differences are unavoidable.
```
MichaelaShtilmanMinkin: Is there a reason for the shorter-than-usual line length here?

pmantica1: No there is not. I changed it to have the normal line length.

PR #720, opened 2020-01-07
Add boolean folded field to OutputMetadata and switch to NamedTuples in compiler frontend

chewselene:
Adding a `folded` boolean field to `OutputMetadata` will make post-processing MSSQL folds (#713) much easier as @bojanserafimov pointed out. This PR adds a `folded` field and updates the tests to be compliant with this new field.

Also changes from using `namedtuple` to `NamedTuple` for `OutputMetadata` and `IrAndMetadata` and removes unused namedtuple `LocationStackEntry`.

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -160,9 +160,9 @@ class OutputMetadata(namedtuple("OutputMetadata", ("type", "optional", "folded")
         # Unfortunately, GraphQL types don't have an equality operator defined,
         # and instead have this "is_same_type" function. Hence, we have to override equality here.
         return (
-            is_same_type(self.type, other.type) and
-            self.optional == other.optional and
-            self.folded == other.folded
+            is_same_type(self.type, other.type)
+            and self.optional == other.optional
+            and self.folded == other.folded
         )
 
     def __ne__(self, other):
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -59,7 +59,7 @@ To get from GraphQL AST to IR, we follow the following pattern:
     step F-2. Emit a type coercion block if appropriate, then recurse into the fragment's selection.
     ***************
 """
-from typing import Dict, List, NamedTuple
+from typing import Dict, Generator, List, NamedTuple, Tuple
 
 from graphql import (
     GraphQLInt,
@@ -130,20 +130,17 @@ from .validation import validate_schema_and_query_ast
 
 
 class OutputMetadata(NamedTuple):
-    """Metadata about a query's outputs.
+    """Metadata about a query's outputs."""
 
-    Containing the following members:
-        type: GraphQL type object, such as GraphQLString or GraphQLInt, describing the type of the
-              output value.
-        optional: boolean, whether the output is part of an optional traversal, which would allow a
-                  value of null.
-        folded: boolean, whether the output is within a fold scope. Note that the "type" is expected
-                to be a GraphQLList unless the output field is an _x_count, in which case the type
-                must be GraphQLInt.
-    """
-
+    # The type of the output value.
     type: GraphQLType
+
+    # Whether the output is part of an optional traversal, which would allow a value of null.
     optional: bool
+
+    # Whether the output is within a fold scope. Note that if the output is within a fold scope,
+    # the `type` is expected to be a GraphQLList unless the output field is an _x_count,
+    # in which case the type must be GraphQLInt.
     folded: bool
 
     def __eq__(self, other):
@@ -162,19 +159,18 @@ class OutputMetadata(NamedTuple):
 
 
 class IrAndMetadata(NamedTuple):
-    """Internal representation (IR) and metadata for a schema/query.
+    """Internal representation (IR) and metadata for a particular schema and query combination."""
 
-    Contains the following members:
-        ir_blocks: List[BasicBlock], a list of IR basic block objects describing a query
-        input_metadata: Dict[str, GraphQLType], mapping
-                        expected input parameters -> inferred GraphQL type
-        output_metadata: Dict[str, OutputMetadata], mapping output name -> output metadata
-        query_metadata_table: QueryMetadataTable, describing location metadata
-    """
-
+    # List of basic block objects describing the query.
     ir_blocks: List[BasicBlock]
+
+    # Mapping of expected input parameters -> inferred GraphQL type.
     input_metadata: Dict[str, GraphQLType]
+
+    # Mapping output name -> output metadata.
     output_metadata: Dict[str, OutputMetadata]
+
+    # Describing the location metadata.
     query_metadata_table: QueryMetadataTable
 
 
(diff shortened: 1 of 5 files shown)


graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -964,7 +969,9 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
     # Based on the outputs context data, add an output step and construct the output metadata.
     basic_blocks.append(_compile_output_step(query_metadata_table))
     output_metadata = {
-        name: OutputMetadata(type=info.type, optional=info.optional)
+        name: OutputMetadata(type=info.type, optional=info.optional, folded=True)
+        if isinstance(info.location, FoldScopeLocation)
+        else OutputMetadata(type=info.type, optional=info.optional, folded=False)
```
obi1kenobi: The indentation came out a bit weird here (the `if` is the same clause as the `OutputMetadata` but not indented under it). I'd either use parens to force a new line after the `:`, or do something like:
```python
name: OutputMetadata(type=info.type, optional=info.optional, folded=isinstance(info.location, FoldScopeLocation))
```

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -153,12 +153,17 @@ def _construct_location_stack_entry(location, num_traverses):
 # - type: a GraphQL type object, like String or Integer, describing the type of that output value
 # - optional: boolean, whether the output is part of an optional traversal and
 #             could therefore have a value of null because it did not exist
-class OutputMetadata(namedtuple("OutputMetadata", ("type", "optional"))):
+# - folded: boolean, whether the output is within a fold scope
+class OutputMetadata(namedtuple("OutputMetadata", ("type", "optional", "folded"))):
```
obi1kenobi: Since we're already editing this class in a non-trivial way, consider making it a new-style `NamedTuple` with proper typed fields. I think it would clean up the documentation comment in addition to adding the new functionality.

obi1kenobi: Consider adding an explicit definition + test of whether outputting `_x_count` (by definition inside a `@fold` scope) counts as a "folded" output.

Also consider adding a sanity check on the `type` field, since `folded=True` implies some pretty strong guarantees about the valid values of `type`. The answer here in part depends on the answer to the above.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -127,34 +129,23 @@
...
+    """Metadata about a query's outputs.
 
+    Containing the following members:
+        type: GraphQL type object, such as GraphQLString or GraphQLInt, describing the type of the
+              output value.
+        optional: boolean, whether the output is part of an optional traversal, which would allow a
+                  value of null.
+        folded: boolean, whether the output is within a fold scope. Note that the "type" is expected
+                to be a GraphQLList unless the output field is an _x_count, in which case the type
+                must be GraphQLInt.
```
obi1kenobi: Thoughts on moving the docs to be inline right next to the field definition itself? That way we won't have to specify the type in the docstring as well as in the type hint (e.g. `folded: boolean` would go away since it would be implied by proximity to the field.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -966,14 +964,48 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
...
+                )
+            # Ensure all other folded outputs are GraphQLList.
+            elif info.location.field != COUNT_META_FIELD_NAME and not isinstance(
+                info.type, GraphQLList
+            ):
+                raise AssertionError(
+                    u"Invalid output: non-{} folded output must have type "
+                    u"GraphQLList. Received type {} for folded output "
+                    u"{}.".format(COUNT_META_FIELD_NAME, info.type, name)
+                )
```
obi1kenobi: This might be good to pull into a helper function, there's quite a bit of complexity in this function at the moment.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -1065,11 +1097,7 @@ def ast_to_ir(schema, ast, type_equivalence_hints=None):
                                 *****
 
     Returns:
-        IrAndMetadata named tuple, containing fields:
-        - ir_blocks: a list of IR basic block objects
-        - input_metadata: a dict of expected input parameters (string) -> inferred GraphQL type
-        - output_metadata: a dict of output name (string) -> OutputMetadata object
-        - query_metadata_table: a QueryMetadataTable object containing location metadata
+        IrAndMetadata for the given schema/AST
```
obi1kenobi: nit: thoughts on having an explicit `and` instead of the `/`? I'm a bit worried it might get misinterpreted as an `or`.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -127,34 +129,23 @@
...
-        raise AssertionError(
-            u"Attempted to create a LocationStackEntry namedtuple with an invalid "
-            u'value for "num_traverses" {}. This is not allowed.'.format(num_traverses)
-        )
-    if not isinstance(location, Location):
-        raise AssertionError(
-            u"Attempted to create a LocationStackEntry namedtuple with an invalid "
-            u'value for "location" {}. This is not allowed.'.format(location)
-        )
-    return LocationStackEntry(location=location, num_traverses=num_traverses)
```
obi1kenobi: Was this dead code and removed intentionally? If so, neat!

chewselene: Seemed dead to me! I went to convert `LocationStackEntry` to a `NamedTuple` and realized nothing is calling this code 😄

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -869,6 +865,50 @@ def _validate_all_tags_are_used(metadata):
         )
 
 
+def _validate_output_types(outputs: Generator[Tuple[str, OutputInfo], None, None]):
```
obi1kenobi: In cases like this consider `Iterable[Tuple[str, OutputInfo]]` as an alternative annotation. It's 100% equivalent from a type point of view, just might be a bit friendlier on the eyes since it doesn't end in `None, None`.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -869,6 +865,50 @@ def _validate_all_tags_are_used(metadata):
...
+
+    Checks that _x_count output is within a fold scope, and that its type is GraphQLInt.
+    Checks that all other outputs within a fold scope have GraphQLList as their type.
+
+    Args:
+        outputs: generator producing (output names, OutputInfo) tuples describing a query's outputs
+
+    Raises:
+        AssertionError if an invalid output type is found
+
```
obi1kenobi: nit: remove this blank line

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -869,6 +865,50 @@ def _validate_all_tags_are_used(metadata):
...
+    """
+    for name, info in outputs:
```
obi1kenobi: It might be a bit more pleasant to pull this loop out of the function. I think the function would then a bit easier to reason about — no more generators in or out etc.

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -967,45 +1007,16 @@ def _compile_root_ast_to_ir(schema, ast, type_equivalence_hints=None):
...
-                    u"Invalid output: non-{} folded output must have type "
-                    u"GraphQLList. Received type {} for folded output "
-                    u"{}.".format(COUNT_META_FIELD_NAME, info.type, name)
-                )
-        # Construct output metadata.
-        output_metadata[name] = OutputMetadata(
+    # Construct the output metadata, ensuring that all folded outputs have a valid type.
+    _validate_output_types(query_metadata_table.outputs)
+    output_metadata = {
+        name: OutputMetadata(
```
obi1kenobi: An alternative setup might be something like:
```
output_metadata = {
    name: make_output_metadata(name, info)
    for name, info in query_metadata_table.outputs
}
```
with all of the validation and output metadata creation in the `make_output_metadata` function. That might make it a bit easier to read and follow along.

PR #722, opened 2020-01-08
Reorganize integration tests code

pmantica1:
The code related to setting up integration tests is extremely disorganized. I have been working with the compiler for a while and still didn't have an intuitive where each piece of functionality is located so I decided to reorganize it. (There is no change in functionality). 

This pull request is hard to review by using github alone so reviewers please pull this branch and look at the `integration_tests` folder is organized.


obi1kenobi:
Can you describe the reorganization and what you attempted to achieve with it? You say `extremely disorganized` but that seems vague and likely subjective.

pmantica1:
Here are some concrete issues with our current layout:

1.  We have a `test_data_tools` directory. In this directory, we have a file to load for loading each of the backend's clients. For `neo4j`, we have `neo4j_graph.py`, for `orientdb` we have `orientdb_graph.py`, etc. However, the code involved in setting up the sql clients is included in `data_tool.py`. This design is inconsistent. 
2. In that directory, we also have a `schema.py` file to load the orientdb schema and a `schema.sql`.  file describing the orientdb schema. The filenames of these files are not great since they do not suggest that they are OrientDB specific. Furthermore, the `.sql` suffix makes it seem that the `schema.sql` file is describing the schema of SQL backends. 
3. I think that the fact that we have a file named `data_tool.py` is by itself an issue. It is really unclear what do I put in this file. It is not clear what is considered a data tool and what isn't.
4. The `snapshot_tests` directory depends on the `integration_backend_config.py` file. This hints that the `snapshot_tests` should be part of the `integration_tests`. It should not be a parallel directory.
5. There is a file named `test_backends.py` listing the names of backends. The main issue that I have with this file is its name. There are around 10 other files named `test_*.py` It is quite difficult to find this file and be able to tell that this is a helper file. In my refactor, I rename it to `backends.py`. I have this same issue with `test_helpers.py`. However, modifying the name of this file is going to lead to much code change so I decided not to change it to limit the size of this diff.
6. The `test_cost_estimation.py` and `test_query_pagination.py` files are part of the `snapshot_tests` directory. When I hear the word snapshot, I think of testing against autogenerated data. All the data that these files use is manually specified. They do so a snapshot client fixture for some reason. I tried changing this and it wasn't trivial to do so we can do this in a later pull request.

These are just some of the issues. I am not mentioning everything. 

My proposed re-design is the following:

integration_tests/
    clients/ # It contains a file for each backend. Each file includes all of the necessary code to load 
                 # the schema and client of the backend.
    data/
           integration_data.py # contains standard integration data and the code necessary to load such 
                                           # data for each backend.
          orientdb_snapshot_data.py # contains the code necessary to load the orientdb snapshot data.
          orientdb_snapshot_data.sql # the OrientDB snapshot data.
   helpers.py # Integration test helpers. Originally named `integration_test_helpers.py`. We are 
                     # already in the integration_tests folder. We don't need to add the integration tests  
                     # prefix.
    test_core_features.py # Originally named `test_backends_integration.py`
    test_cost_estimation.py # I just moved this file.
    test_orientdb_snapshots.py # Originally named test_orientdb_match_query.py
    test_query_pagination.py # I just moved this file.

 I also renamed the `test_backends.py` file to `backends.py` and added some extra constants.

bojanserafimov:
Cost estimation tests are snapshot tests because they don't test against the spec, but test that the expected cost estimate values don't change. It's hard to fully specify what's the correct estimate without rewriting the complete implementation inside the spec.

The correct thing to do is not to move them into integration tests, but remove the dependency from integration data by generating a schema graph without schema generation from orientdb. I don't think it's worth the trouble to refactor unless you're working towards this goal.

pmantica1:
I agree with: 
> The correct thing to do is not to move them into integration tests, but remove the dependency from integration data by generating a schema graph without schema generation from orientdb. 

I also see your point that they are snapshots. However, it doesn't really make sense to have a snapshots directory that covers all of the library's snapshots. Ultimately, I think that it makes much more sense to divide test files by functionality in a way that is parallel to how the library is structured. 

How about if I take these two out of the integration tests repository and add a TODO to remove the dependency on the integration test client? I think that the TODO might already exist.

pmantica1:
Upon giving it some thought and discussing this PR with Bojan, I realized that this PR is mostly a patch to fix some of the symptoms caused by design issues with the compiler's tests, such as the problem that Bojan mentioned above about having cost estimation tests depend on integration data. 

Furthermore, even though this PR is mostly copy-paste, it is still making a good amount of controversial design decisions so it deserves to be split into multiple PRs.

Therefore, I decided to close this PR. In the future, I might submit a part of it as another pull request.

PR #723, opened 2020-01-08
Use general range reasoning in the cost estimator

bojanserafimov:
Context: The paginator returns a query describing the first page, and a query describing the rest of the results. The paginator can then be ran again on the remainder, producing the second page, and the remainder. This process continues until we run out of pages. However, to make sure we eventually run out of pages, the cost estimator must be able to understand any filters produced by the paginator. A useful sanity check is that the estimated number of pages of a paginator-generated page is one page.

To satisfy this property, both the paginator and the cost estimator will use the same range reasoning method, using the `int_value_conversion.py` module. In this PR I integrate this module into the cost estimator. In future PRs, the paginator will be built using the same `int_value_conversion` methods.

Since the generic range reasoning methods can reason about datetimes, the functionality of the cost estimator is implicitly extended, so I add a test with cost estimation on datetimes.

### DIFF
--- a/graphql_compiler/cost_estimation/filter_selectivity_utils.py
+++ b/graphql_compiler/cost_estimation/filter_selectivity_utils.py
@@ -31,14 +31,15 @@ Selectivity = namedtuple(
     ),
 )
 
-# IntegerInterval is used to denote a continuous non-empty interval of integers.
-IntegerInterval = namedtuple(
-    "IntegerInterval",
+# Interval is used to denote a continuous non-empty interval of values.
+Interval = namedtuple(
+    "Interval",
     (
-        "lower_bound",  # int or None, inclusive lower bound of integers in the interval.
+        "lower_bound",  # Any or None, inclusive lower bound of integers in the interval.
         # Intervals that do not have a lower bound denote this by using None.
-        "upper_bound",  # int or None, inclusive upper bound of integers in the interval.
-        # Intervals that do not have an upper bound denote this by using None.
+        "upper_bound",  # Any or None, inclusive upper bound of integers in the interval.
+        # Intervals that do not have an upper bound denote this by using None. The lower
+        # and upper bound must be of the same type.
     ),
 )
 
@@ -79,7 +80,7 @@ def _are_filter_fields_uniquely_indexed(filter_fields, unique_indexes):
 
 
 def _create_integer_interval(lower_bound, upper_bound):
-    """Return IntegerInterval for the given bounds, or None if the interval is empty.
+    """Return Interval for the given bounds, or None if the interval is empty.
 
     Args:
         lower_bound: int or None, describing the inclusive lower bound of the integer interval.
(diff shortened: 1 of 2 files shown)


graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -395,6 +394,33 @@ def _get_selectivity_fraction_of_interval(interval, quantiles):
...
+                    parameters[get_parameter_name(filter_argument)],
+                )
+                for filter_argument in filter_info.args
+            ]
+
+            filter_interval = _get_query_interval_of_integer_inequality_filter(
+                parameter_values, filter_info.op_name
+            )
+            interval = _get_intersection_of_intervals(interval, filter_interval)
+    return interval
```
bojanserafimov: moved from line 428..

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -463,22 +472,31 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
                 selectivity_at_field = _combine_filter_selectivities(
                     [selectivity_at_field, selectivity]
                 )
-            elif is_int_field:
-                # int fields are not uniformly distributed but we have quantile information.
+            else:
+                # Get value interval
+                # HACK(bojanserafimov): Abuse of IntegerInterval spec. Constructed with non-interval
+                #                       endpoints. The type should be renamed to allow this usage.
```
bojanserafimov: Will be addressed in this PR

obi1kenobi:
Seems fine so far.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -22,14 +31,15 @@
     ),
 )
 
-# IntegerInterval is used to denote a continuous non-empty interval of integers.
-IntegerInterval = namedtuple(
-    "IntegerInterval",
+# Interval is used to denote a continuous non-empty interval of values.
```
bojanserafimov: I could make it generic, but then it wouldn't be a namedtuple. Mypy doesn't support generic namedtuples.

bojanserafimov: Doing this in a separate PR.. Looks like it's worth it

bojanserafimov: https://github.com/kensho-technologies/graphql-compiler/pull/726

obi1kenobi: Generic dataclasses are a thing, though — and I think that might be worth looking into.

bojanserafimov:
Waiting on https://github.com/kensho-technologies/graphql-compiler/pull/726

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -353,22 +362,29 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
...
+                lower_bound, upper_bound = None, None
+                if interval.lower_bound is not None:
+                    lower_bound = convert_int_to_field_value(
+                        schema_info, location_name, field_name, interval.lower_bound
+                    )
+                if interval.upper_bound is not None:
+                    upper_bound = convert_int_to_field_value(
+                        schema_info, location_name, field_name, interval.upper_bound
+                    )
+                value_interval = Interval[int](lower_bound, upper_bound)
```
bojanserafimov: This is incorrect, and shouldn't pass type checks. Fixing..

PR #725, opened 2020-01-08
Improve pagination planning

bojanserafimov:
1. Include the pagination field in the `VertexPartition`. Some vertices have multiple fields eligible for pagination, and the choice is not obvious.
2. Raise `PaginationError` instead of returning `None`. The effect is similar, but it allows better error messages.
3. Use general range reasoning instead of special-casing uuid, int, datetime types, etc.
4. When the planner can't make a good plan, now it tries its best effort and returns an ok plan


graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,28 +1,38 @@
...
+
+
+class PaginationPlan(NamedTuple):
+    """The intent to split the query with a combination of VertexPartitions"""
+
+    vertex_partitions: Tuple[VertexPartition, ...]
+
+
+# A field with N quantiles has a pagination capacity N / <QUANTILE_REQUIREMENT_FACTOR>
+QUANTILE_REQUIREMENT_FACTOR = 5
```
obi1kenobi: Could you describe this in a bit more detail? I'm not 100% sure what the intent here is and what problem is being solved this way.

I'd also remove the angle-brackets from `<QUANTILE_REQUIREMENT_FACTOR>`, I think that might improve readability.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,28 +1,38 @@
...
 
 
 # TODO(bojanserafimov): Make this function return a best effort pagination plan
 #                       when a good one is not found instead of returning None.
-def try_get_pagination_plan(schema_info, query_ast, number_of_pages):
+def get_pagination_plan(schema_info, query_ast, number_of_pages):
```
obi1kenobi: Just making a note for myself so I don't forget — this needs type hints.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -36,18 +46,48 @@ def try_get_pagination_plan(schema_info, query_ast, number_of_pages):
...
-            return None
+        if quantiles is None:
+            raise GraphQLPaginationError(
+                u"Cannot paginate because no quantiles exist for {}.{}".format(
+                    pagination_node.name.value, pagination_field
+                )
+            )
+        if len(quantiles) < QUANTILE_REQUIREMENT_FACTOR * number_of_pages:
+            raise GraphQLPaginationError(
+                u"Cannot paginate because there are not enought quantiles "
```
obi1kenobi: typo: `enought`

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -36,18 +46,48 @@ def try_get_pagination_plan(schema_info, query_ast, number_of_pages):
...
         # We make sure there's more than enough quantiles because we don't interpolate.
-        if quantiles is None or len(quantiles) < 10 * number_of_pages:
-            return None
+        if quantiles is None:
+            raise GraphQLPaginationError(
+                u"Cannot paginate because no quantiles exist for {}.{}".format(
+                    pagination_node.name.value, pagination_field
+                )
+            )
+        if len(quantiles) < QUANTILE_REQUIREMENT_FACTOR * number_of_pages:
```
obi1kenobi: I'm not sure about this. On one hand, we say we can't paginate — which is fine. On the other, when that happens, people are still going to want to run the query, and they'll probably run it without pagination.

It seems to me like it might be better to make some sort of interpolated best-effort guess and paginate anyway, rather than refusing to paginate and causing a more serious issue in the big picture.

bojanserafimov: Interpolation takes work and can be avoided by just getting more stats, so I've avoided it so far.

Interpolation implies the cost-estimator should interpolate too. I haven't had the time to see if that's the best idea.

However, you bring up a good point, and the least I can do is set `QUANTILE_REQUIREMENT_FACTOR` to 1, or just remove it.

bojanserafimov: Also, if there's 10 quantiles but the user wants 100 pages, I can also say "I can't do 100, but here's 10 pages". That's easy to do.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,28 +1,38 @@
...
-# The intent to split the query at a certain vertex into a certain number of pages.
-VertexPartition = namedtuple(
-    "VertexPartition",
-    (
-        "query_path",  # Tuple[field name : str] leading to the vertex to be split
-        "number_of_splits",  # The number of subdivisions intended for this vertex
-    ),
-)
+class VertexPartition(NamedTuple):
+    """The intent to split the query at a certain vertex into a certain number of pages."""
```
MichaelaShtilmanMinkin: I'm not sure if this represent an intent as much as something else... Do you think there's a more precise word we could use here?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,28 +1,38 @@
...
-PaginationPlan = namedtuple("PaginationPlan", ("vertex_partitions",))  # List[VertexPartition]
+    # field to use for pagination
+    pagination_field: str
+
+    # The number of subdivisions intended for this vertex
+    number_of_splits: int
+
+
+class PaginationPlan(NamedTuple):
+    """The intent to split the query with a combination of VertexPartitions"""
```
MichaelaShtilmanMinkin: Same here (re: `intent`)

graphql_compiler/query_pagination/pagination_planning.py
obi1kenobi: Also if we need 100 pages on a query that looks like `A -> B`, and the ideal pagination location is `A` but it can support only 10 pages, we may want to produce a plan like "10 pages at `A`, and for each of them, 10 pages at `B`". While this is not the ideal plan, it's better than not paginating at all, and arguably it's better (up to a point) than only producing 10 pages (the ones at `A`). Where it ceases to be better than only producing 10 pages at `A` is if we actually needed say 10000 pages total, and we needed to split `B` into so many pieces that we suspect our pages might be severely unbalanced due to `A-B` correlations, but that scenario ends badly no matter what we do.

Obviously, if we are ever in this situation, the correct long-term answer is to collect more stats at `A`. This discussion is only about the fall-back scenario if we haven't yet collected all the stats we want.

bojanserafimov: Yes. But plans with multiple vertex partitions are complicated because they require a change of API. They return 3 continuation queries. Maybe we should change the api to return a list of continuations, while it's still early.

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: I can say `plan` instead. Does that make it clearer?

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: I made it return the best plan it can find, and also return some warnings. It's up to the caller to decide what to do with those warnings. I think it would make sense to use them as auxiliary information in the slow query log (or badly paginated query log), to inform what action to take.

graphql_compiler/exceptions.py
```diff
@@ -36,4 +36,4 @@ class GraphQLInvalidArgumentError(GraphQLError):
 
 
 class GraphQLPaginationError(GraphQLError):
-    """Exception raised when the pagination planner was unable to find a plan."""
+    """Exception raised when the pagination planner was unable to find a good plan."""
```
MichaelaShtilmanMinkin: Goodness is a bit ambiguous imo; shouldn't we only raise an exception if something is truly broken?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,10 +1,39 @@
...
+        super(PaginationFieldNotSpecified, self).__init__(
+            "Pagination field not specified for vertex {}".format(vertex_name)
+        )
+        self.vertex_name = vertex_name
+
+
+class NotEnoughQuantiles(PaginationWarning):
+    def __init__(self, vertex_name, field_name, current_resolution, desired_resolution):
+        super(NotEnoughQuantiles, self).__init__(
+            "Pagination would have been more successful if more quantiles were provided"
```
MichaelaShtilmanMinkin: Need an extra space after `provided`

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,10 +1,39 @@
...
+            "For {}.{}. Currently there is {}, ideally there should be {}".format(
```
MichaelaShtilmanMinkin: `For` should be lowercased

MichaelaShtilmanMinkin: How about `is/are` instead of `is`? I'm assuming the value is likely to be other than 1.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -26,14 +55,17 @@ class PaginationPlan(NamedTuple):
...
 
 
-# A field with N quantiles has a pagination capacity N / <QUANTILE_REQUIREMENT_FACTOR>
-QUANTILE_REQUIREMENT_FACTOR = 5
+def get_pagination_plan(
+    schema_info: QueryPlanningSchemaInfo, query_ast: DocumentNode, number_of_pages: int
+):
+    """Make a PaginationPlan for the given query and number of desired pages if possible.
 
+    Raises GraphQLPaginationError if the statistics object is misconfigured.
```
MichaelaShtilmanMinkin: I think the `Raises` section generally comes at the end of the docstring per the rest of the project's convention. We should probably also include an `Args` and `Returns`.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -26,14 +55,17 @@ class PaginationPlan(NamedTuple):
...
 
-# TODO(bojanserafimov): Make this function return a best effort pagination plan
-#                       when a good one is not found instead of returning None.
-def get_pagination_plan(schema_info, query_ast, number_of_pages):
-    """Make a PaginationPlan for the given query and number of desired pages if possible."""
+    Might paginate to fewer than the desired number of pages if no good pagination plan
```
MichaelaShtilmanMinkin: `Might` or `will`? Under what circumstance other than the lack of a good pagination plan will this happen?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -42,39 +74,64 @@ def get_pagination_plan(schema_info, query_ast, number_of_pages):
...
+            ),
+            [],
         )
+    elif field_supports_range_reasoning(schema_info, pagination_node, pagination_field):
+        quantiles = schema_info.statistics.get_field_quantiles(pagination_node, pagination_field)
+
+        # If the ratio len(quantiles) / number_of_pages is N, then the largest page is
+        # expected to be (N + 1) / N times larger than the smallest page. So if N is 5,
+        # (N + 1) / N = 1.2, which is not too bad.
+        ideal_quantile_resolution = 5 * number_of_pages + 1
```
MichaelaShtilmanMinkin: Are we checking `number_of_pages != 0` anywhere?

graphql_compiler/query_pagination/pagination_planning.py
MichaelaShtilmanMinkin: Maybe `partitioning plan`? Does that make sense in this context?

graphql_compiler/exceptions.py
bojanserafimov: Good catch, I didn't intend to leave this in the PR.

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: split = partition. I could say `Plan to split the query ...` or `Plan to partition the query results by adding filters ...`.

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: I can put Raises at the end. I don't think there's a need to document args

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: Might. I elaborated more on when this happens.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
+    # field to use for pagination
+    pagination_field: str
 
+    # The number of subdivisions intended for this vertex
+    number_of_splits: int
 
-# The intent to split the query with a combination of VertexPartitions
-PaginationPlan = namedtuple("PaginationPlan", ("vertex_partitions",))  # List[VertexPartition]
 
+class PaginationPlan(NamedTuple):
```
obi1kenobi: Since `get_pagination_plan` might not always be able to paginate to the desired number of pages, it would be nice if the pagination plan specified the number of pages it contains.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
 
 from ..ast_manipulation import get_only_query_definition, get_only_selection_from_ast
-from ..cost_estimation.helpers import is_int_field_type, is_uuid4_type
+from ..cost_estimation.helpers import is_uuid4_type
+from ..cost_estimation.int_value_conversion import field_supports_range_reasoning
 from ..exceptions import GraphQLError
+from ..schema.schema_info import QueryPlanningSchemaInfo
+
+
+class PaginationWarning(object):
```
obi1kenobi: Consider dataclasses for all of these, it would make them easier to serialize and send over the wire.

Also, consider using `advisory` instead of `warning`. Warning is well-defined in Python to be [something quite different](https://docs.python.org/3/library/warnings.html) (e.g. the `Warning` base class is a subclass of `Exception`), so it would be good to avoid confusion. I like `advisory` instead because we are giving the user information and advising (but not requiring) that a certain action be taken, which is the exact meaning of the word per [Merriam Webster](https://www.merriam-webster.com/dictionary/advisory).

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
-            return None
-    else:
-        return None
+    elif field_supports_range_reasoning(schema_info, pagination_node, pagination_field):
+        quantiles = schema_info.statistics.get_field_quantiles(pagination_node, pagination_field)
+
+        # If the ratio len(quantiles) / number_of_pages is N, then the largest page is
+        # expected to be (N + 1) / N times larger than the smallest page. So if N is 5,
+        # (N + 1) / N = 1.2, which is not too bad.
+        ideal_quantile_resolution = 5 * number_of_pages + 1
```
obi1kenobi: Can we make this `5` a named constant?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
-        # We make sure there's more than enough quantiles because we don't interpolate.
-        if quantiles is None or len(quantiles) < 10 * number_of_pages:
-            return None
-    else:
-        return None
+    elif field_supports_range_reasoning(schema_info, pagination_node, pagination_field):
+        quantiles = schema_info.statistics.get_field_quantiles(pagination_node, pagination_field)
+
+        # If the ratio len(quantiles) / number_of_pages is N, then the largest page is
+        # expected to be (N + 1) / N times larger than the smallest page. So if N is 5,
```
obi1kenobi: Is this assuming a particular distribution of data? Would you mind giving one more sentence of explanation here as to how this is derived, just enough to give the reader some intuition about why this holds?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
+class PaginationFieldNotSpecified(PaginationWarning):
+    def __init__(self, vertex_name):
+        """Initialize a PaginationFieldNotSpecified PaginationWarning."""
+        super(PaginationFieldNotSpecified, self).__init__(
+            "Pagination field not specified for vertex {}".format(vertex_name)
+        )
+        self.vertex_name = vertex_name
+
+
+class NotEnoughQuantiles(PaginationWarning):
```
obi1kenobi: Consider `InsufficientQuantiles`.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
+    def __init__(self, vertex_name, field_name, current_resolution, desired_resolution):
+        """Initialize a NotEnoughQuantiles PaginationWarning."""
+        super(NotEnoughQuantiles, self).__init__(
+            "Pagination would have been more successful if more quantiles were provided "
+            "for {}.{}. Currently there are {}, ideally there should be {}".format(
+                vertex_name, field_name, current_resolution, desired_resolution
```
obi1kenobi: Great message, very specific about what the advice is.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
+    def __init__(self, message):
+        """Initialize a PaginationWarning with a custom message."""
+        self.message = message
+
+
+class PaginationFieldNotSpecified(PaginationWarning):
+    def __init__(self, vertex_name):
+        """Initialize a PaginationFieldNotSpecified PaginationWarning."""
+        super(PaginationFieldNotSpecified, self).__init__(
+            "Pagination field not specified for vertex {}".format(vertex_name)
```
obi1kenobi: Consider adding the suggested resolution steps here, similarly to how you've done below.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
+                    (VertexPartition((pagination_node,), pagination_field, number_of_pages),)
+                ),
+                [],
+            )
+    else:
+        type_name = (
+            schema_info.schema.get_type(pagination_node.name.value)
+            .fields[pagination_field]
+            .type.name
+        )
```
obi1kenobi: nit: split this into some intermediate named variables, there's a lot going on here

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,156 @@
...
+        raise AssertionError(
```
obi1kenobi: I'm not sure why this is an error rather than returning a no-pagination plan accompanied by an appropriate warning (an advisory, with my suggested nomenclature).

bojanserafimov: The schema info was not initialized properly, if we designated a pagination field with unsupported type. This is on us.

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: To avoid denormalization problems, I added a function that computes that instead.

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: Good suggestion.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,172 @@
...
+        """Initialize a human-readable message."""
+        self.message = (
+            "Pagination would have been more successful if more quantiles were provided "
+            "for {}.{}. Currently there are {}, ideally there should be {}".format(
+                self.vertex_name, self.field_name, self.current_resolution, self.desired_resolution
+            )
+        )
+
 
+class VertexPartition(NamedTuple):
```
MichaelaShtilmanMinkin: `VertexPartitionPlan`, perhaps, to make it clear it is not the actual partition?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,172 @@
...
-#                       when a good one is not found instead of returning None.
-def try_get_pagination_plan(schema_info, query_ast, number_of_pages):
-    """Make a PaginationPlan for the given query and number of desired pages if possible."""
+class PaginationPlan(NamedTuple):
+    """Plan to split the query with a combination of VertexPartitions."""
+
+    vertex_partitions: Tuple[VertexPartition, ...]
+
+
+def get_num_pages_generated_by_plan(plan: PaginationPlan) -> int:
```
MichaelaShtilmanMinkin: It looks like we use `number` instead of `num` elsewhere (`estimate_number_of_pages`). If it's not too long, let's maybe spell it out?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,172 @@
...
+    Might paginate to fewer than the desired number of pages if no good pagination plan
+    is found. This can happen when there's not enough data, or when the planner is not
+    smart enough to find a good plan. In that case it will return along with the result
+    a list of PaginationAdvisorys that indicate why the pagination was not successful.
+    """
     definition_ast = get_only_query_definition(query_ast, GraphQLError)
 
+    if number_of_pages <= 0:
+        raise AssertionError(
+            u"The number of pages should be at least 1: {}".format(number_of_pages)
```
MichaelaShtilmanMinkin: To be more clear about what happened, maybe include `at least 1, but {} were requested.`?

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: if I say `number` I'll have to say `number_of`, otherwise it's weird. Then it gets long.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,174 @@
...
+
+@dataclass
+class PaginationAdvisory(object):
+    message: str
+
+
+@dataclass
+class PaginationFieldNotSpecified(PaginationAdvisory):
+    vertex_name: str
+    message: str = field(init=False)
```
obi1kenobi: I don't think this is required. `message` exists already due to the parent class, and doesn't need a redefinition. I think this is shadowing the parent class' field.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,174 @@
...
-from ..cost_estimation.helpers import is_int_field_type, is_uuid4_type
+from ..cost_estimation.helpers import is_uuid4_type
+from ..cost_estimation.int_value_conversion import field_supports_range_reasoning
 from ..exceptions import GraphQLError
+from ..schema.schema_info import QueryPlanningSchemaInfo
+
+
+@dataclass
+class PaginationAdvisory(object):
+    message: str
```
obi1kenobi: If you want to set this field in a `__post_init__` then consider making it an abstract base class and set `message: str = field(init=False)` here.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,174 @@
...
+        """Initialize a human-readable message."""
+        self.message = (
+            "Pagination would have been more successful if more quantiles were provided "
+            "for {}.{}. Currently there are {}, ideally there should be {}".format(
+                self.vertex_name, self.field_name, self.current_resolution, self.desired_resolution
+            )
+        )
+
 
+class VertexPartitionPlan(NamedTuple):
```
obi1kenobi: Any reason not to make these namedtuples be dataclasses instead?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,175 @@
...
 from ..exceptions import GraphQLError
+from ..schema.schema_info import QueryPlanningSchemaInfo
+
+
+@dataclass
+class PaginationAdvisory(ABC):
+    message: str = field(init=False)
+
+
+@dataclass
```
obi1kenobi: nit: Any reason we don't want all of these to be frozen? Or does the `__post_init__` not play nicely with that for some reason?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,175 @@
...
-def try_get_pagination_plan(schema_info, query_ast, number_of_pages):
-    """Make a PaginationPlan for the given query and number of desired pages if possible."""
+@dataclass(frozen=True)
+class PaginationPlan:
+    """Plan to split the query with a combination of VertexPartitionPlans."""
+
+    vertex_partitions: Tuple[VertexPartitionPlan, ...]
+
+
+def get_num_pages_generated_by_plan(plan: PaginationPlan) -> int:
```
obi1kenobi: nit: can we avoid the `num` abbreviation somehow? How about `get_plan_page_count` or something along those lines?

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,175 @@
...
+    """Return the number of pages that a PaginationPlan would generate."""
+    number_of_pages = 0
+    for vertex_partition in plan.vertex_partitions:
+        number_of_pages *= vertex_partition.number_of_splits
+    return number_of_pages
+
+
+def get_pagination_plan(
+    schema_info: QueryPlanningSchemaInfo, query_ast: DocumentNode, number_of_pages: int
+) -> Tuple[PaginationPlan, List[PaginationAdvisory]]:
```
obi1kenobi: Consider changing `List[PaginationAdvisory]` to `Tuple[PaginationAdvisory, ...]` since we don't really need it to be mutable / a list at all.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,175 @@
...
+                    (VertexPartitionPlan((pagination_node,), pagination_field, number_of_pages),)
+                ),
+                [
+                    InsufficientQuantiles(
+                        pagination_node, pagination_field, len(quantiles), ideal_quantile_resolution
+                    )
+                ],
+            )
+        else:
+            return (
```
obi1kenobi: Some comments on what is going on in all of these if/elif/else branches would be nice. I eventually figured it out, but I've been staring at it for the last 5 minutes and the general approach we're using is fresh in my mind. In six months, I might not be able to figure out what is going on here.

graphql_compiler/query_pagination/pagination_planning.py
obi1kenobi: Can we make the error message say that the schema info is invalid and give some advice on how to correct it then? I would not have figured this out on the basis of the error message.

graphql_compiler/query_pagination/pagination_planning.py
bojanserafimov: yes, post init doesn't work

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+
+    # Ideally, we paginate into the desired number of pages, and don't advise for any changes.
+    # These two values are mutated in the code below if this vertex is not ideal in any way.
+    page_capacity = number_of_pages
+    advisories: Tuple[PaginationAdvisory, ...] = tuple()
+    if is_uuid4_type(schema_info, pagination_node, pagination_field):
+        # UUIDs are uniformly distributed, but it is not wise to make more pages than
+        # there are values.
+        class_count = schema_info.statistics.get_class_count(pagination_node)
+        page_capacity = min(page_capacity, class_count)
```
bojanserafimov: @obi1kenobi this is the only change in functionality since you last saw this code.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+        vertex_type = schema_info.schema.get_type(pagination_node)
+        field_type_name = vertex_type.fields[pagination_field].type.name
+        raise AssertionError(
+            u"Cannot paginate on {}.{} because pagination on {} is not supported ".format(
+                pagination_node, pagination_field, field_type_name
+            )
+        )
+
+    # Ideally, we paginate into the desired number of pages, and don't advise for any changes.
+    # These two values are mutated in the code below if this vertex is not ideal in any way.
```
obi1kenobi: Nit: the values themselves aren't mutated (they are immutable), instead the variables reference new values.
```suggestion
    # These two variables are updated in the code below if this vertex is not ideal in any way.
```

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+        # expected to be (N + 1) / N times larger than the smallest page. This is because
+        # if we assume that quantiles divide the domain into equal chunks, some pages will
+        # span over N + 1 chunks, and some will span over N chunks. So if N is 5,
+        # (N + 1) / N = 1.2, which is not too bad.
+        ideal_min_num_quantiles_per_page = 5
+        ideal_quantile_resolution = ideal_min_num_quantiles_per_page * number_of_pages + 1
+
+        quantiles = schema_info.statistics.get_field_quantiles(pagination_node, pagination_field)
+        if quantiles is None:
+            # If there's no quantiles, we don't paginate. We could try to assume an uniform
```
obi1kenobi:
```suggestion
            # If there are no quantiles, we don't paginate. We could try to assume an uniform
```

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+            # value distribution and make some pagination plan with more than 1 page, but
+            # the cost estimator would need to match the behavior and assume uniform value
+            # distribution when estimating the selectivity of range filters.
+            page_capacity = 1
+        elif len(quantiles) - 1 < number_of_pages:
+            # If we have some quantiles but not enough, we generate a page for each chunk
+            # of values separated by the quantiles.
```
obi1kenobi:
```suggestion
            # of values separated by the quantiles. N quantiles split the domain into N - 1 chunks:
            # the 0% and 100% quantiles capture the min and max values in the domain, and the 
            # remaining N - 2 describe the distribution in between.
```

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+            page_capacity = len(quantiles) - 1
+
+        # If we have less than the ideal number of quantiles, we request for more.
```
obi1kenobi:
```suggestion
        # If we have fewer than the ideal number of quantiles, we advise creating more.
```

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+def get_pagination_plan(
+    schema_info: QueryPlanningSchemaInfo, query_ast: DocumentNode, number_of_pages: int
+) -> Tuple[PaginationPlan, Tuple[PaginationAdvisory, ...]]:
+    """Make a best-effort PaginationPlan and advise on how to improve statistics.
+
+    Might paginate to fewer than the desired number of pages if no good pagination plan
+    is found. This can happen when there's not enough data, or when the planner is not
+    smart enough to find a good plan. In that case it will return along with the result
+    a list of PaginationAdvisorys that indicate why the pagination was not successful.
+    """
```
obi1kenobi: Since this is more than a one-liner, it needs Args/Returns sections:
```suggestion

    Args:
        schema_info: query planning information, including quantile statistics for pagination
        query_ast: GraphQL AST node describing the query being paginated
        number_of_pages: desired number of pages to attempt to paginate the query into

    Returns:
        tuple including a best-effort pagination plan together with a tuple of advisories describing
        any ways in which the pagination plan was less than ideal and how to resolve them
    """
```

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,184 @@
...
+
+def get_pagination_plan(
+    schema_info: QueryPlanningSchemaInfo, query_ast: DocumentNode, number_of_pages: int
+) -> Tuple[PaginationPlan, Tuple[PaginationAdvisory, ...]]:
+    """Make a best-effort PaginationPlan and advise on how to improve statistics.
+
+    Might paginate to fewer than the desired number of pages if no good pagination plan
+    is found. This can happen when there's not enough data, or when the planner is not
+    smart enough to find a good plan. In that case it will return along with the result
+    a list of PaginationAdvisorys that indicate why the pagination was not successful.
```
obi1kenobi: Suggested rewrite with minor clarifications and fixes (e.g. list -> tuple):
```
    Might paginate to fewer than the desired number of pages if unable to find a more satisfactory
    plan, and may even return plans with only a single page. For example, this can happen when
    the captured statistics are insufficient, or when the planner is not smart enough to find a good 
    plan. In these cases, it will return alongside the result a tuple of PaginationAdvisory objects
    that indicate why the desired pagination was not possible. Each PaginationAdvisory states the
    necessary step that may be taken to avoid it in the future.
```
Unfortunately I couldn't use the inline "suggest changes" tool since that only works for one-liners at the moment.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,203 @@
...
-#                       when a good one is not found instead of returning None.
-def try_get_pagination_plan(schema_info, query_ast, number_of_pages):
-    """Make a PaginationPlan for the given query and number of desired pages if possible."""
+    If the issue can be fixed it will return alongside the result a tuple of PaginationAdvisory
+    objects that indicate why the desired pagination was not possible. Each PaginationAdvisory
+    states the necessary step that may be taken to avoid it in the future.
+
+    If the returned plan has fewer than the desired number of pages and no advisories are returned
+    then this vertex is inherently not a good vertex to paginate on, and the caller of this
+    function should consider other vertices for pagination.
```
bojanserafimov: @obi1kenobi slightly modified your suggested docstring to add this. Some motivation of why it's like this:

Consider the next version of the pagination planner, that considers vertices other than just the root of the query. That planner will have to examine the pagination capacity of each vertex, and maybe combine multiple vertex partition plans to meet the desired number of pages. In this scenario, we can't make the leap that "if no perfect vertex partition plan is found, some action can be taken to improve statistics". We can only say that "If the total pagination capacity of all vertex partition plans we can make are insufficient, some action can be taken to improve statistics".

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -1,53 +1,199 @@
...
+    schema_info: QueryPlanningSchemaInfo, query_ast: DocumentNode, number_of_pages: int
+) -> Tuple[PaginationPlan, Tuple[PaginationAdvisory, ...]]:
+    """Make a best-effort PaginationPlan and advise on how to improve statistics.
+
+    Might paginate to fewer than the desired number of pages if unable to find a more satisfactory
+    plan, and may even return plans with only a single page. For example, this can happen when
+    the captured statistics are insufficient, or when the planner is not smart enough to find a
+    good plan.
+
+    If the issue can be fixed it will return alongside the result a tuple of PaginationAdvisory
```
obi1kenobi: Slight clarification, since it is not clear what the `it` here refers to.
```suggestion
    If the issue can be fixed, the return value will also contain a tuple of PaginationAdvisory
```

PR #726, opened 2020-01-09
Add generic Interval class

bojanserafimov:
There's an `IntegerInterval` class we currently use. I move it to a separate file, and make it generic.


graphql_compiler/cost_estimation/interval.py
```diff
@@ -0,0 +1,91 @@
...
+    ) -> None:
+        """Initialize an interval. A value of None means unbounded."""
+        self.lower_bound: Optional[IntervalDomain] = lower_bound
+        self.upper_bound: Optional[IntervalDomain] = upper_bound
+
+    def is_empty(self) -> bool:
+        """Return whether the interval is empty"""
+        if self.lower_bound is None or self.upper_bound is None:
+            return False
+        return self.lower_bound > self.upper_bound
```
obi1kenobi: Presumably this means that `lower_bound == upper_bound` is an interval with a singular value in it? I.e. both bounds are inclusive? Would be nice to call that out specifically.

graphql_compiler/cost_estimation/interval.py
```diff
@@ -0,0 +1,91 @@
...
+            return True
+        return self.lower_bound == other.lower_bound and self.upper_bound == other.upper_bound
+
+
+def measure_interval(interval: Interval[int]) -> Optional[int]:
+    if interval.lower_bound is None or interval.upper_bound is None:
+        return None
+    if interval.is_empty():
+        return 0
+    return interval.upper_bound - interval.lower_bound + 1
```
obi1kenobi: This might be tricky. I'm not sure if subtracting two datetimes with different timezones is a defined / supported operation.

graphql_compiler/cost_estimation/interval.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def _get_stronger_lower_bound(lower_bound_a, lower_bound_b):
```
obi1kenobi: Prefer type hints rather than `int or None` docstrings. I'm assuming this is copy-pasted from the other file.

graphql_compiler/cost_estimation/interval.py
```diff
@@ -0,0 +1,91 @@
...
+    elif upper_bound_a is not None:
+        stronger_upper_bound = upper_bound_a
+    elif upper_bound_b is not None:
+        stronger_upper_bound = upper_bound_b
+
+    return stronger_upper_bound
+
+
+def intersect_intervals(interval_a, interval_b):
+    """Return the intersection of two IntegerIntervals."""
```
obi1kenobi: `IntegerIntervals` needs to be updated here, and this function needs to be generic.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1324,87 +1323,87 @@ class IntegerIntervalTests(unittest.TestCase):
 
     def test_interval_creation(self) -> None:
         """Test that intervals are created correctly, and that empty intervals are detected."""
-        interval = _create_integer_interval(5, 1000)
-        self.assertTrue(interval is not None)
+        interval = Interval[int](5, 1000)
```
obi1kenobi: This is fine and no need to change it now, though I suspect the `[int]` wasn't actually necessary since the integers in the constructor already eliminate all other possibilities for the type parameter.

graphql_compiler/cost_estimation/.#interval.py
```diff
@@ -0,0 +1 @@
+bojan@kadmin-XPS-15-9570.9333:1577977235
```
obi1kenobi: Consider updating the `.gitignore` or your local exclude file to avoid checking in files like this

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -358,38 +256,41 @@ def _get_selectivity_fraction_of_interval(interval, quantiles):
...
         lower_bound_quantile = bisect.bisect_left(proper_quantiles, interval.lower_bound)
-        interval_size = 0.5 + len(proper_quantiles) - lower_bound_quantile
+        interval_size = 0.5 + float(len(proper_quantiles) - lower_bound_quantile)
     else:
         lower_bound_quantile = bisect.bisect_left(proper_quantiles, interval.lower_bound)
         upper_bound_quantile = bisect.bisect_left(proper_quantiles, interval.upper_bound)
         if lower_bound_quantile == upper_bound_quantile:
             # https://math.stackexchange.com/questions/195245/
-            interval_size = 1.0 / 3
+            interval_size = 1.0 / 3.0
```
obi1kenobi: Let's add a bit more info here. Imagine that link stops working — we still want to have some idea of what to look for.

graphql_compiler/cost_estimation/interval.py
bojanserafimov: That's why it's only defined on `Interval[Int]`

graphql_compiler/cost_estimation/interval.py
bojanserafimov: Called it out in the init spec.

graphql_compiler/cost_estimation/interval.py
obi1kenobi: Ah, sorry, I missed that.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -79,112 +70,17 @@ def _convert_uuid_string_to_int(uuid_string):
...
-    strong_upper_bound = _get_stronger_upper_bound(interval_a.upper_bound, interval_b.upper_bound)
-
-    intersection = _create_integer_interval(strong_lower_bound, strong_upper_bound)
-    return intersection
-
-
-def _get_query_interval_of_binary_integer_inequality_filter(parameter_values, filter_operator):
-    """Return IntegerInterval or None of values passing through a binary integer inequality filter.
+def _get_query_interval_of_binary_integer_inequality_filter(
+    parameter_values: List[int], filter_operator: str
```
obi1kenobi: Maybe `Collection[int]` rather than `List[int]`? I'm not sure the list-iness matters here.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -79,112 +70,17 @@ def _convert_uuid_string_to_int(uuid_string):
...
+) -> Interval[int]:
+    """Return Interval[int] of values passing through a binary integer inequality filter.
 
     Args:
         parameter_values: List[int], describing the parameters for the inequality filter.
```
obi1kenobi: Probably safe to remove the types here.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -349,7 +245,9 @@ def _combine_filter_selectivities(selectivities):
     return Selectivity(kind=combined_selectivity_kind, value=combined_selectivity_value)
 
 
-def _get_selectivity_fraction_of_interval(interval, quantiles):
+def _get_selectivity_fraction_of_interval(
+    interval: Interval[IntervalDomain], quantiles: List[IntervalDomain]
+):
```
obi1kenobi:
```suggestion
) -> float:
```

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -358,38 +256,42 @@ def _get_selectivity_fraction_of_interval(interval, quantiles):
...
+                   into N-1 groups of almost equal size. The first element of the list is the
                    smallest known value, and the last element is the largest known value.
                    The i-th element is a value greater than or equal to i/N of all present values.
                    N has to be at least 2.
 
     Returns:
         float, the fraction of the values contained in the interval.
     """
+    if interval.is_empty():
+        return 0
```
obi1kenobi: It's nice to reinforce the float type of the result for the sake of the reader
```suggestion
        return 0.0
```

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1324,87 +1323,87 @@ class IntegerIntervalTests(unittest.TestCase):
 
     def test_interval_creation(self) -> None:
```
obi1kenobi: Some non-int tests on `intersect_intervals` would be nice, especially with date/datetime types.

graphql_compiler/cost_estimation/interval.py
```diff
@@ -0,0 +1,74 @@
...
+        return self.lower_bound > self.upper_bound
+
+    def __eq__(self, other) -> bool:
+        """Compare two intervals. Empty intervals are considered equal to each other."""
+        if self.is_empty() and other.is_empty():
+            return True
+        return self.lower_bound == other.lower_bound and self.upper_bound == other.upper_bound
+
+
+def measure_interval(interval: Interval[int]) -> Optional[int]:
```
obi1kenobi: Maybe `measure_int_interval` instead?

graphql_compiler/cost_estimation/interval.py
```diff
@@ -0,0 +1,74 @@
+import datetime
+from typing import Generic, Optional, TypeVar
+
+
+IntervalDomain = TypeVar("IntervalDomain", int, str, datetime.date, datetime.datetime)
+
+
+class Interval(Generic[IntervalDomain]):
```
obi1kenobi: Consider making this a dataclass, and also consider making it hashable.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
bojanserafimov: Maybe it doesn't. Is it worth checking? Mypy is optional type hints for a reason, let's not obsess about details.

graphql_compiler/cost_estimation/interval.py
bojanserafimov: Don't need to hash it yet. It's not trivial since all empty intervals are equal.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -212,20 +108,20 @@ def _get_query_interval_of_binary_integer_inequality_filter(parameter_values, fi
             u"binary integer inequality operator {}.".format(filter_operator)
         )
 
-    query_interval = _create_integer_interval(lower_bound, upper_bound)
```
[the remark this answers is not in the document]
bojanserafimov: I think that makes the code more obscure. For n-ary filter, we check that the collection has n-1 values. No reason to special case n=1, as there's also ternary filters.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
bojanserafimov: Restricted the domain of intersection to ints to avoid getting too out of scope for auto-pagination.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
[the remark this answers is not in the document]
obi1kenobi: The function is called `_get_query_interval_of_binary_integer_inequality_filter` though — supplying a ternary (or any non-binary) filter to it would seem like a bug to me. I'm not sure I understand why the code would be made more obscure with a one-liner way to get the only element from the collection, care to elaborate?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -247,20 +143,20 @@ def _get_query_interval_of_ternary_integer_inequality_filter(parameter_values, f
...
-        parameter_values: List[int], describing the parameters for the inequality filter.
-        filter_operator: str, describing the inequality filter operation being performed.
+        parameter_values: the parameters for the inequality filter.
+        filter_operator: the inequality filter operation being performed.
 
     Returns:
-        - IntegerInterval namedtuple, non-empty interval of values that pass through the filter.
-        - None if the interval is empty.
+        interval of values that pass through the filter.
     """
```
obi1kenobi: In terms of readability, it feels like this would be a better place to unpack the values from `parameter_values` — immediately inside each `len(parameter_values)` check. Then, the binary/ternary operator functions could take individual arguments for each parameter value, and we wouldn't need the asserts inside them since that part would be correct by construction.

PR #729, opened 2020-01-09
Add fix lint script

bojanserafimov:
Runs black and isort from project root.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -114,7 +114,8 @@ If the style guides differ on a convention, the PEP 8 style guide is preferred.
 Additionally, any contributions must pass the linter :code:`scripts/lint.sh`
 when executed from a pipenv shell (i.e. after running :code:`pipenv shell`).
 To run the linter on changed files only, commit your changes and run
-:code:`scripts/lint.sh --diff`.
+:code:`scripts/lint.sh --diff`. Some linters can automatically fix errors.
+Use :code:`scripts/fix_lint.sh` to run the automatic fixes.
 
 Finally, all python files in the repository must display the copyright
 of the project, to protect the terms of the license. Please make sure
--- a/scripts/fix_lint.sh
+++ b/scripts/fix_lint.sh
@@ -8,6 +8,13 @@ then
     exit 1
 fi
 
+# Exit non-zero on errors, undefined variables, and errors in pipelines.
+set -euo pipefail
+
+# Ensure that the "**" glob operator is applied recursively.
+# Make globs that do not match return null values.
+shopt -s globstar nullglob
+
 # Make sure the current working directory for this script is the root directory.
 cd "$(git -C "$(dirname "${0}")" rev-parse --show-toplevel )"
 
--- a/scripts/fix_lint.sh
+++ b/scripts/fix_lint.sh
@@ -19,7 +19,7 @@ shopt -s globstar nullglob
 cd "$(git -C "$(dirname "${0}")" rev-parse --show-toplevel )"
 
 # Print each command
-set -o xtrace
+set -x
 
 black .
 isort --recursive .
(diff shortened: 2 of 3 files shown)


scripts/fix_lint.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Copyright 2018-present Kensho Technologies, LLC.
+
+# Make sure the current working directory for this script is the root directory.
+cd "$(git -C "$(dirname "${0}")" rev-parse --show-toplevel )"
+
+# Print each command
+set -o xtrace
```
obi1kenobi: `set -x` is a slightly more common way to do this.
```suggestion
set -x
```

scripts/fix_lint.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Copyright 2018-present Kensho Technologies, LLC.
```
obi1kenobi: This is very important to not get wrong, please be careful with copy-pasting.
```suggestion
# Copyright 2020-present Kensho Technologies, LLC.
```

chewselene: not sure that this matters at all, but how do we pick what year goes here?

scripts/fix_lint.sh
```diff
@@ -0,0 +1,11 @@
...
+
```
obi1kenobi: These are just good practices for ~all shell scripts.
```suggestion

# Exit non-zero on errors, undefined variables, and errors in pipelines.
set -euo pipefail

# Ensure that the "**" glob operator is applied recursively.
# Make globs that do not match return null values.
shopt -s globstar nullglob

```

scripts/fix_lint.sh
obi1kenobi: The year the file was added, and yes — it matters a lot. Happy to chat if you're interested in the details.

bojanserafimov: No idea. I usually use a year between the first commit date and now :D Will update this one to 2020

scripts/fix_lint.sh
```diff
@@ -0,0 +1,18 @@
...
+# Assert script is running inside pipenv shell
+if [[ "$VIRTUAL_ENV" == "" ]]
+then
+    echo "Please run pipenv shell first"
+    exit 1
+fi
```
obi1kenobi: Consider reusing the approach we use elsewhere:
```
# Get into the graphql-compiler virtualenv. "pipenv shell" returns exit code 1 if already there,
# which doesn't matter to us so we avoid crashing the script by adding "|| true".
pipenv shell || true
```

CONTRIBUTING.rst
```diff
@@ -114,7 +114,8 @@ If the style guides differ on a convention, the PEP 8 style guide is preferred.
 Additionally, any contributions must pass the linter :code:`scripts/lint.sh`
 when executed from a pipenv shell (i.e. after running :code:`pipenv shell`).
 To run the linter on changed files only, commit your changes and run
-:code:`scripts/lint.sh --diff`.
+:code:`scripts/lint.sh --diff`. Some linters can automatically fix errors.
+Use :code:`scripts/fix_lint.sh` to run the automatic fixes.
```
obi1kenobi: @pmantica1 is there a second copy of this file that needs to be updated?

pmantica1: `docs/source/about/contributing.rst`

scripts/fix_lint.sh
bojanserafimov: Doesn't work when outside pipenv shell, and prints confusing messages when already in pipenv shell

bojanserafimov:
Maybe this should be a flag on `lint.sh` instead of a separate script. Thoughts?

obi1kenobi:
> Maybe this should be a flag on `lint.sh` instead of a separate script. Thoughts?

Maybe, but there are kind of a lot of flags there already. Let's keep it simple.

PR #734, opened 2020-01-14
Add util for name generation

bojanserafimov:
It's a util I'll need for pagination.

### DIFF
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -1,4 +1,6 @@
 # Copyright 2017-present Kensho Technologies, LLC.
+from typing import Sequence
+
 from graphql import GraphQLList, GraphQLNamedType, GraphQLNonNull
 import six
 
@@ -29,3 +31,14 @@ def is_same_type(left, right):
         return is_same_type(left.of_type, right.of_type)
     else:
         return False
+
+
+def find_new_name(desired_name: str, taken_names: Sequence[str], try_original: bool = False) -> str:
+    """Return a name similar to the desired one that is not already taken."""
+    if try_original and desired_name not in taken_names:
+        return desired_name
+
+    index = 0
+    while "{}_{}".format(desired_name, index) in taken_names:
+        index += 1
+    return "{}_{}".format(desired_name, index)
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -33,9 +33,13 @@ def is_same_type(left, right):
         return False
 
 
-def find_new_name(desired_name: str, taken_names: Sequence[str], try_original: bool = False) -> str:
-    """Return a name similar to the desired one that is not already taken."""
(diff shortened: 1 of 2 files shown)


graphql_compiler/global_utils.py
```diff
@@ -29,3 +31,16 @@ def is_same_type(left, right):
         return is_same_type(left.of_type, right.of_type)
     else:
         return False
+
+
+def generate_new_name(base_name: str, taken_names: Sequence[str]) -> str:
```
obi1kenobi: This approach seems unnecessarily wasteful to me. Imagine that `len(taken_names) == N`. The next call to `generate_new_name` will take `O(N^2)` to generate a single new name:
- The `while` loop will only break out on iteration number `N + 1`.
- Since `taken_names` is a `Sequence`, the `in` check in the `while` loop is `O(N)`.

Why not use an infinite generator of new names, where each `next()` call on the generator produces a new, unique name? Or at the very least, why not make `taken_names` be a `Set`, and make the starting index be `len(taken_names)`?

graphql_compiler/global_utils.py
```diff
@@ -29,3 +31,16 @@ def is_same_type(left, right):
...
+    """Return a name based on the provided string that is not already taken.
+
+    This method tries the following names:
+    {base_name}_0, then {base_name}_1, etc.
+    and returns the first one that's not in taken_names
+    """
+    index = 0
+    while "{}_{}".format(base_name, index) in taken_names:
+        index += 1
+    return "{}_{}".format(base_name, index)
```
obi1kenobi: Deduplicate the `.format()` calls, to ensure that the returned value is guaranteed not to have been present in `taken_names`.

graphql_compiler/global_utils.py
```diff
@@ -29,3 +31,16 @@ def is_same_type(left, right):
...
     else:
         return False
+
+
+def generate_new_name(base_name: str, taken_names: Sequence[str]) -> str:
+    """Return a name based on the provided string that is not already taken.
+
+    This method tries the following names:
+    {base_name}_0, then {base_name}_1, etc.
+    and returns the first one that's not in taken_names
```
obi1kenobi: Reminder about our rule that docstrings that span more than one line always have `Args` and `Returns` sections.

bojanserafimov: I thought we got rid of that rule since mypy. In this case it just adds noise (6 redundant lines).

graphql_compiler/global_utils.py
bojanserafimov: Yes, but that is a wasteful use of this function. The solution is not to use it. I don't see an easy general solution, and I'd rather take this function out of `global_utils.py` than spend more work on it.

obi1kenobi: Sounds good, let's do that then.

PR #737, opened 2020-01-22
Add a function that computes the fingerprint of a schema

pmantica1:
I'll wait until we move to ReadTheDocs to add documentation for this function.


pmantica1:
> Looking at the implementation [here](https://github.com/graphql-python/graphql-core-next/blob/8653ef20f196c55b60522cacfc6691c202857a5e/src/graphql/utilities/lexicographic_sort_schema.py), I'm a bit worried that the order of fields in schema directives is sorted.
> 
> Could you add some test cases that include directives that take multiple arguments, that are used e.g. `FIELD_DEFINITION` locations, and see if the order of arguments to such directives is also sorted? The input would look something like:
> 
> ```
> directive @custom_directive(b: String, a: Int) on FIELD_DEFINITION
> 
> type Foo {
>     my_field @custom_directive(b: "abc", a: 123)
> }
> ```
> 
> and the output should have the directive argument order sorted in both the directive definition and its use in the schema.
> 
> If the `lexicographic_sort_schema` function doesn't already do this, we should open an issue. I think we should also offer to make a PR for it, since the GraphQL core library already does most of the heavy lifting and this should not be difficult for us to add.
> 
> I have a couple of smaller nits to pick at (things like typos in docstrings etc.) but figured I'd start with the big stuff.

I think that `test_directive_defintion_argument_order` and `test_directive_definition_location_order` test are already testing in part what you are referring to. In my latest commit I added `test_argument_order_of_field_definition_directive` and premptively fixed some typos.

graphql_compiler/tests/test_schema_fingerprint.py
```diff
@@ -175,6 +175,23 @@ def test_directive_location_order(self):
...
+    def test_argument_order_of_field_definition_directive(self):
+        schema_text1 = """
+            directive @custom_directive(b: String, a: Int) on FIELD_DEFINITION
+
+            type Foo {
+                my_field: Int @custom_directive(b: "123", a: 1)
+            }
+        """
+        schema_text2 = """
+            directive @custom_directive(b: String, a: Int) on FIELD_DEFINITION
```
obi1kenobi: Should this be `b, a` order or `a, b` order? Might be worth digging into the GraphQL spec to see if there's a reason not to reorder here.

graphql_compiler/tests/test_schema_fingerprint.py
```diff
@@ -0,0 +1,245 @@
...
+            directive @filter(
+                op_name: String!
+                value: [String!]
+            ) repeatable on FIELD | INLINE_FRAGMENT
+        """
+        schema_text2 = """
+            directive @filter(
+                value: [String!]
+                op_name: String!
+            ) repeatable on FIELD | INLINE_FRAGMENT
```
obi1kenobi: What happens if we have a repeatable directive on `FIELD_DEFINITION` and it is applied twice to the same field? Is the lexicographic sorting function going to order those two occurrences of the directive in a specific way?

Presumably, the following:
```graphql
type Foo {
    my_field: Int @custom_directive(b: "456", a: 1) @custom_directive(b: "123", a: 23)
}
```
after lexicographic sorting should turn into
```graphql
type Foo {
    my_field: Int @custom_directive(a: 23, b: "123") @custom_directive(a: 1, b: "456")
}
```
and it would be good to verify that schemas like that have the same fingerprints with a test.

Separately, it might be a good idea to check if the GraphQL-core library has tests like this for the lexicographic ordering function, and contribute them in a PR if not.

graphql_compiler/tests/test_schema_fingerprint.py
pmantica1: I intentionally set this as `b, a` order. I already have a test called `test_directive_definition_argument_order` to test the order of arguments in a directive definition.

pmantica1: Did that answer your question? I am a bit confused of what you are trying to get at.

graphql_compiler/tests/test_schema_fingerprint.py
pmantica1: I decided to remove this test entirely. I realized that the GraphQLSchema object does not capture information about directive located at field definitions, (other than for the deprecated directive). In other words, the GraphQLField class does not have a directives field or anything of the sort.

This made me realize though that I was missing tests where the fingerprint was different. (I would have realized that this sooner if I had added a test that expected that two fingerprints were different based on having different directives on field definitions). I added a number of this style of tests.

graphql_compiler/tests/test_schema_fingerprint.py
```diff
@@ -0,0 +1,398 @@
...
+
+from graphql import build_ast_schema, parse, print_schema
+
+from ..schema import compute_schema_fingerprint
+from .test_helpers import compare_graphql
+
+
+def _compare_schema_fingerprints(
+    test_case: unittest.TestCase, schema_text1: str, schema_text2: str, expect_equality: bool = True
+) -> None:
```
obi1kenobi: I can't say I'm a fan of the name of this function or of `expect_equality = True` as an approach. A lot of the tests end in `_compare_schema_fingerprints(self, schema1, schema2)` which does not look like an assertion, and even if you assume it is one, it's not clear whether it's comparing for equality or inequality.

I'd suggest either having different functions for equality and inequality, e.g. `assert_equal_fingerprints()` and `assert_not_equal_fingerprints()`. If you are worried about the repetition between them, you could also pull `compute_schema_fingerprint(build_ast_schema(parse(x)))` into a function of its own.

graphql_compiler/schema/__init__.py
```diff
@@ -447,3 +451,21 @@ def check_for_nondefault_directive_names(directives):
...
+
+def compute_schema_fingerprint(schema: GraphQLSchema) -> str:
+    """Compute a fingerprint compactly representing all the data in the given schema.
+
+    The fingerprint is not sensitive to things like type or field order. This function is guaranteed
+    to be robust enough that if two GraphQLSchema have the same fingerprint, then they also
+    represent the same schema.
+
+    Args:
+        schema: the schema to use.
```
obi1kenobi:
```suggestion
        schema: the schema for which to compute a fingerprint.
```

graphql_compiler/schema/__init__.py
```diff
@@ -447,3 +451,21 @@ def check_for_nondefault_directive_names(directives):
...
+
+    Returns:
+        a fingerprint compactly representing the data in the schema.
```
obi1kenobi:
```suggestion
        a hexadecimal string fingerprint compactly representing the data in the schema.
```

graphql_compiler/tests/test_schema_fingerprint.py
obi1kenobi: Ignoring directives applied to schema elements such as fields and types is not a robust plan. We cannot promise to calculate semantically meaningful fingerprints over "most" of the data for an ill-defined "most".

At the very least, add the tests you removed back in, but rewrite them in such a way that they point out the problematic behavior and fail, then mark them `xfail` to signal that they are known issues and open an issue on Github to resolve them.

We may also be able to use something like the approach in https://github.com/graphql-python/graphql-core-next/issues/65 to get directive values for types, fields, and other schema elements, and add them into the hash separately.

Hopefully https://github.com/graphql/graphql-js/issues/1789 gets addressed before too long (and perhaps we should help it along if you feel like learning some Javascript) so we can do this more thoroughly via the lexicographic sort function.

graphql_compiler/schema/__init__.py
```diff
@@ -447,3 +451,21 @@ def check_for_nondefault_directive_names(directives):
...
             u"Unsupported directives found: {}".format(nondefault_directives_found)
         )
+
+
+def compute_schema_fingerprint(schema: GraphQLSchema) -> str:
+    """Compute a fingerprint compactly representing all the data in the given schema.
+
+    The fingerprint is not sensitive to things like type or field order. This function is guaranteed
+    to be robust enough that if two GraphQLSchema have the same fingerprint, then they also
+    represent the same schema.
```
obi1kenobi:
```suggestion
    represent the same schema.
    
    Because of internal implementation changes, different versions of this library *may* produce
    different fingerprints for the same schema. Since cross-version fingerprint stability
    is an *explicit non-goal* here, changing a schema's fingerprint will not be considered 
    a breaking change.
```

graphql_compiler/tests/test_schema_fingerprint.py
```diff
@@ -64,7 +65,172 @@ def test_schema_fingerprint_basic(self):
...
+            input Input {
+                a: Int
+            }
+        """
+        _compare_schema_fingerprints(self, schema_text1, schema_text2, expect_equality=False)
+
+    @pytest.mark.xfail(
+        strict=True,
+        reason="Type extension information is lost when a schema is print and parsed. "
+        "See https://github.com/graphql/graphql-js/issues/2386",
```
obi1kenobi: nit: I think it might be a bit more readable if you surround the string literal here in parens: `reason=(...)`

pmantica1:
> My remaining nits. I'd still love to get the "repeated directive on the same field" test case in there as well.

Added `test_reordered_repeatable_directives_at_field_definition`. I missed your latest commits. I will address them in an upcoming commit.

graphql_compiler/schema/__init__.py
```diff
@@ -447,3 +451,29 @@ def check_for_nondefault_directive_names(directives):
...
+    
+    Because of internal implementation changes, different versions of this library *may* produce
+    different fingerprints for the same schema. Since cross-version fingerprint stability
+    is an *explicit non-goal* here, changing a schema's fingerprint will not be considered 
+    a breaking change.
+
+    The fingerprint may not include information about directives at field definitions.
```
obi1kenobi: This is way too specific. There are probably half a dozen other kinds of things that are also not covered (e.g. directives on types, on scalars, on extensions) and yet are not mentioned. Let's not actively mislead people by listing only one of the many caveats present here.

I'd keep it more vague and say that this fingerprint is done on a best-effort basis and has some known issues at the moment, and direct people to the discussion in this PR for more details.

PR #738, opened 2020-01-24
Add cost estimation analysis

bojanserafimov:
Currently there's 3 stages in pagination:
1. Cost estimation: compute the expected cardinality of the final result
2. Pagination planning: See where pagination is possible, and decide on one specific plan
3. Pagination execution: Generate parameters, modify the query ast, return 2 queries

With the current setup, the planner does not have enough information to guarantee the generated plan is feasible (that there is enough data to execute it) (see https://github.com/kensho-technologies/graphql-compiler/pull/735/files#diff-766097efd2c183e1f2db00093827ea46R124). So we need to push some of the execution reasoning earlier in the process.

I propose we push all query analysis to the first phase. The idea is similar to the `metadata_table` output from `compile`, and similar to the LLVM analysis passes: https://llvm.org/docs/Passes.html#analysis-passes The distinction between the two is that the metadata table computes everything in one pass, whereas the LLVM computes different results in separate passes that depend on each other.

I prefer LLVM's approach going forward, and here's why: The fact that currently pagination also depends on the metadata table that is generated from the compiler is evidence that **analysis meant for one purpose is useful for other purposes**. I can see future work that would depend on some but not all of the query analysis we currently have:
- cross-db planning
- SQL-like explain
- cost-based rate limiting

Getting only the analysis that's actually needed is not just a performance optimization. Some analysis passes might require additional data that you don't have, and we want to make sure every functionality only asks for data that is required for it, and nothing more.

In this PR, I've implemented 3 analysis passes:
- `get_field_value_intervals`
- `get_distinct_result_set_estimates`
- `get_pagination_capacities`

The results of these passes are used both in pagination planning, and plan execution. The results can be bundled into a `PaginationAnalysisBundle` for ease of plumbing.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ pyodbc = "==4.0.27"
 pyorient = "==1.5.5"
 redis = ">=3.2.1,<4"
 redisgraph = ">=1.7,<1.9"
+
 # Linters and other development tools
 bandit = ">=1.5.1,<2"
 black = "==19.10b0"  # This is still marked as a beta release, pin it explicitly: https://github.com/pypa/pipenv/issues/1760
@@ -27,6 +28,7 @@ pylint = ">=2.4.4,<3"
 pytest = ">=5.1.3,<6"
 pytest-cov = ">=2.6.1,<3"
 snapshottest = ">=0.5.1,<1"
+
 # Documentation requirements. Keep in sync with docs/requirements.txt.
 # Read the Docs doesn't support pipfiles: https://github.com/readthedocs/readthedocs.org/issues/3181
 sphinx-rtd-theme = ">=0.4.3,<1"
@@ -40,10 +42,11 @@ graphql-core = ">=3,<4"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
+
 # The below is necessary to make a few pylint passes work properly, since pylint expects to be able
 # to run "import graphql_compiler" in the environment in which it runs.
-graphql-compiler = {editable = true,path = "."}
-cached-property = "*"
+graphql-compiler = {editable = true, path = "."}
+cached-property = ">=1.5.1,<2"
 
 [requires]
 python_version = "3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,6 @@ pyodbc = "==4.0.27"
 pyorient = "==1.5.5"
 redis = ">=3.2.1,<4"
 redisgraph = ">=1.7,<1.9"
-
 # Linters and other development tools
 bandit = ">=1.5.1,<2"
 black = "==19.10b0"  # This is still marked as a beta release, pin it explicitly: https://github.com/pypa/pipenv/issues/1760
@@ -28,7 +27,6 @@ pylint = ">=2.4.4,<3"
 pytest = ">=5.1.3,<6"
 pytest-cov = ">=2.6.1,<3"
 snapshottest = ">=0.5.1,<1"
-
 # Documentation requirements. Keep in sync with docs/requirements.txt.
 # Read the Docs doesn't support pipfiles: https://github.com/readthedocs/readthedocs.org/issues/3181
 sphinx-rtd-theme = ">=0.4.3,<1"
@@ -42,10 +40,10 @@ graphql-core = ">=3,<4"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
-
 # The below is necessary to make a few pylint passes work properly, since pylint expects to be able
 # to run "import graphql_compiler" in the environment in which it runs.
-graphql-compiler = {editable = true, path = "."}
+graphql-compiler = {editable = true,path = "."}
+cached-property = "*"
 
 [requires]
 python_version = "3.8"
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -59,7 +59,7 @@ def get_field_value_intervals(
 
     Args:
         schema_info: QueryPlanningSchemaInfo
-        query_metadata: representation of the query
+        query_metadata: info on locations, inputs, outputs, and tags in the query
         parameters: parameters used for the query
 
     Returns:
@@ -108,7 +108,7 @@ def get_distinct_result_set_estimates(
 
     Args:
         schema_info: QueryPlanningSchemaInfo
-        query_metadata: a representation of the query
+        query_metadata: info on locations, inputs, outputs, and tags in the query
         parameters: the query parameters
 
     Returns:
@@ -150,7 +150,7 @@ def get_pagination_capacities(
         schema_info: QueryPlanningSchemaInfo
         field_value_intervals: see get_field_value_intervals
         distinct_result_set_estimates: see get_distinct_result_set_estimates
-        query_metadata: a representation of the query
+        query_metadata: info on locations, inputs, outputs, and tags in the query
         parameters: the query parameters
 
     Returns:
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -129,7 +129,8 @@ def get_pagination_capacities(
 
                         # Since we can't be sure the minimum observed value is the
                         # actual minimum value, we treat values less than it as part
-                        # of the first quantile. That's why we drop the minimum and
+                        # of the first quantile. Similarly, we treat values greater than the known
+                        # maximum as part of the last quantile. That's why we drop the minimum and
                         # maximum observed values from the quantile list.
                         proper_quantiles = quantiles[1:-1]
 
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -63,6 +63,9 @@ def get_pagination_capacities(schema_info, field_value_intervals, query_metadata
     for location, location_info in query_metadata.registered_locations:
         vertex_type_name = location_info.type.name
 
+        # TODO Factor in the distinct_result_set_estimates. For instance, if a field has a
+        #      unique filter on it, it's pagination capacity is 1.
+
         for field_name, field in location_info.type.fields.items():
             key = (location.query_path, field_name)
             if not is_meta_field(field_name):
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -66,7 +66,9 @@ def get_distinct_result_set_estimates(schema_info, query_metadata, parameters):
 
 
 
-def get_pagination_capacities(schema_info, field_value_intervals, query_metadata, parameters):
+def get_pagination_capacities(
+    schema_info, field_value_intervals, distinct_result_set_estimates, query_metadata, parameters
+):
     """Get the pagination capacity for each eligible pagination field.
 
     The pagination capacity of a field is defined as the maximum number of pages we can split
@@ -79,15 +81,11 @@ def get_pagination_capacities(schema_info, field_value_intervals, query_metadata
     for location, location_info in query_metadata.registered_locations:
         vertex_type_name = location_info.type.name
 
-        # TODO Factor in the distinct_result_set_estimates. For instance, if a field has a
-        #      unique filter on it, it's pagination capacity is 1.
-
         for field_name, field in location_info.type.fields.items():
             key = (location.query_path, field_name)
             if not is_meta_field(field_name):
                 if is_uuid4_type(schema_info, vertex_type_name, field_name):
-                    # TODO multiply by filter selectivity
-                    pagination_capacities[key] = schema_info.statistics.get_class_count(vertex_type_name)
+                    pagination_capacities[key] = distinct_result_set_estimates[location.query_path]
                 elif field_supports_range_reasoning(schema_info, vertex_type_name, field_name):
                     # TODO write test for this case
                     field_value_interval = field_value_intervals.get(key, Interval(None, None))
@@ -111,6 +109,9 @@ def get_pagination_capacities(schema_info, field_value_intervals, query_metadata
 
                         # TODO take into account duplicate quantile values
 
-                        pagination_capacities[key] = len(relevant_quantiles) + 1
+                        pagination_capacities[key] = min(
+                            len(relevant_quantiles) + 1,
+                            distinct_result_set_estimates[location.query_path]
+                        )
 
     return pagination_capacities
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -154,7 +154,7 @@ def get_pagination_capacities(
         parameters: the query parameters
 
     Returns:
-        The pagination capacity of each Property path
+        the pagination capacity of each PropertyPath
     """
     pagination_capacities = {}
     for location, location_info in query_metadata.registered_locations:
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -176,11 +176,10 @@ def get_pagination_capacities(
                     )
                     if quantiles is not None:
 
-                        # Since we can't be sure the minimum observed value is the
-                        # actual minimum value, we treat values less than it as part
-                        # of the first chunk. Similarly, we treat values greater than the known
-                        # maximum as part of the last chunk. That's why we drop the minimum and
-                        # maximum observed values from the quantile list.
+                        # The first and last values of the quantiles are the minimum and maximum
+                        # observed values. We call all other values the proper quantiles. We don't
+                        # directly use the minimum and maximum values as page boundaries since we
+                        # will most likely generate empty pages.
                         proper_quantiles = quantiles[1:-1]
 
                         # Get the relevant quantiles (ones inside the field_value_interval)
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -63,7 +63,7 @@ def get_field_value_intervals(
         parameters: parameters used for the query
 
     Returns:
-        Dict mapping some PropertyPaths to the interval of filtered values at that field.
+        dict mapping some PropertyPath objects to their interval of allowed values
     """
     field_value_intervals = {}
     for location, location_info in query_metadata.registered_locations:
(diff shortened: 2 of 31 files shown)


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,201 @@
...
+                            len(relevant_quantiles) + 1,
+                            distinct_result_set_estimates[location.query_path],
+                        )
+
+    return pagination_capacities
+
+
+@dataclass
+class QueryAnalysis:
+    """A cache for analysis passes over a fixed query and fixed schema_info."""
```
bojanserafimov: @obi1kenobi this class can simplify the codebase by a lot, and might change some APIs if fully propagated. Let me know if you see any red flags before I go on using it, since it would be a large refactor to reverse it.

obi1kenobi: It seems fine, sure.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,201 @@
...
+                elif field_supports_range_reasoning(schema_info, vertex_type_name, field_name):
+                    field_value_interval = field_value_intervals.get(key, Interval(None, None))
+                    quantiles = schema_info.statistics.get_field_quantiles(
+                        vertex_type_name, field_name
+                    )
+                    if quantiles is not None:
+
+                        # Since we can't be sure the minimum observed value is the
+                        # actual minimum value, we treat values less than it as part
+                        # of the first quantile. That's why we drop the minimum and
```
obi1kenobi:
```suggestion
                        # of the first quantile. Similarly, we treat values greater than the known
                        # maximum as part of the last quantile. That's why we drop the minimum and
```

graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
```diff
@@ -0,0 +1,221 @@
...
+from ...schema_generation.graphql_schema import get_graphql_schema_from_schema_graph
+from ..test_helpers import generate_schema_graph
+
+
+# The following TestCase class uses the 'snapshot_orientdb_client' fixture
+# which pylint does not recognize as a class member.
+# pylint: disable=no-member
+@pytest.mark.slow
+class CostEstimationAnalysisTests(unittest.TestCase):
+    """Test the cost estimation module using standard input data when possible."""
```
obi1kenobi: avoid `standard input` here, it would be confused for `stdin`

Pipfile
```diff
@@ -12,7 +12,6 @@ pyodbc = "==4.0.27"
 pyorient = "==1.5.5"
 redis = ">=3.2.1,<4"
 redisgraph = ">=1.7,<1.9"
-
```
MichaelaShtilmanMinkin: Did the linter remove the newlines for you? I find that they help with readability

Pipfile
```diff
@@ -42,10 +40,10 @@ graphql-core = ">=3,<4"
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
-
 # The below is necessary to make a few pylint passes work properly, since pylint expects to be able
 # to run "import graphql_compiler" in the environment in which it runs.
-graphql-compiler = {editable = true, path = "."}
+graphql-compiler = {editable = true,path = "."}
+cached-property = "*"
```
MichaelaShtilmanMinkin: We probably want to specify some version range here

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+    distinct_result_set_estimates = {}
+    for location, location_info in query_metadata.registered_locations:
+        vertex_type_name = location_info.type.name
+        filter_infos = query_metadata.get_filter_infos(location)
+        class_count = schema_info.statistics.get_class_count(vertex_type_name)
+        distinct_result_set_estimates[location.query_path] = adjust_counts_for_filters(
+            schema_info, filter_infos, parameters, vertex_type_name, class_count
+        )
+
+    # TODO transfer along many-to-one edges
```
MichaelaShtilmanMinkin: Could you please make this TODO a bit more clear? Not sure I understand what it means. Thanks!

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+    schema_info: QueryPlanningSchemaInfo,
+    query_metadata: QueryMetadataTable,
+    parameters: Dict[str, Any],
+) -> Dict[PropertyPath, Interval[Any]]:
+    """Map each PropertyPath in the query to its field value interval if it can be computed.
+
+    Args:
+        - schema_info: QueryPlanningSchemaInfo
+        - query_metadata: a representation of the query
+        - parameters: the parameters used for the query
```
MichaelaShtilmanMinkin: (nit) it would be nice to avoid articles (`a`/`an`/`the`) as the first word of the descriptions of Args/Returns as they don't add much value & make things more verbose / harder to skim.

Pipfile
bojanserafimov: I used `pipenv install` instead of manually editing the pipfile

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+from ..schema.schema_info import QueryPlanningSchemaInfo
+from .filter_selectivity_utils import (
+    adjust_counts_for_filters,
+    get_integer_interval_for_filters_on_field,
+)
+
+
+def _convert_int_interval_to_field_value_interval(
+    schema_info: QueryPlanningSchemaInfo, vertex_type: str, field: str, interval: Interval[int]
+) -> Interval[Any]:
```
pmantica1: Add docstring. This should probably give you a linting error.

Pipfile
MichaelaShtilmanMinkin: Gotcha - can you put them back 😇 ?

graphql_compiler/cost_estimation/analysis.py
pmantica1: Remove preceding `-`

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+    return Interval(lower_bound, upper_bound)
+
+
+def get_field_value_intervals(
+    schema_info: QueryPlanningSchemaInfo,
+    query_metadata: QueryMetadataTable,
+    parameters: Dict[str, Any],
+) -> Dict[PropertyPath, Interval[Any]]:
+    """Map each PropertyPath in the query to its field value interval if it can be computed.
```
pmantica1: This is not really each PropertyPath. It is more like every PropertyPath with a dict. Also, you already passed the 1-line barrier. Please explain what if it can be computed means in the next line.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+            ]
+
+            integer_interval = get_integer_interval_for_filters_on_field(
+                schema_info, filters_on_field, vertex_type_name, field_name, parameters
+            )
+            field_value_interval = _convert_int_interval_to_field_value_interval(
+                schema_info, vertex_type_name, field_name, integer_interval
+            )
+
+            field_value_intervals[(location.query_path, field_name)] = field_value_interval
```
pmantica1: Any reason not to make these two codes blocks helper functions?

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Group filters by field
+        single_field_filters: Dict[str, List[FilterInfo]] = {}
+        for filter_info in filter_infos:
+            if len(filter_info.fields) == 0:
+                raise AssertionError(f"Got filter on 0 fields {filter_info} on {vertex_type_name}")
+            elif len(filter_info.fields) == 1:
+                single_field_filters.setdefault(filter_info.fields[0], []).append(filter_info)
+            else:
+                pass  # We don't do anything for multi-field filters yet
```
pmantica1: Lets raise `NotImplementedError` instead?

graphql_compiler/cost_estimation/analysis.py
bojanserafimov: will add, but there's no linting error

graphql_compiler/cost_estimation/analysis.py
bojanserafimov: We don't raise `NotImplementedError` when the cost estimator is imperfect. We'd have to raise on every query.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+def get_field_value_intervals(
+    schema_info: QueryPlanningSchemaInfo,
+    query_metadata: QueryMetadataTable,
+    parameters: Dict[str, Any],
+) -> Dict[PropertyPath, Interval[Any]]:
+    """Map each PropertyPath in the query to its field value interval if it can be computed.
+
+    Args:
+        - schema_info: QueryPlanningSchemaInfo
+        - query_metadata: a representation of the query
```
pmantica1: How about: "container with info on locations, inputs, outputs, and tags in the query"? (Stolen from the docstring). I don't like having the exact same documentation in two places, but I think that this is substantially better than the docstrings descriptions about the `query_metadata_table` that I've found through the compiler, i.e. "QueryMetadataTable object that captures information about the query".

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+        - parameters: the parameters used for the query
+
+    Returns:
+        Dict mapping some PropertyPaths to the interval of filtered values at that field.
```
pmantica1: nits:
`Dict` to `dict`
`PropertyPaths` to `PropertyPath objects`
It is not the interval of filtered values. It is more like a `dict mapping properties with a filter directive to their interval of allowed values.`

graphql_compiler/cost_estimation/analysis.py
bojanserafimov: This looks simple to me. The helper function wouldn't be reused.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,238 @@
...
+
+def get_distinct_result_set_estimates(
+    schema_info: QueryPlanningSchemaInfo,
+    query_metadata: QueryMetadataTable,
+    parameters: Dict[str, Any],
+) -> Dict[VertexPath, float]:
+    """Map each VertexPath in the query to its distinct result set estimate.
+
+    The distinct result set estimate for a query node is the expected number of different
+    vertices that will appear under it in the result of the query.
```
pmantica1: How about:
`The distinct result set estimate for vertex query node is the expected number of different instances of the vertex type that will appear in the result set of the query. For instance, suppose a query that included an edge traversal from A  to B that also included a unique filter on A. In this case, the distinct result estimate for A is 1 even though the cardinality of the result set might be quite large.`

graphql_compiler/global_utils.py
```diff
@@ -1,11 +1,18 @@
...
 
 from graphql import DocumentNode, GraphQLList, GraphQLNamedType, GraphQLNonNull
 import six
 
 
+# A path starting with a vertex and continuing with edges from that vertex
+VertexPath = Tuple[str, ...]
+
+# A VertexPath with a property on the final vertex of the path
+PropertyPath = Tuple[VertexPath, str]
```
pmantica1: Can you make this a namedtuple?

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,251 @@
...
+                    quantiles = schema_info.statistics.get_field_quantiles(
+                        vertex_type_name, field_name
+                    )
+                    if quantiles is not None:
+
+                        # Since we can't be sure the minimum observed value is the
+                        # actual minimum value, we treat values less than it as part
+                        # of the first quantile. Similarly, we treat values greater than the known
+                        # maximum as part of the last quantile. That's why we drop the minimum and
+                        # maximum observed values from the quantile list.
```
pmantica1: "We treat values less than it part of the quantile". This comment doesn't make any sense. A quantile is a precise value not a collection of values. I know that this is a technicality but I think that it's really important to be as precise as possible here. I know this comment is repeated multiple times throughout the code base so feel free to modify it in those places too. I won't block you though if you don't feel it is worth it at the moment to fix this comment in the other places.

pmantica1: I think that the code would be much clearer if you explained in this comment instead how you used quantiles to compute pages. Maybe through a simple example like "suppose that we have quantiles [1,2,3] then we would have 4 pages [[-inf, 1), [1,2), [2,3), [3, inf]] ". I don't know if this example is right but you probably get the point.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,251 @@
...
+        )
+    return Interval(lower_bound, upper_bound)
+
+
+def get_field_value_intervals(
+    schema_info: QueryPlanningSchemaInfo,
+    query_metadata: QueryMetadataTable,
+    parameters: Dict[str, Any],
+) -> Dict[PropertyPath, Interval[Any]]:
+    """Map the PropertyPath of each field in the query to its field value interval.
```
pmantica1: Its not of each field in the query. It's each field in the query with filter directives.

graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
```diff
@@ -0,0 +1,226 @@
...
+        class_counts = {"Animal": 1000}
+        statistics = LocalStatistics(class_counts)
+        schema_info = QueryPlanningSchemaInfo(
+            schema=graphql_schema,
+            type_equivalence_hints=type_equivalence_hints,
+            schema_graph=schema_graph,
+            statistics=statistics,
+            pagination_keys=pagination_keys,
+            uuid4_fields=uuid4_fields,
+        )
```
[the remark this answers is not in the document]
bojanserafimov: The data initialized here is different for each test case. The bigger problem here is that we rely on an orientdb client. I'd rather work on that than solve test deduplication.

graphql_compiler/global_utils.py
```diff
@@ -1,11 +1,19 @@
...
 
 from graphql import DocumentNode, GraphQLList, GraphQLNamedType, GraphQLNonNull
 import six
 
 
+# A path starting with a vertex and continuing with edges from that vertex
+VertexPath = Tuple[str, ...]
+
+# A VertexPath with a property on the final vertex of the path.
+PropertyPath = namedtuple("PropertyPath", ["vertex_path", "field_name"])
```
pmantica1: Can do you `from typing import NamedTuple` instead to add so nice type annotations?

graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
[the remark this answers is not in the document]
pmantica1: I agree that the bigger problem here is the reliance on the OrientDB client, but that is much more difficult to solve. There is still a decent amount of shared data. How about `get_test_query_planing_schema_info(schema_graph, statistics, pagination_keys)`? The rest of the information is constant or can be derived from the `schema_graph`. You could put in in the `test_helpers` file and add a TODO above like "Use this function to remove code duplication in cost estimation and pagination tests".

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,256 @@
...
+                        max_quantile = len(proper_quantiles)
+                        if field_value_interval.lower_bound is not None:
+                            min_quantile = bisect.bisect_left(
+                                proper_quantiles, field_value_interval.lower_bound
+                            )
+                        if field_value_interval.upper_bound is not None:
+                            max_quantile = bisect.bisect_left(
+                                proper_quantiles, field_value_interval.upper_bound
+                            )
+                        relevant_quantiles = proper_quantiles[min_quantile:max_quantile]
```
pmantica1: This code still doesn't make sense to me based on the comment above. Based on your comment, above we should make 4 pages out of the [1,2,3]  quantiles:  [[-inf, 1), [1,2), [2,3), [3, inf]]. However, assuming no filter bounds and a distinct result estimate of infinity, we would get a pagination capacity of 2 no?

graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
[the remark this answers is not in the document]
pmantica1: Resolved offline.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,255 @@
...
+        parameters: the query parameters
+
+    Returns:
+        the pagination capacity of each PropertyPath
+    """
+    pagination_capacities = {}
+    for location, location_info in query_metadata.registered_locations:
+        vertex_type_name = location_info.type.name
+
+        for field_name, _ in location_info.type.fields.items():
```
obi1kenobi: Use `.keys()` instead since you don't need the value.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,255 @@
...
+                    ]
+                elif field_supports_range_reasoning(schema_info, vertex_type_name, field_name):
+                    field_value_interval = field_value_intervals.get(
+                        property_path, Interval(None, None)
+                    )
+                    quantiles = schema_info.statistics.get_field_quantiles(
+                        vertex_type_name, field_name
+                    )
+                    if quantiles is not None:
+
```
obi1kenobi: nit: Remove this unnecessary blank line.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -0,0 +1,255 @@
...
+                            )
+                        relevant_quantiles = proper_quantiles[min_quantile:max_quantile]
+
+                        # TODO(bojanserafimov): If the relevant quantiles contain duplicates, the
+                        #                       pagination capacity would be lower.
+
+                        pagination_capacities[property_path] = min(
+                            len(relevant_quantiles) + 1,
+                            distinct_result_set_estimates[location.query_path],
+                        )
```
obi1kenobi: This should probably be extracted into one or more functions. For example, you could make the `elif` block be a function call, and then outside that `if/elif` (i.e. at the same indentation level as line 164) you could write something like: 
```python
if pagination_capacity_for_path is not None:
    pagination_capacities[property_path] = pagination_capacity_for_path
```

I think that would make this much cleaner and easier to read.

PR #739, opened 2020-01-27
Implement pagination parameterizer

bojanserafimov:
1. Implement the parameterizer: It modifies the query ast to add filters
2. Remove traces of the old method, and update the `query_splitter` module to work with the new APIs. The one end-to-end test we have now passes, but we still need more end-to-end tests.
3. Add tests for the parameterizer

### DIFF
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -1,6 +1,20 @@
 # Copyright 2017-present Kensho Technologies, LLC.
-from graphql import GraphQLList, GraphQLNamedType, GraphQLNonNull
+from dataclasses import dataclass
+from graphql import GraphQLList, GraphQLNamedType, GraphQLNonNull, DocumentNode
+from typing import Dict, Any
 import six
+
+
+@dataclass
+class QueryStringWithParameters:
+    query_string: str
+    parameters: Dict[str, Any]
+
+
+@dataclass
+class ASTWithParameters:
+    query_ast: DocumentNode
+    parameters: Dict[str, Any]
 
 
 def merge_non_overlapping_dicts(merge_target, new_data):
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -1,13 +1,15 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 from dataclasses import dataclass
-from graphql import GraphQLList, GraphQLNamedType, GraphQLNonNull, DocumentNode
-from typing import Dict, Any
+from typing import Any, Dict
+
+from graphql import DocumentNode, GraphQLList, GraphQLNamedType, GraphQLNonNull
 import six
 
 
 @dataclass
 class QueryStringWithParameters:
     """A query string and parameters that validate against the query."""
+
     query_string: str
     parameters: Dict[str, Any]
 
@@ -15,6 +17,7 @@ class QueryStringWithParameters:
 @dataclass
 class ASTWithParameters:
     """A query AST and parameters that validate against the query."""
+
     query_ast: DocumentNode
     parameters: Dict[str, Any]
 
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -7,12 +7,14 @@ import six
 
 @dataclass
 class QueryStringWithParameters:
+    """A query string and parameters that validate against the query."""
     query_string: str
     parameters: Dict[str, Any]
 
 
 @dataclass
 class ASTWithParameters:
+    """A query AST and parameters that validate against the query."""
     query_ast: DocumentNode
     parameters: Dict[str, Any]
 
--- a/graphql_compiler/query_pagination/parameter_generator.py
+++ b/graphql_compiler/query_pagination/parameter_generator.py
@@ -238,29 +238,3 @@ def generate_parameters_for_vertex_partition(
         return _compute_parameters_for_non_uuid_field(
             schema_info, field_value_interval, vertex_partition, vertex_type, pagination_field
         )
-
-
-def generate_parameters_for_parameterized_query(
-    schema_info, parameterized_pagination_queries, num_pages
-):
-    """Generate parameters for the given parameterized pagination queries.
-
-    Args:
-        schema_info: QueryPlanningSchemaInfo
-        parameterized_pagination_queries: ParameterizedPaginationQueries namedtuple, parameterized
-                                          queries for which parameters are being generated.
-        num_pages: int, number of pages to split the query into.
-
-    Returns:
-        two dicts:
-            - dict, parameters with which to execute the page query. The next page query's
-              parameters are generated such that only a page of the original query's result data is
-              produced when executed.
-            - dict, parameters with which to execute the remainder query. The remainder query
-              parameters are generated such that they produce the remainder of the original query's
-              result data when executed.
-    """
-    # TODO(bojanserafimov): Replace this method with generate_parameters_for_vertex_partition. Its
-    #                       api is simpler (no dependence on parameterization), and it uses the
-    #                       VertexPartitionPlan.
-    raise NotImplementedError()
--- a/graphql_compiler/query_pagination/parameter_generator.py
+++ b/graphql_compiler/query_pagination/parameter_generator.py
@@ -1,7 +1,7 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 import bisect
 import itertools
-from typing import Any, Dict, Iterable, cast
+from typing import Any, Dict, Iterator, cast
 
 from graphql import DocumentNode
 from graphql.language.printer import print_ast
@@ -20,7 +20,7 @@ from ..schema.schema_info import QueryPlanningSchemaInfo
 from .pagination_planning import VertexPartitionPlan
 
 
-def _sum_partition(number: int, num_splits: int) -> Iterable[int]:
+def _sum_partition(number: int, num_splits: int) -> Iterator[int]:
     """Represent an integer as a sum of N almost equal integers, sorted in descending order.
 
     Example: _sum_partition(5, 3) = [2, 2, 1]
@@ -43,7 +43,7 @@ def _sum_partition(number: int, num_splits: int) -> Iterable[int]:
     )
 
 
-def _deduplicate_sorted_generator(generator: Iterable[Any]) -> Iterable[Any]:
+def _deduplicate_sorted_generator(generator: Iterator[Any]) -> Iterator[Any]:
     """Return a generator that skips repeated values in the given sorted generator."""
     prev = object()
     for i in generator:
@@ -75,7 +75,7 @@ def _compute_parameters_for_uuid_field(
     vertex_partition: VertexPartitionPlan,
     vertex_type: str,
     field: str,
-) -> Iterable[Any]:
+) -> Iterator[Any]:
     """Return a generator of parameter values for the vertex partition at a uuid field.
 
     See generate_parameters_for_vertex_partition for more details.
@@ -118,7 +118,7 @@ def _compute_parameters_for_non_uuid_field(
     vertex_partition: VertexPartitionPlan,
     vertex_type: str,
     field: str,
-) -> Iterable[Any]:
+) -> Iterator[Any]:
     """Return a generator of parameter values for the vertex partition at a non-uuid field.
 
     See generate_parameters_for_vertex_partition for more details.
@@ -179,7 +179,7 @@ def generate_parameters_for_vertex_partition(
     query_ast: DocumentNode,
     parameters: Dict[str, Any],
     vertex_partition: VertexPartitionPlan,
-) -> Iterable[Any]:
+) -> Iterator[Any]:
     """Return a generator of parameter values that realize the vertex partition.
 
     Composability guarantee: The values returned can be used to create
--- a/graphql_compiler/query_pagination/query_parameterizer.py
+++ b/graphql_compiler/query_pagination/query_parameterizer.py
@@ -63,6 +63,7 @@ def _add_pagination_filter(
     query_path: Tuple[str, ...],
     pagination_field: str,
     directive_to_add: DirectiveNode,
+    extended_parameters: Dict[str, Any],
 ) -> Tuple[DocumentNode, List[str]]:
     """Add the filter to the target field, returning a query and the names of removed filters.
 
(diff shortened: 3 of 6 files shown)


graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,145 @@
...
+    while "{}_{}".format(base_name, index) in taken_names:
+        index += 1
+    return "{}_{}".format(base_name, index)
+
+
+def _add_pagination_filters(query_ast, query_path, pagination_field, directive_to_add):
+    if not isinstance(query_ast, (FieldNode, InlineFragmentNode, OperationDefinitionNode)):
+        raise AssertionError(
+            u'Input AST is of type "{}", which should not be a selection.'
+            u"".format(type(query_ast).__name__)
```
obi1kenobi: Why not just use an f-string?

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,145 @@
...
+            field_name = get_ast_field_name(selection_ast)
+            if field_name == pagination_field:
+                found_field = True
+                new_selection_ast = copy(selection_ast)
+                new_selection_ast.directives = copy(selection_ast.directives)
+                found_directive = False
+
+                new_directives = []
+                for directive in selection_ast.directives:
+                    if (directive.arguments[0].value.value == directive_to_add.arguments[0].value.value):
```
obi1kenobi: Maybe use some local variables to make this a bit less complicated?

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,145 @@
...
         query_ast: Document, query that is being paginated.
         parameters: dict, list of parameters for the given query.
+        vertex_partition: pagination plan
 
     Returns:
-        ParameterizedPaginationQueries namedtuple
+        next_page_ast: Ast for the first page. Includes an additional filter.
+        next_page_removed_params: ???
+        remainder_ast: Ast for the remainder query. Includes an additional filter.
+        remainder_removed_params: ???
```
obi1kenobi: Bad refactor?

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,145 @@
...
-    controlled.
+        if not found_field:
+            raise AssertionError()
+
+    new_ast = copy(query_ast)
+    new_ast.selection_set = SelectionSetNode(selections=new_selections)
+    return new_ast, removed_argument
+
+
+def _make_directive(op_name, param_name):
```
obi1kenobi: Let's name this something a bit more descriptive.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,194 @@
...
+        query_type,
+        vertex_partition.query_path,
+        vertex_partition.pagination_field,
+        _make_binary_filter_directive_node(">=", param_name),
+    )
+
+    page_parameters = {k: v for k, v in parameters.items() if k not in next_page_removed_parameters}
+    remainder_parameters = {
+        k: v for k, v in parameters.items() if k not in remainder_removed_parameters
+    }
```
bojanserafimov: Here we remove the parameters corresponding to any redundant filters that were replaced with stronger filters. The new filters should be stronger since the parameter generator takes the existing filters into account, and makes sure to generate values in between the existing range of possible values. But still, it would be nice to assert that this new filter is indeed stronger. However, we can't do that because we don't have the values to be inserted yet.

@obi1kenobi This is the biggest doubt I have about the current API. Let me know if you see a better way to guarantee correctness.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,194 @@
...
+    """Add the filter to the target field, returning a query and the names of removed filters.
+
+    Args:
+        query_ast: The query in which we are adding a filter
+        query_path: The path to the pagination vertex
+        pagination_field: The field on which we are adding a filter
+        directive_to_add: The filter directive to add
+
+    Returns:
+        new_ast: A query with the filter inserted, and any filters on the same location with
```
obi1kenobi:
```suggestion
        tuple (new_ast, removed_parameters)
        new_ast: A query with the filter inserted, and any filters on the same location with
```

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,194 @@
...
+                    else:
+                        new_directives.append(directive)
+                new_directives.append(directive_to_add)
+                new_selection_ast.directives = new_directives
+            new_selections.append(new_selection_ast)
+        if not found_field:
+            new_selections.insert(
+                0, FieldNode(name=NameNode(value=pagination_field), directives=[directive_to_add])
+            )
+    else:
```
obi1kenobi: Do you think there are potential refactorings that might make this function a bit easier on the eyes? There's a lot going on here, and it's not easy to follow.

bojanserafimov: yes, working on it

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,194 @@
...
-        parameters: dict, list of parameters for the given query.
+        query: the query to parameterize
+        vertex_partition: pagination plan
 
     Returns:
-        ParameterizedPaginationQueries namedtuple
+        # TODO use ASTWithParameters struct to simplify API
+        next_page: Ast and params for next page.
+        remainder_ast: Ast and params for remainder.
+        param_name: The parameter name used in the new filters.
```
bojanserafimov: Currently this function is independent of what value will be inserted to define the boundary between two pages. This function outputs a `param_name`, and the caller is responsible for inserting the desired value.

This flexibility comes at a cost: The caller is expected to insert a value that makes the filter stronger than any of the filters that were removed in its place. For example, if there's an existing `x > $val, where val = 5` filter, the parameterizer will remove it and in it's place put `x > $__paged_param_0`. The caller will probably use a value greater than `5` for the `__paged_param_0` parameter, since that is in the spec of the parameter generator.

However, it would be nice to assert that all filters that were removed are replaced by stronger filters. One way to do that is to pass the desired parameter value into this function. That will add a cross-dependency on parameter generation, but it's a dependency that's very easy to mock out in tests, so we can still test the modules separately.

@obi1kenobi thoughts?

obi1kenobi: It seems reasonable to make this code assert its correctness, especially a bug there would be quite tricky to notice and fix. Let's go for adding the dependency, especially if you think it's easy to mock out.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,223 @@
...
+                new_selection_ast.directives = new_directives
+            new_selections.append(new_selection_ast)
+        if not found_field:
+            new_selections.insert(
+                0, FieldNode(name=NameNode(value=pagination_field), directives=[directive_to_add])
+            )
+    else:
+        # Recurse until the target vertex is reached.
+        if query_ast.selection_set is None:
+            raise AssertionError()
```
bojanserafimov: Working on filling all the error messages

graphql_compiler/query_pagination/parameter_generator.py
```diff
@@ -20,7 +20,7 @@
 from .pagination_planning import VertexPartitionPlan
 
 
-def _sum_partition(number: int, num_splits: int) -> Iterable[int]:
+def _sum_partition(number: int, num_splits: int) -> Iterator[int]:
```
bojanserafimov: `Iterable` doesn't have a `next` method

graphql_compiler/global_utils.py
```diff
@@ -1,8 +1,27 @@
...
+    query_string: str
+    parameters: Dict[str, Any]
+
+
+@dataclass
+class ASTWithParameters:
+    """A query AST and parameters that validate against the query."""
+
+    query_ast: DocumentNode
+    parameters: Dict[str, Any]
```
bojanserafimov: Moved from pagination module

graphql_compiler/query_pagination/query_splitter.py
```diff
@@ -42,21 +33,23 @@ def split_into_page_query_and_remainder_query(schema_info, query_ast, parameters
         raise AssertionError(
             u"Could not split query {} into pagination queries for the next page"
             u" of results, as the number of pages {} must be greater than 1: {}".format(
-                query_ast, num_pages, parameters
+                query.query_ast, num_pages, query.parameters
             )
         )
 
-    parameterized_queries = generate_parameterized_queries(schema_info, query_ast, parameters)
+    # TODO propagate advisories to top-level
```
bojanserafimov: This will require an API change in the query splitter, which is a bit out of scope for this PR.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,288 @@
...
+                    parameter_name = _get_binary_filter_node_parameter(directive)
+                    parameter_value = new_parameters[parameter_name]
+                    if not _is_new_filter_stronger(
+                        operation, new_directive_parameter_value, parameter_value
+                    ):
+                        raise AssertionError(
+                            f"Pagination filter {directive_to_add} on "
+                            f"{pagination_field} is not stronger than "
+                            f"an existing filter {directive}. This is "
+                            f"likely a bug in parameter generation."
```
obi1kenobi: Maybe print the query string in addition to all this, so we can reproduce more easily?

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,288 @@
...
+    parameter_value: Any,
+) -> Tuple[ASTWithParameters, ASTWithParameters]:
+    """Generate two parameterized queries that can be used to paginate over a given query.
 
     Args:
         schema_info: QueryPlanningSchemaInfo
-        query_ast: Document, query that is being paginated.
-        parameters: dict, list of parameters for the given query.
+        query: the query to parameterize
+        vertex_partition: pagination plan
```
obi1kenobi: add a bit more context here

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,288 @@
...
+        parameter_value: the value of the parameter used for pagination
 
     Returns:
-        ParameterizedPaginationQueries namedtuple
+        next_page: Ast and params for next page.
```
obi1kenobi:
```suggestion
        tuple (next_page, remainder)
        next_page: AST and params for next page.
```

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,288 @@
...
+        remainder_ast: Ast and params for remainder.
```
obi1kenobi:
```suggestion
        remainder: AST and params for the remainder query that returns all results not on the next page.
```

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,288 @@
...
     """
-    raise NotImplementedError()
+    query_type = get_only_query_definition(query.query_ast, GraphQLError)
```
obi1kenobi: Is `query_type` an appropriate name here? I don't think this is a type based on how it's used, it looks more like the root AST definition node for the query.

graphql_compiler/query_pagination/query_splitter.py
obi1kenobi: I assume next PR?

graphql_compiler/query_pagination/query_splitter.py
```diff
@@ -42,21 +33,23 @@ def split_into_page_query_and_remainder_query(schema_info, query_ast, parameters
...
+    pagination_plan, _ = get_pagination_plan(schema_info, query.query_ast, num_pages)
+    if len(pagination_plan.vertex_partitions) != 1:
+        raise NotImplementedError(
+            u"We only support pagination plans with one vertex partition. "
+            u"Reveived {}".format(pagination_plan)
```
obi1kenobi:
```suggestion
            u"Received {}".format(pagination_plan)
```

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,288 @@
...
+    parameter_name = get_parameter_name(argument_name)
+    return parameter_name
+
+
+def _get_filter_node_operation(filter_directive: DirectiveNode) -> str:
+    """Return the @filter's op_name as a string."""
+    return cast(StringValueNode, filter_directive.arguments[0].value).value
+
+
+def _is_new_filter_stronger(operation: str, new_filter_value: Any, old_filter_value: Any) -> bool:
```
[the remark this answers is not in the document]
bojanserafimov: That type will be inferred from the caller, and in practice will turn out to be `Any`. I don't think mypy is reliable for checks like these.

graphql_compiler/global_utils.py
bojanserafimov: With https://github.com/kensho-technologies/graphql-compiler/pull/738 it might be a good idea to add something like `QueryWithPaginationAnalysis` which contains the query string, query ast, the cost estimate, the pagination capacity map, the metadata table, etc. This way we:
- avoid parsing to AST and printing again back to string
- make plumbing easier, reducing the number of parameters in most functions in this module

graphql_compiler/query_pagination/query_parameterizer.py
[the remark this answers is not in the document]
obi1kenobi: I agree — but I also don't see a reason why not to make explicit the constraint that the docstring promises, even if in practice it ends up not being able to catch anything.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -1,61 +1,314 @@
...
-    raise NotImplementedError()
+    query_string = print_ast(query.query_ast)
+    query_root = get_only_query_definition(query.query_ast, GraphQLError)
+
+    # Create extended parameters that include the pagination parameter value
+    param_name = _generate_new_name("__paged_param", set(query.parameters.keys()))
+    extended_parameters = dict(query.parameters)
+    extended_parameters[param_name] = parameter_value
+
+    next_page_root, next_page_parameters = _add_pagination_filter_recursively(
```
obi1kenobi: Any reason not to push the `ASTWithParameters` creation into this function? It seems like we get a raw tuple out and then immediately turn it into a `ASTWithParameters` at all call sites, as far as I can tell.

bojanserafimov: It's a recursive function. If you make it create a `DocumentNode` on output, you'll get the wrong result. I can wrap it in a helper function if that's what you're suggesting

graphql_compiler/query_pagination/query_parameterizer.py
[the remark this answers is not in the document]
bojanserafimov: Because It's not worth the effort. If I add a `TypeVar` mypy will complain that it doesn't have a `<=` operation. Then I'll have to specify `FieldValueType = TypeVar("FieldValueType", int, str, datetime.date, datetime.datetime)`, but then how do I know it's only one of these types? Because we only do range reasoning on them. Then I'll have to put that in the spec here, and in all the intermediate fnctions. It's just not worth the review cycles. It's self-explanatory that a field we are using the `<=` operator on has `<=` functionality in python.

PR #740, opened 2020-01-29
Change pagination api

bojanserafimov:
Making 3 changes to the public api:
1. Return the list of advisories
2. Return remainder as list of queries instead of as optionally one query. Currently the list has either 0 or 1 elements, but that will change soon. I'd rather not have to change the API once we document it in the README.
3. Use the `QueryStringWithParameters` and `ASTWithParameters` dataclasses
4. Add mypy type hints

### DIFF
--- a/graphql_compiler/query_pagination/__init__.py
+++ b/graphql_compiler/query_pagination/__init__.py
@@ -81,7 +81,7 @@ def paginate_query(
         page_size: int, describes the desired number of result rows per page.
 
     Returns:
-        tuple containing queries for going over the original query in a paginated fashion:
+        tuple containing two elements:
             - page_and_remainder such that:
               - page_and_remainder.whole_query == query
               - page_and_remainder.page_size == page_size
--- a/graphql_compiler/query_pagination/__init__.py
+++ b/graphql_compiler/query_pagination/__init__.py
@@ -1,14 +1,10 @@
 # Copyright 2019-present Kensho Technologies, LLC.
-from collections import namedtuple
-
 from graphql.language.printer import print_ast
 
-from ..global_utils import QueryStringWithParameters, ASTWithParameters
 from ..ast_manipulation import safe_parse_graphql
 from ..cost_estimation.cardinality_estimator import estimate_number_of_pages
-from ..query_pagination.query_splitter import (
-    split_into_page_query_and_remainder_query,
-)
+from ..global_utils import ASTWithParameters, QueryStringWithParameters
+from ..query_pagination.query_splitter import split_into_page_query_and_remainder_query
 
 
 def paginate_query_ast(schema_info, query_ast, parameters, page_size):
--- a/graphql_compiler/query_pagination/__init__.py
+++ b/graphql_compiler/query_pagination/__init__.py
@@ -27,7 +27,7 @@ def paginate_query_ast(
 
     Returns:
         tuple containing two elements:
-            - ASTWithParameters namedtuple, describing a query expected to return roughly a page
+            - ASTWithParameters, describing a query expected to return roughly a page
               of result data of the original query.
             - Tuple of ASTWithParameters, describing queries that return the rest of the result
               data of the original query. If the original query is expected to return only a page or
@@ -81,7 +81,7 @@ def paginate_query(
 
     Returns:
         tuple containing queries for going over the original query in a paginated fashion:
-            - QueryStringWithParameters namedtuple, query expected to return roughly a page of
+            - QueryStringWithParameters, query expected to return roughly a page of
               result data of the original query.
             - Tuple of QueryStringWithParameters, describing queries that return the rest of the
               result data of the original query. If the original query is expected to return only
--- a/graphql_compiler/query_pagination/__init__.py
+++ b/graphql_compiler/query_pagination/__init__.py
@@ -1,13 +1,19 @@
 # Copyright 2019-present Kensho Technologies, LLC.
+from typing import Tuple
+
 from graphql.language.printer import print_ast
 
 from ..ast_manipulation import safe_parse_graphql
 from ..cost_estimation.cardinality_estimator import estimate_number_of_pages
 from ..global_utils import ASTWithParameters, QueryStringWithParameters
-from ..query_pagination.query_splitter import split_into_page_query_and_remainder_query
+from ..schema.schema_info import QueryPlanningSchemaInfo
+from .pagination_planning import PaginationAdvisory
+from .query_splitter import split_into_page_query_and_remainder_query
 
 
-def paginate_query_ast(schema_info, query_ast, parameters, page_size):
+def paginate_query_ast(
+    schema_info: QueryPlanningSchemaInfo, query: ASTWithParameters, page_size: int
+) -> Tuple[ASTWithParameters, Tuple[ASTWithParameters, ...], Tuple[PaginationAdvisory, ...]]:
     """Generate a query fetching a page of results and the remainder query for a query AST.
 
     Since the cost estimator may underestimate or overestimate the actual number of pages, you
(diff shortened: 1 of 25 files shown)


bojanserafimov:
If the api looks good I'll start working on end-to-end tests

bojanserafimov:
Tests in separate PR https://github.com/kensho-technologies/graphql-compiler/pull/741/files

graphql_compiler/query_pagination/__init__.py
```diff
@@ -63,36 +75,32 @@ def paginate_query(schema_info, query_string, parameters, page_size):
...
-        schema_graph: SchemaGraph instance.
-        statistics: Statistics object.
-        query_string: str, valid GraphQL query to be paginated.
+        query: QueryStringWithParameters
         parameters: dict, parameters with which query will be estimated.
         page_size: int, describes the desired number of result rows per page.
 
     Returns:
         tuple containing queries for going over the original query in a paginated fashion:
             - QueryStringWithParameters namedtuple, query expected to return roughly a page of
```
obi1kenobi: `QueryStringWithParameters namedtuple` is it still a namedtuple?

graphql_compiler/query_pagination/__init__.py
```diff
@@ -63,36 +75,32 @@ def paginate_query(schema_info, query_string, parameters, page_size):
...
-            print_ast(remainder_ast_with_parameters.query_ast),
-            remainder_ast_with_parameters.parameters,
-        )
+    remainder_queries_with_parameters = tuple(
+        QueryStringWithParameters(print_ast(query.query_ast), query.parameters)
+        for query in remainder_ast_with_parameters
+    )
 
-    return page_query_with_parameters, remainder_query_with_parameters
+    return page_query_with_parameters, remainder_queries_with_parameters, advisories
```
obi1kenobi: Thoughts on making some pleasant dataclass for these?

graphql_compiler/query_pagination/__init__.py
```diff
@@ -1,13 +1,19 @@
...
+from ..schema.schema_info import QueryPlanningSchemaInfo
+from .pagination_planning import PaginationAdvisory
+from .query_splitter import split_into_page_query_and_remainder_query
 
 
-def paginate_query_ast(schema_info, query_ast, parameters, page_size):
+def paginate_query_ast(
+    schema_info: QueryPlanningSchemaInfo, query: ASTWithParameters, page_size: int
+) -> Tuple[ASTWithParameters, Tuple[ASTWithParameters, ...], Tuple[PaginationAdvisory, ...]]:
     """Generate a query fetching a page of results and the remainder query for a query AST.
```
pmantica1: nit: remainder queries.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -16,45 +22,51 @@ def paginate_query_ast(schema_info, query_ast, parameters, page_size):
...
         tuple containing two elements:
             - ASTWithParameters namedtuple, describing a query expected to return roughly a page
               of result data of the original query.
-            - ASTWithParameters namedtuple or None, describing a query that returns the rest of the
-              result data of the original query. If the original query is expected to return only a
-              page or less of results, then this element will have value None.
+            - Tuple of ASTWithParameters, describing queries that return the rest of the result
+              data of the original query. If the original query is expected to return only a page or
+              less of results, then this element will be an empty tuple.
+            - Tuple of PaginationAdvisories that communicate what can be done to improve pagination
```
pmantica1: nit: Tuple of PaginationAdvisory objects

graphql_compiler/query_pagination/__init__.py
```diff
@@ -16,45 +22,51 @@ def paginate_query_ast(schema_info, query_ast, parameters, page_size):
...
+    return page_query, remainder_queries, advisories
 
 
-def paginate_query(schema_info, query_string, parameters, page_size):
+def paginate_query(
+    schema_info: QueryPlanningSchemaInfo, query: QueryStringWithParameters, page_size: int
+) -> Tuple[
+    QueryStringWithParameters, Tuple[QueryStringWithParameters, ...], Tuple[PaginationAdvisory, ...]
+]:
     """Generate a query fetching a page of results and the remainder query for a query string.
```
pmantica1: nit: remainder queries.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -63,36 +75,32 @@ def paginate_query(schema_info, query_string, parameters, page_size):
...
             - QueryStringWithParameters namedtuple, query expected to return roughly a page of
               result data of the original query.
-            - QueryStringWithParameters namedtuple or None. If the given query was estimated to
-              return more than a page of results, this element is a QueryStringWithParameters
-              namedtuple describing a query for the rest of the result data. Otherwise, this element
-              is None.
+            - Tuple of QueryStringWithParameters, describing queries that return the rest of the
+              result data of the original query. If the original query is expected to return only
+              a page or less of results, then this element will be an empty tuple.
+            - Tuple of PaginationAdvisories that communicate what can be done to improve pagination
```
pmantica1: nit: PaginationAdvisory objects.

graphql_compiler/query_pagination/__init__.py
bojanserafimov: What's wrong with `Tuple`? I don't see a need for an extra layer of nesting here.

graphql_compiler/query_pagination/__init__.py
pmantica1: nit: QueryStringWithParameters is not a namedtuple.

graphql_compiler/query_pagination/query_splitter.py
```diff
@@ -23,11 +25,13 @@ def split_into_page_query_and_remainder_query(
         num_pages: int, number of pages to split the query into.
 
     Returns:
-        tuple of (ASTWithParameters namedtuple, ASTWithParameters namedtuple), describing two
-        queries: the first query when executed will return roughly a page of results of the original
-        query; the second query will return the rest of the results of the original query. The union
-        of the two queries' result data is equivalent to the given query and parameter's result
-        data. There are no guarantees on the order of the result rows for the two generated queries.
+        tuple containing three elements:
+            - ASTWithParameters namedtuple, describing a query expected to return roughly a page
```
pmantica1: nit: ASTWithParameters is not a namedtuple. Could you please fix this everywhere else?

graphql_compiler/query_pagination/__init__.py
bojanserafimov: I guess grouping the first page with the remainder queries into a dataclass will create a nice space to document their relationship once, and cut on documentation needs throughout the codebase each time we use them. But each time we invoke this concept, it's a bit different (sometimes it's query strings, sometime ASTs, etc), that I think we won't get much from it. Not sure if that's what you're looking for.

bojanserafimov:
@pmantica1 We return multiple remainder queries when the remainder can't be described with just one query. Consider for instance a pagination plan with multiple vertex partitions. To describe the remainder in one page you need to `OR` two filters. I explained this in the docstring now.

graphql_compiler/query_pagination/__init__.py
obi1kenobi: You might be able to make it a generic container that allows it to work the same for both ASTs and for query strings by taking in the type it should use to represent `(query, args)` combinations. And yes, the motivation would be to have a single place that is well-documented and intuitive, rather than a bunch of non-named destructuring at each call site + verbose docstrings everywhere.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -1,98 +1,114 @@
...
-            - QueryStringWithParameters namedtuple, query expected to return roughly a page of
+            - QueryStringWithParameters, query expected to return roughly a page of
               result data of the original query.
-            - QueryStringWithParameters namedtuple or None. If the given query was estimated to
-              return more than a page of results, this element is a QueryStringWithParameters
-              namedtuple describing a query for the rest of the result data. Otherwise, this element
-              is None.
+            - Tuple of QueryStringWithParameters, describing queries that return the rest of the
+              result data of the original query. If the original query is expected to return only
+              a page or less of results, then this element will be an empty tuple. If the pagination
```
pmantica1: Bojan thanks for the explanation. I now understand the purpose of this diff.

I would remove the reference to the internal pagination internals: `VertexPartitionPlans`. If you think that it is too hard to explain the reasoning without this reference, then feel free to delete this explanation.

graphql_compiler/query_pagination/__init__.py
bojanserafimov: Added `PageAndRemainder` dataclass. Not fully propagated to the parameterizer. That doesn't change the API, and I'd rather do it in parallel with https://github.com/kensho-technologies/graphql-compiler/pull/741

PR #741, opened 2020-01-29
Add end to end pagination tests, fix timezone issues

bojanserafimov:
Wrote some tests, fixed some bugs. Notable bug fixes:
1. Timezone naive and aware datetimes are not comparable. There's 2 sources of input datetimes: input query args and quantile statistics. The easiest way to make them agree on a format is coerce input datetimes into the correct format.
2. Fixed assertion error message
3. Fix critical typo in parameter generator

### DIFF
--- a/graphql_compiler/cost_estimation/int_value_conversion.py
+++ b/graphql_compiler/cost_estimation/int_value_conversion.py
@@ -102,6 +102,7 @@ def convert_field_value_to_int(
     if is_int_field_type(schema_info, vertex_class, property_field):
         return value
     elif is_datetime_field_type(schema_info, vertex_class, property_field):
+        # TODO is astimezone the right thing to use? Maybe value.replace(tzinfo=pytz.utc)?
         return (value.astimezone(pytz.utc) - DATETIME_EPOCH_UTC) // datetime.timedelta(
             microseconds=1
         )
--- a/graphql_compiler/cost_estimation/int_value_conversion.py
+++ b/graphql_compiler/cost_estimation/int_value_conversion.py
@@ -102,8 +102,7 @@ def convert_field_value_to_int(
     if is_int_field_type(schema_info, vertex_class, property_field):
         return value
     elif is_datetime_field_type(schema_info, vertex_class, property_field):
-        # TODO is astimezone the right thing to use? Maybe value.replace(tzinfo=pytz.utc)?
-        return (value.astimezone(pytz.utc) - DATETIME_EPOCH_UTC) // datetime.timedelta(
+        return (value.replace(tzinfo=pytz.utc) - DATETIME_EPOCH_UTC) // datetime.timedelta(
             microseconds=1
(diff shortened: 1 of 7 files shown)


graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -102,7 +102,7 @@ def convert_field_value_to_int(
     if is_int_field_type(schema_info, vertex_class, property_field):
         return value
     elif is_datetime_field_type(schema_info, vertex_class, property_field):
-        return (value.astimezone(pytz.utc) - DATETIME_EPOCH_UTC) // datetime.timedelta(
+        return (value.replace(tzinfo=pytz.utc) - DATETIME_EPOCH_UTC) // datetime.timedelta(
```
bojanserafimov: We used to treat naive times as local times. However, local times lose their meaning over RPC, so it's simpler to treat them as utc times.

graphql_compiler/query_pagination/parameter_generator.py
```diff
@@ -169,7 +169,7 @@ def _compute_parameters_for_non_uuid_field(
     #                       that are known to be empty. This can cause the number of generated
     #                       pages to be less than the desired number of pages.
     return _deduplicate_sorted_generator(
-        proper_quantiles[index]
+        relevant_quantiles[index]
```
bojanserafimov: Bug wasn't caught before since we didn't have a test paginating recursively on a non-uuid field

graphql_compiler/cost_estimation/int_value_conversion.py
pmantica1: It seems that we are currently losing timezone information when converting datetime to ints. Can we not store the timezone information and use to it decode the value of datetime fields? Filtering with a timezone aware datetime on a timezone naive datetime field will likely lead to an error.

If you agree this is an issue feel free to address it in a separate PR or maybe just open an issue. I am also willing to help with it.

bojanserafimov: That's not the issue. The stats are either in tz-aware or tz-naive format. Consider a query where the user gives argument in the opposite format (tz aware/naive). How do you compare these two datetimes?

If everything is happening on the same server (there's no RPC involved), then a tz-naive datetime means a datetime in local time. That can be converted to the tz of the tz-aware datetime, and they can be compared.

However, if you receive a request over RPC with tz-naive datetimes, you don't know the local timezone of the sender, and all you can do is assume it's utc.

bojanserafimov: So yes, we are losing timezone information by assuming tz-naive datetimes are utc. However, it is an assumption we have to make, unless you have any suggestion.

bojanserafimov:
Replaced with https://github.com/kensho-technologies/graphql-compiler/pull/744

PR #744, opened 2020-01-30
Fix pagination timezone bug

bojanserafimov:
There's two situations in pagination where we deal with datetimes of different tz-awareness:
1. The quantile statistics are tz-aware, but an existing filter is tz-naive
2. The quantile statistics are tz-naive, but an existing filter is tz-aware

To simplify the problem, I disregard the first case now, since tz-aware database fields are not common. I disallow quantiles to be tz-aware.

For the second problem, no good solution exists. If a value is tz-naive, the user should not be able to use tz-aware filter, and the type system should communicate this. Since we don't do that yet, we have two options on what to do in this case:
1. Error on pagination
2. Somehow coerce the value into a tz-naive datetime on pagination

The first option is bad because we will have a query that compiles successfully but fails to paginate.

The second option is fine because the worst that can happen is we generate uneven page sizes.

In this PR i implement the second option:
- I assert that quantiles are tz-naive
- I coerce tz-naive to tz-aware datetimes by ignoring the tzinfo
- I add tests

### DIFF
--- a/graphql_compiler/cost_estimation/int_value_conversion.py
+++ b/graphql_compiler/cost_estimation/int_value_conversion.py
@@ -17,9 +17,6 @@ import datetime
 from typing import Any
 from uuid import UUID
 
-import pytz
-
-from ..exceptions import GraphQLError
 from ..schema.schema_info import QueryPlanningSchemaInfo
 from .helpers import is_date_field_type, is_datetime_field_type, is_int_field_type, is_uuid4_type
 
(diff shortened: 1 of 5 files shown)


pmantica1:
I disagree with this approach. The datetime field in MS SQL is always timezone naive. datetimeoffset field is timezone aware. Because of the naming, I suspect that people use the datetime field most frequently. For people targetting an MS SQL backend it will still be highly counterintuitive that the time they inputted for the query is somehow modified.  

https://docs.microsoft.com/en-us/sql/t-sql/functions/date-and-time-data-types-and-functions-transact-sql?view=sql-server-ver15

PR #746, opened 2020-01-30
Fix bug in parameter generator

bojanserafimov:
There was an off-by-one error in a poorly documented and tested helper function.

I isolated the quantile choosing logic into a separate function, tested that, and fixed the off-by-one errors in the expected results of other tests.

### DIFF
--- a/graphql_compiler/tests/snapshot_tests/test_query_pagination.py
+++ b/graphql_compiler/tests/snapshot_tests/test_query_pagination.py
@@ -514,6 +514,43 @@ class QueryPaginationTests(unittest.TestCase):
         self.assertEqual(expected_parameters, list(generated_parameters))
 
     @pytest.mark.usefixtures("snapshot_orientdb_client")
+    def test_parameter_value_generation_int_few_quantiles(self):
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -513,6 +513,43 @@ def test_parameter_value_generation_int(self):
...
+            }
+        }"""
+        args = {}
+        query_ast = safe_parse_graphql(query)
+        vertex_partition = VertexPartitionPlan(("Species",), "limbs", 3)
+        generated_parameters = generate_parameters_for_vertex_partition(
+            schema_info, query_ast, args, vertex_partition
+        )
+
+        expected_parameters = [26, 51, 76]
```
obi1kenobi: I don't understand this bit. Hopefully you can explain when you put the fix together.

bojanserafimov: It's incorrect, but the code has an `IndexError` anyway

PR #750, opened 2020-02-12
Simplify cardinality estimation

bojanserafimov:
1. Remove `graphql_to_ir` call from `estimate_query_result_cardinality`. It doesn't belong there. During pagination the query metadata table is already available, and it's a waste to regenerate it.
2. Move `_estimate_number_of_pages` from cost estimation to pagination module, and remove unnecessary parameters from it.

### DIFF
--- a/graphql_compiler/query_pagination/__init__.py
+++ b/graphql_compiler/query_pagination/__init__.py
@@ -13,14 +13,7 @@ from .typedefs import PageAndRemainder
 
 
 def _estimate_number_of_pages(query: ASTWithParameters, result_size: float, page_size: int) -> int:
-    """Estimate how many pages of results will be generated for a given query.
-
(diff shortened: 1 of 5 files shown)


graphql_compiler/query_pagination/__init__.py
```diff
@@ -4,14 +4,55 @@
...
+        query: ASTWithParameters
+        result_size: The estimated result size of a query
+        page_size: The desired page size of a query
+
+    Returns:
+        int, estimated number of pages if the query were executed.
+
+    Raises:
+        ValueError if page_size is below 1.
+    """
```
pmantica1: Please rewrite docstring to account for functionality change and remove unused `query` parameter

bojanserafimov: Will do. But `query` is used in error message

PR #753, opened 2020-02-13
Use pagination capacities

bojanserafimov:
1. Use pagination analysis in pagination planning instead of doing analysis on the spot.
2. The new method improves planning, so I add a test that verifies the new behavior

### DIFF
--- a/graphql_compiler/query_pagination/pagination_planning.py
+++ b/graphql_compiler/query_pagination/pagination_planning.py
@@ -1,15 +1,12 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from abc import ABC
 from dataclasses import dataclass, field
-from typing import Optional, Tuple
+from typing import Tuple
 
-from ..global_utils import PropertyPath
 from ..ast_manipulation import get_only_query_definition, get_only_selection_from_ast
 from ..cost_estimation.analysis import QueryPlanningAnalysis
-from ..cost_estimation.helpers import is_uuid4_type
-from ..cost_estimation.int_value_conversion import field_supports_range_reasoning
 from ..exceptions import GraphQLError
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -70,6 +70,45 @@ def test_pagination_planning_basic(self) -> None:
...
+        schema_info = QueryPlanningSchemaInfo(
+            schema=graphql_schema,
+            type_equivalence_hints=type_equivalence_hints,
+            schema_graph=schema_graph,
+            statistics=statistics,
+            pagination_keys=pagination_keys,
+            uuid4_fields=uuid4_fields,
+        )
+
+        # Check that the correct plan is generated when it's obvious (page the root)
```
pmantica1: Delete this comment.
```suggestion
```

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -185,21 +108,38 @@ def get_pagination_plan(
...
+    #                       pagination when uuid pagination is also available.
+    pagination_field = query_analysis.schema_info.pagination_keys.get(pagination_node)
+    if pagination_field is None:
+        return PaginationPlan(tuple()), (PaginationFieldNotSpecified(pagination_node),)
+
+    # Get the pagination capacity
+    vertex_path = (root_node,)
+    property_path = PropertyPath(vertex_path, pagination_field)
+    capacity = query_analysis.pagination_capacities.get(property_path)
+    if capacity is None:
```
pmantica1:
```suggestion
    # If the pagination capacity is None, then there must be no quantiles for this property.
    if capacity is None:
```

PR #754, opened 2020-02-14
Add edge constraints

bojanserafimov:
Needed for better pagination planning. It's not good to paginate on a vertex reachable by many-to-one edges from a uniquely filtered vertex. To prevent this we need many-to-one edge information

### DIFF
--- a/graphql_compiler/cost_estimation/helpers.py
+++ b/graphql_compiler/cost_estimation/helpers.py
@@ -1,5 +1,8 @@
 # Copyright 2019-present Kensho Technologies, LLC.
+from typing import cast
+
 from graphql import GraphQLInt
+from graphql.type.definition import GraphQLObjectType
 
 from ..global_utils import is_same_type
 from ..schema import GraphQLDate, GraphQLDateTime
@@ -10,7 +13,9 @@ def is_datetime_field_type(
     schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
 ) -> bool:
     """Return whether the field is of type GraphQLDateTime."""
-    field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
+    field_type = (
+        cast(GraphQLObjectType, schema_info.schema.get_type(vertex_name)).fields[field_name].type
+    )
     return is_same_type(GraphQLDateTime, field_type)
 
 
@@ -18,7 +23,9 @@ def is_date_field_type(
     schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
 ) -> bool:
     """Return whether the field is of type GraphQLDate."""
-    field_type = schema_info.schema.get_type(vertex_name).fields[field_name].type
+    field_type = (
+        cast(GraphQLObjectType, schema_info.schema.get_type(vertex_name)).fields[field_name].type
+    )
     return is_same_type(GraphQLDate, field_type)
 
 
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema/schema_info.py
```diff
@@ -324,46 +327,62 @@ def make_sqlalchemy_schema_info(
...
+    #                       during schema generation.
+    pagination_keys: Dict[str, str]
+
+    # Dict mapping vertex names in the graphql schema to a set of property names that
+    # are known to contain uniformly distributed uppercase uuid values. The types of those
+    # fields are expected to be ID or String.
+    uuid4_fields: Dict[str, Set[str]]
+
+    # Map edge names to constraints inferred for them.
+    edge_constraints: Dict[str, Set[EdgeConstraint]] = field(default_factory=dict)
```
bojanserafimov: The default value here is to please tests. Once the design is accepted, I'll remove the default and do the necessary plumbing.

pmantica1: We discussed this offline already, but, just to make this discussion public, I wanted to mention that one issue here is that the `SchemaGraph` reflects similar information already through its `IndexDefinition`. However, I think that in the long run we want to move away from having the `SchemaGraph` be part of the query planning schema info. It is a very heavyweight object and we use only a very small part of it for query planning.

graphql_compiler/schema/schema_info.py
```diff
@@ -324,46 +327,62 @@ def make_sqlalchemy_schema_info(
...
-    ),
-)
+@unique
+class EdgeConstraint(Enum):
+    AtLeastOneSource = auto()
+    AtMostOneSource = auto()
+    AtLeastOneDestination = auto()
+    AtMostOneDestination = auto()
+    NoParallelEdges = auto()
+    NoLoops = auto()
```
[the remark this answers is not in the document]
bojanserafimov: Parallel: two edges with matching source and destination
Loop: edge from x to x

graphql_compiler/schema/schema_info.py
bojanserafimov: Yes. Some databases enforce edge constraints with indexes (OrientDB), and some databases impose constraints by construction (SQL foreign keys are many-to-one). Therefore these edge constraints are on a higher level than the SchemaGraph.

graphql_compiler/cost_estimation/helpers.py
```diff
@@ -1,33 +1,51 @@
...
 
 
+def _get_field_type(
+    schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
+) -> Union[GraphQLList, GraphQLScalarType]:
+    """Get the GraphQL type of the field on the specified vertex."""
+    return (
+        cast(
+            Union[GraphQLObjectType, GraphQLInterfaceType], schema_info.schema.get_type(vertex_name)
+        )
```
obi1kenobi: This shouldn't be a cast, instead it should be an `isinstance` + assertion. Technically, `get_type()` can return union types, enums, etc. and while `vertex_name` _should_ only refer to interfaces and objects, it might not.

graphql_compiler/cost_estimation/helpers.py
```diff
@@ -1,33 +1,51 @@
...
+)
 
 from ..global_utils import is_same_type
 from ..schema import GraphQLDate, GraphQLDateTime
 from ..schema.schema_info import QueryPlanningSchemaInfo
 
 
+def _get_field_type(
+    schema_info: QueryPlanningSchemaInfo, vertex_name: str, field_name: str
+) -> Union[GraphQLList, GraphQLScalarType]:
```
obi1kenobi: The return type here is (long-term) problematic as well. In principle, the type of a field can also be an interface, object, or union. In fact, some of the edge constraints are expressible in GraphQL SDL in ways that can cause this — for example, many-to-one edges are represented by `field_name: NeighborType` instead of `field_name: [NeighborType]`.

This hack is fine for now, but in the not-too-distant future, we should reconsider how we actually represent these ideas and make our approach as close as possible to vanilla GraphQL because that will improve our compatibility with the rest of the ecosystem.

PR #757, opened 2020-02-19
Use many-to-one edges in pagination planning

bojanserafimov:
Make sure there is no unique filter on query nodes reachable from the pagination node by many-to-one edges.

### DIFF
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -8,7 +8,7 @@ from graphql.language.printer import print_ast
 
 from ..ast_manipulation import safe_parse_graphql
 from ..compiler.compiler_frontend import graphql_to_ir
-from ..compiler.helpers import get_edge_direction_and_name, Location
+from ..compiler.helpers import Location, get_edge_direction_and_name
 from ..compiler.metadata import FilterInfo, QueryMetadataTable
 from ..cost_estimation.cardinality_estimator import estimate_query_result_cardinality
 from ..cost_estimation.int_value_conversion import (
@@ -19,7 +19,7 @@ from ..cost_estimation.int_value_conversion import (
 from ..cost_estimation.interval import Interval
 from ..global_utils import ASTWithParameters, PropertyPath, QueryStringWithParameters, VertexPath
 from ..schema import is_meta_field
-from ..schema.schema_info import QueryPlanningSchemaInfo, EdgeConstraint
+from ..schema.schema_info import EdgeConstraint, QueryPlanningSchemaInfo
 from .filter_selectivity_utils import (
     adjust_counts_for_filters,
     get_integer_interval_for_filters_on_field,
@@ -142,8 +142,8 @@ def get_distinct_result_set_estimates(
             from_path = location.query_path[:-1]
             to_path = location.query_path
             edge_direction, edge_name = get_edge_direction_and_name(location.query_path[-1])
-            edge_constraints = schema_info.edge_constraints.get(edge_name, set())
-            if edge_direction == 'in':
+            edge_constraints = schema_info.edge_constraints.get(edge_name, EdgeConstraint(0))
+            if edge_direction == "in":
                 from_path, to_path = to_path, from_path
 
             if EdgeConstraint.AtMostOneDestination in edge_constraints:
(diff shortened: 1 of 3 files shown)


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -125,10 +126,28 @@ def get_distinct_result_set_estimates(
...
 
-    # TODO(bojanserafimov): If there's a many-to-one edge from A to B in the query, the
-    #                       distinct result set estimate for B cannot be greater than
-    #                       the one for A. Taking this into account would make the results
-    #                       more accurate.
+    single_destination_traversals = set()
+    for location, _ in query_metadata.registered_locations:
+        if len(location.query_path) > 1:
+            from_path = location.query_path[:-1]
+            to_path = location.query_path
```
obi1kenobi: These are not the same type -- one is the entire query path, and one is just the last element in it. That doesn't seem intended..?

graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -108,6 +108,53 @@ def test_pagination_planning_unique_filter(self) -> None:
...
+        number_of_pages = 10
+        analysis = analyze_query_string(schema_info, query)
+        pagination_plan, warnings = get_pagination_plan(analysis, number_of_pages)
+
+        # This is a white box test. We check that we don't paginate on the root when it has a
+        # unique filter on its many-to-one neighbor. A better plan is to paginate on a
+        # different vertex, but that is not implemented.
+        expected_plan = PaginationPlan(tuple())
+        expected_warnings: Tuple[PaginationAdvisory, ...] = tuple()
+        self.assertEqual([w.message for w in expected_warnings], [w.message for w in warnings])
```
obi1kenobi: Warnings are something else in Python, let's stick with the "advisory" terminology: https://docs.python.org/3/library/warnings.html

graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -108,6 +108,53 @@ def test_pagination_planning_unique_filter(self) -> None:
...
+            type_equivalence_hints=type_equivalence_hints,
+            schema_graph=schema_graph,
+            statistics=statistics,
+            pagination_keys=pagination_keys,
+            uuid4_fields=uuid4_fields,
+            edge_constraints=edge_constraints,
+        )
+
+        query = QueryStringWithParameters(
+            """{
```
obi1kenobi: The formatting here is a bit weird.

graphql_compiler/cost_estimation/analysis.py
bojanserafimov: It's all but the last element

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -130,10 +130,32 @@ def get_distinct_result_set_estimates(
...
+            # TODO(bojanserafimov): We currently ignore FoldScopeLocations. However, a unique
+            #                       filter on a FoldScopeLocation also uniquely filters the
+            #                       enclosing scope.
+            continue
+
+        if len(location.query_path) > 1:
+            from_path = location.query_path[:-1]
+            to_path = location.query_path
+            edge_direction, edge_name = get_edge_direction_and_name(location.query_path[-1])
+            edge_constraints = schema_info.edge_constraints.get(edge_name, EdgeConstraint(0))
```
pmantica1: I am not sure what this syntax `EdgeConstraint(0)` does, but this seems sketchy. Do you mind explaining or fixing?

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -130,10 +130,32 @@ def get_distinct_result_set_estimates(
...
+            if edge_direction == "in":
+                from_path, to_path = to_path, from_path
+
+            if EdgeConstraint.AtMostOneDestination in edge_constraints:
+                single_destination_traversals.add((from_path, to_path))
+            if EdgeConstraint.AtMostOneSource in edge_constraints:
+                single_destination_traversals.add((to_path, from_path))
+
+    for _ in single_destination_traversals:  # This is on purpose (TODO explain why)
```
pmantica1: Can you address this TODO? I understand what you are doing here so don't feel the need to go into a lot of detail. Maybe a link to a similar algorithm would be sufficient.

graphql_compiler/cost_estimation/analysis.py
bojanserafimov: Flags are represented with bits. 0 means all bits unset. Can't find another way to get this value.

PR #759, opened 2020-02-19
Fix query analysis with FoldScopeLocations

bojanserafimov:
I added a test with a query that has a `@fold` and it raised an error. Then I fixed it.

### DIFF
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -8,7 +8,7 @@ from graphql.language.printer import print_ast
 
 from ..ast_manipulation import safe_parse_graphql
 from ..compiler.compiler_frontend import graphql_to_ir
-from ..compiler.helpers import FoldScopeLocation
+from ..compiler.helpers import Location
 from ..compiler.metadata import FilterInfo, QueryMetadataTable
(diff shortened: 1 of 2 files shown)


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -118,6 +121,8 @@ def get_distinct_result_set_estimates(
     """
     distinct_result_set_estimates = {}
     for location, location_info in query_metadata.registered_locations:
+        if not isinstance(location, Location):
+            continue  # Might be a FoldScopeLocation
```
pmantica1:
```suggestion
            continue  # We don't paginate inside folds.
```

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -161,6 +166,8 @@ def get_pagination_capacities(
     pagination_capacities = {}
     for location, location_info in query_metadata.registered_locations:
         vertex_type_name = location_info.type.name
+        if not isinstance(location, Location):
+            continue  # We don't consider FoldScopeLocations for pagination for simplicity
```
pmantica1:
```suggestion
            continue  # We don't paginate inside folds.
```

PR #761, opened 2020-02-24
Sync setup.py with Pipfile

bojanserafimov:


setup.py
```diff
@@ -57,10 +57,12 @@ def find_long_description():
     install_requires=[  # Make sure to keep in sync with Pipfile requirements.
         "arrow>=0.15.0,<1",
         "funcy>=1.7.3,<2",
+        "dataclasses>=0.6",
```
obi1kenobi: This is unnecessary, see `extras_require` below.

PR #762, opened 2020-02-25
Fix bug with impossible pagination

bojanserafimov:
When pagination is needed but not possible we get an `AssertionError`. Instead we should return a non-paginated query.

In this diff I fix that bug, and resolve an unrelated TODO (advisory plumbing) on the way.


graphql_compiler/query_pagination/query_splitter.py
```diff
@@ -61,4 +51,4 @@ def split_into_page_query_and_remainder_query(
         pagination_plan.vertex_partitions[0],
         first_param,
     )
-    return page_query, remainder_query, advisories
+    return page_query, remainder_query
```
obi1kenobi: Update the docstring for this function

PR #763, opened 2020-02-26
Test pagination with compiler tests

bojanserafimov:
Checking that pagination does not raise errors on any of the 140 compiler input data tests. The implementation of this with `dir` and `getattr` is not great, but I'll address that after I address the bugs.

Bugs discovered and addressed:
1. Attempting to do range reasoning on fields that don't support it
2. Attempting to reason about the range of tagged parameters as if they were runtime parameters
3. FoldScopeLocation not expected
4. Attempting range reasoning on meta field

Remaining limitations:
1. Bad things will happen if we paginate on a field that has an inequality filter with a tagged parameter. Not sure what yet, will address in separate PR.

### DIFF
--- a/graphql_compiler/tests/snapshot_tests/test_query_pagination.py
+++ b/graphql_compiler/tests/snapshot_tests/test_query_pagination.py
@@ -1205,12 +1205,6 @@ class QueryPaginationTests(unittest.TestCase):
         from ..test_input_data import CommonTestData
 
         failing_tests = (
-            "complex_optional_variables",
-            "filter_count_with_tagged_optional_parameter_in_fold_scope",
-            "filter_count_with_tagged_parameter_in_fold_scope",
-            "filter_field_with_tagged_optional_parameter_in_fold_scope",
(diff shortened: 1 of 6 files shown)


graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -1182,3 +1184,45 @@ def test_impossible_pagination(self):
...
+            "Int": 5,
+            "Date": datetime.date(2000, 1, 1),
+            "DateTime": datetime.datetime(2000, 1, 1),
+            "Decimal": 5.3,
+            "[String]": ["lol", "lel"],
+        }
+
+        for test_name in dir(test_input_data):
+            method = getattr(test_input_data, test_name)
+            if hasattr(method, "__annotations__"):
```
obi1kenobi: Use `get_test_function_names_from_class` from `test_testing_invariants.py` for this — let's have one way to do this. Maybe also move it to the `helpers` file as well.

PR #764, opened 2020-03-03
Disable pagination with tags

bojanserafimov:
Currently we might paginate on a field that has an inequality filter using a tagged parameter. Our code is not smart enough to deal with that, so whatever output we produce is bound to be wrong in many ways. In this PR I disable pagination on such fields.

Breakdown of changes:
1. Add a `get_fields_eligible_for_pagination` analysis pass. The pass makes sure to exclude meta fields, fields inside `@fold`, and fields with tagged parameters in inequality filters.
2. Add a `get_single_field_filters` analysis. This is a refactor that materializes analysis that is currently used by cost estimation, pagination, and the new pagination field eligibility test.
3. Add a test for  `get_single_field_filters`
4. Use the new analysis passes in computation of pagination capacities.

There is no end-to-end test for this right now, since we only paginate on the root, and tagged parameters cannot appear on filters in the root since they have to be declared before they are used, and property fields come before vertex fields.

### DIFF
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -87,7 +87,7 @@ def get_single_field_filters(
                     filter_info
                 )
             else:
-                pass  # We don't do anything for multi-field filters yet
+                pass
 
         for field_name, filters in single_field_filters_for_vertex.items():
             property_path = PropertyPath(location.query_path, field_name)
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -57,7 +57,18 @@ def _convert_int_interval_to_field_value_interval(
 def get_single_field_filters(
     schema_info: QueryPlanningSchemaInfo, query_metadata: QueryMetadataTable,
 ) -> Dict[PropertyPath, Set[FilterInfo]]:
-    """Find the single field filters for each field. Filters like name_or_alias are excluded."""
+    """Find the single field filters for each field.
+
+    Filters that apply to multiple fields, like name_or_alias, are ignored.
+    Filters inside fold scopes are not considered.
+
+    Args:
+        schema_info: QueryPlanningSchemaInfo
+        query_metadata: info on locations, inputs, outputs, and tags in the query
+
+    Returns:
+        A set of filters for each field.
+    """
     single_field_filters = {}
     for location, location_info in query_metadata.registered_locations:
         if not isinstance(location, Location):
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -1,7 +1,7 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 import bisect
 from dataclasses import dataclass
-from typing import Any, Dict, List, Set
+from typing import Any, Dict, Set
 
 from cached_property import cached_property
 from graphql.language.printer import print_ast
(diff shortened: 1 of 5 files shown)


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -204,46 +258,48 @@ def get_pagination_capacities(
...
-                        # TODO(bojanserafimov): If the relevant quantiles contain duplicates, the
-                        #                       pagination capacity would be lower.
-
-                        pagination_capacities[property_path] = min(
-                            len(relevant_quantiles) + 1,
-                            int(distinct_result_set_estimates[location.query_path]),
+            if property_path not in fields_eligible_for_pagination:
+                continue
+
+            if is_uuid4_type(schema_info, vertex_type_name, field_name):
```
bojanserafimov: This 40-line change is just indentation change.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -53,9 +54,81 @@ def _convert_int_interval_to_field_value_interval(
...
+        schema_info: QueryPlanningSchemaInfo
+        query_metadata: info on locations, inputs, outputs, and tags in the query
+
+    Returns:
+        A set of filters for each field.
+    """
+    single_field_filters = {}
+    for location, location_info in query_metadata.registered_locations:
+        if not isinstance(location, Location):
+            continue  # We don't paginate inside folds.
```
pmantica1: Can you remove this block and remove the "filters inside fold scopes are not considered" bit from the documentation? We already perform this check in `get_fields_eligible_for_pagination`. I don't see why we need it here and it makes the function docstring more complicated.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -53,9 +54,81 @@ def _convert_int_interval_to_field_value_interval(
...
+        single_field_filters_for_vertex: Dict[str, Set[FilterInfo]] = {}
+        for filter_info in filter_infos:
+            if len(filter_info.fields) == 0:
+                raise AssertionError(f"Got filter on 0 fields {filter_info} on {vertex_type_name}")
+            elif len(filter_info.fields) == 1:
+                single_field_filters_for_vertex.setdefault(filter_info.fields[0], set()).add(
+                    filter_info
+                )
+            else:
+                pass  # We don't do anything for multi-field filters yet
```
pmantica1: Remove this comment.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -53,9 +54,81 @@ def _convert_int_interval_to_field_value_interval(
...
+
+    Filters that apply to multiple fields, like name_or_alias, are ignored.
+    Filters inside fold scopes are not considered.
+
+    Args:
+        schema_info: QueryPlanningSchemaInfo
+        query_metadata: info on locations, inputs, outputs, and tags in the query
+
+    Returns:
+        A set of filters for each field.
```
pmantica1: nit: There won't be a key-value pair if the field doesn't have any filters.

```suggestion
        dict mapping fields to their set of filters.
```

graphql_compiler/cost_estimation/analysis.py
bojanserafimov: The function is not called `get_single_field_filters_for_pagination`, and in fact it could be used by cost estimation if it did look inside fold scopes.

bojanserafimov: I would extend it to include folded locations, but it's not as simple as removing this block.

PR #765, opened 2020-03-04
Use db-agnostic uuids for pagination parameter generation

bojanserafimov:
MSSQL considers the last 6 bytes of the uuid most significant, which is different from what most databases do and what we expect https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/comparing-guid-and-uniqueidentifier-values

This causes problems with paginating on uuid fields. To fix the problem, all uuid parameters we generate for pagination should have their first 6 bytes and last 6 bytes equal.

### DIFF
--- a/graphql_compiler/query_pagination/parameter_generator.py
+++ b/graphql_compiler/query_pagination/parameter_generator.py
@@ -98,7 +98,7 @@ def _convert_int_interval_to_field_value_interval(
     return Interval(lower_bound, upper_bound)
 
 
-def _make_uuid_mssql_compatible(uuid_value: str) -> str:
+def _get_nearest_db_agnostic_uuid(uuid_value: str) -> str:
     """Return the nearest db-agnostic uuid.
 
     Most databases compare uuids by considering the leftmost digist most significant.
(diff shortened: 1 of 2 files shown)


bojanserafimov:
> With this code will the union of the pagination filters will cover the entire range of UUID values? What if there is already an existing filter on the field?

Good catch. If there's an existing uuid inequality filter that is not db-agnostic bad things will happen. This is a more general problem with cost estimation. We have two choices:
1. Disable range reasoning with non-db-agnostic uuid inequality filters
2. Include uuid comparison method information in the `QueryPlanningSchemaInfo`

pmantica1:
> > With this code will the union of the pagination filters will cover the entire range of UUID values? What if there is already an existing filter on the field?
> 
> Good catch. If there's an existing uuid inequality filter that is not db-agnostic bad things will happen. This is a more general problem with cost estimation. We have two choices:
> 
> 1. Disable range reasoning with non-db-agnostic uuid inequality filters
> 2. Include uuid comparison method information in the `QueryPlanningSchemaInfo`

I'd say let's go with 2. It seems simple to implement. It's not great API-wise, but I think we can revert this decision easily in the future if we come up with something bettter since the API for `QueryPlanningSchemaInfo` is far from being stable.

obi1kenobi:
So should this PR be closed without merging then, since #767 got merged already?

PR #767, opened 2020-03-10
Consider uuid comparison method during query planning

bojanserafimov:
MSSQL considers the last 6 bytes of the uuid most significant, which is different from what most databases do and what we expect https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/comparing-guid-and-uniqueidentifier-values

This causes problems with cost estimation and pagination on uuid fields. To fix the problem, I make the `QueryPlanningSchemaInfo` aware of the uuid comparison method used by the database.

This approach is different and likely better than https://github.com/kensho-technologies/graphql-compiler/pull/765 (see Pedro's comment that points out some flaws).

### DIFF
--- a/graphql_compiler/tests/snapshot_tests/test_query_pagination.py
+++ b/graphql_compiler/tests/snapshot_tests/test_query_pagination.py
@@ -1013,6 +1013,48 @@ class QueryPaginationTests(unittest.TestCase):
         self.assertEqual(expected_parameters, list(generated_parameters))
 
     @pytest.mark.usefixtures("snapshot_orientdb_client")
+    def test_parameter_value_generation_mssql_uuid_with_existing_filter(self):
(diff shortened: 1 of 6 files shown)


graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -919,13 +973,55 @@ def test_parameter_value_generation_uuid(self):
         ]
         self.assertEqual(expected_parameters, list(generated_parameters))
 
+    @pytest.mark.usefixtures("snapshot_orientdb_client")
+    def test_parameter_value_generation_mssql_uuid(self):
```
pmantica1: In the spirit of regression testing, can you include at least one test where you test that pagination works correctly if there is an existing filter on an uuid field?

PR #768, opened 2020-03-10
Write UUID flip test

bojanserafimov:
Give the function a more descriptive name and test it.

### DIFF
--- a/graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
+++ b/graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
@@ -20,6 +20,7 @@ from ...cost_estimation.filter_selectivity_utils import (
 from ...cost_estimation.int_value_conversion import (
     convert_field_value_to_int,
     convert_int_to_field_value,
+    swap_uuid_prefix_and_suffix
 )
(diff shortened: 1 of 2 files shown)


graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1602,3 +1603,14 @@ def test_int_value_conversion_date(self):
...
             )
             self.assertEqual(date_value, recovered_date)
+
+    def test_swap_uuid_prefix_and_suffix(self):
+        # Test with a uuid with lots of different characters
+        uuid_string = "01234567-89ab-cdef-0123-456789abcdef"
+        flipped_uuid = swap_uuid_prefix_and_suffix(uuid_string)
+        self.assertEqual("456789ab-cdef-cdef-0123-0123456789ab", flipped_uuid)
+
+        # Test with a different uuid in case repeating characters hid bugs
```
pmantica1: I don't understand what you mean here.

bojanserafimov: I was explaining why I chose that particular input, but on second thought I see that my way of thinking about it is more confusing than just not saying anything.

PR #770, opened 2020-03-12
Fix parameterization assertion

bojanserafimov:
When the parameterizer replaces one filter with another, it asserts the new filter is stronger, testing that the parameter generator generated the correct parameters. However, this implies:
1. The check needs to be as complicated as the parameter generator module is
2. The end-to-end tests need to be as thorough as the parameter generator specialized tests

Changes in this pr:
- Add analysis pass
- Refactor to give the parameter generator access to query analysis
- Fix false-positive trigger of the assertion when paginating on mssql uuid.
- Downgrade the assertion to a logging error to prevent false-positives from breaking everything
- Add regression test


graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -158,12 +162,13 @@ def _add_pagination_filter_at_node(
...
-                            f"Pagination filter {directive_to_add} on "
+                        logger.error(
+                            f"Pagination filter {print_ast(directive_to_add)} on "
                             f"{pagination_field} is not stronger than "
-                            f"an existing filter {directive}. This is "
+                            f"an existing filter {print_ast(directive)}. This is "
                             f"likely a bug in parameter generation. "
-                            f"Query string: {query_string}"
+                            f"Query string: {query_string}, parameters: "
+                            f"{extended_parameters}."
```
obi1kenobi: Can you include the old/new parameter values adjacent to the directives they correspond to?

.pylintrc
```diff
@@ -139,7 +139,6 @@ enable=abstract-class-instantiated,
     invalid-slots-object,
     invalid-star-assignment-target,
     invalid-unary-operand-type,
-    logging-format-interpolation,
```
obi1kenobi: remove this please

obi1kenobi:
I'll override the `codecov/patch` check because this is a line we hope is never hit, so testing it has a somewhat lower utility.

bojanserafimov:
codecov it's not blocking, if you accept yout can merge

bojanserafimov:
Working on this actively (wish I could go back to draft PR).

bojanserafimov:
Ready for review. See updated PR description.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -72,21 +84,18 @@ def _is_new_filter_stronger(operation: str, new_filter_value: Any, old_filter_va
     Returns:
         whether the old filter can be removed with no change in query meaning.
     """
-    if type(new_filter_value) != type(old_filter_value):
-        raise AssertionError(
-            f"Expected {new_filter_value} and {old_filter_value} "
-            f"to have the same type, but got {type(new_filter_value)} "
-            f"and {type(old_filter_value)}."
-        )
+    vertex_type = query_analysis.types[property_path.vertex_path]
```
obi1kenobi: Isn't this just `query_analysis.metadata_table.get_location_info(...).type.name` or something like that?

The new property and function to compute it don't seem particularly useful to me, I feel like I'm missing something.

bojanserafimov: The metadata table uses `BaseLocation` for the key. For this purpose it's fine to make up `visit_counter = 1`, but in general I don't want to deal with gremlin artifacts in this module. In fact, I suspect there's a bug somewhere because of the way we handle `visit_counter`, and I'm looking for it.

bojanserafimov: See https://github.com/kensho-technologies/graphql-compiler/pull/775

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -60,10 +64,18 @@ def _get_filter_node_operation(filter_directive: DirectiveNode) -> str:
     return cast(StringValueNode, filter_directive.arguments[0].value).value
 
 
-def _is_new_filter_stronger(operation: str, new_filter_value: Any, old_filter_value: Any) -> bool:
+def _is_new_filter_stronger(
+    query_analysis: QueryPlanningAnalysis,
+    property_path: PropertyPath,
+    operation: str,
+    new_filter_value: Any,
+    old_filter_value: Any,
```
obi1kenobi: nit: These are expected to be the same type, right? That's what I gather from the docstring at least. If so, `new_filter_value: T, old_filter_value: T` may be more appropriate.

bojanserafimov: Not sure if that's changing anything. Can't mypy just plug in `object` for `T`? Strong typing with inheritance is weird.

PR #774, opened 2020-03-16
Add missing installation instructions

pmantica1:
In this diff, I add missing installation instructions and remove sudo from all installation commands.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -27,7 +27,8 @@ or on Ubuntu with:
 
 .. code:: bash
 
-   apt-get install python-mysqldb
+   sudo apt-get install libmysqlclient-dev
+   sudo apt-get install python-mysqldb
 
 For more details on other systems please refer to `MySQL dialect
 information <https://docs.sqlalchemy.org/en/latest/dialects/mysql.html>`__.
@@ -46,6 +47,7 @@ Or Ubuntu with:
 
 .. code:: bash
 
+    sudo apt-get install unixodbc unixodbc-dev
     wget -qO- https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
     sudo add-apt-repository "$(wget -qO- https://packages.microsoft.com/config/ubuntu/"$(lsb_release -r -s)"/prod.list)"
     sudo apt-get update
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -27,8 +27,7 @@ or on Ubuntu with:
 
 .. code:: bash
 
-   apt-get install libmysqlclient-dev
-   apt-get install python-mysqldb
+   apt-get install libmysqlclient-dev python-mysqldb
 
 For more details on other systems please refer to `MySQL dialect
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -27,6 +27,7 @@ or on Ubuntu with:
 
 .. code:: bash
 
+   apt-get install libmysqlclient-dev
    apt-get install python-mysqldb
```
obi1kenobi: Why not combine these into the same `apt-get install` command?

CONTRIBUTING.rst
```diff
@@ -46,6 +47,7 @@ Or Ubuntu with:
 
 .. code:: bash
 
+    sudo apt-get install unixodbc unixodbc-dev
```
obi1kenobi: See line 55. Also, I'm pretty sure that `unixodbc` is a prerequisite of `unixodbc-dev` because development packages tend to depend on the non-development ones.

obi1kenobi:
On the "sometimes with sudo, sometimes without", I'd say we shouldn't have sudo in our shown commands, and we should add a note telling users that they may need to sudo depending on their system setup. Over-using sudo is dangerous, and when to use it is best left to users' judgment.

CONTRIBUTING.rst
pmantica1: Removed `unixodbc`. Based on our `.travis.yml` file, I think that only `unixodbc-dev` is required.

pmantica1:
> On the "sometimes with sudo, sometimes without", I'd say we shouldn't have sudo in our shown commands, and we should add a note telling users that they may need to sudo depending on their system setup. Over-using sudo is dangerous, and when to use it is best left to users' judgment.

Changed commands to never use sudo. I don't think we need that note though. I think that users will be able to tell that they need to add sudo to the commands based on the "permission denied" errors.

obi1kenobi:
> > On the "sometimes with sudo, sometimes without", I'd say we shouldn't have sudo in our shown commands, and we should add a note telling users that they may need to sudo depending on their system setup. Over-using sudo is dangerous, and when to use it is best left to users' judgment.
> 
> Changed commands to never use sudo. I don't think we need that note though. I think that users will be able to tell that they need to add sudo to the commands based on the "permission denied" errors.

They might or might not. Adding a brief note that `sudo` may be required costs us nothing, though, and may reduce the friction that some potential contributors might experience, so it still seems overwhelmingly net-positive to me.

CONTRIBUTING.rst
```diff
@@ -46,11 +46,11 @@ Or Ubuntu with:
 
 .. code:: bash
 
-    wget -qO- https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
```
obi1kenobi: Maybe keep this sudo — this one will definitely be needed, and properly adding `sudo` is more complex here than everywhere else since it's not just "prefix the line with `sudo`".

Either that or maybe split this command into two steps:
```
wget -qO- https://packages.microsoft.com/keys/microsoft.asc > /tmp/microsoft.asc
apt-key add /tmp/microsoft.asc
```
where only the second one needs `sudo`.

PR #775, opened 2020-03-17
Fix revisit_count handling

bojanserafimov:
The `revisit_count` field exists for gremlin, so any developer unfamiliar with the gremlin backend is not likely to think of it when dealing with `Location` objects. As one of those developers, I've written some buggy code that fails on queries with optionals.

Changes in this PR:
1. Add test that fails
2. Change the return type of `get_types` from string type name to grahpql type
3. Implement `get_filters` analysis pass that gets filters at a location across all revisits
4. Remove direct use of `Location` objects, resolving the bug


bojanserafimov:
Checking the sql backend for `revisit_count` bugs is also a good idea.

PR #777, opened 2020-03-18
Fix pagination planning edge case

bojanserafimov:
The pagination planner sometimes generates an impossible plan (not enough quantiles to execute it). The parameter generator has to generate fewer than the planned number of pages. This is fine in most cases, but when it generates parameters for 0 pages we get errors.

Possible solutions:
1. Deal with duplicate quantiles during computation of pagination capacities, making sure that all generated plans are possible to execute.
2. Deal with empty generator produced by parameter generator.

Solution 1 looks right for the long term, but does not fit in a single PR. To deal with duplicate quantiles before we generate a plan, we have to:
- iterate the quantiles for all fields defined in the query. This iteration can easily take seconds of execution time.
- deal with uneven page sizes. We've so far had the assumption that a capacity of N means the ability to split into N equally sized pages. To deal with duplicates at this stage we have to come up with a definition of pagination capacity that allows for uneven page sizes.
- reconsider how we store quantiles


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -272,7 +272,6 @@ def get_pagination_capacities(
     fields_eligible_for_pagination: Set[PropertyPath],
     field_value_intervals: Dict[PropertyPath, Interval[Any]],
     distinct_result_set_estimates: Dict[VertexPath, float],
-    parameters: Dict[str, Any],
```
bojanserafimov: Just noticed an unused parameter. Irrelevant to the PR.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -81,17 +82,32 @@ def paginate_query_ast(
...
     )
     if num_pages > 1:
         pagination_plan, advisories = get_pagination_plan(query_analysis, num_pages)
-        if pagination_plan.vertex_partitions:
-            page_query, remainder_query = split_into_page_query_and_remainder_query(
-                query_analysis, pagination_plan
+        if len(pagination_plan.vertex_partitions) == 0:
+            pass
+        elif len(pagination_plan.vertex_partitions) == 1:
+            parameter_generator = generate_parameters_for_vertex_partition(
```
obi1kenobi: nit: mind extracting `pagination_plan.vertex_partitions[0]` somewhere here and assigning it to a local? The repeated `[0]` dereferences keep making me look back for evidence that `[0]` is defined. The code is correct  right now but I think this would make it easier for the reader to convince themselves of that.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -81,17 +82,32 @@ def paginate_query_ast(
...
+                query_analysis.schema_info,
+                query_analysis.ast_with_parameters,
+                pagination_plan.vertex_partitions[0],
+            )
+            first_param = next(parameter_generator, None)
+            if first_param:
```
obi1kenobi: Do we want some sort of advisory if we are unable to paginate as requested? Or is that not worth it / not the right idea for some reason?

bojanserafimov: We use advisories to see what is worth fixing. We know this is worth fixing, so I'd rather fix it than spend more time on getting more signal.

PR #778, opened 2020-03-19
Add selectivity analysis

bojanserafimov:
Computing the selectivity of all filters on a given vertex is currently a helper method used both in cost estimation and in pagination separately. I'm making it an analysis pass.

Working towards 3 goals:
- Making the cost estimation process transparent. Not quite like `EXPLAIN`, but if it's composed of independent thin analysis passes, printing out the results of the passes would be enough information to see what's going on without having to `pdb` into the library.
- Making cost estimation more granularly testable. Currently most of its tests are white-box end-to-end tests.
- Fixing rotation invariance: https://github.com/kensho-technologies/graphql-compiler/pull/593 When cost estimation is simpler, it's easier to make it rotation invariant.

In this PR:
- Make the `get_types` pass go into folded locations.
- Add a `get_fold_scope_roots` pass
- Make `get_fields_eligible_for_pagination` explicitly avoid folds, instead of relying on a limitation of `get_types`
- Add a `get_selectivities` pass
- Add some tests for new passes

### DIFF
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -66,7 +66,9 @@ def _convert_int_interval_to_field_value_interval(
 def _get_location_vertex_path(location: BaseLocation) -> VertexPath:
     """Get the VertexPath for a BaseLocation pointing at a vertex."""
     if location.field is not None:
-        raise AssertionError(f"Location {location} represents a field.")
+        raise AssertionError(
+            f"Location {location} represents a field. Expected a location pointing at a vertex."
+        )
 
     if isinstance(location, Location):
         return location.query_path
@@ -74,8 +76,7 @@ def _get_location_vertex_path(location: BaseLocation) -> VertexPath:
         return location.base_location.query_path + tuple(
             "{}_{}".format(direction, name) for direction, name in location.fold_path
         )
-    else:
-        raise AssertionError("Unexpected location encountered: {}".format(location))
+    raise AssertionError("Unexpected location encountered: {}".format(location))
 
 
 def get_types(
@@ -233,7 +234,7 @@ def get_selectivities(
     for vertex_path, vertex_type in types.items():
         vertex_type_name = vertex_type.name
         filter_infos = filters[vertex_path]
-        # TODO(bojanserafimov) use analysis.field_value_intervals
+        # TODO(bojanserafimov) use precomputed field_value_intervals
         #                      inside this method instead of recomputing it
         selectivity = get_selectivity_of_filters_at_vertex(
             schema_info, filter_infos, parameters, vertex_type_name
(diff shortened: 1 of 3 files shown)


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -56,6 +63,21 @@ def _convert_int_interval_to_field_value_interval(
     return Interval(lower_bound, upper_bound)
 
 
+def _get_location_vertex_path(location: BaseLocation) -> VertexPath:
+    """Get the VertexPath for a BaseLocation pointing at a vertex."""
+    if location.field is not None:
+        raise AssertionError(f"Location {location} represents a field.")
```
obi1kenobi: nit: This error is factually correct but doesn't actually describe the precondition that was violated.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -56,6 +63,21 @@ def _convert_int_interval_to_field_value_interval(
...
+
+    if isinstance(location, Location):
+        return location.query_path
+    elif isinstance(location, FoldScopeLocation):
+        return location.base_location.query_path + tuple(
+            "{}_{}".format(direction, name) for direction, name in location.fold_path
+        )
+    else:
+        raise AssertionError("Unexpected location encountered: {}".format(location))
```
obi1kenobi: I'd probably pull this out of the `else`.

The issue is that if one of the `if-elif` branches is taken but doesn't end in a return (not today, not tomorrow, but sometime in the future), we might fall through to the end of the function and get an implicit `return None`. With the last statement of the function being `raise AssertionError()` that will be impossible.

The other alternative is to construct a result object in the `if-elif` branches, and then return that object as the last line of the function. That might look even cleaner.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -204,11 +222,31 @@ def get_field_value_intervals(
...
     types: Dict[VertexPath, Union[GraphQLObjectType, GraphQLInterfaceType]],
     filters: Dict[VertexPath, Set[FilterInfo]],
     parameters: Dict[str, Any],
+) -> Dict[VertexPath, Selectivity]:
+    """Get the combined selectivities of filters at each vertex."""
+    selectivities = {}
+    for vertex_path, vertex_type in types.items():
+        vertex_type_name = vertex_type.name
+        filter_infos = filters[vertex_path]
+        # TODO(bojanserafimov) use analysis.field_value_intervals
```
obi1kenobi: What is `analysis` referring to here?

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -450,6 +450,16 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
     return combined_selectivity
 
 
+def adjust_counts_with_selectivity(counts, selectivity):
+    """Apply the selectivity to the unfiltered vertex count and return the result."""
```
obi1kenobi: Can you give some more context here, through the name of the function, the documentation, or the naming of the input variables (and ideally all of them)? This is all quite opaque.

Also, missing type hints.

PR #779, opened 2020-03-19
Remove first pipe character from folded MSSQL results as a part of postprocessing

chewselene:
XML path base array aggregation for MSSQL folds places a `|` before every result in the list. This PR removes the `|` from the first element of the string representation of the list so that every result does not have "" as the first element of the list after postprocessing.

### DIFF
--- a/graphql_compiler/post_processing/sql_post_processing.py
+++ b/graphql_compiler/post_processing/sql_post_processing.py
@@ -24,23 +24,32 @@ def _mssql_xml_path_string_to_list(
     if xml_path_result == "":
         return []
 
+    # Some of the special characters involved in XML path array aggregation.
+    delimiter = "|"
+    null = "~"
+
     # Remove the "|" from the first result in the string representation of the list.
+    if xml_path_result[0] != delimiter:
+        raise AssertionError(
+            f"Unexpected fold result. All XML path array aggregated lists must start with a '|'. "
+            f"Received a result beginging with a {xml_path_result[0]}: {xml_path_result}"
+        )
     xml_path_result = xml_path_result[1:]
 
     # Split the XML path result on "|".
-    list_result: Sequence[Optional[str]] = xml_path_result.split("|")
+    list_result: Sequence[Optional[str]] = xml_path_result.split(delimiter)
 
     # Convert "~" to None. Note that this must be done before "^n" -> "~".
-    list_result = [None if result == "~" else result for result in list_result]
+    list_result = [None if result == null else result for result in list_result]
 
     # Convert "^d" to "|".
     list_result = [
-        result.replace("^d", "|") if result is not None else None for result in list_result
+        result.replace("^d", delimiter) if result is not None else None for result in list_result
     ]
 
     # Convert "^n" to "~".
     list_result = [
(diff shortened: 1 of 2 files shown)


graphql_compiler/post_processing/sql_post_processing.py
```diff
@@ -24,6 +24,9 @@ def _mssql_xml_path_string_to_list(
     if xml_path_result == "":
         return []
 
+    # Remove the "|" from the first result in the string representation of the list.
+    xml_path_result = xml_path_result[1:]
```
obi1kenobi: Can we assert that we're removing what we think we're removing?

graphql_compiler/post_processing/sql_post_processing.py
```diff
@@ -24,6 +24,9 @@ def _mssql_xml_path_string_to_list(
...
+
     # Split the XML path result on "|".
     list_result: Sequence[Optional[str]] = xml_path_result.split("|")
```
obi1kenobi: Yikes, we don't have constants for these? Mind making a constant for the separator character?

obi1kenobi:
Perhaps in a separate PR, can we set up an integration test that checks this end-to-end including the real database? Or at least can we open a tracking issue for it that we can resolve during our next test-a-thon?

graphql_compiler/tests/test_post_processing.py
```diff
@@ -340,3 +340,24 @@ def test_convert_complex(self):
...
+        """Test invalid result throws error.
+
+                {
+                    Animal {
+                        in_Animal_ParentOf @fold{
+                            name @output(out_name: "child_names")
+                        }
+                    }
+                }
+                """
```
obi1kenobi: This indentation seems a bit strange, assuming intentional?

Also perhaps a bit more context on the significance of the query might be useful.

graphql_compiler/post_processing/sql_post_processing.py
```diff
@@ -24,20 +24,32 @@ def _mssql_xml_path_string_to_list(
...
         return []
 
+    # Some of the special characters involved in XML path array aggregation.
+    delimiter = "|"
+    null = "~"
+
+    # Remove the "|" from the first result in the string representation of the list.
+    if xml_path_result[0] != delimiter:
+        raise AssertionError(
+            f"Unexpected fold result. All XML path array aggregated lists must start with a '|'. "
```
obi1kenobi: nit: `must start with a '{delimiter}'` instead of another hardcoded `|`,  in case it changes in the future

graphql_compiler/tests/test_post_processing.py
chewselene: This was bad copy pasting! The query isn't really important to the test - I just thought it was a bit easier to understand what was going on if there was an example query that could have produced those results.

obi1kenobi: Could we clarify in the docstring that this query is intended to be an example input that could have produced those results? I had a guess that was what was going on, but explicit is better than implicit.

obi1kenobi: Also, if you don't mind, `@fold{` should be `@fold {` per the pretty-printer.

PR #786, opened 2020-03-30
Optimize SQL recurse with CTEs

bojanserafimov:
In some cases Postgres and MSSQL compute the recurrence before they compute the base case, which can be millions of times slower. To force the backend to compute the base case first, it can be wrapped in a cte.

In this PR I wrap the query contents before the recurse (as seen in the IR) into a cte if they are complicated enough (contain at least one traversal or filter).

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -758,7 +758,7 @@ class CompilationState(object):
         """Execute a Backtrack Block."""
         self._relocate(previous_location)
 
-    def traverse(self, vertex_field, optional):
+    def traverse(self, vertex_field: str, optional: bool) -> None:
         """Execute a Traverse Block."""
         self._recurse_needs_cte = True
 
@@ -887,7 +887,7 @@ class CompilationState(object):
             raise AssertionError("CompilationState is already in global scope.")
         self._current_location = None
 
-    def filter(self, predicate):
+    def filter(self, predicate: Expression) -> None:
         """Execute a Filter Block."""
         self._recurse_needs_cte = True
 
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -873,6 +873,9 @@ class CompilationState(object):
             .where(base.c[CTE_DEPTH_NAME] < literal_depth)
         )
 
+        # Instead of joining to the current _from_clause, we make this alias the _from_clause.
+        # If the existing _from_clause had any information in it, then the _current_alias would
+        # be a cte that contains all that information.
         self._from_clause = self._current_alias
 
     def start_global_operations(self):
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -702,6 +702,7 @@ class CompilationState(object):
         self._from_clause = self._current_alias  # the main sqlalchemy Selectable
         self._outputs = []  # sqlalchemy Columns labelled correctly for output
         self._filters = []  # sqlalchemy Expressions to be used in the where clause
+        self._filters_at_location = {}
 
         self._current_fold = None  # SQLFoldObject to collect fold info and guide output query
         self._fold_vertex_location = None  # location in the IR tree where the fold starts
@@ -785,6 +786,8 @@ class CompilationState(object):
 
     def recurse(self, vertex_field, depth):
         """Execute a Recurse Block."""
+        predicates_on_this_node = self._filters_at_location[self._current_location.query_path]
+
         if self._current_fold is not None:
             raise AssertionError("Recurse inside a fold is not allowed.")
         previous_alias = self._current_alias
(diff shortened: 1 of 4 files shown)


graphql_compiler/tests/test_compiler.py
```diff
@@ -1958,32 +1958,30 @@ def test_simple_recurse(self):
...
+                            ON anon_1.uuid = [Animal_2].parent
                     WHERE anon_1.__cte_depth < 1
             )
             SELECT
                 anon_1.name AS relation_name
             FROM
-                db_1.schema_1.[Animal] AS [Animal_1]
-                JOIN anon_1
-                    ON [Animal_1].uuid = anon_1.__cte_key
+                anon_1
```
bojanserafimov, line 1984: The query is simplified since the unnecessary join into the CTE was removed.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -640,6 +653,22 @@ def generate_subquery(self):
...
+class ColumnRouter:
+    """Container for columns selected from a variety of selectables.
+
+    ContextFields selecting from locations inside a CTE need to be redirected to get those columns
+    from the corresponding columns exposed by the CTE instead. A ColumnRouter can be used to store
+    these column name mappings.
+
+    The Selectable.c property is the only property of selectables used by ContextFields, so that's
+    the only property this class needs to implement to serve as a Selectable.
+    TODO(bojanserafimov): make an abstract class instead of duck-typing this.
```
bojanserafimov, line 666: @obi1kenobi want your opinion on this solution before I go ahead typing.

The alternative is to:
1. Create a `class GeneratedLocation(BaseLocation)`
2. Make a `CreateSubquery` IR block that will be used for CTEs
3. Put all used fields of all locations before the `CreateSubquery` block into the `CreateSubquery` block with some name disambiguation
4. Rewrite all following locations to point to the subquery instead of the locations it's wrapping.

Let me know if you see a cleaner solution.

obi1kenobi: Yeah this seems the best option at the moment. Maybe if/when we re-do the IR we can come up with something better. The alternative approach seems like a heck of a lot of work, only to produce a solution that is still held together by string and duct tape.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -96,6 +98,17 @@ def _find_columns_used_outside_folds(sql_schema_info, ir):
...
+    # Columns used in the base case of CTE recursions should be made available from parent scope
+    for location, _ in ir.query_metadata_table.registered_locations:
+        if isinstance(location, FoldScopeLocation):
+            continue
+        for recurse_info in ir.query_metadata_table.get_recurse_infos(location):
+            traversal = "{}_{}".format(recurse_info.edge_direction, recurse_info.edge_name)
+            used_columns[location.query_path] = used_columns.get(location.query_path, set()).union(
+                used_columns[location.query_path + (traversal,)]
+            )
+            used_columns.setdefault(location.query_path, set()).add(edge.from_column)
```
obi1kenobi, line 110: I'm confused by these two consecutive statements. A few things that are confusing:
- `used_columns` has no type hint, so I don't know what to expect inside it (the function itself is also missing type hints)
- I'm not sure why both statements are assigning to what appears to be the same entry in the dict, `used_columns[location.query_path]`.
- Furthermore, I'm confused why the second statement uses `setdefault()` when the first one guarantees that the value at that key is already present and already a set.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -731,6 +763,36 @@ def traverse(self, vertex_field, optional):
         self._relocate(self._current_location.navigate_to_subpath(vertex_field))
         self._join_to_parent_location(previous_alias, edge.from_column, edge.to_column, optional)
 
+    def _wrap_into_cte(self):
```
obi1kenobi, line 766: missing type hints throughout this function -- let's make sure all new code is type-hinted

graphql_compiler/tests/test_compiler.py
```diff
@@ -9268,75 +9315,91 @@ def test_simple_optional_recurse(self):
...
-                    ON [Animal_2].parent = [Animal_1].uuid
-                LEFT OUTER JOIN anon_1 ON [Animal_1].uuid = anon_1.__cte_key
-            WHERE
-                anon_1.__cte_key IS NOT NULL OR [Animal_1].uuid IS NULL
+                anon_1.[Animal_in_Animal_ParentOf__name] AS child_name,
+                anon_1.[Animal__name] AS name,
+                anon_2.name AS self_and_ancestor_name
+            FROM anon_1, anon_2
         """
+        # TODO Add an integration test for this query to make sure the recurse preserves
```
obi1kenobi, line 9358: Please use the `TODO(<author>):` format so we can track TODO counts and locations with standard tools.

obi1kenobi (commented):
Please make sure to add type hints for all new code and for all existing functions you edited.

obi1kenobi (approved):

hhjmmmmmmmm (approved):

PR #789, opened 2020-04-01
Add integration test for folded GraphQLDecimals

chewselene:
Note: this has a few failing tests in `test_compiler.py` that are fixed as a part of #788. 

`GraphQLDecimal`s are not properly handled during postprocessing for MSSQL folds. The `parse_value` function of `GraphQLDecimal` returned as a part of the `OutputMetadata` incorrectly uses the basic pass through `parse_value` function of `GraphQLScalar` rather than converting to `Decimal` (i.e. `GraphQLDecimal`'s `parse_value` function). This causes the list output of `child_net_worths` to be of type `string` rather than type `Decimal`.

@obi1kenobi pointed out that this may have to do with @pmantica1 's PR #717 

edit: the above issue has been fixed through other PRs. This PR just adds an integration test for outputting folded `GraphQLDecimal`s.


pmantica1:
If the theory abovie is correct, this should be easily fixable once the following pr is merged:
https://github.com/kensho-technologies/graphql-compiler/pull/821

The `convert_scalar_value_to_graphql_type` does not assume that the input `GraphQLType` has the `.parse_value` correctly set.

pmantica1:
I think that this should be fixed now since the `deserialize_value` function uses the original definition of the `parse_value` function of the inputted graphql type.

pmantica1:
@chewselene Do you mind closing or merging this pr whenever you can so that we don't have that many prs open?

PR #791, opened 2020-04-07
Simplify sql fold

bojanserafimov:
Trying to understand and simplify the fold code, I removed some lines and no tests broke. I don't understand the lines so I'd appreciate any help understanding the original purpose. @chewselene are you familiar with this code? What kind of test would catch this?

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -38,7 +38,7 @@ sphinx = ">=1.8,<2"
 arrow = ">=0.15.0,<1"
 funcy = ">=1.7.3,<2"
 dataclasses = ">=0.6"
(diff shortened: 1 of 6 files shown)


bojanserafimov:
My understanding is the goal is to prevent selection from the aliases created inside the fold subquery. I don't see how that would happen, and I also don't see how the current code prevents it. The code doesn't seem to have any effect since the `outer_from_table` value is passed into the function and returned unmodified. It's likely this code is solving a problem that used to exist.

Since we have good test coverage of this, I'd remove the code and continue simplifying.

chewselene:
Yep, I think you are right! Should be okay, to remove.

PR #792, opened 2020-04-08
Add PostgreSQL tests to test_compiler

chewselene:
Most of our PostgreSQL tests in test_compiler were `SKIP_TEST`. In this PR, PostgreSQL compilation is tested separately from MSSQL. 

Would appreciate some eyes on `test_no_op_coercion_inside_fold` in particular, since the produced result is slightly different than the commented out expected result from the TODO (hopefully, the commented expected SQL is just out of date).

### DIFF
--- a/graphql_compiler/tests/test_compiler.py
+++ b/graphql_compiler/tests/test_compiler.py
@@ -36,7 +36,7 @@ def check_test_data(
     expected_gremlin,
     expected_mssql,
     expected_cypher,
-    expected_postgresql=SKIP_TEST,
+    expected_postgresql,
 ):
     """Assert that the GraphQL input generates all expected output queries data."""
     if test_data.type_equivalence_hints:
@@ -143,7 +143,7 @@ class CompilerTests(unittest.TestCase):
                 animal_name: m.Animal___1.name
             ])}
         """
-        expected_sql = """
+        expected_mssql = """
             SELECT
                 [Animal_1].name AS animal_name
             FROM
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/test_compiler.py
```diff
@@ -5298,9 +6087,17 @@ def test_multiple_outputs_in_same_fold(self):
               [x IN collected_Animal__out_Animal_ParentOf___1 | x.name] AS `child_names_list`,
               [x IN collected_Animal__out_Animal_ParentOf___1 | x.uuid] AS `child_uuids_list`
         """
+        # TODO: implement ID type for postgresql folds
```
obi1kenobi, line 6090: `ID` type in the GraphQL schema just means "GraphQL int or GraphQL string", but the handling on the Postgres side should be identical. Hopefully not too much work for a future PR?

graphql_compiler/tests/test_compiler.py
```diff
@@ -5867,27 +6690,65 @@ def test_no_op_coercion_inside_fold(self):
...
-        #     SELECT
-        #         array_agg([Entity_1].name) AS fold_output_1,
-        #         [Animal_2].uuid AS uuid
-        #     FROM db_1.schema_1.[Animal] AS [Animal_2]
-        #     JOIN db_1.schema_1.[Entity] AS [Entity_1]
-        #     ON [Animal_2].uuid = [Entity_1].related_entity
-        #     GROUP BY [Animal_2].uuid
-        # ) AS folded_subquery_1
-        # ON [Animal_1].uuid = folded_subquery_1.uuid'''
+        # TODO(selene): ensure this sql produces expected results
```
obi1kenobi, line 6693: I think the SQL appears valid and appears to do the right thing. Another good sanity-check would be to make sure this SQL is the same as the one generated by the same GraphQL query without that coercion present.

graphql_compiler/tests/test_compiler.py
chewselene: yep, it is simple https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/compiler/expressions.py#L927
dates/datetimes and decimals should be added too.

graphql_compiler/tests/test_compiler.py
chewselene: sweet, confirmed that this produces the same results with/without the coercion.

PR #793, opened 2020-04-08
Pin graphql core <3.1

chewselene:
3.1 contains a breaking change.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -38,7 +38,7 @@ sphinx = ">=1.8,<2"
 arrow = ">=0.15.0,<1"
 funcy = ">=1.7.3,<2"
 dataclasses = ">=0.6"
-graphql-core = ">=3,<4"
+graphql-core = ">=3,<3.1" # 3.1 contains breaking change
 pytz = ">=2017.2"
 six = ">=1.10.0"
 sqlalchemy = ">=1.3.0,<2"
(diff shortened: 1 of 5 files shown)


bojanserafimov:
Should also update setup.py.

Sad to see the semver https://semver.org/ principles broken.

obi1kenobi:
3.1 is the Python equivalent of the Javascript v15 library.
3.0 is the Python equivalent of the Javascript v14 library.

The Javascript library seems to be following semver. I'm not sure what we were using is internal, so it may or may not be on us. We should maybe bring this up with the GraphQL-core maintainers, just to understand what our expectations should be around version bumps. In the meantime, I suggest we pin to minor versions instead of majors.

obi1kenobi:
@bojanserafimov can you take a look at the `query_parameterizer.py` lint issues? I think the version bump allowed mypy to do some deeper analysis and it's now unhappy.

obi1kenobi:
The `helpers.py` mypy issue seems like it might be an instance of https://github.com/python/mypy/issues/5374 so let's `# type: ignore` it.

bojanserafimov:
Sure. I'll send a diff for that file soon.

bojanserafimov:
https://github.com/kensho-technologies/graphql-compiler/pull/794

PR #794, opened 2020-04-08
Fix mypy issues in pagination parameterizer

bojanserafimov:


graphql_compiler/compiler/helpers.py
```diff
@@ -315,7 +315,7 @@ def get_parameter_name(argument: str) -> str:
 LocationT = TypeVar("LocationT", bound="BaseLocation")
 
 
-@total_ordering
+@total_ordering  # type: ignore
```
obi1kenobi, line 318: Maybe link to the mypy issue, so we know why this is ignored and when it might be safe to remove the override?

bojanserafimov:
@chewselene maybe just merge this branch into yours instead of waiting for it to land, since both branches depend on each other.

bojanserafimov:
merged into https://github.com/kensho-technologies/graphql-compiler/pull/793

Issue #745, opened 2020-01-30
Tz-aware inputs for tz-naive fields are allowed


bojanserafimov:
If a database field is tz-naive, using tz-aware filter values should be disallowed since there's no sane way to interpret them that is not a footgun.

One way to solve this is to make tz-naive and tz-aware datetimes separate GraphQL types.

PR #796, opened 2020-04-15
Add more scalar types to postgresql folds

chewselene:
Previously lists of strings or ints were the only allowable fold output types. This PR implements the rest of the supported scalar types (bool, ID, date, datetime, decimal, float) for folded output.


graphql_compiler/compiler/expressions.py
```diff
@@ -928,11 +937,22 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
...
             inner_type = strip_non_null_from_type(self.field_type.of_type)
-            if is_same_type(GraphQLInt, inner_type):
-                sql_array_type = "INT"
-            elif is_same_type(GraphQLString, inner_type):
-                sql_array_type = "VARCHAR"
-            else:
+            graphql_type_to_sql_array_type_dict = {
+                GraphQLInt: "INT",
+                GraphQLString: "VARCHAR",
+                GraphQLID: "UUID",
```
obi1kenobi, line 943: This one isn't valid, unfortunately. The GraphQL `ID` type can map to SQL's various int and string types, as well as `UUID`. Probably safest to coerce to string.

graphql_compiler/compiler/expressions.py
```diff
@@ -928,11 +937,22 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
             # PostgreSQL
             # coalesce to an empty array of the corresponding type
             inner_type = strip_non_null_from_type(self.field_type.of_type)
-            if is_same_type(GraphQLInt, inner_type):
-                sql_array_type = "INT"
-            elif is_same_type(GraphQLString, inner_type):
-                sql_array_type = "VARCHAR"
-            else:
+            graphql_type_to_sql_array_type_dict = {
+                GraphQLInt: "INT",
```
obi1kenobi, line 941: Are we confident `INT` and not `BIGINT` or something else like it is the appropriate choice here?

graphql_compiler/compiler/expressions.py
```diff
@@ -928,11 +937,22 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
...
+                GraphQLString: "VARCHAR",
+                GraphQLID: "UUID",
+                GraphQLBoolean: "BOOL",
+                GraphQLFloat: "FLOAT",
```
obi1kenobi, line 945: Similar to the above, not `DOUBLE` or another type?

graphql_compiler/compiler/expressions.py
```diff
@@ -928,11 +937,22 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
...
+                GraphQLDate: "DATE",
+                GraphQLDateTime: "TIMESTAMP",
+                GraphQLDecimal: "FLOAT",
```
obi1kenobi, line 948: This is a lossy conversion, which we probably don't want if we can avoid it. Is there a `DECIMAL` type in Postgres?

graphql_compiler/compiler/expressions.py
```diff
@@ -928,11 +937,22 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
...
-                sql_array_type = "VARCHAR"
-            else:
+            graphql_type_to_sql_array_type_dict = {
+                GraphQLInt: "INT",
+                GraphQLString: "VARCHAR",
+                GraphQLID: "UUID",
+                GraphQLBoolean: "BOOL",
+                GraphQLFloat: "FLOAT",
+                GraphQLDate: "DATE",
+                GraphQLDateTime: "TIMESTAMP",
```
obi1kenobi, line 947: This also might be problematic down the road. In GraphQL, we don't currently distinguish between datetimes with and without timezones, but Postgres has separate `timestamp` and `timestamptz` types. Not much to be done about it right now, other than maybe opening an issue so we don't forget.

graphql_compiler/compiler/expressions.py
```diff
@@ -928,11 +937,22 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
...
+                GraphQLDecimal: "FLOAT",
+            }
+            sql_array_type = None
+            for graphql_type, type_name in graphql_type_to_sql_array_type_dict.items():
+                if is_same_type(graphql_type, inner_type):
+                    sql_array_type = type_name
+                    break
```
obi1kenobi, line 954: This function is getting pretty long. How about pulling this type-matching logic into its own helper function, so that instead of `break` here you could have `return type_name`?

graphql_compiler/compiler/expressions.py
chewselene: FYI `BIGINT` returns as a string on db-fiddle.

https://www.db-fiddle.com/f/fJtQSrZqZn8tqd1S3ZfQWF/0

Seems like it might be a javascript limitation in db fiddle? ex: https://stackoverflow.com/questions/39168501/pg-promise-returns-integers-as-strings

obi1kenobi: Fascinating! It's probably exactly the Javascript issue you found, and doesn't seem to be a concern in Python:
> PostgreSQL bigint and oid are converted to long in Python 2 and to int in Python 3.
https://www.postgresql.org/docs/current/plpython-data.html#id-1.8.11.11.3

graphql_compiler/compiler/expressions.py
chewselene: Looks like we have #745. Do you think that covers it or should I make a separate issue?

obi1kenobi: It's probably fine, thanks for finding it.

graphql_compiler/compiler/expressions.py
```diff
@@ -802,6 +823,26 @@ def __init__(
         self.field_type = field_type
         self.validate()
 
+    def _get_sql_array_type(self, graphql_type) -> Optional[str]:
```
obi1kenobi, line 826:
```suggestion
    @staticmethod
    def _get_sql_array_type(graphql_type: GraphQLType) -> Optional[str]:
```

graphql_compiler/compiler/expressions.py
```diff
@@ -926,15 +967,12 @@ def to_sql(self, dialect: Any, aliases: AliasesDictType, current_alias: AliasTyp
...
-            # coalesce to an empty array of the corresponding type
-            inner_type = strip_non_null_from_type(self.field_type.of_type)
-            if is_same_type(GraphQLInt, inner_type):
-                sql_array_type = "INT"
-            elif is_same_type(GraphQLString, inner_type):
-                sql_array_type = "VARCHAR"
-            else:
+            # Coalesce to an empty array of the corresponding type.
+            graphql_type = self.field_type.of_type
+            sql_array_type = self._get_sql_array_type(graphql_type)
```
obi1kenobi, line 972:
```suggestion
            sql_array_type = FoldedContextField._get_sql_array_type(graphql_type)
```

obi1kenobi:
I'm not sure why `black` is going haywire like this, but I think @bojanserafimov might have merged into `master` a fix removing all explicit `u` string prefixes. So hopefully merging `master` will resolve most if not all of the issue.

bojanserafimov:
> I'm not sure why `black` is going haywire like this, but I think @bojanserafimov might have merged into `master` a fix removing all explicit `u` string prefixes. So hopefully merging `master` will resolve most if not all of the issue.

Just merged the unicode string fix. Merging from master should address the problem hopefully.

PR #798, opened 2020-04-16
Simplify emit sql error reporting logic

bojanserafimov:
Refactor improving readability and error message quality. Changes:
1. Introduce a `_get_current_primary_key_name` function. Not the most general helper, but covers use cases with better error messages that we had before.
2. Use the helper in `recurse` and `fold`, simplifying some of the longest functions in this module. Now `fold` will raise a proper error if the primary key is missing.


bojanserafimov:
When reviewing, I recommend looking at the two commits separately, since the second one just adds clutter. Not sure why black reformats only the changed files.

PR #800, opened 2020-04-20
Make class count statistics optional

bojanserafimov:
Now that we have advisories it makes more sense to raise one for missing class counts than to raise an `AssertionError`.

Changes in this PR:
1. Add end-to-end pagination tests with class counts missing
2. Make class counts optional
3. Disable pagination completely if any relevant class count is missing
3. Return advisories for misssing class counts

### DIFF
--- a/graphql_compiler/cost_estimation/analysis.py
+++ b/graphql_compiler/cost_estimation/analysis.py
@@ -409,16 +409,17 @@ class QueryPlanningAnalysis:
         return get_types(self.metadata_table)
 
     @cached_property
-    def has_class_counts_data(self) -> bool:
+    def classes_with_missing_counts(self) -> Set[str]:
         """Return whether class counts are available for all vertices and edges used."""
+        classes_with_missing_counts = set()
         for vertex_path, vertex_type in self.types.items():
             if self.schema_info.statistics.get_class_count(vertex_type.name) is None:
-                return False
+                classes_with_missing_counts.add(vertex_type.name)
             if len(vertex_path) > 1:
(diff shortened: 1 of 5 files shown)


graphql_compiler/cost_estimation/analysis.py
```diff
@@ -408,6 +408,19 @@ def types(self) -> Dict[VertexPath, Union[GraphQLObjectType, GraphQLInterfaceTyp
         """Find the type at each VertexPath."""
         return get_types(self.metadata_table)
 
+    @cached_property
+    def classes_with_missing_counts(self) -> Set[str]:
+        """Return whether class counts are available for all vertices and edges used."""
```
obi1kenobi, line 413: This seems imprecise: `whether` makes me expect a boolean, but the result seems to be the list of classes whose counts are missing. The boolean is perhaps `len(result) > 0`?

This sentence is valuable, as it describes the intent / scenario when I'd use this property. But it does not reflect the actual return value, so perhaps rephrase or switch to a long docstring.

graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -14,6 +14,18 @@ class PaginationAdvisory(ABC):
...
 
 
+@dataclass
+class MissingClassCount(PaginationAdvisory):
+    class_name: str
+
+    def __post_init__(self):
+        """Initialize a human-readable message."""
+        self.message = (
+            f"Class count statistics for the vertices and edges mentioned in the queries "
```
obi1kenobi, line 24: `queries` or `query`? Can it be plural?

PR #802, opened 2020-04-22
Add a function that deserializes a dict of json arguments

pmantica1:
I add a function that makes it slightly easier to deserialize json serialized arguments by using the input metadata.

I also add type hints in the surrounding code and fix some documentation issues that I found while adding them.


graphql_compiler/query_formatting/common.py
```diff
@@ -134,17 +141,37 @@ def deserialize_json_argument(name, expected_type, value):
...
+
+    Args:
+        arguments: dict mapping argument names to json serialized values.
+        expected_types: dict mapping argument names to expected GraphQL types.
+
+    Returns:
+        a dict mapping the argument names to their deserialized values. See the docstring of
+        deserialize_json_argument for more info on how arguments are deserialized.
+    """
+    validate_that_dicts_have_the_same_keys(arguments, expected_types)
```
bojanserafimov, line 168: The way I expect this function to be used, this should be a user error. The error raised from inside this function is an `AssertionError` with a developer-friendly but not user-friendly message. I'd rather see "epxected arguments for ..." or "unnecessary arguments provided", with a `InvalidArgumentError`, or something like that.

graphql_compiler/global_utils.py
```diff
@@ -59,3 +59,25 @@ def is_same_type(left, right):
         return is_same_type(left.of_type, right.of_type)
     else:
         return False
+
+
+_KeyType = TypeVar("_KeyType")
+
+
+def validate_that_dicts_have_the_same_keys(
+    dict1: Dict[_KeyType, Any], dict2: Dict[_KeyType, Any]
```
bojanserafimov, line 68: I don't think this asserts the types are the same. It only asserts that if the key type in dict1 is one of the allowed types (any type), and the key type in dict2 is one of the allowed types (any type), then the returned value is None. So I think {1: 5} and {True: 8} would be considered to have the same keys. Might as well make the key `str` until there's a use case for a generic function.

graphql_compiler/query_formatting/common.py
```diff
@@ -44,37 +53,35 @@ def _deserialize_anonymous_json_argument(expected_type, value):
...
         it is already the base case of exceptions raised by the GraphQL parsers.
     """
     allowed_types_for_graphql_type = {
-        GraphQLDate.name: (six.string_types,),
-        GraphQLDateTime.name: (six.string_types,),
-        GraphQLFloat.name: (six.string_types, float, six.integer_types),
-        GraphQLDecimal.name: (six.string_types, float, six.integer_types),
-        GraphQLInt.name: (six.integer_types, six.string_types),
-        GraphQLString.name: (six.string_types,),
+        GraphQLDate.name: (str,),
```
pmantica1, line 66: Mypy was going haywire with the nested type tuples of varying depth. So I just got rid of "six.string_types" and "six.string_types" since we longer support python2 and replaced them with the python3 only equivalents.

graphql_compiler/query_formatting/common.py
```diff
@@ -31,11 +41,11 @@ def _raise_invalid_type_error(name, expected_python_type_name, value):
     )
 
 
-def _deserialize_anonymous_json_argument(expected_type, value):
+def _deserialize_anonymous_json_argument(expected_type: GraphQLScalarType, value: Any) -> Any:
     """Deserialize argument. See docstring of deserialize_json_argument.
 
     Args:
-        expected_type: GraphQLType we expect. All GraphQLNonNull type wrappers are stripped.
```
pmantica1, line 38: This function already expects null wrappers to be stripped.

graphql_compiler/query_formatting/common.py
```diff
@@ -216,7 +250,9 @@ def validate_argument_type(name, expected_type, value):
         )
 
 
-def ensure_arguments_are_provided(expected_types, arguments):
+def ensure_arguments_are_provided(
+    expected_types: Mapping[str, GraphQLType], arguments: Mapping[str, Any]
+) -> None:
     """Ensure that all arguments expected by the query were actually provided."""
```
pmantica1, line 256: This function was doing more than it promised.

bojanserafimov: Code in macro validation was relying on that. It should use `validate_arguments` now

pmantica1: Addressed.

PR #804, opened 2020-04-23
Handle list types during argument deserialization

pmantica1:
In this PR I modify the code to handle list types during argument deserialization. I also simplify the deserialization code and make some minor other improvements.


graphql_compiler/query_formatting/common.py
```diff
@@ -34,37 +35,18 @@
...
-            GraphQLString: str
-            GraphQLBoolean: bool
-            GraphQLID: str
-
-    Raises:
-        ValueError if the value is not appropriate for the type. ValueError is chosen because
-        it is already the base case of exceptions raised by the GraphQL parsers.
-    """
-    allowed_types_for_graphql_type = {
+_ALLOWED_JSON_SCALAR_TYPES = MappingProxyType(
```
pmantica1, line 48: Raymond Hettinger recommended wrapping a anonymous python dict with the `MappingProxyType` as the easiest way of making dictionaries immutable. He is a one of the core python contributors.

graphql_compiler/query_formatting/common.py
```diff
@@ -152,16 +133,25 @@ def deserialize_json_argument(
...
+        ]
+    else:
+        return _deserialize_json_scalar_argument(name, stripped_type, value)
 
 
 def deserialize_multiple_json_arguments(
     arguments: Mapping[str, Any],
     expected_types: Mapping[str, Union[GraphQLNonNull[GraphQLScalarType], GraphQLScalarType]],
-) -> Mapping[str, Any]:
+) -> Dict[str, Any]:
```
pmantica1, line 154: No reason to not be more specific about the return type.

graphql_compiler/query_formatting/common.py
```diff
@@ -152,16 +133,25 @@ def deserialize_json_argument(
...
+            _deserialize_json_scalar_argument(name, inner_stripped_type, element)
+            for element in value
+        ]
+    else:
+        return _deserialize_json_scalar_argument(name, stripped_type, value)
 
 
 def deserialize_multiple_json_arguments(
     arguments: Mapping[str, Any],
     expected_types: Mapping[str, Union[GraphQLNonNull[GraphQLScalarType], GraphQLScalarType]],
```
pmantica1, line 153: Got to fix the type hint here and above.

graphql_compiler/query_formatting/common.py
```diff
@@ -74,49 +56,55 @@ def _deserialize_anonymous_json_argument(expected_type: GraphQLScalarType, value
...
 def deserialize_json_argument(
     name: str,
-    expected_type: Union[GraphQLNonNull[GraphQLScalarType], GraphQLScalarType],
+    expected_type: Union[
+        GraphQLScalarType,
+        GraphQLList[GraphQLScalarType],
+        GraphQLList[GraphQLNonNull[GraphQLScalarType]],
+        GraphQLNonNull[GraphQLScalarType],
+        GraphQLNonNull[GraphQLList[GraphQLScalarType]],
+        GraphQLNonNull[GraphQLList[GraphQLNonNull[GraphQLScalarType]]],
```
pmantica1, line 106: Thought of making this type annotation a variable but got a mypy error that pointed me to the following link:
https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases

pmantica1: I got the equivalent of the "error: Variable "__main__.tp" is not valid as a type" error you can see in the link.

graphql_compiler/global_utils.py
```diff
@@ -59,3 +59,22 @@ def is_same_type(left, right):
...
+    diff1 = mapping1_keys.difference(mapping2_keys)
+    diff2 = mapping2_keys.difference(mapping1_keys)
+
+    if diff1 or diff2:
+        error_message_list = ["Expected mappings to have the same keys."]
+        if diff1:
+            error_message_list.append(f" Keys in the first mapping but not the second: {diff1}.")
+        if diff2:
+            error_message_list.append(f" Keys in the second mapping but not the first: {diff2}.")
+        raise AssertionError(" ".join(error_message_list))
```
chewselene, line 80: I think you'll end up with double spacing here since you're joining with a `" "` and started your error messages with a space as well.

graphql_compiler/query_formatting/common.py
```diff
@@ -74,49 +56,55 @@ def _deserialize_anonymous_json_argument(expected_type: GraphQLScalarType, value
...
-        return name_to_custom_type[expected_type.name].parse_value(value)
-    else:
-        return expected_type.parse_value(value)
+
+def _deserialize_json_scalar_argument(name, expected_type: GraphQLScalarType, value: Any) -> Any:
+    """Deserialize the json serialized scalar argument."""
+    _validate_json_scalar_argument(name, expected_type, value)
+    try:
+        return _SCALAR_DESERIALIZATION_FUNCTIONS[expected_type.name](value)
+    except (ValueError, TypeError) as e:
```
obi1kenobi, line 94: How about `KeyError` because of `_SCALAR_DESERIALIZATION_FUNCTIONS[expected_type.name]`? Is that a bug in the code, or a user input error?

You could avoid this by applying a variant of the "parse, don't validate" principle: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

You have chosen to call `_validate_json_scalar_argument()` which returns no arguments, and then use that as an implicit "proof" that `_SCALAR_DESERIALIZATION_FUNCTIONS[expected_type.name]` won't raise `KeyError`. In addition to being implicit and therefore worse than explicit, this code is also not particularly stable in the face of refactoring and code evolution.

Instead, make a new function instead of `_validate_json_scalar_argument()` that has to either raise an error or return a deserializer function, which you can then call inside the `try-except`. That way, this new function has to do the potentially-KeyError-producing step, but in that case KeyError is an obvious bug due to a failure to appropriately check the user's input.

graphql_compiler/query_formatting/common.py
obi1kenobi: You don't want to make it a variable, you want to make it a type alias — that's precisely the distinction that `mypy` attempted to teach you at that link. A type alias will do what you want here, so please give the link one more shot and then let's discuss if it's still confusing.

graphql_compiler/query_formatting/common.py
pmantica1: Thanks for the article link. It was really good. Also thanks for the suggestion, I think that the code is much cleaner now.

graphql_compiler/query_formatting/common.py
pmantica1: In general, I agree with this suggestion, but I don't think that in this specific scenario it makes sense to include the comment.

I don't think that I should just include the comment in this specific place. The commenting methodology would be inconsistent. (We sometimes add a comment explaining why we use `MappingProxyType` and sometimes we don't). 

I also don't think that I should include the comment every single time that I use the type. That would clutter the codebase and I think that the `MappingProxyType` docstring is already pretty well written.

graphql_compiler/query_formatting/common.py
pmantica1: I am not sure about what I was doing wrong earlier, but I was able to fix this.

graphql_compiler/global_utils.py
```diff
@@ -59,3 +59,17 @@ def is_same_type(left, right):
         return is_same_type(left.of_type, right.of_type)
     else:
         return False
+
+
+def assert_set_equality(set1: Set[Any], set2: Set[Any]) -> None:
```
pmantica1, line 64: I was tempted to write `Set[_Element]` annotations for this functions, but it doesn't seem to have the desired effect. 

https://stackoverflow.com/questions/40033093/mypy-ensure-that-arguments-have-same-type-without-listing-the-types-explicitly
`

graphql_compiler/query_formatting/common.py
```diff
@@ -34,91 +35,84 @@
...
-    # Check for long integers, bypassing the GraphQLInt parser
-    if is_same_type(GraphQLInt, expected_type):
-        if isinstance(value, int):
-            return value
-        elif isinstance(value, str):
-            return int(value)
-        else:
-            raise ValueError(
-                "Unexpected type {}. Expected one of {}.".format(type(value), (int, str))
+    def _json_scalar_deserialization_function(name: str, value: Any) -> Any:
```
obi1kenobi, line 58: Is there any benefit to this being defined as a function-inside-a-function, instead of being a proper top-level function that takes all the args it needs, and then being used by using `partial` to provide the fixed arguments such as `graphql_type` and `expected_python_types`?

graphql_compiler/query_formatting/common.py
```diff
@@ -34,91 +35,84 @@
...
-        "Invalid type for argument {}. Expected {}. Got value {} of "
-        "type {}.".format(name, expected_python_type_name, value, type(value).__name__)
+        f"Invalid type for argument {name}. Expected one of {expected_python_types}. Got value "
+        f"{value} of type {type(value).__name__} instead."
     )
 
 
-def _deserialize_anonymous_json_argument(expected_type: GraphQLScalarType, value: Any) -> Any:
-    """Deserialize argument. See docstring of deserialize_json_argument.
+_CUSTOM_SCALAR_DESERIALIZATION_FUNCTIONS = MappingProxyType(
```
obi1kenobi, line 48: Can you add a comment here explaining why this is necessary and what makes these "custom"? There's nothing obviously custom about "parse ints as ints, and floats as floats", so it must be something non-obvious and worth explaining.

graphql_compiler/query_formatting/common.py
```diff
@@ -34,91 +35,84 @@
...
-    elif expected_type.name in name_to_custom_type:
-        # Since we cannot serialize the parse_value function of custom scalar types when
-        # serializing a schema, it is possible that the parse_value is incorrectly set. We must,
-        # therefore, use the parse_value function of the original scalar type definition.
-        return name_to_custom_type[expected_type.name].parse_value(value)
-    else:
-        return expected_type.parse_value(value)
+
+def _deserialize_json_scalar_argument(name, expected_type: GraphQLScalarType, value: Any) -> Any:
+    """Deserialize the json serialized scalar argument."""
```
obi1kenobi, line 106: Can you go through all the docstrings in this module (including the ones you merged in prior PRs) and reword them to be more descriptive and helpful for an unknowledgeable reader? Right now, many of them (e.g. this one) look like this:
```python
def do_thing():
    """Do the thing"""
```
This docstring provides no additional useful information that couldn't already be inferred from the function name. After reading this, I probably don't know what the function was actually intended for, what problem it solves, and when I should or should not be calling it.

Please avoid docstrings like that. It's better to not have docstrings than to have completely unhelpful ones.

graphql_compiler/query_formatting/common.py
obi1kenobi: I realize this is not part of the PR, but `Deserialize GraphQL arguments parsed from a json file.` seems odd, specifically the `json file` bit. Why does it have to be a file? This function doesn't do anything with files, as far as I can tell.

graphql_compiler/query_formatting/common.py
```diff
@@ -134,6 +128,7 @@ def deserialize_json_argument(
         GraphQLString: "Hello"
         GraphQLBoolean: True
         GraphQLID: "13d72846-1777-6c3a-5743-5d9ced3032ed"
+        GraphQLList: [1,2,3]
```
obi1kenobi, line 131: Please add spaces after comma. All the other lines have spaces after commas.

pmantica1:
Reduced the number of changes in this PR.

graphql_compiler/query_formatting/common.py
```diff
@@ -90,27 +94,36 @@ def _deserialize_anonymous_json_argument(expected_type: GraphQLScalarType, value
     if not isinstance(value, expected_python_types):
         correct_type = False
     if not correct_type:
-        raise ValueError(
-            "Unexpected type {}. Expected one of {}.".format(type(value), expected_python_types)
-        )
+        raise _raise_invalid_type_error(name, expected_python_types, value)
+
+    custom_deserialization_functions = merge_non_overlapping_dicts(
```
pmantica1, line 99: I think that there is opportunity for clean up here, but I tried to keep the number of changes minimal to not make this PR any longer.

pmantica1: This is what previously did in this function:
1. First we access a dict to check if the GraphQL type is one of the supported type and get the expected python types for the argument.
2. Then we check if the argument value has the expected python type.
3. Then we run through a series of if-statements to determine the deserialization function. (This might involve another dict access).

I think that it makes more sense to just create a dict that maps the supported GraphQL types to their expected python types and their deserialization function.

graphql_compiler/query_formatting/common.py
```diff
@@ -90,27 +94,36 @@ def _deserialize_anonymous_json_argument(expected_type: GraphQLScalarType, value
...
-    # Use the default GraphQL parser to parse the value
-    elif expected_type.name in name_to_custom_type:
-        # Since we cannot serialize the parse_value function of custom scalar types when
-        # serializing a schema, it is possible that the parse_value is incorrectly set. We must,
-        # therefore, use the parse_value function of the original scalar type definition.
-        return name_to_custom_type[expected_type.name].parse_value(value)
-    else:
-        return expected_type.parse_value(value)
+    deserialization_function = custom_deserialization_functions.get(
+        expected_type.name, expected_type.parse_value
```
obi1kenobi, line 120: Under what circumstances is the `.get()` expected to go to the "backup" `expected_type.parse_value` option? Can we make those situations explicitly handled by the dict or otherwise, and raise `AssertionError` for everything else? I'm worried this kind of catch-all can cause difficult-to-track-down errors down the road.

graphql_compiler/query_formatting/common.py
```diff
@@ -133,6 +146,7 @@ def deserialize_json_argument(
         GraphQLString: "Hello"
         GraphQLBoolean: True
         GraphQLID: "13d72846-1777-6c3a-5743-5d9ced3032ed"
+        GraphQLList: [1, 2, 3]
```
obi1kenobi, line 149: I thought `GraphQLList` is not a subclass of `GraphQLScalarType`? If so, either the type hint for `expected_type` is incorrect, or this example is incorrect. Also, `GraphQLList` is an incomplete type, perhaps here you meant `GraphQLList(GraphQLInt)`?

graphql_compiler/query_formatting/common.py
```diff
@@ -177,6 +204,8 @@ def deserialize_multiple_json_arguments(
...
+
```
obi1kenobi, line 208: nit: doubled new line before `Args`

graphql_compiler/query_formatting/common.py
pmantica1: Addressed. Decided to go for a full refactor.

graphql_compiler/query_formatting/common.py
```diff
@@ -51,6 +51,47 @@ def _raise_invalid_type_error(
...
+        # strings.
+        GraphQLFloat.name: float,
+        # Bypass the GraphQLInt parser and allow long ints and strings as input. The JSON spec
+        # allows only for 64-bit floating point numbers, so large ints might have to be
+        # represented as strings.
+        GraphQLInt.name: int,
+    }
+)
+
+_JSON_TYPES_AND_DESERIALIZATION_FUNCTIONS = MappingProxyType(
```
obi1kenobi, line 84: Missing type hints on each of these new additions.

pmantica1:
Changed:
`MappingProxyType[str, Callable[[Any], Any]]` to `Mapping[str, Callable[[Any], Any]]`
The first is not valid.

PR #805, opened 2020-04-28
Fix sql recurse bug

bojanserafimov:
See test case.

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -105,17 +105,6 @@ def _find_columns_used_outside_folds(
             if child_location_info.recursive_scopes_depth > location_info.recursive_scopes_depth:
                 used_columns.setdefault(child_location.query_path, set()).add(edge.from_column)
 
-    # Columns used in the base case of CTE recursions should be made available from parent scope
-    for location, _ in ir.query_metadata_table.registered_locations:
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -130,6 +119,17 @@ def _find_columns_used_outside_folds(
...
+    for location, _ in ir.query_metadata_table.registered_locations:
+        if isinstance(location, FoldScopeLocation):
+            continue
+        for recurse_info in ir.query_metadata_table.get_recurse_infos(location):
+            traversal = "{}_{}".format(recurse_info.edge_direction, recurse_info.edge_name)
+            used_columns[location.query_path] = used_columns.get(location.query_path, set()).union(
+                used_columns[location.query_path + (traversal,)]
+            )
+            used_columns[location.query_path].add(edge.from_column)
+
```
bojanserafimov, line 132: Moved this code down so that fields used in tags and outputs would be picked up.

obi1kenobi (approved):

graphql_compiler/compiler/emit_sql.py
hhjmmmmmmmm: Awesome

hhjmmmmmmmm (approved):

PR #803, opened 2020-04-23
Traversals in folds for PostgreSQL

chewselene:
This PR implements traversals inside of folds for PostgreSQL. The `SQLFoldObject` methods `visit_traversed_vertex` and `visit_output_vertex` have been combined into a single method `visit_vertex`, which performs a traversal and checks if there are any outputs at the new location. Since outputs can only come from a single location within a fold, `visit_vertex` throws an error if the user attempts to output from multiple locations within the fold.

PR #806, opened 2020-04-28
Create codecov.yml slightly relaxing patch coverage rule

obi1kenobi:
Inspired by the situation of #803, where the addition of thoroughly-tested code caused a negative coverage result because of a coverage change in an unrelated module.

Our project coverage is already very high, to the point where a single line missing coverage will currently make the coverage check fail. However, while our coverage is high, we don't have a high coverage _depth_ — any given line is covered relatively few times. This makes it somewhat likely that an unrelated change may cause some line to transition from covered to uncovered. Due to our high coverage, this makes it very likely that the newly-uncovered line will cause Codecov to report a negative coverage result, even though the code change may have been only distantly related to the coverage change.

The 0.03% threshold allows a few lines (1 < N < 10) to transition from covered to uncovered, without reporting a negative patch coverage result. If this PR is merged, I think #803 would get a positive patch coverage result.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    patch:
+      default:
+        target: auto
+        threshold: 0.03%
+        base: auto


chewselene (approved):
lgtm

obi1kenobi:
Going to leave this open for a bit longer, reaching out to Codecov support for help on why this PR caused a coverage change to be detected. I just checked out the `master` branch and edited an unrelated, non-Python file, so there shouldn't have been any coverage impact at all.

obi1kenobi:
> Going to leave this open for a bit longer, reaching out to Codecov support for help on why this PR caused a coverage change to be detected. I just checked out the `master` branch and edited an unrelated, non-Python file, so there shouldn't have been any coverage impact at all.

Opened https://community.codecov.io/t/reported-change-in-coverage-when-no-code-changes-made/1346 to seek advice from the Codecov maintainers.

obi1kenobi:
@bojanserafimov could you take a look at our tests and see if any non-determinism has crept in somewhere?

In the Codecov thread linked above, there is some decently compelling evidence that the generated coverage reports we supply to Codecov have a coverage discrepancy — and since those are generated before Codecov does anything, it would seem likely that the issue is on our end. Since the diff between this PR's commit and `0cf5414` doesn't include any code changes, non-deterministic tests are the only way I can see this happening.

bojanserafimov:
> @bojanserafimov could you take a look at our tests and see if any non-determinism has crept in somewhere?
> 
> In the Codecov thread linked above, there is some decently compelling evidence that the generated coverage reports we supply to Codecov have a coverage discrepancy — and since those are generated before Codecov does anything, it would seem likely that the issue is on our end. Since the diff between this PR's commit and `0cf5414` doesn't include any code changes, non-deterministic tests are the only way I can see this happening.

Does codecov run on this branch, or on this branch with master merged into it?

obi1kenobi:
> Does codecov run on this branch, or on this branch with master merged into it?

It runs on this branch, and also runs on `master` separately, then compares this branch to `master` (the "project" job) and this branch to the merge-base with `master` (the latest common ancestor commit, the "patch" job).

bojanserafimov:
> erge-ba



> > Does codecov run on this branch, or on this branch with master merged into it?
> 
> It runs on this branch, and also runs on `master` separately, then compares this branch to `master` (the "project" job) and this branch to the merge-base with `master` (the latest common ancestor commit, the "patch" job).

I don't see any non-determinacy on first glance. Changes on master seem like a more likely cause of this weird report.

bojanserafimov:
Actually, if set iteration order was not deterministic, I can see this exact report being produced. I'll read a bit about what's the exact guarantee python gives.

obi1kenobi:
See the codecov support link, there weren't any changes on `master` when this ran — this PR was branched off from the tip of `master` and `master` didn't change before this PR was opened.

bojanserafimov:
I couldn't find docs on this, but ran experiments locally and found some strange behaviors. If you make a set of ints and print the first element, you will get the same result consistently across multiple runs of that program. If you make a set of objects though, you'll get different ones sometimes.

Looks like the set iteration order depends on the order of elements, and the order of objects with respect to the values they store is not deterministic.

obi1kenobi:
I know Python 3 dicts preserve insertion order, but I don't know if sets do the same thing. We should assume non-deterministic order unless something guarantees determinism.

bojanserafimov:
Agree. But there's also another unrelated problem in this case: we have a set storing dataclasses with eq=False. This can cause other problems like unintuitive behavior of membership tests. My takeaway is let's try to have eq=True for all things that go in sets and dicts. I'll see if there's a way to lint this.

obi1kenobi:
@bojanserafimov reproduced the non-determinism locally, working on tracking it down. Merging the PR since it is no longer necessary to keep it open.

jerrode:
Hey @obi1kenobi, Jerrod from Codecov here, did you all ever find the root cause of the non-determinism in the test results? I'd be really curious to hear.

Thanks!

obi1kenobi:
Hey! Thanks for reaching out! I don't believe we ever found a conclusive answer beyond "the coverage file is not always the same, and Codecov correctly points that out" — right @bojanserafimov?

One of the reasons we had a hard time tracking this down is that our test setup changed when we migrated from Travis CI to GitHub Actions. Now we run separate unit and integration tests, which means there are multiple partial coverage files uploaded to Codecov, and that makes debugging the problem even trickier.

Also, unfortunately the Codecov integration adds a comment (which also triggers email notifications) as soon as it gets _any_ coverage files. Since unit tests always finish first, that means that the first Codecov post (and therefore the email about coverage) always claims an ~8% coverage drop. When integration tests finish, Codecov correctly updates the post in the PR to reflect the total coverage, but this doesn't send a new email and there's no way to un-send the inaccurate prior email either. As a result, I suspect that most of us suffer from the condition known as "normalization of deviance", where we've become desensitized to Codecov reports stating spuriously lower coverage numbers.

So I think this "coverage comes is multiple parts but is reported as soon as the first part is submitted" issue is probably a higher priority item that needs attention. Looking through the docs, I wasn't able to find a way to configure Codecov to expect multiple coverage reports and not report after the first one, but it's possible I wasn't looking in the right place.

Any help would be highly appreciated, though I recognize that we're not currently a paying customer and you probably have to prioritize paying customers' issues over this 😄

jerrode:
@obi1kenobi 

1 thing I'd definitely recommend to deal with your notifications issue: `after_n_builds` in the .yml so Codecov knows to wait to notify until everything is uploaded

https://docs.codecov.io/docs/codecovyml-reference#codecovnotifyafter_n_builds

As far as running separate unit and integration tests, have you checked out our `flags` feature where you can tag separate builds? Fo example, tagging unit and integration tests separately? 

https://docs.codecov.io/docs/flags

Lastly, if you are only uploading partial runs of your CI suite, you should check out Carryforward Flags so we can intelligently pick up only the new files uploaded: https://blog.codecov.io/2020/04/21/new-test-only-what-you-change-with-carryforward-flags/

obi1kenobi:
Wow, thank you — much appreciated! I will read up on the functionality you mentioned and try to put something together very soon.

ChuVienTran:
> Inspired by the situation of #803, where the addition of thoroughly-tested code caused a negative coverage result because of a coverage change in an unrelated module.
> 
> 
> 
> Our project coverage is already very high, to the point where a single line missing coverage will currently make the coverage check fail. However, while our coverage is high, we don't have a high coverage _depth_ — any given line is covered relatively few times. This makes it somewhat likely that an unrelated change may cause some line to transition from covered to uncovered. Due to our high coverage, this makes it very likely that the newly-uncovered line will cause Codecov to report a negative coverage result, even though the code change may have been only distantly related to the coverage change.
> 
> 
> 
> The 0.03% threshold allows a few lines (1 < N < 10) to transition from covered to uncovered, without reporting a negative patch coverage result. If this PR is merged, I think #803 would get a positive patch coverage result.

ChuVienTran (commented):
Code

Dpalazue (commented):
Hi

Newman1515 (commented):
Gat it

Heheehd (commented):
0cf5414

kakar930 (commented):
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

CantStandAFakeMofo2002 (approved):

scottwad (approved):

hhjmmmmmmmm (approved):

PR #807, opened 2020-04-28
Add missing type hints to query_formatting module

pmantica1:
I am adding type hints to `query_formatting/common.py` to reduce the size of this PR.
https://github.com/kensho-technologies/graphql-compiler/pull/804

### DIFF
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -2,7 +2,7 @@
 """Safely insert runtime arguments into compiled GraphQL queries."""
 import datetime
 import decimal
-from typing import Any, Mapping, Union
+from typing import Any, Dict, Mapping, Union
 
 import arrow
 from graphql import (
@@ -18,11 +18,18 @@ from graphql import (
 )
 import six
 
-from ..compiler import CYPHER_LANGUAGE, GREMLIN_LANGUAGE, MATCH_LANGUAGE, SQL_LANGUAGE
+from ..compiler import (
+    CYPHER_LANGUAGE,
+    GREMLIN_LANGUAGE,
+    MATCH_LANGUAGE,
+    SQL_LANGUAGE,
+    CompilationResult,
+)
 from ..compiler.helpers import strip_non_null_from_type
 from ..exceptions import GraphQLInvalidArgumentError
 from ..global_utils import is_same_type
 from ..schema import CUSTOM_SCALAR_TYPES, GraphQLDate, GraphQLDateTime, GraphQLDecimal
+from ..typedefs import GraphQLQueryArgumentType
 from .cypher_formatting import insert_arguments_into_cypher_query_redisgraph
 from .gremlin_formatting import insert_arguments_into_gremlin_query
 from .match_formatting import insert_arguments_into_match_query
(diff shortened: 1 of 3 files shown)


graphql_compiler/query_formatting/common.py
```diff
@@ -268,15 +275,15 @@ def ensure_arguments_are_provided(
...
+    expected_types: Mapping[str, GraphQLQueryArgumentType], arguments: Mapping[str, Any]
 ) -> None:
     """Ensure that all arguments are provided and that they are of the expected type."""
     ensure_arguments_are_provided(expected_types, arguments)
     for name in expected_types:
         validate_argument_type(name, expected_types[name], arguments[name])
 
 
-def insert_arguments_into_query(compilation_result, arguments):
+def insert_arguments_into_query(compilation_result: CompilationResult, arguments: Dict[str, Any]):
```
pmantica1, line 286: I would use the more general `Map` type here, but the documentation of this function and of the inner helper functions clearly specify dict.

graphql_compiler/typedefs.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2020-present Kensho Technologies, LLC.
+from typing import Union
+
+from graphql import GraphQLList, GraphQLNonNull, GraphQLScalarType
+
+
+# The compiler's supported GraphQL types for query arguments. The GraphQL type of a query argument
+# is the type of the field that the argument references. Not to be confused with the GraphQLArgument
+# class in the GraphQL core library.
+GraphQLQueryArgumentType = Union[
```
pmantica1, line 10: I think that it would make sense to move `QueryStringWithParameters` and `ASTWithParameters` from global_utils.py to here in a subsequent diff. Thoughts?

graphql_compiler/tests/test_end_to_end.py
```diff
@@ -231,7 +232,7 @@ def test_argument_types(self) -> None:
                     validate_argument_type(arbitrary_argument_name, graphql_type, invalid_value)
 
     def test_non_null_types_pass_validation(self) -> None:
-        type_and_value = [
+        type_and_value: List[Tuple[GraphQLQueryArgumentType, Any]] = [
```
obi1kenobi, line 235: Should that `Any` be `GraphQLQueryArgumentType` instead?

pmantica1: No it should not. I verified that you get a linting error if you try to do that. 

`Any` refers to the type of value, e.g `"abc"`, `["a", "b", "c"]`. The values in this test case are all either strings or list of strings, but I decided to use the `Any` annotation here because it matches the `value` type annotation of `validate_argument_type`.

obi1kenobi: Ah I see, I misunderstood what `GraphQLQueryArgumentType` is meant to be — how about renaming to `QueryArgumentGraphQLType` instead? `GraphQLQueryArgumentType` sounds like the "type which GraphQL query arguments are", and not the "GraphQL type of query arguments".

Also, thoughts on making a type alias for where this `Any` sits? It may or may not be a good idea, just curious what you think.

pmantica1: Sounds great. 

Regarding your second point, I think that it would be a good idea to have a `QueryArgumentType` to indicate what are the allowed python types for queries. I think that it would make it clearer what the allowed python types are. (It isn't obvious that we use `datetime` for `GraphQLDatetime` arguments instead of isoformatted strings).

However, since we mainly use `Any` as the `QueryArgumentType`. I got a bunch of mypy errors when I tried to change it something more restrictive.

So I think that for this PR, I can write a basic type alias like `QueryArgumentType = Any` which I don't think has that much value or I can make the several times longer by writing `QueryArgumentType = Union[bool, ...]`.  

So I think that it's best if I open an issue for this.

PR #808, opened 2020-04-28
Simplify emit_sql by trimming state

bojanserafimov:


graphql_compiler/compiler/emit_sql.py
```diff
@@ -994,7 +990,7 @@ def fold(self, fold_scope_location):
     def unfold(self):
         """Complete the execution of a Fold Block."""
         # 1. Relocate to outside of the fold.
-        self._relocate(self._outside_fold_location)
+        self._relocate(self._current_location.base_location)
```
chewselene, line 993: Might be nice to add a comment clarifying why the `_current_location`'s `base_location` is the location outside of the fold.

bojanserafimov: Isn't that the definition of `base_location`? What kind of comment would help?

chewselene: Oh yeah, was thinking that it would be updated during traversals, but that just updates the `fold_path` not the `base_location`

PR #811, opened 2020-04-29
Copy deserialization information over to deserialize_multiple_json_arguments

pmantica1:
Currently, one can't understand `deserialize_multiple_json_arguments`'s docstring without having to look at `deserialize_json_argument`'s docstring.

This means that the documentation that pops up when is using an IDE or text editor and hovers over the function's definition is way less useful. 

So in this PR, I change it so that we copy instead of reference the documentation. In general, I dislike copying documentation multiple times, but in this case I think that it's worth it.

An additional goal of this PR is to make this one shorter:
https://github.com/kensho-technologies/graphql-compiler/pull/804/files

### DIFF
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -119,7 +119,7 @@ def deserialize_json_argument(
     expected_type: Union[GraphQLNonNull[GraphQLScalarType], GraphQLScalarType],
     value: Any,
 ) -> Any:
-    """Deserialize a GraphQL argument parsed from a json file.
+    """Deserialize a json serialized GraphQL argument.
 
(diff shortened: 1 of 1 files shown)


graphql_compiler/query_formatting/common.py
```diff
@@ -119,7 +119,7 @@ def deserialize_json_argument(
     expected_type: Union[GraphQLNonNull[GraphQLScalarType], GraphQLScalarType],
     value: Any,
 ) -> Any:
-    """Deserialize a GraphQL argument parsed from a json file.
+    """Deserialize a json serialized GraphQL argument.
```
pmantica1, line 122: There is no reason that the argument has to be parsed from a file.

PR #813, opened 2020-04-29
Refactor SQLFoldObject

bojanserafimov:
The goal is to remove any reference to the `CompilationState` class in `SQLFoldObject`, rename it to `FoldSubqueryBuilder`, write a more tight specification for it, and write dedicated tests for it.

In this PR I move the code that joins the produced fold subquery to the main query into the `CompilationState` class so that `FoldSubqueryBuilder` would just need to do what it does (make the subquery) and not be concerned with how its results are used.

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -996,9 +996,7 @@ class CompilationState(object):
             fold_subquery_alias,
             onclause=(
                 self._current_alias.c[outer_vertex_primary_key_name]
-                == fold_subquery_alias.c[
-                    outer_vertex_primary_key_name
-                ]
+                == fold_subquery_alias.c[outer_vertex_primary_key_name]
             ),
             isouter=False,
         )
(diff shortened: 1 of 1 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -278,6 +278,7 @@ def _get_xml_path_clause(output_column, predicate_expression):
 )
 
 
+# TODO(bojanserafimov): Rename to FoldSubqueryBuilder and simplify usage and spec.
```
bojanserafimov, line 281: @chewselene I'll get to this when you give me the green light. Not excited for a massive merge-conflict :)

chewselene: Good idea! Should be fine to do once #814 is merged

graphql_compiler/compiler/emit_sql.py
```diff
@@ -998,17 +981,27 @@ def unfold(self):
 
         # 2. End the fold, collecting the folded subquery, the new from clause for the main
         # selectable, and the location of the folded outputs.
```
chewselene, line 983: Steps 2 and 3 of unfold should get a comment update. And possibly there should be an additional step between 3 and 4 for joining the fold subquery to the main selectable.

PR #814, opened 2020-04-29
Update emit_sql with typehints, fstrings, NamedTuples

chewselene:
This PR:
- adds typehints (and more error checking to make mypy happy)
- removes most instances of string `format` and replaces with fstrings
- updates a `namedtuple` to a `NamedTuple`
- updates some punctuation and grammar of comments

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -740,7 +740,7 @@ class CompilationState(object):
         self._outputs: List[Label] = []  # SQLAlchemy Columns labelled correctly for output.
         self._filters: List[
             BinaryExpression
-        ] = []  # SQLAlchemy Expressions to be used in the where clause.
+        ] = []  # SQLAlchemy Expressions to be used in the WHERE clause.
 
         # Generates aliases for fold subqueries.
         self._alias_generator: UniqueAliasGenerator = UniqueAliasGenerator()
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -528,7 +528,9 @@ class SQLFoldObject(object):
         # Sort to make select order deterministic.
         return sorted(self._outputs, key=lambda column: column.name, reverse=True)
 
-    def visit_vertex(self, join_descriptor, from_table, to_table, current_fold_scope_location, all_folded_fields):
+    def visit_vertex(
+        self, join_descriptor, from_table, to_table, current_fold_scope_location, all_folded_fields
+    ):
         """Add a new traversal descriptor and add outputs, if visiting an output vertex."""
         if self._ended:
             raise AssertionError(
@@ -680,7 +682,7 @@ class CompilationState(object):
         if self._current_fold is not None and isinstance(new_location, FoldScopeLocation):
             alias_key = (
                 self._current_location.base_location.query_path,
-                self._current_location.fold_path
+                self._current_location.fold_path,
             )
         # Regardless of whether self._current_fold is None or not, if the new location is a
         # Location, relocation should occur based on a Location alias key.
@@ -760,9 +762,17 @@ class CompilationState(object):
         edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
         self._relocate(self._current_location.navigate_to_subpath(vertex_field))
         if self._current_fold is not None:
-            self._current_fold.visit_vertex(edge, previous_alias, self._current_alias, self._current_location, self._all_folded_fields)
+            self._current_fold.visit_vertex(
+                edge,
+                previous_alias,
+                self._current_alias,
+                self._current_location,
+                self._all_folded_fields,
+            )
         else:
-            self._join_to_parent_location(previous_alias, edge.from_column, edge.to_column, optional)
+            self._join_to_parent_location(
+                previous_alias, edge.from_column, edge.to_column, optional
+            )
 
     def _get_current_primary_key_name(self, directive_name: str) -> str:
         """Return the name of the single-column primary key at the current location.
@@ -906,8 +916,13 @@ class CompilationState(object):
 
         # 4. Relocate to inside the fold scope and visit the first vertex.
         self._relocate(fold_scope_location)
-        self._current_fold.visit_vertex(join_descriptor, outer_alias, self._current_alias,
-                                        fold_scope_location, self._all_folded_fields)
+        self._current_fold.visit_vertex(
+            join_descriptor,
+            outer_alias,
+            self._current_alias,
+            fold_scope_location,
+            self._all_folded_fields,
+        )
 
     def unfold(self):
         """Complete the execution of a Fold Block."""
@@ -938,10 +953,7 @@ class CompilationState(object):
         # If the current location is the beginning of a fold, the current alias
         # will eventually be replaced by the resulting fold subquery during Unfold.
         self._aliases[
-            (
-                self._current_location.base_location.query_path,
-                self._current_location.fold_path,
-            )
+            (self._current_location.base_location.query_path, self._current_location.fold_path,)
             if self._current_fold is not None
             else (self._current_location.query_path, None)
         ] = self._current_alias
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -2,7 +2,7 @@
 """Transform a SqlNode tree into an executable SQLAlchemy query."""
 from collections import namedtuple
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Set
+from typing import Dict, List, Optional, Set, Tuple, Union
 
 import six
 import sqlalchemy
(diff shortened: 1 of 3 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -259,23 +271,19 @@ def _get_xml_path_clause(output_column, predicate_expression):
...
-        "join_descriptor",
-        # SQLAlchemy table corresponding to corresponding to the outside vertex of the traversal,
-        # appears on the left side of the join.
-        "from_table",
-        # SQLAlchemy table corresponding to corresponding to the inside vertex of the traversal,
-        # appears on the right side of the join.
-        "to_table",
-    ),
-)
+class SQLFoldTraversalDescriptor(NamedTuple):
```
obi1kenobi, line 274: (this is fine) In principle, we could and probably should move away over time from `NamedTuple` and toward `dataclass`, since dataclasses are more dev-friendly, with essentially everything `NamedTuple` has + automatic repr methods etc.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -338,52 +346,55 @@ class SQLFoldObject(object):
...
+        # List of SQLFoldTraversalDescriptors describing each traversal in the fold
         # starting with the join from the vertex immediately outside the fold to the folded vertex.
-        self._traversal_descriptors = []
-        self._outputs = []  # Output columns for folded subquery.
-        self._filters = []  # SQLAlchemy Expressions to be used in the where clause.
+        self._traversal_descriptors: List[SQLFoldTraversalDescriptor] = []
+        self._outputs: List[Label] = []  # Output columns for folded subquery.
+        self._filters: List[
+            BinaryExpression
+        ] = []  # SQLAlchemy Expressions to be used in the where clause.
```
obi1kenobi, line 374:
```suggestion
        ] = []  # SQLAlchemy Expressions to be used in the WHERE clause.
```

obi1kenobi: Highlighting that we're referring to the SQL keyword and not just the regular English word.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -440,18 +451,28 @@ def _construct_fold_subquery(self, subquery_from_clause):
...
 
-    def _get_array_agg_column(self, intermediate_fold_output_name, fold_output_field):
+    def _get_array_agg_column(
+        self, intermediate_fold_output_name: str, fold_output_field: str
+    ) -> Label:
         """Select an array_agg of the fold output field, labeled as requested."""
+        if self._output_vertex_alias is None:
+            raise AssertionError(
+                "Attempted to perform array aggregation while the "
+                "_output_vertex_alias was set to None."
```
obi1kenobi, line 464: In situations like this, I'd suggest adding as much of the current state into the error message as possible, to ease reproducing and debugging the issue. That would include things like function arguments and relevant `self` attributes. Future-us will be grateful to you for saving them a painful and frustrating debugging step :)

graphql_compiler/compiler/emit_sql.py
```diff
@@ -477,58 +498,64 @@ def _get_mssql_xml_path_column(
...
                             used in subquery.
 
         Returns:
             Selectable for XML PATH aggregation subquery
         """
-        # use join info tuple for most recent traversal to set WHERE clause for XML PATH subquery
+        if self._output_vertex_alias is None:
+            raise AssertionError(
+                "Attempted to aggregate with XML PATH before an _output_vertex_alias had been "
+                "found."
```
obi1kenobi, line 512: Same as above, I'd include as much info as possible about what happened here.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -477,58 +498,64 @@ def _get_mssql_xml_path_column(
...
-    def _get_fold_output_column_clause(self, fold_output_field):
+    def _get_fold_output_column_clause(self, fold_output_field: str) -> Label:
         """Get the SQLAlchemy column expression corresponding to the fold output field."""
         if fold_output_field == COUNT_META_FIELD_NAME:
             return sqlalchemy.func.coalesce(
                 sqlalchemy.func.count(), sqlalchemy.literal_column("0")
             ).label(FOLD_OUTPUT_FORMAT_STRING.format(COUNT_META_FIELD_NAME))
         else:
-            # force column to have explicit label as opposed to anon_label
+            # Force column to have explicit label as opposed to anon_label.
```
obi1kenobi, line 529: Is `anon_label` defined somewhere? I'm not familiar with it. If not, could we make this say `an anonymous label` assuming that's what it's referring to?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -477,58 +498,64 @@ def _get_mssql_xml_path_column(
...
             else:
                 raise NotImplementedError(
                     "Fold only supported for MSSQL and PostgreSQL, "
-                    "dialect set to {}".format(self._dialect.name)
+                    f"dialect set to {self._dialect.name}."
                 )
 
-        # We should have either triggered a not implemented error, or returned earlier
         raise AssertionError(
             "Reached end of function without returning a value, this code should be unreachable."
```
obi1kenobi, line 550: I know this isn't part of the review, so feel free to ignore — just wanted to mention another place where outputting more info would make debugging the resulting error much less frustrating.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -590,21 +629,25 @@ def add_filter(self, predicate, aliases):
         sql_expression = predicate.to_sql(self._dialect, aliases, self._output_vertex_alias)
         self._filters.append(sql_expression)
 
-    def end_fold(self, alias_generator, from_clause, outer_from_table):
+    def end_fold(
+        self,
+        alias_generator: "UniqueAliasGenerator",
```
obi1kenobi, line 634: Would it be feasible to move the definition of `UniqueAliasGenerator` above this class, to avoid needing this forward reference? Happy to wait for a separate PR if so, no need to make this one more complex.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -590,21 +629,25 @@ def add_filter(self, predicate, aliases):
...
+        if self._output_vertex_alias is None or self._output_vertex_location is None:
             raise AssertionError(
-                "No output vertex visited. Invalid state encountered during fold {}".format(self)
+                f"No output vertex visited. Invalid state encountered during fold {self}."
             )
         if len(self._traversal_descriptors) == 0:
             raise AssertionError(
-                "No traversed vertices visited. "
-                "Invalid state encountered during fold {}".format(self)
+                "No traversed vertices visited. " f"Invalid state encountered during fold {self}."
```
obi1kenobi, line 650: This seems like a `black`-ism :) Perhaps merge the two strings into one f-string?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -732,8 +796,13 @@ def _relocate(self, new_location):
                 self._current_classname
             ].alias()
 
-    def _join_to_parent_location(self, parent_alias, from_column, to_column, optional):
+    def _join_to_parent_location(
+        self, parent_alias: Alias, from_column: str, to_column: str, optional: bool
+    ):
         """Join the current location to the parent location using the column names specified."""
+        if self._current_alias is None:
+            raise AssertionError("Cannot join to parent location when _current_alias is None.")
```
obi1kenobi, line 804: One more error where more info would be helpful.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -758,34 +827,38 @@ def _join_to_parent_location(self, parent_alias, from_column, to_column, optiona
...
 
-    def backtrack(self, previous_location):
+    def backtrack(self, previous_location: BaseLocation) -> None:
         """Execute a Backtrack Block."""
         self._relocate(previous_location)
 
     def traverse(self, vertex_field: str, optional: bool) -> None:
         """Execute a Traverse Block."""
+        if self._current_location is None:
+            raise AssertionError("Cannot traverse when the _current_location is None.")
```
obi1kenobi, line 860: Another constant error message, more info would be helpful if we ever hit it.

graphql_compiler/compiler/emit_sql.py
bojanserafimov: @obi1kenobi see if you agree with intent of this upcoming PR https://github.com/kensho-technologies/graphql-compiler/pull/813 since it makes this irrelevant

graphql_compiler/compiler/emit_sql.py
```diff
@@ -673,50 +716,71 @@ class ColumnRouter:
...
         # Dict mapping (some_location.query_path, fold_scope_location.fold_path) tuples to
-        # corresponding table _Aliases. some_location is either self._current_location
-        # or the base location of an open FoldScopeLocation.
+        # corresponding table Aliases. some_location is either self._current_location
+        # or the base location of an open FoldScopeLocation. For Locations, the second argument of
+        # the tuple will be None.
         # Note: for tables with an _x_count column, that column will always
         # be named "fold_output__x_count".
-        self._aliases = {}
+        self._aliases: Dict[Tuple[QueryPath, Optional[FoldPath]], Alias] = {}
```
bojanserafimov, line 752: This is a bit more complicated. The value is sometimes a `ColumnRouter`. I'm thinking about making a common base class for both, but we could leave it untyped or have a union type meanwhile.

graphql_compiler/compiler/emit_sql.py
obi1kenobi: Sounds good to me, happy with #813 and also generally happy to delegate such decisions to you two.

graphql_compiler/compiler/emit_sql.py
chewselene: made it a union for now

PR #818, opened 2020-05-01
Raise NotImplementedError for MSSQL _x_count and multiple outputs within the same fold scope

chewselene:
The current implementation of `_x_count` does not actually count the number of results returned with the fold. Using `COUNT(*)` outside the `XML PATH` statement causes the count to always be 1 for the 1 string result returned through `XML PATH` aggregation. 

The current implementation for multiple outputs within a fold produces an individual `XML PATH` statement per output. This does not guarantee that the outputs from within the same fold will be returned in the same order.

Both these things should raise `NotImplementedError` for now, but should be implemented in the future.

### DIFF
--- a/graphql_compiler/tests/integration_tests/test_backends_integration.py
+++ b/graphql_compiler/tests/integration_tests/test_backends_integration.py
@@ -436,9 +436,12 @@ class IntegrationTests(TestCase):
     @use_all_backends(except_backends=(test_backend.REDISGRAPH,))  # Not implemented yet
     @integration_fixtures
     def test_fold_basic(self, backend_name: str) -> None:
-        # (query, args, expected_results) tuples.
-        # The queries are ran in the order specified here.
-        queries: List[Tuple[str, Dict[str, Any], List[Dict[str, Any]]]] = [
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -436,9 +436,12 @@ def test_recurse(self) -> None:
     @use_all_backends(except_backends=(test_backend.REDISGRAPH,))  # Not implemented yet
     @integration_fixtures
     def test_fold_basic(self, backend_name: str) -> None:
-        # (query, args, expected_results) tuples.
-        # The queries are ran in the order specified here.
-        queries: List[Tuple[str, Dict[str, Any], List[Dict[str, Any]]]] = [
+        # (query, args, expected_results, excluded_backends) tuples.
+        # Note: excluded_backends is distinct from `@use_all_backends(expect_backends=(...)) because
+        # some backends such as MSSQL haven most, but not all, fold functionality implemented.
```
bojanserafimov, line 441:
```suggestion
        # some backends such as MSSQL have most, but not all, fold functionality implemented.
```

PR #820, opened 2020-05-06
Show all tables missing primary keys

### LINKED ISSUE #650: Schema autogeneration from SQL complains about missing primary keys in tables one at a time
Related to #649. Assume multiple tables in a database are missing primary keys, and the user is attempting to autogenerate a schema from that database.

The `MissingPrimaryKeyError` that will get raised will contain the information of only a single table that is missing a primary key. That means that the user will end up re-running the autogeneration process N times to discover N tables with missing primary keys, which is incredibly inconvenient.

The error message should contain all the problematic tables with missing primary keys, so the user can avoid all of them in one go.

LWprogramming:
Used for schema autogeneration from SQL. Resolves #650

### DIFF
--- a/graphql_compiler/schema_generation/sqlalchemy/utils.py
+++ b/graphql_compiler/schema_generation/sqlalchemy/utils.py
@@ -4,15 +4,22 @@ from ..exceptions import MissingPrimaryKeyError
 
 def validate_that_tables_have_primary_keys(tables):
     """Validate that each SQLAlchemy Table object has a primary key."""
+    tables_missing_primary_keys = set()
     for table in tables:
         if not table.primary_key:
-            raise MissingPrimaryKeyError(
-                "SQLAlchemy Table with name {} and schema {} is missing a "
-                "primary key. Note that the primary keys in SQLAlchemy "
-                "Table objects do not have to match the primary keys in "
-                "the underlying row. They must simple be unique and "
-                "non-null identifiers of each row.".format(table.name, table.schema)
-            )
+            tables_missing_primary_keys.add(table)
+    if tables_missing_primary_keys:
+        error_message = (
+            "At least one SQLAlchemy Table is missing a "
+            "primary key. Note that the primary keys in SQLAlchemy "
+            "Table objects do not have to match the primary keys in "
+            "the underlying row. They must simply be unique and "
+            "non-null identifiers of each row. The tables missing primary "
+            "keys have names and schemas as follows: "
+        )
+        for table in tables_missing_primary_keys:
+            error_message += "name: {} schema: {} ".format(table.name, table.schema)
+        raise MissingPrimaryKeyError(error_message)
 
 
 def validate_that_tables_belong_to_the_same_metadata_object(tables):
--- a/graphql_compiler/schema_generation/sqlalchemy/utils.py
+++ b/graphql_compiler/schema_generation/sqlalchemy/utils.py
@@ -1,8 +1,9 @@
 # Copyright 2019-present Kensho Technologies, LLC.
-from typing import Set, List, Iterable
+from typing import Iterable, List, Set
+
+from sqlalchemy import Table
 
 from ..exceptions import MissingPrimaryKeyError
-from sqlalchemy import Table
 
 
 def validate_that_tables_have_primary_keys(tables: Iterable[Table]):
@@ -20,7 +21,10 @@ def validate_that_tables_have_primary_keys(tables: Iterable[Table]):
             "non-null identifiers of each row. The tables missing primary "
             "keys have names and schemas as follows: "
         )
-        faulty_tables: List[str] = ["name: {} schema: {} ".format(table.name, table.schema) for table in tables_missing_primary_keys]
+        faulty_tables: List[str] = [
+            "name: {} schema: {} ".format(table.name, table.schema)
+            for table in tables_missing_primary_keys
+        ]
         error_message += " ".join(faulty_tables)
         raise MissingPrimaryKeyError(error_message)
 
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_generation/sqlalchemy/utils.py
```diff
@@ -4,15 +4,22 @@
 
 def validate_that_tables_have_primary_keys(tables):
     """Validate that each SQLAlchemy Table object has a primary key."""
+    tables_missing_primary_keys = set()
```
obi1kenobi, line 7: We are currently in the process of type-hinting the entire compiler codebase, so our general rule of thumb is that if you edit a function, since you already have to understand what it does, you should also add type hints to it — including arguments, return values, and any intermediate values. For example, `tables_missing_primary_keys = set()` could be `tables_missing_primary_keys: Set[Table] = set()` or something along those lines.

graphql_compiler/schema_generation/sqlalchemy/utils.py
```diff
@@ -4,15 +4,22 @@
...
+        error_message = (
+            "At least one SQLAlchemy Table is missing a "
+            "primary key. Note that the primary keys in SQLAlchemy "
+            "Table objects do not have to match the primary keys in "
+            "the underlying row. They must simply be unique and "
+            "non-null identifiers of each row. The tables missing primary "
+            "keys have names and schemas as follows: "
+        )
+        for table in tables_missing_primary_keys:
+            error_message += "name: {} schema: {} ".format(table.name, table.schema)
```
obi1kenobi, line 21: Use `" ".join(<stuff>)` instead of `error_message +=` — the `+=` approach ends up being the Shlemiel the Painter's algorithm: https://www.joelonsoftware.com/2001/12/11/back-to-basics/

graphql_compiler/tests/schema_generation_tests/test_sqlalchemy_schema_generation.py
```diff
@@ -308,3 +308,21 @@ def test_missing_primary_key(self):
         faulty_vertex_name_to_table = {table_without_primary_key.name: table_without_primary_key}
         with self.assertRaises(MissingPrimaryKeyError):
             get_sqlalchemy_schema_info(faulty_vertex_name_to_table, {}, dialect())
+
+    def test_missing_multiply_primary_keys(self):
```
pmantica1, line 312:
```suggestion
    def test_missing_multiple_primary_keys(self):
```

graphql_compiler/schema_generation/sqlalchemy/utils.py
```diff
@@ -1,18 +1,32 @@
...
+        error_message: str = (
+            "At least one SQLAlchemy Table is missing a "
+            "primary key. Note that the primary keys in SQLAlchemy "
+            "Table objects do not have to match the primary keys in "
+            "the underlying row. They must simply be unique and "
+            "non-null identifiers of each row. The tables missing primary "
+            "keys have names and schemas as follows: "
+        )
+        faulty_tables: List[str] = [
+            "name: {} schema: {} ".format(table.name, table.schema)
```
pmantica1, line 25: Combined with the " ".join() below this would lead to a double space.

graphql_compiler/schema_generation/sqlalchemy/utils.py
```diff
@@ -1,18 +1,32 @@
 # Copyright 2019-present Kensho Technologies, LLC.
+from typing import Iterable, List, Set
+
+from sqlalchemy import Table
+
 from ..exceptions import MissingPrimaryKeyError
 
 
-def validate_that_tables_have_primary_keys(tables):
+def validate_that_tables_have_primary_keys(tables: Iterable[Table]):
```
pmantica1, line 9: Missing return type annotation: 
```suggestion
def validate_that_tables_have_primary_keys(tables: Iterable[Table]) -> None:
```

graphql_compiler/schema_generation/sqlalchemy/utils.py
```diff
@@ -1,18 +1,32 @@
...
     """Validate that each SQLAlchemy Table object has a primary key."""
+    tables_missing_primary_keys: Set[Table] = set()
```
pmantica1, line 11: It doesn't hurt and now that you added you don't have to remove them, but you don't need to add type annotations to all variables within a function. I think that our convention is to add type annotations in:
1. Function signatures. 
2. Global variables

Mypy might also sometimes require additional type annotations to function properly, (you'll know when this is the case because mypy will raise some sort of error or warning).

pmantica1: Suggestion instead of making this a set of tables, make this a set of strings.

Then write `table_missing_primary_keys.add(table.fullname)` below. The fullname includes both the name and the schema information. 

Then at the bottom you can write simply write:
`raising MiisingPrimaryKeyError(
... 
f"Tables missing primary keys: {tables_missing_primary_keys}"
)`

Notice that I used an f-string. This is a new python feature that we are starting to use as a replacement for `format()`. 
https://realpython.com/python-f-strings/

I won't block your review if you decide to use `format()` instead, but f-strings are more convenient to use so I think it'd be a good idea if you got acquainted with f-strings.

obi1kenobi: +1 to @pmantica1's points. This is fine, working code, but it is slightly unidiomatic: we're targeting Python 3.6+ and not using the best available features in the target language. Let's go for one more round of polish since this is code that our users are likely to read — it raises a non-obvious user-facing error, for which it's not unreasonable to dig into the source.

LWprogramming: It turns out that this particular annotation would be good (to clarify that it's a set of strings in particular rather than a set of some unknown type), but that's a good point in general-- thanks for the heads-up!

The f-string thing is really handy, I've updated it.

PR #821, opened 2020-05-06
MSSQL fold postprocessing tests for all scalar types

chewselene:
Adding MSSQL fold postprocessing tests for all builtin graphQL scalar types.  

Fold postprocessing relies on converting from an XML PATH aggregated string. However, the GraphQLType `parse_value` function does not take in string representations for all types. For example, `"1"` would not be parsed to `1` using GraphQLInt's `parse_value`. 

Our json deserialization code already handled most of these types of conversions. I have refactored the json deserialization to be more generic, and have allowed more representations for booleans (such as `0`, `"True"`, `"true"`, and `"0"`).

### DIFF
--- a/graphql_compiler/deserialization.py
+++ b/graphql_compiler/deserialization.py
@@ -1,7 +1,7 @@
 # Copyright 2020-present Kensho Technologies, LLC.
 """Convert values to their underlying GraphQLType."""
 from types import MappingProxyType
-from typing import Any, Callable, Dict, Mapping, Tuple, Type
+from typing import Any, Callable, Mapping, Tuple, Type
 
 from graphql import (
     GraphQLBoolean,
@@ -13,16 +13,15 @@ from graphql import (
     GraphQLString,
 )
 
-from graphql_compiler import GraphQLInvalidArgumentError
 from graphql_compiler.compiler.helpers import strip_non_null_from_type
 from graphql_compiler.global_utils import is_same_type
-from graphql_compiler.typedefs import QueryArgumentGraphQLType
 
 from .global_utils import assert_set_equality
 from .schema import SUPPORTED_SCALAR_TYPES, GraphQLDate, GraphQLDateTime, GraphQLDecimal
+from .typedefs import QueryArgumentGraphQLType
 
 
-_ALLOWED_JSON_SCALAR_TYPES: Mapping[str, Tuple[Type, ...]] = MappingProxyType(
+_ALLOWED_SCALAR_TYPES: Mapping[str, Tuple[Type, ...]] = MappingProxyType(
     {
         GraphQLDate.name: (str,),
         GraphQLDateTime.name: (str,),
@@ -35,7 +34,7 @@ _ALLOWED_JSON_SCALAR_TYPES: Mapping[str, Tuple[Type, ...]] = MappingProxyType(
     }
 )
 assert_set_equality(
-    set(_ALLOWED_JSON_SCALAR_TYPES.keys()),
+    set(_ALLOWED_SCALAR_TYPES.keys()),
     {graphql_type.name for graphql_type in SUPPORTED_SCALAR_TYPES},
 )
 
@@ -50,8 +49,8 @@ def _custom_boolean_deserialization(value: Any) -> bool:
         return False
     else:
         raise ValueError(
-            f"Received unexpected GraphQLBoolean value {value} ({type(value)}). Expected one "
-            f"of the following {true_values + false_values}."
+            f"Received unexpected GraphQLBoolean value {value} of type {type(value)}. Expected one "
+            f"of: {true_values + false_values}."
         )
 
 
@@ -75,7 +74,7 @@ _JSON_TYPES_AND_DESERIALIZATION_FUNCTIONS: Mapping[
 ] = MappingProxyType(
     {
         scalar_type.name: (
-            _ALLOWED_JSON_SCALAR_TYPES[scalar_type.name],
+            _ALLOWED_SCALAR_TYPES[scalar_type.name],
             _CUSTOM_SCALAR_DESERIALIZATION_FUNCTIONS.get(scalar_type.name, scalar_type.parse_value),
         )
         for scalar_type in SUPPORTED_SCALAR_TYPES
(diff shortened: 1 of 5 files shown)


chewselene:
@pmantica1 made a great point that it didn't make sense to be importing from `query_formatting`. Since converting values to `GraphQLType`s is something that is useful for both argument deserialization and MSSQL fold postprocessing, I've split that functionality into its own file (currently called `deserialization.py`, but open to better names). Hopefully this new structure makes things a bit clearer.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+        scalar_type.name: (
+            _ALLOWED_JSON_SCALAR_TYPES[scalar_type.name],
+            _CUSTOM_SCALAR_DESERIALIZATION_FUNCTIONS.get(scalar_type.name, scalar_type.parse_value),
+        )
+        for scalar_type in SUPPORTED_SCALAR_TYPES
+    }
+)
+
+
+def convert_scalar_value_to_graphql_type(expected_type: GraphQLScalarType, value: Any) -> Any:
```
pmantica1, line 86: How about `deserialize_scalar_value` instead?  It is shorter and `convert_scalar_value_to_graphql_type` is not really accurate since this function does not return a GraphQL type.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+            GraphQLID: str
+
+    Raises:
+        TypeError: if the value was not of the expected type.
+    """
+    types_and_deserialization = _JSON_TYPES_AND_DESERIALIZATION_FUNCTIONS.get(expected_type.name)
+    if types_and_deserialization is None:
+        raise TypeError(
+            f"Unexpected GraphQLType {expected_type}. No deserialization function known."
+        )
```
pmantica1, line 111: I would raise an `AssertionError` here since this is error is due because we are attempting to deserialize the value of a `GraphQLScalarType` that is not one of the `GraphQLScalarTypes` that is mentioned in the docstring. It is also more indicative of a compiler/server side bug instead than an user error.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+            GraphQLString: str
+            GraphQLBoolean: bool
+            GraphQLID: str
+
+    Raises:
+        GraphQLInvalidArgumentError: if the argument value was not of the expected type.
+    """
+    # Explicitly disallow passing boolean values for non-boolean types.
+    if isinstance(value, bool) and not is_same_type(GraphQLBoolean, expected_type):
+        raise GraphQLInvalidArgumentError("")
```
pmantica1, line 144: I would move this code to the function above and add an error message.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+        TypeError: if the value was not of the expected type.
+    """
+    types_and_deserialization = _JSON_TYPES_AND_DESERIALIZATION_FUNCTIONS.get(expected_type.name)
+    if types_and_deserialization is None:
+        raise TypeError(
+            f"Unexpected GraphQLType {expected_type}. No deserialization function known."
+        )
+    expected_python_types, deserialization_function = types_and_deserialization
+    if not isinstance(value, expected_python_types):
+        raise TypeError(
```
pmantica1, line 114: I think that it makes more sense to raise a value error here to be consistent with the `.parse_value` functions of the `GraphQLScalarType` objects. They usually raise a `ValueError` if the input is of an invalid type or value.

Also, please mention that why we choose to raise a `ValueError` in the docstring. Maybe write the following:
```
ValueError: if the value is not appropriate for the type. ValueError is chosen because
            it is already the base case of exceptions raised by the GraphQL parsers.
``` 
(This is almost exactly the same comment Bojan wrote before. I added some indentation).

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+            GraphQLInt: int
+            GraphQLString: str
+            GraphQLBoolean: bool
+            GraphQLID: str
+            GraphQLList: list of the inner type
+
+    Raises:
+        GraphQLInvalidArgumentError: if any of the argument values was not of the expected type.
+    """
+    return {
```
pmantica1, line 250: Can you re-add the `ensure_arguments_are_provided` functionality here?

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+            GraphQLBoolean: bool
+            GraphQLID: str
+            GraphQLList: list of the inner type
+
+    Raises:
+        GraphQLInvalidArgumentError: if the argument value was not of the expected type.
+    """
+    stripped_type = strip_non_null_from_type(expected_type)
+    if isinstance(stripped_type, GraphQLList):
+        if not isinstance(value, list):
```
pmantica1, line 193: What do you think of moving this logic "downward" and creating a more generic `convert_value_to_graphql_type` function and expose that function as public API instead of `convert_scalar_value_to_graphql_type`? (If you follow my suggestion above this would be `deserialize_value`).

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+        )
+    expected_python_types, deserialization_function = types_and_deserialization
+    if not isinstance(value, expected_python_types):
+        raise TypeError(
+            f"{value} ({type(value)} cannot be deserialized to GraphQL type {expected_type}."
+        )
+    return deserialization_function(value)
+
+
+def deserialize_scalar_argument(name: str, expected_type: GraphQLScalarType, value: Any) -> Any:
```
pmantica1, line 120: Lets keep this function as private API: "_deserialize_scalar_argument".

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+        GraphQLFloat: 4.3, "5.0", 5
+        GraphQLDecimal: "5.00000000000000000000000000001"
+        GraphQLInt: 4, "3803330000000000000000000000000000000000000000000"
+        GraphQLString: "Hello"
+        GraphQLBoolean: True, 1, "1", "True", "true"
+        GraphQLID: "13d72846-1777-6c3a-5743-5d9ced3032ed"
+        GraphQLList(GraphQLInt): [1, 2, 3]
+
+    Args:
+        name: he name of the argument. It will be used to provide a more descriptive error
```
pmantica1, line 171:
```suggestion
        name: the name of the argument. It will be used to provide a more descriptive error
```

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+            GraphQLDateTime: datetime.datetime with tzinfo=pytz.utc
+            GraphQLFloat: float
+            GraphQLDecimal: decimal.Decimal
+            GraphQLInt: int
+            GraphQLString: str
+            GraphQLBoolean: bool
+            GraphQLID: str
+
+    Raises:
+        GraphQLInvalidArgumentError: if the argument value was not of the expected type.
```
pmantica1, line 140: 4x docstring repetition is a bit much. If you agree that we should make this function private, do you mind making this docstring a one-liner?

graphql_compiler/deserialization.py
chewselene: I opted to remove because I didn't want to import from `query_formatting`. This function was only used in 1 instance in `test_end_to_end` so I just moved the call to `ensure_arguments_are_provided` to that test. Alternatively, I could make a function in `query_formatting/common.py` called `validate_and_deserialize_multiple_arguments` or something similar that calls `ensure_arguments_are_provided` and then this function. Since the function was only used once, I didn't think it was worthwhile to create such a function, but happy to put one in if you think it would be beneficial.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+    true_values = [1, "1", "true", "True", True]
+    false_values = [0, "0", "false", "False", False]
+    if value in true_values:
+        return True
+    elif value in false_values:
+        return False
+    else:
+        raise ValueError(
+            f"Received unexpected GraphQLBoolean value {value} ({type(value)}). Expected one "
+            f"of the following {true_values + false_values}."
```
pmantica1, line 54: nit
```suggestion
            f"of: {true_values + false_values}."
```

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,253 @@
...
+    """Deserialize a boolean, allowing for common string or int representations."""
+    true_values = [1, "1", "true", "True", True]
+    false_values = [0, "0", "false", "False", False]
+    if value in true_values:
+        return True
+    elif value in false_values:
+        return False
+    else:
+        raise ValueError(
+            f"Received unexpected GraphQLBoolean value {value} ({type(value)}). Expected one "
```
pmantica1, line 53:
```suggestion
            f"Received unexpected GraphQLBoolean value {value} of type {type(value)}. Expected one "
```

graphql_compiler/deserialization.py
pmantica1: I don't think we should add a `validate_and_deserialize_multiple_arguments` since it doesn't make sense to have two functions are almost the same except for the one difference that one performs more validation. (When would someone ever call `deserialize_multiple_arguments`). Also, I think that some sort of validation is already implied whenever a function is parsing/deserializing something. 

However, I do think that it's extremely important to have `ensure_arguments_are_provided_here`.  Suppose that the user inputs additional or misspelled arguments. Then this function will raise a `KeyError` instead of a `GraphQLInvalidArgumentError`. While it is safe to regard `GraphQLInvalidArgumentError`  as being due to invalid user input, a `KeyError` will most likely generate some sort of alarm if this code is used in production. This alarm will probably waste developer productivity since this is most likely a user error and nothing can be done. 

The missing arguments case is probably worse because there will still be some sort of error/alarm raised but it won't be as obvious that it was raised because of faulty user input.

I agree though that we should probably not import this function from `query_formatting/common.py`. I think that it makes sense to move this function elsewhere. (Maybe `global_utils.py`). I know that this function is part of the public API of that module but I am not sure if we've released a new compiler version since that function was introduced. Do you mind checking? If it is part of the public API then we can still move it elsewhere and import it to `query_formatting/common.py` to avoid breaking API.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,174 @@
...
+            GraphQLDecimal: decimal.Decimal
+            GraphQLInt: int
+            GraphQLString: str
+            GraphQLBoolean: bool
+            GraphQLID: str
+
+    Raises:
+        AssertionError: if the expected_type does not have a known deserialization function.
+        ValueError: if the value is not appropriate for the type. ValueError is chosen because
+            it is already the base case of exceptions raised by the GraphQL parsers.
```
pmantica1, line 116: nit: Can you fix the indentation here so that "it" is aligned with "if"? Sorry, I think that this indentation issue comes from the code I attached to the comment I made earlier. Github code formatting was being kind of weird.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    stripped_type = strip_non_null_from_type(expected_type)
+    if isinstance(stripped_type, GraphQLList):
+        if not isinstance(value, list):
+            raise AssertionError(f"Cannot deserialize non-list value {value} to GraphQLList type.")
```
pmantica1, line 169: I think that it makes more sense to raise a ValueError here.

graphql_compiler/query_formatting/common.py
```diff
@@ -377,4 +183,23 @@ def insert_arguments_into_query(compilation_result: CompilationResult, arguments
         )
 
 
+def deserialize_argument(name: str, expected_type: QueryArgumentGraphQLType, value: Any,) -> Any:
+    """Deserialize a GraphQL argument, raising a GraphQLInvalidArgumentError if invalid."""
+    try:
+        return deserialize_value(expected_type, value)
+    except (ValueError, TypeError, AssertionError) as e:
```
pmantica1, line 190: I would not catch an `AssertionError` here. We want to raise `GraphQLInvalidArgumentError` if there is an "user error", An `AssertionError` is more indicative of a server-side/compiler bug.

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,174 @@
...
+            GraphQLDateTime: datetime.datetime with tzinfo=pytz.utc
+            GraphQLFloat: float
+            GraphQLDecimal: decimal.Decimal
+            GraphQLInt: int
+            GraphQLString: str
+            GraphQLBoolean: bool
+            GraphQLID: str
+
+    Raises:
+        AssertionError: if the expected_type does not have a known deserialization function.
```
pmantica1, line 114: nit: I would prefer if we actually didn't document assertion errors. I don't think documenting them does much harm here, but I don't want to encourage this practice in the compiler.

http://google.github.io/styleguide/pyguide.html#doc-function-raises

graphql_compiler/deserialization.py
```diff
@@ -0,0 +1,174 @@
...
+            GraphQLInt: int
+            GraphQLString: str
+            GraphQLBoolean: bool
+            GraphQLID: str
+            GraphQLList: list of the inner type
+
+    Raises:
+        AssertionError: if the expected_type does not have a known deserialization function.
+        ValueError: if the value is not appropriate for the type. ValueError is chosen because
+            it is already the base case of exceptions raised by the GraphQL parsers.
```
pmantica1, line 164: Same comments as above.

PR #822, opened 2020-05-11
Fix non-determinism in cost estimation coverage

bojanserafimov:
Our cost estimation is deterministic, but it's coverage is not, since a commutative function is being called with undetermined order of arguments.

TODO: Fix all similar issues in this file.


bojanserafimov:
Not worth for now.

PR #827, opened 2020-05-13
Make GraphQLDateTime tz naive

### LINKED ISSUE #715: Validate that OrientDB datetime arguments are tz aware
We currently allow tz naive datetime arguments for OrientDB, but OrientDB will throw a very cryptic error if we pass in tz naive datetime arguments. So we should have explicit validation in the compiler requiring that OrientDB datetime arguments are tz aware.

### LINKED ISSUE #745: Tz-aware inputs for tz-naive fields are allowed

pmantica1:
Change the `GraphQLDateTime` scalar type from being timezone-aware aware to being timezone-naive to follow the usual database convention of naming the timezone-naive type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive types.

Resolves #715 by disallowing timezone aware arguments.
Resolves #745

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,9 +7,9 @@ Current development version
 v2.0.0
 ------
 
-- **BREAKING** Change the GraphQLDateTime scalar type from being timezone-aware aware to being
-  timezone-naive. This change will break MATCH and gremlin queries filtering on or outputting
-  datetime fields.
+- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
+  being timezone-naive in order to follow the usual database convention of naming the timezone-naive
+  type "datetime". This change breaks OrientDB queries filtering on or outputting datetime fields.
 
 v1.11.0
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,8 +4,8 @@ Changelog
 Current development version
 ---------------------------
 
-v2.0.0
-------
+v2.0.0 (upcoming release)
+-------------------------
 
 - **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
   being timezone-naive to follow the usual database convention of naming the timezone-naive
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,8 +8,9 @@ v2.0.0
 ------
 
 - **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
-  being timezone-naive in order to follow the usual database convention of naming the timezone-naive
-  type "datetime". This change breaks OrientDB queries filtering on or outputting datetime fields.
+  being timezone-naive to follow the usual database convention of naming the timezone-naive
+  type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
+  types.
 
 v1.11.0
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ v2.0.0
   being timezone-naive to follow the usual database convention of naming the timezone-naive
   type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
   types.
+  `#827 <https://github.com/kensho-technologies/graphql-compiler/pull/827>`__
 
 v1.11.0
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ Current development version
 v2.0.0 (upcoming release)
 -------------------------
 
-- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
+- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware to
   being timezone-naive to follow the usual database convention of naming the timezone-naive
   type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
   types.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 Current development version
 ---------------------------
 
+v2.0.0
+------
+
+- **BREAKING** Change the GraphQLDateTime scalar type from being timezone aware to timezone naive.
+  In the future, we plan to add timezone
+
 v1.11.0
 -------
 
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -4,6 +4,13 @@ Changelog
 Current development version
 ---------------------------
 
+v2.0.0
+------
+
+- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
+  being timezone-naive in order to follow the usual database convention of naming the timezone-naive
+  type "datetime". This change breaks OrientDB queries filtering on or outputting datetime fields.
+
 v1.11.0
 -------
 
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -4,8 +4,8 @@ Changelog
 Current development version
 ---------------------------
 
-v2.0.0
-------
+v2.0.0 (upcoming release)
+-------------------------
 
 - **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
   being timezone-naive to follow the usual database convention of naming the timezone-naive
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -8,8 +8,9 @@ v2.0.0
 ------
 
 - **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
-  being timezone-naive in order to follow the usual database convention of naming the timezone-naive
-  type "datetime". This change breaks OrientDB queries filtering on or outputting datetime fields.
+  being timezone-naive to follow the usual database convention of naming the timezone-naive
+  type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
+  types.
 
 v1.11.0
 -------
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -11,6 +11,7 @@ v2.0.0
   being timezone-naive to follow the usual database convention of naming the timezone-naive
   type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
   types.
+  `#827 <https://github.com/kensho-technologies/graphql-compiler/pull/827>`__
 
 v1.11.0
 -------
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -7,7 +7,7 @@ Current development version
 v2.0.0 (upcoming release)
 -------------------------
 
-- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
+- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware to
   being timezone-naive to follow the usual database convention of naming the timezone-naive
   type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
   types.
--- a/graphql_compiler/cost_estimation/int_value_conversion.py
+++ b/graphql_compiler/cost_estimation/int_value_conversion.py
@@ -88,8 +88,6 @@ def convert_int_to_field_value(
     See module docstring for details. The int_value is expected to be in the range of
     convert_field_value_to_int.
 
-    Datetimes are returned in the UTC timezone.
-
     Args:
         schema_info: QueryPlanningSchemaInfo
         vertex_class: str, name of vertex class to which the property field belongs.
--- a/graphql_compiler/schema/__init__.py
+++ b/graphql_compiler/schema/__init__.py
@@ -303,27 +303,28 @@ def _serialize_datetime(value: Any) -> str:
     # Python datetime.datetime is a subclass of datetime.date, but in this case, the two are not
     # interchangeable. Rather than using isinstance, we will therefore check for exact type
     # equality.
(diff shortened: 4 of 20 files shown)


pmantica1:
Resolves https://github.com/kensho-technologies/graphql-compiler/issues/715 by disallowing timezone aware arguments.

pmantica1:
Also resolves https://github.com/kensho-technologies/graphql-compiler/issues/745

graphql_compiler/cost_estimation/int_value_conversion.py
```diff
@@ -88,8 +88,6 @@ def convert_int_to_field_value(
     See module docstring for details. The int_value is expected to be in the range of
     convert_field_value_to_int.
 
-    Datetimes are returned in the UTC timezone.
-
```
pmantica1, line 92: The docstring here was already incorrect. So I didn't need to change any of the code.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -633,6 +633,26 @@ def test_filter_on_date(self, backend_name: str) -> None:
         ]
         self.assertResultsEqual(graphql_query, parameters, backend_name, expected_results)
 
+    # RedisGraph doesn't support temporal types, so DateTime types aren't supported.
+    @use_all_backends(except_backends=(test_backend.REDISGRAPH,))
+    @integration_fixtures
+    def test_filter_on_datetime(self, backend_name: str) -> None:
```
pmantica1, line 639: Added a datetime integration test.

graphql_compiler/tests/snapshot_tests/snapshots/snap_test_orientdb_match_query.py
```diff
@@ -119,7 +119,7 @@
...
+snapshots['OrientDBParameterizedMatchQueryTests::test_complex_optional_variables_with_starting_filter 1'] = frozenset({(frozenset({('other_parent_fed_at', '2008-09-23T00:00:00'), ('grandparent_fed_at', '2012-07-02T00:00:00')}), 2), (frozenset({('other_parent_fed_at', '2009-08-12T00:00:00'), ('child_fed_at', '2012-07-02T00:00:00'), ('grandparent_fed_at', '2012-07-02T00:00:00')}), 1), (frozenset({('child_fed_at', '2012-07-02T00:00:00'), ('grandparent_fed_at', '2012-07-02T00:00:00')}), 2), (frozenset({('other_parent_fed_at', '2009-08-12T00:00:00'), ('grandparent_fed_at', '2012-07-02T00:00:00')}), 5), (frozenset({('other_parent_fed_at', '2008-09-23T00:00:00'), ('grandparent_fed_at', '2008-09-23T00:00:00')}), 2), (frozenset({('other_parent_fed_at', '2003-12-03T00:00:00'), ('grandparent_fed_at', '2008-09-23T00:00:00')}), 2), (frozenset({('grandparent_fed_at', '2012-07-02T00:00:00')}), 5), (frozenset({('other_parent_fed_at', '2003-12-03T00:00:00'), ('grandparent_fed_at', '2012-07-02T00:00:00')}), 2), (frozenset({('grandparent_fed_at', '2008-09-23T00:00:00')}), 5)})
```
pmantica1, line 122: The snapshot test change because the returned datetimes are now timezone naive instead of timezone aware because of the change in `STANDARD_DATETIME_FORMAT` which specifies the datetime format of returned datetime values. 

To be extra clear, the returned datetime values aren't empty now. They just lack the "Z" that indicated a UTC timezone.

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
```diff
@@ -156,7 +154,6 @@
     mysqltypes.SMALLINT: GraphQLInt,
     mysqltypes.REAL: GraphQLFloat,
     mysqltypes.TEXT: GraphQLString,
-    mysqltypes.TIMESTAMP: GraphQLDateTime,
```
pmantica1, line 159: We remove support for `TIMESTAMP` types since these are actually tz-aware.

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
```diff
@@ -175,15 +176,26 @@
...
+    # HACK(pmantica1): The code above implies that the GraphQL type is just a factor of the class
+    #                  of column_type. However, it is also a function of the attributes of
+    #                  column_type. We should probably refactor this file to better reflect that.
+    if isinstance(column_type, sqltypes.DateTime) and column_type.timezone:
         warnings.warn(
-            'Ignoring column "{}" with unsupported SQL datatype: {}'.format(
-                column_name, type(column_type).__name__
-            )
+            f'Ignoring column "{column_name}". Timezone aware datetime types are '
+            f"currently not supported."
```
pmantica1, line 189: OrientDB datetimes can also be tz-aware, but I don't know how to tell if they are. Even if I knew how to extract this information, I think that it would probably require a non-trivial refactor to extract it and I am not sure this is worth effort at the moment.

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
```diff
@@ -175,15 +176,26 @@
...
+    column_name: str, column_type: TypeEngine
+) -> Optional[GraphQLScalarType]:
     """Return the matching GraphQLScalarType for the SQL datatype or None if none is found."""
-    maybe_graphql_type = SQL_CLASS_TO_GRAPHQL_TYPE.get(type(column_type), None)
-    if maybe_graphql_type is None:
-        # Trying to get the string representation of the SQLAlchemy JSON and ARRAY types
-        # will lead to an error. We therefore use repr instead.
+    # HACK(pmantica1): The code above implies that the GraphQL type is just a factor of the class
+    #                  of column_type. However, it is also a function of the attributes of
+    #                  column_type. We should probably refactor this file to better reflect that.
```
bojanserafimov, line 185: I don't see the problem with the code, or why a refactor is necessary.

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
bojanserafimov: Agree, doesn't seem to be worth it.

graphql_compiler/schema_generation/sqlalchemy/scalar_type_mapper.py
pmantica1: I was thinking then this could be hacky in the future once we add a timezone aware GraphQL datetime type. Then it wouldn't be possible to map SQL types to zero-or-one GraphQL types. So I don't think that a dictionary would be the right construct. However, this code is not really hacky right now so I'll remove the comment. (Also, I don't know if I explained myself above, but I don't think the issue I described is relevant yet, so let's not discuss it any further).

CHANGELOG.rst
```diff
@@ -4,6 +4,15 @@ Changelog
...
 
+v2.0.0
+------
+
+- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
+  being timezone-naive to follow the usual database convention of naming the timezone-naive
+  type "datetime" and avoid confusion after we've added both timezone-aware and timezone-naive
+  types.
+  `#827 <https://github.com/kensho-technologies/graphql-compiler/pull/827>`__
+
```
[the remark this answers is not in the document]
pmantica1: I think that we should just write "v.2.0.0 (upcoming release)" here.

I don't love modifying the milestone notes. One reason is that by modifying the rst text directly you can see exactly how the changelog is going to look like. Also, with the milestone notes we don't have a great way of reviewing other people's work. So if we want to distribute the changelog writing among the contributors, there is no great way to enforce the high changelog writing standards.

I think that changelog for dev release is a bit inaccurate since we've had 20 dev releases so far so it is is a bit late for that :). 

I investigated what other libraries did. The changelog entry for SQLAlchemy 2.0 already exists, but the SQLAlchemy doesn't exist yet.

docs/source/about/changelog.rst
```diff
@@ -4,6 +4,15 @@ Changelog
 Current development version
 ---------------------------
 
+v2.0.0 (upcoming release)
+-------------------------
+
+- **BREAKING** Change the :code:`GraphQLDateTime` scalar type from being timezone-aware aware to
```
chewselene, line 10: nit: `timezone-aware aware`

What's the difference between this changelog and the other `CHANGELOG.rst`? Are they always kept in sync with one another?

graphql_compiler/schema/__init__.py
```diff
@@ -301,26 +301,33 @@ def _parse_date_value(value):
...
         raise ValueError(
-            "Expected argument to be a python datetime object. "
-            "Got {} of type {} instead.".format(value, type(value))
+            f"Expected a timezone naive datetime object. Got {value} of type {type(value)} instead."
         )
-    return value.isoformat()
 
 
-def _parse_datetime_value(value):
+def _parse_datetime_value(value: Any) -> datetime:
```
bojanserafimov, line 321: Can we say `value` here is a string?

graphql_compiler/schema/__init__.py
```diff
@@ -301,26 +301,33 @@ def _parse_date_value(value):
     return arrow.get(value, "YYYY-MM-DD").date()
 
 
-def _serialize_datetime(value):
+def _serialize_datetime(value: Any) -> str:
```
bojanserafimov, line 304: Do we also want to enforce no tzinfo in the `validate_argument_type` function in `query_formatting/common.py`? I know deserialization can't produce a datetime with tzinfo, but not all arguments come from deserialization. They could be hard-coded or generated in python code, or come from a custom deserializer.

docs/source/about/changelog.rst
pmantica1: Thanks for pointing the typo.

They have the same content and they should be kept in sync with each other. `changelog.rst` is meant to be the changelog for readthedocs.  There is still no link in the `README.rst` pointing to the readtheddocs, so the readthedocs documentation is not really discoverable. 
That's why we still have an identical `CHANGELOG.rst` in the top level directory. 

So unfortunately until we fully release readthedocs,  most of the documentation is going to be duplicated in to places.

This is what the readthedocs documentation looks like right now:
https://graphql-compiler.readthedocs.io/en/latest/

Right now we are only missing the documentation that would go in the `README.rst` once we move to readthedocs.

graphql_compiler/schema/__init__.py
pmantica1: I think that it makes sense that tzinfo is None for the following reasons:
1. It is easier to stop enforcing it in the future than to start enforcing it.
2. Without enforcing this, it would not be safe to start paginating previously un-paginated queries since pagination currently removes timezone info. 
3.  There would be no way to fulfill the docstring of `validate_argument_type` that the validation done in that function is stricter than the validation done in another other backend. 

Overall the way I look at it is that this gives us some nice guarantees that make the code easier to maintain, so I would prefer to keep enforce this until we have concrete reasons not to.

graphql_compiler/schema/__init__.py
pmantica1: Changing the type annotation to `str` does not give us a mypy type error, but I think that it might make more sense to have `Any` here and in `_serialize_datetime`

GraphQL core next has an `Any` type annotation for the input values of `parse_value` and `serialize` functions and GraphQLJS has a `mixed` type annotation for these values, which seems to be the JS equivalent of `Any`. 
https://flow.org/en/docs/types/mixed/


I get the feeling that this more likely going to lead to unnecessary `coerce_type` calls than to prevent any bugs. What do you think? 

A specific case I am thinking about is what if we had:

```
def deserialize_json_value(json_value: Union[str, int, ...], expected_type: GraphQLType):
      ....
```
Since you can't pass  a `Union[str, ...]` argument into a function with a `str` type annotation, we would probably have to have a lot of `coerce_type` internally within this function. 

I am thinking of a similar scenario for the `serialize` function. We use the `serialize` value for validation. In the future, we might update the `validate_argument_type` type annotation to:

```
ALLOWED_PYTHON_TYPES: Union[str, int, ...]

def validate_argument_type(
       name: str, 
       value: ALLOWED_PYTHON_TYPES, 
       graphql_type: GraphQLType
)  -> None:
```

For a similar reason, we would have to a lot of `coerce_type` calls internally to make this function pass mypy linting.

graphql_compiler/schema/__init__.py
bojanserafimov: I think you misunderstood what I'm saying: Let's add extra validation in `validate_argument_type`, checking that `tzinfo is None`.

pmantica1: Oh my bad. We already have this validation though:
```
    elif is_same_type(GraphQLDateTime, stripped_type):
        if not isinstance(value, (datetime.date, arrow.Arrow)):
            _raise_invalid_type_error(name, (datetime.date, arrow.Arrow), value)
        try:
            GraphQLDateTime.serialize(value)
        except ValueError as e:
            raise GraphQLInvalidArgumentError(e)
```
```
def _serialize_datetime(value: Any) -> str:
    """Serialize a DateTime object to its proper ISO-8601 representation."""
    # Python datetime.datetime is a subclass of datetime.date, but in this case, the two are not
    # interchangeable. Rather than using isinstance, we will therefore check for exact type
    # equality.
    #
    # We don't allow Arrow objects as input since it seems that Arrow objects are always tz aware.
    # This is supported by the fact that the `.naive` Arrow method returns a datetime object instead
    # of an Arrow object.
    if type(value) == datetime and value.tzinfo is None:
        return value.isoformat()
    else:
        raise ValueError(
            f"Expected a timezone naive datetime object. Got {value} of type {type(value)} instead."
        )
```

PR #831, opened 2020-05-20
_x_count Error Messaging and Test Updates

chewselene:
- Added an integration test for `_x_count`
- Updated error messaging for some `NotImplementedError`s in `compiler/expressions.py`
- Switched cypher's `SKIP_TEST` to `NotImplementedError` for `_x_count` compiler tests since we should ensure that the compiler correctly raises an error rather that just skipping the test

 Not sure when we prefer to use `SKIP_TEST` rather than `NotImplementedError` or a compiled query, but we have quite a few instances of `SKIP_TEST`. I suspect a number of them could be changed to `NotImplementedError` or a compiled query if that is desirable.


bojanserafimov:
Looks good, just linter is failing.

PR #834, opened 2020-05-28
Schema rename suppress types

LWprogramming:
First part of the implementation described in pull request #828 and issue #810, covering just 1-0 renaming for types (aka suppressing types). I'm going to read over the changes here first and then when it's ready, I'll mark it as ready for review

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -1,20 +1,22 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from collections import namedtuple
-from typing import Optional, Dict
+from typing import Dict, Optional
 
-from graphql import build_ast_schema, DocumentNode
-from graphql.language.visitor import Visitor, visit, REMOVE
+from graphql import DocumentNode, build_ast_schema
+from graphql.language.visitor import REMOVE, Visitor, visit
 import six
 
+from ..ast_manipulation import get_ast_with_non_null_and_list_stripped
 from .utils import (
+    CascadingSuppressionError,
     SchemaNameConflictError,
     check_ast_schema_is_valid,
     check_type_name_is_valid,
     get_copy_of_node_with_new_name,
     get_query_type_name,
-    get_scalar_names, CascadingSuppressionError,
+    get_scalar_names,
 )
-from ..ast_manipulation import get_ast_with_non_null_and_list_stripped
+
 
 RenamedSchemaDescriptor = namedtuple(
     "RenamedSchemaDescriptor",
@@ -27,7 +29,9 @@ RenamedSchemaDescriptor = namedtuple(
 )
 
 
-def rename_schema(ast: DocumentNode, renamings: Dict[str, Optional[str]]) -> RenamedSchemaDescriptor:
+def rename_schema(
+    ast: DocumentNode, renamings: Dict[str, Optional[str]]
+) -> RenamedSchemaDescriptor:
     """Create a RenamedSchemaDescriptor; types and query type fields are renamed using renamings.
 
     Any type, interface, enum, or fields of the root type/query type whose name
@@ -75,6 +79,7 @@ def rename_schema(ast: DocumentNode, renamings: Dict[str, Optional[str]]) -> Ren
 
     # Rename query type fields
     from graphql.language.printer import print_ast
+
     ast = _rename_fields(ast, renamings, query_type)
     return RenamedSchemaDescriptor(
         schema_ast=ast,
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -1,8 +1,9 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from collections import namedtuple
+from typing import Optional, Dict
 
-from graphql import build_ast_schema
-from graphql.language.visitor import Visitor, visit
+from graphql import build_ast_schema, DocumentNode
+from graphql.language.visitor import Visitor, visit, REMOVE
 import six
 
 from .utils import (
@@ -11,9 +12,9 @@ from .utils import (
     check_type_name_is_valid,
     get_copy_of_node_with_new_name,
     get_query_type_name,
-    get_scalar_names,
+    get_scalar_names, CascadingSuppressionError,
 )
-
+from ..ast_manipulation import get_ast_with_non_null_and_list_stripped
 
 RenamedSchemaDescriptor = namedtuple(
     "RenamedSchemaDescriptor",
@@ -26,7 +27,7 @@ RenamedSchemaDescriptor = namedtuple(
 )
 
 
-def rename_schema(ast, renamings):
+def rename_schema(ast: DocumentNode, renamings: Dict[str, Optional[str]]) -> RenamedSchemaDescriptor:
     """Create a RenamedSchemaDescriptor; types and query type fields are renamed using renamings.
 
     Any type, interface, enum, or fields of the root type/query type whose name
@@ -38,9 +39,8 @@ def rename_schema(ast, renamings):
         ast: Document, representing a valid schema that does not contain extensions, input
              object definitions, mutations, or subscriptions, whose fields of the query type share
              the same name as the types they query. Not modified by this function
-        renamings: Dict[str, str], mapping original type/field names to renamed type/field names.
-                   Type or query type field names that do not appear in the dict will be unchanged.
-                   Any dict-like object that implements get(key, [default]) may also be used
+        renamings: Dict[str, Optional[str]], mapping original type/field names to renamed type/field names or None.
+                   Type or query type field names that do not appear in the dict will be unchanged. If a type is mapped to None, it will be suppressed in the renamed schema and queries will not be able to access it. Any dict-like object that implements get(key, [default]) may also be used
 
     Returns:
         RenamedSchemaDescriptor, a namedtuple that contains the AST of the renamed schema, and the
@@ -74,7 +74,8 @@ def rename_schema(ast, renamings):
     }
 
     # Rename query type fields
-    ast = _rename_query_type_fields(ast, renamings, query_type)
+    from graphql.language.printer import print_ast
+    ast = _rename_fields(ast, renamings, query_type)
     return RenamedSchemaDescriptor(
         schema_ast=ast,
         schema=build_ast_schema(ast),
@@ -113,8 +114,8 @@ def _rename_types(ast, renamings, query_type, scalars):
     return renamed_ast, visitor.reverse_name_map
 
 
-def _rename_query_type_fields(ast, renamings, query_type):
-    """Rename all fields of the query type.
+def _rename_fields(ast, renamings, query_type):
+    """Rename fields, including query type fields.
 
     The input AST will not be modified.
 
@@ -127,7 +128,7 @@ def _rename_query_type_fields(ast, renamings, query_type):
     Returns:
         DocumentNode, representing the modified version of the input schema AST
     """
-    visitor = RenameQueryTypeFieldsVisitor(renamings, query_type)
+    visitor = RenameFieldsVisitor(renamings, query_type)
     renamed_ast = visit(ast, visitor)
     return renamed_ast
 
@@ -232,6 +233,9 @@ class RenameSchemaTypesVisitor(Visitor):
             return node
 
         new_name_string = self.renamings.get(name_string, name_string)  # Default use original
+        if new_name_string is None:
+            # suppress the type
+            return REMOVE
         check_type_name_is_valid(new_name_string)
 
         if (
@@ -268,16 +272,16 @@ class RenameSchemaTypesVisitor(Visitor):
             renamed_node = self._rename_name_and_add_to_record(node)
             if renamed_node is node:  # Name unchanged, continue traversal
                 return None
-            else:  # Name changed, return new node, `visit` will make shallow copies along path
+            else:  # Name changed or suppressed, return new node, `visit` will make shallow copies along path
                 return renamed_node
         else:
             # All Node types should've been taken care of, this line should never be reached
             raise AssertionError('Unreachable code reached. Missed type: "{}"'.format(node_type))
 
 
-class RenameQueryTypeFieldsVisitor(Visitor):
+class RenameFieldsVisitor(Visitor):
     def __init__(self, renamings, query_type):
-        """Create a visitor for renaming fields of the query type in a schema AST.
+        """Create a visitor for renaming fields in a schema AST.
 
         Args:
             renamings: Dict[str, str], from original field name to renamed field name. Any
@@ -285,7 +289,7 @@ class RenameQueryTypeFieldsVisitor(Visitor):
             query_type: str, name of the query type (e.g. RootSchemaQuery)
         """
         # Note that as field names and type names have been confirmed to match up, any renamed
-        # field already has a corresponding renamed type. If no errors, due to either invalid
+        # query type field already has a corresponding renamed type. If no errors, due to either invalid
         # names or name conflicts, were raised when renaming type, no errors will occur when
         # renaming query type fields.
         self.in_query_type = False
@@ -299,16 +303,47 @@ class RenameQueryTypeFieldsVisitor(Visitor):
 
     def leave_object_type_definition(self, node, key, parent, path, ancestors):
         """If the node's name matches the query type, record that we left the query type."""
+        if len(node.fields) == 0:
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} suppressed every type in the schema so it will be impossible to query for anything. To fix this, check why the `renamings` argument of `rename_schema` mapped every type to None."
+            )
         if node.name.value == self.query_type:
             self.in_query_type = False
 
     def enter_field_definition(self, node, *args):
-        """If inside the query type, rename field and add the name pair to reverse_field_map."""
+        """Update fields as needed.
+
+        If inside the query type, rename field and add the name pair to reverse_field_map.
+        If not inside the query type, check that no field depends on a type that was suppressed.
+
+        Raises:
+            CascadingSuppressionError if a field that was not suppressed depends on a type that was suppressed.
+        """
+        field_name = node.name.value
         if self.in_query_type:
-            field_name = node.name.value
             new_field_name = self.renamings.get(field_name, field_name)  # Default use original
             if new_field_name == field_name:
                 return None
+            if new_field_name is None:
+                # suppress the type
+                return REMOVE
             else:  # Make copy of node with the changed name, return the copy
                 field_node_with_new_name = get_copy_of_node_with_new_name(node, new_field_name)
                 return field_node_with_new_name
+        else:
+            # at a field of a type that is not the query type
+            type = get_ast_with_non_null_and_list_stripped(node.type)
+            if type == REMOVE:
+                # then this field depends on a type that was suppressed, which is illegal
+                type_name = args[-1][-1].name.value
+                raise CascadingSuppressionError(
+                    f"Type renamings {self.renamings} attempted to suppress a type, but type {type_name}'s field {field_name} still depends on that type. Suppressing individual fields hasn't been implemented yet, but when it is, you can fix this error by suppressing the field as well."
+                )
+            return None
+
+    def enter_union_type_definition(self, node, *args):
+        union_name = node.name.value
+        if len(node.types) == 0:
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} suppressed all types belonging to the union {union_name}. To fix this, you can suppress the union as well by adding {union_name}: None to the `renamings` argument of `rename_schema`."
+            )
--- a/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
+++ b/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
@@ -8,7 +8,11 @@ from graphql.language.visitor import QUERY_DOCUMENT_KEYS
 from graphql.pyutils import snake_to_camel
 
 from ...schema_transformation.rename_schema import RenameSchemaTypesVisitor, rename_schema
-from ...schema_transformation.utils import InvalidTypeNameError, SchemaNameConflictError, CascadingSuppressionError
+from ...schema_transformation.utils import (
+    CascadingSuppressionError,
+    InvalidTypeNameError,
+    SchemaNameConflictError,
+)
 from .input_schema_strings import InputSchemaStrings as ISS
 
 
@@ -352,14 +356,11 @@ class TestRenameSchema(unittest.TestCase):
         )
         self.assertEqual(renamed_schema_string, print_ast(renamed_schema.schema_ast))
         self.assertEqual(
-            {"NewDroid": "Droid"},
-            renamed_schema.reverse_name_map,
+            {"NewDroid": "Droid"}, renamed_schema.reverse_name_map,
         )
 
     def test_union_member_suppress(self):
-        renamed_schema = rename_schema(
-            parse(ISS.union_schema), {"Droid": None}
-        )
+        renamed_schema = rename_schema(parse(ISS.union_schema), {"Droid": None})
         renamed_schema_string = dedent(
             """\
             schema {
@@ -379,8 +380,7 @@ class TestRenameSchema(unittest.TestCase):
         )
         self.assertEqual(renamed_schema_string, print_ast(renamed_schema.schema_ast))
         self.assertEqual(
-            {},
-            renamed_schema.reverse_name_map,
+            {}, renamed_schema.reverse_name_map,
         )
 
     def test_list_rename(self):
@@ -693,7 +693,9 @@ class TestRenameSchema(unittest.TestCase):
 
     def test_field_still_depends_on_suppressed_type(self):
         with self.assertRaises(CascadingSuppressionError):
-            rename_schema(parse(ISS.multiple_fields_schema), {"Dog": None})  # a field in Human is of type Dog.
+            rename_schema(
+                parse(ISS.multiple_fields_schema), {"Dog": None}
+            )  # a field in Human is of type Dog.
 
     def test_field_in_list_still_depends_on_suppressed_type(self):
         with self.assertRaises(CascadingSuppressionError):
--- a/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
+++ b/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
@@ -404,48 +404,6 @@ class TestRenameSchema(unittest.TestCase):
             renamed_schema.reverse_name_map,
         )
 
-    def test_list_suppress(self):
-        renamed_schema = rename_schema(
-            parse(ISS.list_schema),
-            {
-                "Droid": "NewDroid",
-                "Character": "NewCharacter",
-                "Height": None,
-                "Date": "NewDate",
-                "id": "NewId",
-                "String": "NewString",
-            },
-        )
-        renamed_schema_string = dedent(
-            """\
-            schema {
-              query: SchemaQuery
-            }
-
-            type NewDroid implements NewCharacter {
-              id: String
-              dates: [Date]
-              friends: [NewDroid]
-              enemies: [NewCharacter]
-            }
-
-            type SchemaQuery {
-              NewDroid: [NewDroid]
-            }
-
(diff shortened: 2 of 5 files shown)


LWprogramming:
Summary:

* Implement type suppressions (so no field renamings, 1-many, suppressions etc just yet)

* Checking for any illegal schema state that can arise just from suppressing types (see the new Visitor class I wrote)

* Tests for both successful type suppression & illegal schema state errors

* Small adjustments to test input strings and other tests that use these input strings to make sure the tests say what we think they say (e.g. confirming that suppressing all types in a union is why it raised an error, and not because we suppressed every type in the schema, all of which happen to be a member of that union).

* Type hinting `rename_schema`

Not sure what's going on with the failing CI tests-- is this a known issue? Doesn't seem like anything in this PR would prevent them from running and the internet hasn't brought up much yet.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +303,128 @@ def enter(self, node, key, parent, path, ancestors):
...
+    def enter_field_definition(self, node: FieldDefinitionNode, *args):
+        """If not at query type, check that no field depends on a type that was suppressed."""
+        if self.in_query_type:
+            return None
+        # at a field of a type that is not the query type
+        field_name = node.name.value
+        node_type = get_ast_with_non_null_and_list_stripped(node.type)
+        if node_type == REMOVE:
+            # then this field depends on a type that was suppressed, which is illegal
+            type_name = args[-1][-1].name.value
```
LWprogramming, line 405: This is an awkward way to get the type that this field belongs to. I haven't found a better way to get this information, but it'd be really useful to figure out how to access both the type in a more natural way.

Also would be really nice to figure out what the type of the field was before the suppression happened, so we could put that information in the error message, but I haven't thought of a clean way to implement that.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +303,128 @@ def enter(self, node, key, parent, path, ancestors):
...
-    def leave_object_type_definition(self, node, key, parent, path, ancestors):
+    def leave_object_type_definition(
+        self, node: ObjectTypeDefinitionNode, key, parent, path, ancestors
+    ):
         """If the node's name matches the query type, record that we left the query type."""
+        if not node.fields:
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} suppressed every type in the schema so it will be"
+                f" impossible to query for anything. To fix this, check why the `renamings` "
+                f"argument of `rename_schema` mapped every type to None. Note that adding "
```
LWprogramming, line 345: The other two existing suppression error messages have an easily actionable step: add more suppressions so that the field/union doesn't end up in an illegal state. It seems really unlikely that someone would want to suppress every type in practice, because then you'd have a useless (empty) schema, so I'm not sure what makes the most sense to write here.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -232,6 +264,9 @@ def _rename_name_and_add_to_record(self, node):
             return node
 
         new_name_string = self.renamings.get(name_string, name_string)  # Default use original
+        if new_name_string is None:
+            # suppress the type
+            return REMOVE
```
LWprogramming, line 270: With this additional line of code, type-hinting this function is really tricky because now you can return a node or `REMOVE`. Here, `REMOVE` is `None` but that could change so just saying something like `Optional[Node]` doesn't necessarily work.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -75,14 +90,19 @@ def rename_schema(ast, renamings):
 
     # Rename query type fields
     ast = _rename_query_type_fields(ast, renamings, query_type)
+
+    # check for fields or unions that depend on types that were suppressed
```
obi1kenobi, line 94: Please try to keep code style in line with the surrounding code. Here, the comment a few lines above is capitalized, but your newly-added one is not. This particular instance is very minor, but every little bit counts and we want to keep the codebase as clean and polished-looking as possible.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -132,6 +154,14 @@ def _rename_query_type_fields(ast, renamings, query_type):
     return renamed_ast
 
 
+def _check_for_cascading_type_suppression(
+    ast: DocumentNode, renamings: Dict[str, Optional[str]], query_type: str
+) -> DocumentNode:
```
obi1kenobi, line 159: Please make sure all your functions have docstrings. Our rule for docstrings is that if the docstring does not fit on one line, it must have `Args/Returns` sections, and ideally a `Raises` section as well if it raises any notable errors.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -186,19 +216,21 @@ class RenameSchemaTypesVisitor(Visitor):
...
+    def __init__(
+        self, renamings: Dict[str, Optional[str]], query_type: str, scalar_types: Set[str]
+    ):
         """Create a visitor for renaming types in a schema AST.
 
         Args:
-            renamings: Dict[str, str], mapping from original type name to renamed type name.
-                       Any name not in the dict will be unchanged
+            renamings: Dict[str, Optional[str]], mapping from original type name to renamed type
+            name. Any name not in the dict will be unchanged
```
obi1kenobi, line 226: Similarly to the above, please make sure your code edits match and preserve the style of the surrounding code. Here, line 226 gets de-indented, even though the line it replaced and the next few lines have a different style.

It's often very useful to give your own code a self-review in the PR review viewer before having others review it — these things are much more obvious when looking at a diff, and I'm sure you'll be able to catch most if not all of them by yourself. That will speed up the review process drastically, and make you and the team more productive.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -218,8 +250,9 @@ def _rename_name_and_add_to_record(self, node):
                   NameNode.
 
         Returns:
-            Node object, identical to the input node, except with possibly a new name. If the
-            name was not changed, the returned object is the exact same object as the input
+            Node object or REMOVE. If it returns REMOVE, then the type was suppressed. Otherwise, it
```
obi1kenobi, line 253: I'd make a note about what `REMOVE` is here, because that wasn't obvious to me. It's important for the reader to know that this is a constant defined by the GraphQL library, and not something this library invents.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -232,6 +265,9 @@ def _rename_name_and_add_to_record(self, node):
             return node
 
         new_name_string = self.renamings.get(name_string, name_string)  # Default use original
+        if new_name_string is None:
+            # suppress the type
```
obi1kenobi, line 269: Style-matching: comments are generally capitalized in this file (and likely elsewhere in this repo)

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: Good catch! Let's ignore this for now and type it as `Optional[Node]`, and leave a comment explaining the above in the code. That way, if `mypy` starts complaining one day, we'll know that the type hint may need to change to match the new value of `REMOVE`.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -257,7 +293,7 @@ def _rename_name_and_add_to_record(self, node):
             node_with_new_name = get_copy_of_node_with_new_name(node, new_name_string)
             return node_with_new_name
 
-    def enter(self, node, key, parent, path, ancestors):
+    def enter(self, node: Node, key, parent, path, ancestors):
```
obi1kenobi, line 296: Are you able to type the rest of the args and the return value?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
-    def enter_object_type_definition(self, node, *args):
+    def enter_object_type_definition(self, node: ObjectTypeDefinitionNode, *args):
         """If the node's name matches the query type, record that we entered the query type."""
         if node.name.value == self.query_type:
             self.in_query_type = True
 
-    def leave_object_type_definition(self, node, key, parent, path, ancestors):
+    def leave_object_type_definition(
+        self, node: ObjectTypeDefinitionNode, key, parent, path, ancestors
+    ):
```
obi1kenobi, line 340: Similarly, are you able to type the remainder of the args and the return value?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
-        # names or name conflicts, were raised when renaming type, no errors will occur when
+        # query type field already has a corresponding renamed type. If no errors, due to either
+        # invalid names or name conflicts, were raised when renaming type, no errors will occur when
         # renaming query type fields.
         self.in_query_type = False
         self.renamings = renamings
         self.query_type = query_type
 
-    def enter_object_type_definition(self, node, *args):
+    def enter_object_type_definition(self, node: ObjectTypeDefinitionNode, *args):
```
obi1kenobi, line 333: Is `*args` actually the correct function signature here, or are there specific arguments this function is always called with?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
 
-    def leave_object_type_definition(self, node, key, parent, path, ancestors):
+    def leave_object_type_definition(
+        self, node: ObjectTypeDefinitionNode, key, parent, path, ancestors
+    ):
         """If the node's name matches the query type, record that we left the query type."""
+        if not node.fields:
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} suppressed every type in the schema so it will be"
+                f" impossible to query for anything. To fix this, check why the `renamings` "
```
obi1kenobi, line 345: By convention, we leave the whitespace at ends of string literals, and not at the beginning. If the line above will be too long when you move the space up there, then move one more word down to the next line, and still end the previous line with a space.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
+            # then this field depends on a type that was suppressed, which is illegal
+            type_name = args[-1][-1].name.value
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} attempted to suppress a type, but type "
+                f"{type_name}'s field {field_name} still depends on that type. Suppressing "
+                f"individual fields hasn't been implemented yet, but when it is, you can fix "
+                f"this error by suppressing the field as well. Note that adding suppressions "
+                f"may lead to other types, fields, unions, etc. requiring suppression so you "
+                f"may need to iterate on this before getting a legal schema."
+            )
```
obi1kenobi, line 414: Perhaps we can add a separate helper (separate PR, though perhaps build it and merge it first?) that given a list of types we'd _like to_ suppress, produces the list of fields etc. that must be suppressed as well for the type suppression to be legal. That enables a few nice things:
- The workflow becomes more pleasant: "iterate until you stop getting errors" becomes "use this helper, review its output, and apply it if it looks good".
- The invariant we check for here simply becomes "the helper finds no fields that need suppression", enabling code reuse.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
+        if self.in_query_type:
+            return None
+        # at a field of a type that is not the query type
+        field_name = node.name.value
+        node_type = get_ast_with_non_null_and_list_stripped(node.type)
+        if node_type == REMOVE:
+            # then this field depends on a type that was suppressed, which is illegal
+            type_name = args[-1][-1].name.value
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} attempted to suppress a type, but type "
```
obi1kenobi, line 408: This seems like it would raise an error separately for each field it runs into, one at a time. That means fixing everything wrong with a schema requires _many_ iterations where the code complains about one wrong thing at a time, instead of giving you everything that is wrong all at once. So this is working code, but not particularly ergonomic from an end-user point of view. See my comment below for an alternative suggestion that would make raising an error about all the issues at once easier.

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: Are you sure that `*args` is actually intended to be variadic? I think I commented on this above as well — I feel like we may be mis-defining the function signature in a way that just happens to work, but then leads to problems like these.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
+        self.query_type = query_type
+
+    def enter_object_type_definition(self, node: ObjectTypeDefinitionNode, *args):
+        """If the node's name matches the query type, record that we entered the query type."""
+        if node.name.value == self.query_type:
+            self.in_query_type = True
+
+    def leave_object_type_definition(
+        self, node: ObjectTypeDefinitionNode, key, parent, path, ancestors
+    ):
```
obi1kenobi, line 392: Remember to complete type hints on all functions, including `-> None` if the function doesn't return anything. Also remember that `__init__` does not return anything, so it should always be typed `-> None`.

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: Yeah this is fair, I wouldn't worry about it too much. The thing I wanted to ask about is this: why is this a `Cascading SuppressionError`? I don't see the cascading aspect of it, so why not use the more general `SchemaTransformError`?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +304,128 @@ def enter(self, node, key, parent, path, ancestors):
...
         """If the node's name matches the query type, record that we left the query type."""
+        if not node.fields:
+            raise CascadingSuppressionError(
+                f"Type renamings {self.renamings} suppressed every type in the schema so it will be"
+                f" impossible to query for anything. To fix this, check why the `renamings` "
+                f"argument of `rename_schema` mapped every type to None. Note that adding "
+                f"suppressions may lead to other types, fields, unions, etc. requiring suppression "
+                f"so you may need to iterate on this before getting a legal schema."
```
obi1kenobi, line 348: I'm a bit confused about why this last bit is here describing the workflow for additional suppressions. Don't we want the user to _remove_ suppressions to fix the error? This feels like we're giving the user some mixed messages.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Hm, I just re-read the source code for visitors and it turns out that although other subclasses use `*args`, the superclass `Visitor` actually explicitly lists out all of the arguments so I'll add the arguments explicitly here.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Oops, yes-- adding suppressions was true for cascading suppression issues but this is a `SchemaTransformError` so we can remove that.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Right now we only use `SchemaTransformError` as a parent class for more specific errors like `CascadingSuppressionError`-- the code raises only the specific errors and not `SchemaTransformError`. Given that there's only one way to suppress every type, does it make more sense to make a whole new error class just for this or just use the existing `SchemaTransformError` parent class? I suppose it's reasonably clear from the error message what the problem is, but I suppose it might depend on what's more convenient to see at a glance?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -257,7 +296,14 @@ def _rename_name_and_add_to_record(self, node):
             node_with_new_name = get_copy_of_node_with_new_name(node, new_name_string)
             return node_with_new_name
 
-    def enter(self, node, key, parent, path, ancestors):
+    def enter(
+        self,
+        node: Node,
+        key: Optional[Any],
```
LWprogramming, line 302: type annotations like these might seem strange since Any includes None, but [it's a bit more complicated](https://github.com/python/mypy/issues/3138) as a consequence of how unions work

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -268,47 +314,178 @@ def enter(self, node, key, parent, path, ancestors):
...
+        node_type = get_ast_with_non_null_and_list_stripped(node.type)
+        if node_type == REMOVE:
+            # Then this field depends on a type that was suppressed, which is illegal
+            # Note that we refer to the last item in ancestors, which is not the same as parent. The
+            # last item in ancestors is actually the grandparent of node, specifically the
+            # ObjectTypeDefinitionNode that has this field. We use the grandparent node instead of
+            # the parent because the parent of a FieldDefinitionNode is simply a list of
+            # FieldDefinitionNodes, which doesn't contain the name of the type containing this node
+            # (which we need for the error message).
+            type_name = ancestors[-1].name.value
```
LWprogramming, line 459: Better than before with `*args` but still not great (see preceding comment)

Also, although it's impossible for ancestors to be `None` inside a `FieldTypeDefinitionNode` I'm going to add an assertion in code for stuff like this tomorrow

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: That's a really good point-- now that you mention it, we can do something similar with unions as well.

i.e. Instead of raising an error here, we could add the information about the field to a data structure belonging to this visitor object. Then, after `visit` returns, we could perform a single pass through this data structure and raise errors as needed for any fields/ unions, showing everything in a single pass _before_ `rename` calls `_rename_types` at all. 

I think this actually shouldn't be all that much work since most of it is already done with this visitor class so I can probably get it in this PR.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -186,25 +217,27 @@ class RenameSchemaTypesVisitor(Visitor):
...
-        self.reverse_name_map = {}  # Dict[str, str], from renamed type name to original type name
-        # reverse_name_map contains all types, including those that were unchanged
+        self.reverse_name_map: Dict[str, str] = {}  # From renamed type name to original type name
+        # reverse_name_map contains all non-suppressed types, including those that were unchanged
         self.query_type = query_type
         self.scalar_types = frozenset(scalar_types)
         self.builtin_types = frozenset({"String", "Int", "Float", "Boolean", "ID"})
 
-    def _rename_name_and_add_to_record(self, node):
+    def _rename_name_and_add_to_record(self, node: Node) -> Union[ellipsis, Optional[Node]]:
```
LWprogramming: Stuck on type-hinting this function signature, specifically the return type.

This function can return a `Node`, `None`, or `REMOVE`. It turns out the GraphQL library [sets](https://github.com/graphql-python/graphql-core/blob/master/src/graphql/language/visitor.py#L34) `REMOVE` to the Python literal `Ellipsis`, not `None`-- I think PyCharm got confused because there's a stub file that sets `Ellipsis` to `None`, but when this code actually runs `Ellipsis != None`.

The current error says that this type-hint isn't valid and `note: Suggestion: use type[...] instead of type(...)`.

* type[ellipsis] and type[type(Ellipsis)] don't work because of the square brackets.

* Although `type(Ellipsis)` is the class `ellipsis`, adding `ellipsis` to the union in place of `type(REMOVE)` makes both tests and linter to fail because `ellipsis` is undefined.

* Replacing `type(REMOVE)` with `Type[type(REMOVE)]` also makes the linter fail and produces a link to [this documentation](https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases). It seems unrelated to what the type hint we're trying to do here because I a) haven't been able to find the `ellipsis` class and b) it makes more sense to depend on the type of `REMOVE` as much as possible rather than rely on the fact that it's set to `Ellipsis`.

obi1kenobi: Try `Literal[Ellipsis]` instead? So `Union[Literal[Ellipsis], Optional[Node]]` for the return type hint.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -186,25 +238,28 @@ class RenameSchemaTypesVisitor(Visitor):
...
-        self.reverse_name_map = {}  # Dict[str, str], from renamed type name to original type name
-        # reverse_name_map contains all types, including those that were unchanged
+        self.reverse_name_map: Dict[str, str] = {}  # From renamed type name to original type name
+        # reverse_name_map contains all non-suppressed types, including those that were unchanged
         self.query_type = query_type
         self.scalar_types = frozenset(scalar_types)
         self.builtin_types = frozenset({"String", "Int", "Float", "Boolean", "ID"})
 
-    def _rename_name_and_add_to_record(self, node):
+    def _rename_name_and_add_to_record(self, node: Node) -> Union[type(REMOVE), Optional[Node]]:
```
LWprogramming, line 262: This is the other lint error we're seeing, which is because the [enter](https://github.com/kensho-technologies/graphql-compiler/pull/834/files#diff-3b859f46e97b0af834db0b3b7de08658R321) function has `node` type as `Node`, but this function will only be called with a node of type in `self.rename_types`, and all of those nodes have names. Not sure how to write that in the type-hint though-- putting anything except `Node` as the type-hint makes mypy complain in the `enter` function that calls `_rename_name_and_add_to_record`

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Nope :(

```
error: Parameter 1 of Literal[...] is invalid
error: Variable "builtins.Ellipsis" is not valid as a type
```

obi1kenobi: I went through:
https://github.com/python/typing/issues/684
https://github.com/python/mypy/issues/2180
https://github.com/python/mypy/issues/7818

The consensus seems to be that using `Ellipsis` as an "easy-to-understand, easy-to-import sentinel value" is not recommended and will lead to type issues exactly like we're seeing here. The issues are closed as "won't fix", so I'd recommend checking out the latest version of GraphQL-core (I think we're a minor version behind or so), and if the use of `Ellipsis` is still present, opening a GitHub issue. You could link back to this PR (ideally to this comment thread, by referencing the new issue from here), and ask the maintainers there to create a separate singleton sentinel value per the BDFL's guidance in the GitHub issues I listed above.

In the meantime, I'd recommend making a type alias like `VisitorReturnType = Union[Optional[Node], Any]`, documenting and explaining the necessity of using `Any`, linking to the GraphQL-core issue you open, and using that type alias everywhere it's needed.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -26,20 +37,27 @@
...
+    Scalars, directives, enum values, and fields not belonging to the root/query type will never be
+    renamed.
 
     Args:
         ast: Document, representing a valid schema that does not contain extensions, input
              object definitions, mutations, or subscriptions, whose fields of the query type share
              the same name as the types they query. Not modified by this function
-        renamings: Dict[str, str], mapping original type/field names to renamed type/field names.
-                   Type or query type field names that do not appear in the dict will be unchanged.
+        renamings: Dict[str, Optional[str]], mapping original type/root type field names to renamed
```
chewselene, line 59: Consider removing the types from the `Args` section. Now that we have typehints, adding the type here is redundant and could accidentally get out of date.

graphql_compiler/schema_transformation/rename_schema.py
chewselene, line 262: Can this function return `None`? The `Returns` section of the docstring seems to say otherwise.

LWprogramming: Thanks for the catch! I've looked more closely into how the library defines special values other than `REMOVE`. It seems like everywhere the code previously returned `None`, we should return `IDLE` instead (which happens to be set to `None` right now).

In that case, based on the issue I opened in GraphQL-core today, I'll adjust `VisitorReturnType` (as mentioned in a previous comment above) to be the union of `Node` and `Any`, and when the special return values get collected into an enum it'll simply be a matter of changing the `Any` instead of messing around with optional types.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Sounds good-- this PR is already pretty big so after this gets merged I'll be removing these in a separate PR

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Done! One note: In a separate PR, I'll be updating all the existing `return None`s to `return IDLE` where it makes sense since the GraphQL library defines that for doing nothing. See [here](https://github.com/kensho-technologies/graphql-compiler/pull/834#discussion_r436889940) explaining why the `VisitorReturnType` won't have the `Optional` part.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I've confirmed that the following workaround makes the linter happy, but is quite hacky:

Add `type_hint_rename_types` to `RenameSchemaTypesVisitor` as another class attribute (parallels the [`rename_types`](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema_transformation/rename_schema.py#L179) frozenset that already exists)

```
type_hint_rename_types = Union[
    EnumTypeDefinitionNode,
    InterfaceTypeDefinitionNode,
    NamedTypeNode,
    ObjectTypeDefinitionNode,
    UnionTypeDefinitionNode,
]
```

Then we can use this `Union` as the type in the function signature for `_rename_name_and_add_to_record` and replace the call [`self._rename_name_and_add_to_record(node)`](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema_transformation/rename_schema.py#L268) with 

```
renamed_node = self._rename_name_and_add_to_record(
    cast(RenameSchemaTypesVisitor.type_hint_rename_types, node)
)
```

Trying to turn the existing set to a mypy `Union` doesn't work syntactically (e.g. iterating over these 5 types don't work) as far as I can tell, but I'd be very happy to be proven wrong!

LWprogramming:
That makes sense-- since this'll break it into several smaller PRs, I'll also be able to incorporate some of the comments that I hadn't touched previously.

Here's how I plan to split it up:

- [ ] type-hints for existing code, no functionality changes. Also removes type-hints from `Args` section in comments) as Selene mentioned.

- [ ] Replace `return None` with `return IDLE` in the visitor functions because the GraphQL-core library defines that constant for us. Also would involve creating the `VisitorReturnType` which we'll need later anyways. No functionality changes.

- [ ] Actually implement type suppression: possibly including the helper function mentioned [here](https://github.com/kensho-technologies/graphql-compiler/pull/834#discussion_r432562122) if it's not too long and plays nice with the existing tests here. Functionality change

- [ ] Tests for type suppression will probably be in a separate PR opened at the same time as the type-suppression implementation.

obi1kenobi:
Sounds great!

LWprogramming:
Now that PR 858 and 859 are out in draft form, all the significant changes described here have either already been merged or exist in one of those two PRs so I'll close this PR and delete this branch when we're done. For the time being, though, I'll keep this open until those two get merged jus tin case there's something here that I've missed by mistake.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -183,6 +240,156 @@ def test_interface_rename(self):
             {"NewKid": "Kid", "NewCharacter": "Character"}, renamed_schema.reverse_name_map
         )
 
+    # def test_suppress_interface_implementation(self):
```
bojanserafimov, line 243: Any reason this is commented? If not implemented, you can use `xfail` (search for examples in this repo).

LWprogramming:
Converting to draft because PRs 858 and 859 (plus a bit more work with enums and interfaces) will supersede the work here.

PR #835, opened 2020-05-28
Add missing pytest.mark.slow annotations

pmantica1:
These tests actually query the OrientDB database to generate the schema graph.

### DIFF
--- a/graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
+++ b/graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
@@ -1026,6 +1026,7 @@ def _make_schema_info_and_get_filter_selectivity(
     )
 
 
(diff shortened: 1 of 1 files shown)


LWprogramming:
Hmm, this is strange-- you're getting the same CI error as I am. Is an issue we've dealt with in the past? The internet hasn't been super helpful when I looked this up

obi1kenobi:
`pipenv` released a new version in the last day or two, and we have a line that says `pip install --upgrade pipenv` in the script. Perhaps that's worth looking into as a potential culprit? It seems like `pipenv` may have a bug that causes it to incorrectly handle package extras.

If that does turn out to be the problem, I'd poke around [pipenv's GitHub](https://github.com/pypa/pipenv) for an existing open issue, or open one if none exist already.

PR #841, opened 2020-06-03
Type hint batch: compiler/ir_lowering_common/

MichaelaShtilmanMinkin:
Adding type hints per instructions in #688 

Most of the manual changes involved changing the paths in the imports and `Location` -> `LocationT` and `Any` -> `BasicBlockT`.

### DIFF
--- a/graphql_compiler/compiler/ir_lowering_common/common.py
+++ b/graphql_compiler/compiler/ir_lowering_common/common.py
@@ -1,9 +1,9 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Language-independent IR lowering and optimization functions."""
-from typing import Any, Dict, List, Optional, Set, Tuple, Union
-from typing_extensions import TypedDict
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 import six
+from typing_extensions import TypedDict
 
 from ...compiler.compiler_entities import BasicBlockT
 from ..blocks import (
@@ -28,7 +28,7 @@ from ..expressions import (
     TernaryConditional,
     TrueLiteral,
 )
-from ..helpers import BaseLocation, FoldScopeLocation, LocationT, validate_safe_string
+from ..helpers import BaseLocation, FoldScopeLocation, validate_safe_string
 from ..metadata import QueryMetadataTable
 
 
@@ -194,9 +194,9 @@ def optimize_boolean_expression_comparisons(ir_blocks: List[BasicBlock]) -> List
         elif expression.right == identity_literal and left_is_binary_composition:
             return expression.left
         elif expression.left == inverse_literal and right_is_binary_composition:
-            expression_to_rewrite = expression.right  # type: ignore
+            expression_to_rewrite = cast(BinaryComposition, expression.right)
         elif expression.right == inverse_literal and left_is_binary_composition:
-            expression_to_rewrite = expression.left  # type: ignore
+            expression_to_rewrite = cast(BinaryComposition, expression.left)
 
         if expression_to_rewrite is None:
             # We couldn't find anything to rewrite, return the expression as-is.
@@ -222,8 +222,7 @@ def optimize_boolean_expression_comparisons(ir_blocks: List[BasicBlock]) -> List
 def extract_folds_from_ir_blocks(
     ir_blocks: List[BasicBlock],
 ) -> Tuple[
-    Dict[FoldScopeLocation, List[BasicBlock]],
-    List[BasicBlock],
+    Dict[FoldScopeLocation, List[BasicBlock]], List[BasicBlock],
 ]:
     """Extract all @fold data from the IR blocks, and cut the folded IR blocks out of the IR.
 
@@ -343,11 +342,7 @@ def extract_optional_location_root_info(
 
 
 SimpleOptionalLocationInfo = TypedDict(
-    "SimpleOptionalLocationInfo",
-    {
-        "inner_location": BaseLocation,
-        "edge_field": str,
-    }
+    "SimpleOptionalLocationInfo", {"inner_location": BaseLocation, "edge_field": str,}
 )
 
 
--- a/graphql_compiler/compiler/ir_lowering_common/common.py
+++ b/graphql_compiler/compiler/ir_lowering_common/common.py
@@ -163,8 +163,8 @@ def optimize_boolean_expression_comparisons(ir_blocks: List[BasicBlock]) -> List
         if not isinstance(expression, BinaryComposition):
             return expression
 
-        left_is_binary_composition = isinstance(expression.left, BinaryComposition)
-        right_is_binary_composition = isinstance(expression.right, BinaryComposition)
+        left_is_binary_composition: Optional[BinaryComposition] = expression.left if isinstance(expression.left, BinaryComposition) else None
+        right_is_binary_composition: Optional[BinaryComposition] = expression.right if isinstance(expression.right, BinaryComposition) else None
 
         if not left_is_binary_composition and not right_is_binary_composition:
             # Nothing to rewrite, return the expression as-is.
@@ -194,9 +194,9 @@ def optimize_boolean_expression_comparisons(ir_blocks: List[BasicBlock]) -> List
         elif expression.right == identity_literal and left_is_binary_composition:
             return expression.left
         elif expression.left == inverse_literal and right_is_binary_composition:
-            expression_to_rewrite = cast(BinaryComposition, expression.right)
+            expression_to_rewrite = right_is_binary_composition
         elif expression.right == inverse_literal and left_is_binary_composition:
-            expression_to_rewrite = cast(BinaryComposition, expression.left)
+            expression_to_rewrite = left_is_binary_composition
 
         if expression_to_rewrite is None:
             # We couldn't find anything to rewrite, return the expression as-is.
--- a/graphql_compiler/compiler/ir_lowering_common/common.py
+++ b/graphql_compiler/compiler/ir_lowering_common/common.py
@@ -1,6 +1,6 @@
 # Copyright 2017-present Kensho Technologies, LLC.
 """Language-independent IR lowering and optimization functions."""
-from typing import Any, Dict, List, Optional, Tuple, cast
+from typing import Any, Dict, List, Optional, Tuple
 
 import six
 from typing_extensions import TypedDict
@@ -186,13 +186,6 @@ def optimize_boolean_expression_comparisons(ir_blocks: List[BasicBlock]) -> List
             return expression
 
         expression_to_rewrite: Optional[BinaryComposition] = None
-        # N.B. cast statements occur in the last 2 elif statements because of the following:
-        # ```
-        # Incompatible types in assignment (expression has type "Expression", variable has type
-        # "Optional[BinaryComposition]")
-        # ```
-        # Given the logic at the time of writing this comment, we know that
-        # `right_is_binary_composition` says that right is BinaryComposition, not just Expression
         if expression.left == identity_literal and right_is_binary_composition:
             return expression.right
         elif expression.right == identity_literal and left_is_binary_composition:
--- a/graphql_compiler/compiler/ir_lowering_common/common.py
+++ b/graphql_compiler/compiler/ir_lowering_common/common.py
@@ -82,7 +82,7 @@ def lower_context_field_existence(
 ) -> List[BasicBlock]:
     """Lower ContextFieldExistence expressions into lower-level expressions."""
 
-    def regular_visitor_fn(expression: Any) -> Any:
+    def regular_visitor_fn(expression: Expression) -> Expression:
         """Expression visitor function that rewrites ContextFieldExistence expressions."""
         if not isinstance(expression, ContextFieldExistence):
             return expression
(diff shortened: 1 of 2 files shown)


graphql_compiler/compiler/ir_lowering_common/common.py
```diff
@@ -204,7 +210,15 @@ def visitor_fn(expression):
...
 
 
-def extract_folds_from_ir_blocks(ir_blocks):
+def extract_folds_from_ir_blocks(
+    ir_blocks: List[BasicBlockT],
+) -> Tuple[
+    Union[
+        Dict[FoldScopeLocation, List[BasicBlockT]],
+        Dict[FoldScopeLocation, List[Union[CoerceType, Filter, MarkLocation]]],
+    ],
```
obi1kenobi, line 219: This is a common thing the auto-generated type hints make, and I don't think it's correct here — it's overly specific.  `Dict[FoldScopeLocation, List[BasicBlockT]]` should be the first thing in the tuple, and not a union.

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -43,10 +50,10 @@ def translate_potential_location(location_translations, potential_location):
         return potential_location
 
 
-def make_location_rewriter_visitor_fn(location_translations):
+def make_location_rewriter_visitor_fn(location_translations: Dict[LocationT, LocationT]) -> Callable:
```
obi1kenobi, line 53: `Callable` is insufficiently specific, I think we should be able to tighten it up by specifying what args it takes and what its return type is.

graphql_compiler/compiler/ir_lowering_common/common.py
```diff
@@ -70,10 +75,12 @@ def to_match(self):
...
 
 
-def lower_context_field_existence(ir_blocks, query_metadata_table):
+def lower_context_field_existence(
+    ir_blocks: List[BasicBlockT], query_metadata_table: QueryMetadataTable
+) -> List[BasicBlockT]:
     """Lower ContextFieldExistence expressions into lower-level expressions."""
 
-    def regular_visitor_fn(expression):
+    def regular_visitor_fn(expression: Any) -> Any:
```
LWprogramming, line 83: Would narrowing this from `Any` to `Expression` work? It seems like the only place that uses this is `visit_and_update_expressions` which takes a `visitor_fn: Callable[[Expression], Expression]`

(similar logic might apply to other functions type-hinted as `Any`->`Any`)

MichaelaShtilmanMinkin:
Some stuff is still failing (will take a bit more time) but wanted to change the big changes thus far

graphql_compiler/compiler/ir_lowering_common/common.py
```diff
@@ -173,15 +181,22 @@ def visitor_fn(expression):
...
         if expression.left == identity_literal and right_is_binary_composition:
             return expression.right
         elif expression.right == identity_literal and left_is_binary_composition:
             return expression.left
         elif expression.left == inverse_literal and right_is_binary_composition:
-            expression_to_rewrite = expression.right
+            expression_to_rewrite = expression.right  # type: ignore
         elif expression.right == inverse_literal and left_is_binary_composition:
-            expression_to_rewrite = expression.left
+            expression_to_rewrite = expression.left  # type: ignore
```
MichaelaShtilmanMinkin, line 199: The way this is written made typing a tad weird here. I would consider re-writing some fo the logic here, but I think it is outside the scope of this diff.

obi1kenobi: Would a `cast` be appropriate instead of `type: ignore`? `type: ignore` prevents a good chunk of type checking onward from this point, whereas `cast` suppresses this typing issue but preserves the stated new type, and type checking may resume onward from here.

MichaelaShtilmanMinkin: Brilliant, didn't know about this.

graphql_compiler/compiler/ir_lowering_common/common.py
```diff
@@ -173,15 +181,22 @@ def visitor_fn(expression):
...
         if expression.left == identity_literal and right_is_binary_composition:
             return expression.right
         elif expression.right == identity_literal and left_is_binary_composition:
             return expression.left
         elif expression.left == inverse_literal and right_is_binary_composition:
-            expression_to_rewrite = expression.right
+            expression_to_rewrite = cast(BinaryComposition, expression.right)
         elif expression.right == inverse_literal and left_is_binary_composition:
-            expression_to_rewrite = expression.left
+            expression_to_rewrite = cast(BinaryComposition, expression.left)
```
obi1kenobi, line 199: This is fine, feel free to leave as is. I just wanted to give you an option to use an alternative approach, if you are dissatisfied with the current one, and would like to avoid the casts and the need for the comment.

The crux of the problem is that `mypy`'s static analysis is insufficiently powerful to see that `right_is_binary_composition == True` iff `isinstance(expression.right, BinaryComposition)`. This is true because:
- `right_is_binary_composition` is initialized with that precise statement on line 167 (in the new numbering), and
- `right_is_binary_composition` is never altered anywhere between that point and this point.

The easiest way to address this limitation of `mypy`'s is to make `left_binary_expression: Optional[BinaryComposition]` and `right_binary_expression: Optional[BinaryComposition]` variables, which are set to `expression.left` and `expression.right` if that side is a `BinaryComposition`, and `None` otherwise. Essentially, this is a type coercion operation, and in many languages, this operation is implemented via an `as` operator: `left_binary_expression = expression.left as BinaryComposition` would be the pseudo-Python way to express it here.

Then, the assignments on lines 197/199 don't need a cast, because they would read `expression_to_rewrite = left_binary_expression` which is valid no matter what, as the types of the two variables are definitionally the same.

Feel free to use your best judgment in deciding whether this alternative would be more or less readable than the current version.

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -43,10 +52,12 @@ def translate_potential_location(location_translations, potential_location):
...
 
 
-def make_location_rewriter_visitor_fn(location_translations):
+def make_location_rewriter_visitor_fn(
+    location_translations: Dict[Type[BaseLocation], Type[BaseLocation]]
+) -> Callable[[Expression], Expression]:
     """Return a visitor function that is able to replace locations with equivalent locations."""
 
-    def visitor_fn(expression):
+    def visitor_fn(expression: Any) -> Any:
```
obi1kenobi, line 60: This is also not a good type hint, even though `mypy` will accept it due to `Any` special-casing.

Here, we have the guarantee that the `expression` input type is of the same type as the return value, and they are both bounded from above by `Expression`. That means this is `ExpressionT -> ExpressionT`, and even line 57 could be updated to state that, to be even more clear.

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -43,10 +52,12 @@ def translate_potential_location(location_translations, potential_location):
         return potential_location
 
 
-def make_location_rewriter_visitor_fn(location_translations):
+def make_location_rewriter_visitor_fn(
+    location_translations: Dict[Type[BaseLocation], Type[BaseLocation]]
```
obi1kenobi, line 56: I think this is incorrect. The dict does not map types to types, it maps values of that type to values of the other type. The `Type` here means that things like `BaseLocation` or `Location` should be put into this dict, not their instances (note I didn't use parens next to the names).

Also, as above, `location_translations` can only hold `Location`-typed keys and values, though there was no way to know this from this function or module alone.
```suggestion
    location_translations: Dict[Location, Location]
```

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -17,7 +23,10 @@ def make_revisit_location_translations(query_metadata_table):
     return location_translations
 
 
-def translate_potential_location(location_translations, potential_location):
+def translate_potential_location(
+    location_translations: Dict[Type[BaseLocation], Type[BaseLocation]],
+    potential_location: Type[BaseLocation],
+) -> Type[BaseLocation]:
```
obi1kenobi, line 29:
```suggestion
    location_translations: Dict[Location, Location],
    potential_location: LocationT,
) -> LocationT:
```
See below why the `Type` bit is invalid. `LocationT` is because `potential_location` and the return value are guaranteed to be the same type.

It's also not clear in this zoomed-in context, but `location_translations` cannot contain keys or values of type `FoldScopeLocation` — only `Location` ones.

graphql_compiler/compiler/ir_lowering_common/common.py
```diff
@@ -249,7 +268,9 @@ def extract_folds_from_ir_blocks(ir_blocks):
     return folds, remaining_ir_blocks
 
 
-def extract_optional_location_root_info(ir_blocks):
+def extract_optional_location_root_info(
+    ir_blocks: List[BasicBlock],
+) -> Tuple[List[BaseLocation], Dict[BaseLocation, Tuple[BaseLocation, ...]]]:
```
LWprogramming, line 273: Selene raised a good point in my PR [here](https://github.com/kensho-technologies/graphql-compiler/pull/834#discussion_r436807755) that when type-hinting functions, we can also revise the Args and Returns descriptions in the docstring to describe only what the args and return values do (if it's not particularly inconvenient), rather than explicitly rewriting their type in the docstring.

graphql_compiler/compiler/ir_lowering_common/common.py
MichaelaShtilmanMinkin: I'm not sure why it's showing up as Any now, I was pretty sure I changed this locally and pushed the changes 🤔 ... I'll watch out for this next time I push to this PR.

graphql_compiler/compiler/ir_lowering_common/common.py
```diff
@@ -70,10 +77,12 @@ def to_match(self):
...
 
 
-def lower_context_field_existence(ir_blocks, query_metadata_table):
+def lower_context_field_existence(
+    ir_blocks: List[BasicBlock], query_metadata_table: QueryMetadataTable
+) -> List[BasicBlock]:
     """Lower ContextFieldExistence expressions into lower-level expressions."""
 
-    def regular_visitor_fn(expression):
+    def regular_visitor_fn(expression: Any) -> Any:
```
MichaelaShtilmanMinkin, line 85:
```suggestion
    def regular_visitor_fn(expression: Expression) -> Expression:
```

graphql_compiler/compiler/ir_lowering_common/common.py
MichaelaShtilmanMinkin: Makes sense. I think I'm going to avoid updating the Args/Returns sections and only focus on adding type hints in this PR, though, especially since the signature is staying functionally the same.

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
MichaelaShtilmanMinkin, line 60:
```suggestion
    def visitor_fn(expression: ExpressionT) -> ExpressionT:
```

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -43,10 +52,12 @@ def translate_potential_location(location_translations, potential_location):
         return potential_location
 
 
-def make_location_rewriter_visitor_fn(location_translations):
+def make_location_rewriter_visitor_fn(
+    location_translations: Dict[Type[BaseLocation], Type[BaseLocation]]
+) -> Callable[[Expression], Expression]:
```
MichaelaShtilmanMinkin, line 57:
```suggestion
) -> Callable[[ExpressionT], ExpressionT]:
```

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -1,11 +1,17 @@
...
-from ..helpers import FoldScopeLocation, Location
+from ...compiler.expressions import ExpressionT
+from ..helpers import FoldScopeLocation, Location, LocationT
+from ..metadata import QueryMetadataTable
 
 
-def make_revisit_location_translations(query_metadata_table):
+def make_revisit_location_translations(
+    query_metadata_table: QueryMetadataTable,
+) -> Dict[LocationT, LocationT]:
```
obi1kenobi, line 14:
```suggestion
) -> Dict[Location, Location]:
```
This is not apparent here, but will become apparent once that `from ..metadata` import up there is typed.

graphql_compiler/compiler/ir_lowering_common/location_renaming.py
```diff
@@ -1,11 +1,22 @@
...
 import six
 
+from ...compiler.expressions import ExpressionT
 from ..helpers import FoldScopeLocation, Location
+from ..metadata import QueryMetadataTable
+
+
+# This type exists in order to reduce the scope of what is allowed to be translated
+# since LocationT was not sufficiently specific.
+TranslatedLocationT = TypeVar("TranslatedLocationT", Location, FoldScopeLocation)
```
obi1kenobi, line 14: If you do end up opening a `mypy` issue about this, please link to the issue here as a courtesy to the next person that tries to do the same thing :)

PR #844, opened 2020-06-05
Implement filter inside a fold for MSSQL

chewselene:
I also updated the query for `filter_within_fold_scope` because the old query also included multiple outputs. We have separate tests for multiple folded outputs and it seemed like a test called `filter_within_fold_scope` should test just that functionality. I can add in a test for both filtering and having multiple folded outputs if that is desirable, but we definitely want separate tests for each these functionalities since multiple folded outputs are not implemented for MSSQL. Note that updating this test is causing a snapshot test to fail.

### DIFF
--- a/graphql_compiler/tests/test_compiler.py
+++ b/graphql_compiler/tests/test_compiler.py
@@ -7092,6 +7092,95 @@ class CompilerTests(unittest.TestCase):
             expected_postgresql,
         )
 
+    def test_filter_and_multiple_outputs_within_fold_scope(self):
+        test_data = test_input_data.filter_and_multiple_outputs_within_fold_scope()
+
+        expected_match = """
+                    SELECT
+                        $Animal___1___out_Animal_ParentOf.description AS `child_descriptions`,
+                        $Animal___1___out_Animal_ParentOf.name AS `child_list`,
+                        Animal___1.name AS `name`
+                    FROM (
+                        MATCH {{
+                            class: Animal,
+                            as: Animal___1
+                        }}
+                        RETURN $matches
+                    ) LET
+                        $Animal___1___out_Animal_ParentOf =
+                            Animal___1.out("Animal_ParentOf")[(name = {desired})].asList()
+                """
+        expected_gremlin = """
+            g.V('@class', 'Animal')
+            .as('Animal___1')
+            .transform{it, m -> new com.orientechnologies.orient.core.record.impl.ODocument([
+                child_descriptions: (
+                    (m.Animal___1.out_Animal_ParentOf == null) ? [] : (
+                        m.Animal___1.out_Animal_ParentOf
+                         .collect{entry -> entry.inV.next()}
+                         .findAll{entry -> (entry.name == $desired)}
+                         .collect{entry -> entry.description}
+                    )
+                ),
+                child_list: (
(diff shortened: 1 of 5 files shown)


LWprogramming:
In this case it'd be ok to update the snapshot test, right? We expected the test results to change since the input changed.

chewselene:
> In this case it'd be ok to update the snapshot test, right? We expected the test results to change since the input changed.

Yep, the snapshot will need to be updated. Just wanted to get an idea if we want to keep a test with multiple outputs and filters in a fold before I go and attempt to update the snapshot. If keeping such a test is desirable, I can make this snapshot test be for `test_filter_and_multiple_outputs_within_fold_scope` without updating the rows returned from the query. If we're eliminating this test entirely, I'll update the snapshot of rows returned.

obi1kenobi:
> Just wanted to get an idea if we want to keep a test with multiple outputs and filters in a fold before I go and attempt to update the snapshot. If keeping such a test is desirable, I can make this snapshot test be for `test_filter_and_multiple_outputs_within_fold_scope` without updating the rows returned from the query.

Since returning multiple outputs from inside a fold can be something that is or is not implemented depending on the particular choice of database, I'd say let's keep a test for it around. That way, if we support this in the future for MSSQL, we'll be able to use that test to verify we implemented it correctly.

chewselene:
I updated the snapshot test name in `tests/snapshot_tests/test_orientdb_match_query.py`, which caused me to need to update the snapshot in `tests/snapshot_tests/snapshots/snap_test_orientdb_match_query.py` with the new name. Then I ran our linting script. It auto formatted a bunch of stuff and now the diff for `snap_test_orientdb_match_query.py` is **huge**. Not sure if that is a problem/how to avoid that autoformatting?

graphql_compiler/tests/test_compiler.py
```diff
@@ -7006,8 +7005,111 @@ def test_filter_within_fold_scope(self):
...
+                    FROM (
+                        MATCH {{
+                            class: Animal,
+                            as: Animal___1
+                        }}
+                        RETURN $matches
+                    ) LET
+                        $Animal___1___out_Animal_ParentOf =
+                            Animal___1.out("Animal_ParentOf")[(name = {desired})].asList()
+                """
```
obi1kenobi, line 7112: nit: Any particular reason for the extra indentation, compared to other `expected_match` assignments?

graphql_compiler/tests/test_input_data.py
```diff
@@ -2344,11 +2344,37 @@ def filter_within_fold_scope() -> CommonTestData:  # noqa: D103
...
+def filter_and_multiple_outputs_within_fold_scope() -> CommonTestData:  # noqa: D103
+    graphql_input = """{
+            Animal {
+                name @output(out_name: "name")
+                out_Animal_ParentOf @fold {
+                    name @filter(op_name: "=", value: ["$desired"]) @output(out_name: "child_list")
+                    description @output(out_name: "child_descriptions")
+                }
+            }
+        }"""
```
obi1kenobi, line 2377: same here, seems like there is one extra step of indentation compared to equivalent code in surrounding test cases

graphql_compiler/tests/test_compiler.py
chewselene: this is some unintentional copy/pasting... #846 to fix.

PR #847, opened 2020-06-11
Type-hint rename_schema.py

LWprogramming:
Step 1 as outlined in [PR 834](https://github.com/kensho-technologies/graphql-compiler/pull/834#issuecomment-641549327)

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -199,7 +199,7 @@ class RenameSchemaTypesVisitor(Visitor):
             "UnionTypeDefinitionNode",
         }
     )
-    type_hint_rename_types = Union[
+    RenameTypes = Union[
         EnumTypeDefinitionNode,
         InterfaceTypeDefinitionNode,
         NamedTypeNode,
@@ -211,14 +211,14 @@ class RenameSchemaTypesVisitor(Visitor):
         # introspection for Unions. Its maintainers recently merged a PR that allows for this, but
         # the changes are more recent than the latest mypy release. PR here:
         # https://github.com/python/mypy/pull/8779
-        type_hint_rename_types_names = frozenset(
-            cls.__name__ for cls in get_args(type_hint_rename_types)
+        RenameTypes_names = frozenset(
+            cls.__name__ for cls in get_args(RenameTypes)
         )
-        if type_hint_rename_types_names != rename_types:
+        if RenameTypes_names != rename_types:
             raise AssertionError(
-                f"type_hint_rename_types and rename_types don't contain the same "
-                f"types. type_hint_rename_types is "
-                f"{type_hint_rename_types_names} and rename_types is "
+                f"RenameTypes and rename_types don't contain the same "
+                f"types. RenameTypes is "
+                f"{RenameTypes_names} and rename_types is "
                 f"{rename_types}"
             )
 
@@ -239,7 +239,7 @@ class RenameSchemaTypesVisitor(Visitor):
         self.scalar_types = frozenset(scalar_types)
         self.builtin_types = frozenset({"String", "Int", "Float", "Boolean", "ID"})
 
-    def _rename_name_and_add_to_record(self, node: type_hint_rename_types) -> Node:
+    def _rename_name_and_add_to_record(self, node: RenameTypes) -> Node:
         """Change the name of the input node if necessary, add the name pair to reverse_name_map.
 
         Don't rename if the type is the query type, a scalar type, or a builtin type.
@@ -301,7 +301,7 @@ class RenameSchemaTypesVisitor(Visitor):
         elif node_type in self.rename_types:
             # Rename node, put name pair into record
             renamed_node = self._rename_name_and_add_to_record(
-                cast(RenameSchemaTypesVisitor.type_hint_rename_types, node)
+                cast(RenameSchemaTypesVisitor.RenameTypes, node)
             )
             if renamed_node is node:  # Name unchanged, continue traversal
                 return None
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -1,6 +1,6 @@
 # Copyright 2019-present Kensho Technologies, LLC.
 from collections import namedtuple
-from typing import Any, Dict, List, Optional, Set, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union, cast, get_args
 
 from graphql import (
     DocumentNode,
@@ -206,6 +206,21 @@ class RenameSchemaTypesVisitor(Visitor):
         ObjectTypeDefinitionNode,
         UnionTypeDefinitionNode,
     ]
+    if not TYPE_CHECKING:
+        # Unfortunately, the current version of mypy (version >=0.750,<1) doesn't allow us to do
+        # introspection for Unions. Its maintainers recently merged a PR that allows for this, but
+        # the changes are more recent than the latest mypy release. PR here:
+        # https://github.com/python/mypy/pull/8779
+        type_hint_rename_types_names = frozenset(
+            cls.__name__ for cls in get_args(type_hint_rename_types)
+        )
+        if type_hint_rename_types_names != rename_types:
+            raise AssertionError(
+                f"type_hint_rename_types and rename_types don't contain the same "
+                f"types. type_hint_rename_types is "
+                f"{type_hint_rename_types_names} and rename_types is "
+                f"{rename_types}"
+            )
 
     def __init__(self, renamings: Dict[str, str], query_type: str, scalar_types: Set[str]) -> None:
         """Create a visitor for renaming types in a schema AST.
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -37,7 +37,7 @@ RenamedSchemaDescriptor = namedtuple(
 )
 
 
-def rename_schema(ast: DocumentNode, renamings: Dict[str, str]) -> RenamedSchemaDescriptor:
+def rename_schema(schema_ast: DocumentNode, renamings: Dict[str, str]) -> RenamedSchemaDescriptor:
     """Create a RenamedSchemaDescriptor; types and query type fields are renamed using renamings.
 
     Any type, interface, enum, or fields of the root type/query type whose name
@@ -46,9 +46,9 @@ def rename_schema(ast: DocumentNode, renamings: Dict[str, str]) -> RenamedSchema
     belonging to the root/query type will never be renamed.
 
     Args:
-        ast: Represents a valid schema that does not contain extensions, input object definitions,
-             mutations, or subscriptions, whose fields of the query type share the same name as the
-             types they query. Not modified by this function
+        schema_ast: Represents a valid schema that does not contain extensions, input object
+                    definitions, mutations, or subscriptions, whose fields of the query type share
+                    the same name as the types they query. Not modified by this function
         renamings: Maps original type/field names to renamed type/field names. Type or query type
                    field names that do not appear in the dict will be unchanged. Any dict-like
                    object that implements get(key, [default]) may also be used
@@ -70,14 +70,14 @@ def rename_schema(ast: DocumentNode, renamings: Dict[str, str]) -> RenamedSchema
         - SchemaNameConflictError if there are conflicts between the renamed types or fields
     """
     # Check input schema satisfies various structural requirements
-    check_ast_schema_is_valid(ast)
+    check_ast_schema_is_valid(schema_ast)
 
-    schema = build_ast_schema(ast)
+    schema = build_ast_schema(schema_ast)
     query_type = get_query_type_name(schema)
     scalars = get_scalar_names(schema)
 
     # Rename types, interfaces, enums
-    ast, reverse_name_map = _rename_types(ast, renamings, query_type, scalars)
+    schema_ast, reverse_name_map = _rename_types(schema_ast, renamings, query_type, scalars)
     reverse_name_map_changed_names_only = {
         renamed_name: original_name
         for renamed_name, original_name in six.iteritems(reverse_name_map)
@@ -85,25 +85,25 @@ def rename_schema(ast: DocumentNode, renamings: Dict[str, str]) -> RenamedSchema
     }
 
     # Rename query type fields
-    ast = _rename_query_type_fields(ast, renamings, query_type)
+    schema_ast = _rename_query_type_fields(schema_ast, renamings, query_type)
     return RenamedSchemaDescriptor(
-        schema_ast=ast,
-        schema=build_ast_schema(ast),
+        schema_ast=schema_ast,
+        schema=build_ast_schema(schema_ast),
         reverse_name_map=reverse_name_map_changed_names_only,
     )
 
 
 def _rename_types(
-    ast: DocumentNode, renamings: Dict[str, str], query_type: str, scalars: Set[str]
+    schema_ast: DocumentNode, renamings: Dict[str, str], query_type: str, scalars: Set[str]
 ) -> Tuple[DocumentNode, Dict[str, str]]:
     """Rename types, enums, interfaces using renamings.
 
     The query type will not be renamed. Scalar types, field names, enum values will not be renamed.
 
-    The input AST will not be modified.
+    The input schema AST will not be modified.
 
     Args:
-        ast: The schema that we're returning a modified version of
+        schema_ast: The schema that we're returning a modified version of
         renamings: Maps original type/interface/enum name to renamed name. Any name not in the dict
                    will be unchanged
         query_type: Name of the query type, e.g. 'RootSchemaQuery'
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -185,37 +199,41 @@ class RenameSchemaTypesVisitor(Visitor):
             "UnionTypeDefinitionNode",
         }
     )
-
-    def __init__(self, renamings, query_type, scalar_types):
+    type_hint_rename_types = Union[
```
LWprogramming, line 202: this looks weird since we just defined `rename_types` above, but so far it's the only way I've been able to make the type-hints work out. [See here](https://github.com/kensho-technologies/graphql-compiler/pull/834#discussion_r436979070)

bojanserafimov: One alternative I can think of is:
```
from typing import get_args
rename_types = {cls.__name__ for cls in get_args(type_hint_rename_types)}
```

Not sure if that's any better

obi1kenobi: This is better than the `get_args()` alternative because this is resolvable without running the code. That means `mypy` will be able to check it, unlike @bojanserafimov's offered alternative — remember that `mypy` reads the Python AST but doesn't actually execute the code.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -82,25 +93,26 @@ def rename_schema(ast, renamings):
...
     The input AST will not be modified.
 
     Args:
-        ast: Document, the schema that we're returning a modified version of
-        renamings: Dict[str, str], mapping original type/interface/enum name to renamed name. If
-                   a name does not appear in the dict, it will be unchanged
-        query_type: str, name of the query type, e.g. 'RootSchemaQuery'
-        scalars: Set[str], the set of all scalars used in the schema, including user defined
-                 scalars and and used builtin scalars, excluding unused builtins
+        ast: The schema that we're returning a modified version of
```
obi1kenobi, line 106: Interesting, I forgot this is a schema AST — can we rename it to `schema_ast` while we're at it?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -185,37 +199,41 @@ class RenameSchemaTypesVisitor(Visitor):
...
         self.renamings = renamings
-        self.reverse_name_map = {}  # Dict[str, str], from renamed type name to original type name
+        self.reverse_name_map: Dict[str, str] = {}  # from renamed type name to original type name
         # reverse_name_map contains all types, including those that were unchanged
         self.query_type = query_type
         self.scalar_types = frozenset(scalar_types)
         self.builtin_types = frozenset({"String", "Int", "Float", "Boolean", "ID"})
 
-    def _rename_name_and_add_to_record(self, node):
+    def _rename_name_and_add_to_record(self, node: type_hint_rename_types) -> Node:
```
obi1kenobi, line 227: The `Returns` section of the docstring (`Node object, identical to the input node, except with possibly a new name`) seems to suggest a tighter type bound here:
```
NodeT = TypeVar("NodeT", bound=Node)
def _rename_name_and_add_to_record(self, node: NodeT) -> NodeT:
```

Thoughts?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -185,37 +199,41 @@ class RenameSchemaTypesVisitor(Visitor):
...
         Don't rename if the type is the query type, a scalar type, or a builtin type.
 
         The input node will not be modified. reverse_name_map may be modified.
 
         Args:
-            node: EnumTypeDefinitionNode, InterfaceTypeDefinitionNode, NamedTypeNode,
-                  ObjectTypeDefinitionNode, or UnionTypeDefinitionNode. An object representing an
-                  AST component, containing a .name attribute corresponding to an AST node of type
-                  NameNode.
+            node: An object representing an AST component, containing a .name attribute
```
obi1kenobi, line 235: Double-check the rest of the codebase, I'm not sure we lead with `a/an` and whether we capitalize argument and return value definitions unless they are types whose names are capitalized. Let's aim to conform to the bulk of the code.

graphql_compiler/schema_transformation/rename_schema.py
bojanserafimov: I agree that it's better, just not for that reason. Mypy can still infer the type of `rename_types`, since the return type of `__name__` is `str`.

obi1kenobi: Fair point.

@LWprogramming btw, let's avoid the `cast()` call that you linked to, and instead explicitly assert that the type of `node` is indeed one of the types in the union:
```
if not isinstance(node, <union types here>):
    raise AssertionError(<appropriate error message>)
# Here, mypy knows that if this line is reached, the type of "node" must have been appropriate, and no cast is necessary.
```

Remember that `cast()` is implicitly a "mypy, trust me, I know what I'm doing" statement. It is almost always better to use an explicit assertion rather than a mypy suppression like this.

LWprogramming: I tried a bunch of different ways to implement this, but ran into problems with deciding which properties I wanted to prioritize. Here are the ones that I'm thinking about:

* define the rename-types with as little repetition as possible because they basically refer to the same idea

* ensure that it's quick to check if a node is in `noop_types` or `rename_types`. If, for instance, they weren't `frozenset`s but rather lists instead, there might be performance issues with manipulating very large schemas because we'd need to check many node types for each node in the schema AST.

* type-hint `_rename_name_and_add_to_record` so that `mypy` knows the necessary attributes exist (i.e. `node` has  `name` attribute)

* avoid needing to call `cast` by raising an assertion error if the node is the wrong type

The problem is that avoiding `cast` while being able to type-hint properly seems to require even more repetition of the same idea in code. I've tried several different ways to specify `<union types here>` so that we can write the assertion: passing in the `type_hint_rename_types` Union, passing in a set of the classes involved in `rename_types` (not the class names but the classes themselves), and passing in a tuple of the classes involved. This is the only thing that worked:

```
rename_types_classes = (
    EnumTypeDefinitionNode,
    InterfaceTypeDefinitionNode,
    NamedTypeNode,
    ObjectTypeDefinitionNode,
    UnionTypeDefinitionNode,
)
...
elif node_type in self.rename_types:
    # Rename node, put name pair into record
    if not isinstance(node, self.rename_types_classes):
        raise AssertionError
```

This didn't:
```
rename_types_classes = tuple(rename_types)
```

And this didn't either:
```
if not isinstance(node, tuple(self.rename_types)):
```

In other words, the only way I've found that satisfies `mypy` while avoiding `cast` involves making another representation of the same concept.

I could attempt to reduce this duplication by using Bojan's suggestion to dynamically compute `rename_types`, but that doesn't solve the problem of needing

* a `Union` for the type hints

* a tuple to avoid `cast()`

* a set to make membership checking quick

obi1kenobi: To be clear, I'm not opposed to using `cast` if it's accompanied by a dynamically-checked assertion of the property that `cast` is used to assure `mypy` of. In that case, `cast` is simply reiterating to `mypy` a fact that if `mypy` may in the future be able to deduce from the runtime check via more sophisticated code analysis.

My concern was simply around using `cast` in a situation where we "know" a certain fact to be "true" but aren't actually stating our certainty under penalty of raised `AssertionError`.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I've picked a few files in various places and it varies-- some lines lead with `a/an/the` while others don't and I haven't found a clear pattern. Generally, the first word isn't capitalized unless it's a proper noun like `GraphQL` so I'll update that

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: According to the bound documentation, it seems like this code just says the argument and return have to be a subclass of `Node` which doesn't seem to be a tighter bound. Depending on what we end up deciding on in the [discussion above](https://github.com/kensho-technologies/graphql-compiler/pull/847#discussion_r438832613), it looks the same as just using the type from there. Am I missing something here?

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: In this case it might be fine to not have the assertion, since the [line of code](https://github.com/kensho-technologies/graphql-compiler/pull/847/files#diff-3b859f46e97b0af834db0b3b7de08658R286) where we use `cast` is inside a conditional saying that the node type is in `rename_types`. In a sense, the conditional functions as the assertion because we won't enter the path unless it's a type to rename. We wouldn't need to assert that `node_type` is in `rename_types` because the condition says exactly that already.

Unfortunately, this does depend on `rename_types` and `type_hint_rename_types` changing in parallel, but `mypy` complains about types at runtime (like how we described earlier in this thread) when I try to avoid the duplication by dynamically defining `rename_types`.

There might be some way to guarantee that `rename_types` and `type_hint_rename_types` have the same information-- maybe something like [this](https://stackoverflow.com/questions/45957615/check-a-variable-against-union-type-at-runtime-in-python-3-6), although it does look pretty hacky. If we do this check in `__init__`, we should be good to go, right?

obi1kenobi: I think `mypy` should at least in part enforce that those two change together, otherwise there would be a type mismatch, no?

Dynamic check at module level is probably okay too, if you feel that's necessary.

If I may nit a bit, `type_hint_rename_types` doesn't sound like the name of a type alias though :) How about `RenameTypes` instead, or something like that?

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: It's unfortunate that we have files that lead with `a/an/the` 😄  Let's try to not add them, since they generally pad the line length without communicating much useful info.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Unfortunately `mypy` doesn't enforce this automatically.

[Right now](https://github.com/python/mypy/issues/5354) it effectively doesn't allow us to do any introspection for `Unions`, but its maintainers did merge a PR [very recently](https://github.com/python/mypy/pull/8779) that lets us avoid the error `The type alias to Union is invalid in runtime context`. Unfortunately it's more recent than the latest `mypy` release so for now I'll tell `mypy` to ignore the dynamic check.

LWprogramming: Unfortunate... just discovered `get_args` is only available in Python 3.8, not 3.6 or 3.7

bojanserafimov: Can you still use `RenameTypes.__args__`? Gives me the same result so far, but not sure if the spec is the same.

LWprogramming: That seems to work in [3.8](https://github.com/python/cpython/blob/3.8/Lib/typing.py#L1304) and [3.7](https://github.com/python/cpython/blob/3.7/Lib/typing.py#L618) (although not documented as far as I can tell), but the code for [3.6](https://github.com/python/cpython/blob/3.6/Lib/typing.py#L686) seems significantly different and I haven't been able to tell if this would work.

obi1kenobi: Let's just define them twice and add a comment to keep them the same. While it's important to deduplicate where possible and makes sense, in this case I think we've put in more than a reasonable amount of work to try to make that happen.

In the end, we have to make sure the cure isn't worse than the disease, and special-casing by Python version / implementing our own `get_args` / similar options seem much worse than just a tiny bit of duplication in a place that we hopefully won't need to change anytime soon — the GraphQL AST is unlikely to get new AST nodes anytime soon.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -190,18 +190,32 @@ class RenameSchemaTypesVisitor(Visitor):
...
+    # rename_types must be a set of strings corresponding to the names of the classes in
+    # RenameTypes. The duplication exists because introspection for Unions via typing.get_args()
+    # doesn't exist until Python 3.8. In Python 3.8, this would be a valid way to define
+    # rename_types:
+    # rename_types = frozenset(cls.__name__ for cls in get_args(RenameTypes))  # type: ignore
+    # Note: even with Python 3.8, the mypy version at the time of writing (version 0.770) doesn't
+    # allow for introspection for Unions. mypy's maintainers recently merged a PR
+    # (https://github.com/python/mypy/pull/8779) that permits this line of code, but did so after
+    # the mypy 0.770 release. If we do end up removing the duplication at a later point but not
+    # update the mypy version, we'd need to ignore it (as shown in the in-line comment).
```
obi1kenobi, line 202: Love the detailed comment, this is great!

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -35,12 +46,12 @@ def rename_schema(ast, renamings):
...
-             the same name as the types they query. Not modified by this function
-        renamings: Dict[str, str], mapping original type/field names to renamed type/field names.
-                   Type or query type field names that do not appear in the dict will be unchanged.
-                   Any dict-like object that implements get(key, [default]) may also be used
+        schema_ast: represents a valid schema that does not contain extensions, input object
+                    definitions, mutations, or subscriptions, whose fields of the query type share
+                    the same name as the types they query. Not modified by this function
+        renamings: maps original type/field names to renamed type/field names. Type or query type
+                   field names that do not appear in the dict will be unchanged. Any dict-like
+                   object that implements get(key, [default]) may also be used
```
obi1kenobi, line 54: The type hint of `Dict[str, str]` will actually prohibit this "dict-like" bit from being true. I think you are looking for `Mapping` instead of `Dict` here: https://docs.python.org/3/library/typing.html#typing.Mapping

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: We can tighten this by setting the return type to be `RenameTypes`. Unfortunately, the current linter version complains about this [and it was fixed recently](https://github.com/PyCQA/pyflakes/pull/448), so I'm currently looking into updating our dependencies. It seems like in the process of updating the Pipfile, we updated `pylint` and the new version can take a really long time to run.

I've committed the change here to see if linting works with Travis CI on Github-- if it does, I'll revert the change here and split the version bump into a separate PR.

LWprogramming: Turns out making it a module-level attribute did indeed work! Updating that now and then the PR will be good to go.

LWprogramming: Will update with generic typing--just read the discussion on the difference between this and the existing type-hint

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -59,77 +90,80 @@ def rename_schema(ast, renamings):
...
-                   does not appear in the dict, it will be unchanged
-        query_type: string, name of the query type, e.g. 'RootSchemaQuery'
+        schema_ast: schema that we're returning a modified version of
+        renamings: maps original query type field name to renamed name. Any name not in the dict
+                   will be unchanged
+        query_type: name of the query type, e.g. 'RootSchemaQuery'
 
     Returns:
-        DocumentNode, representing the modified version of the input schema AST
+        Modified version of the input schema AST
```
obi1kenobi, line 162:
```suggestion
        modified version of the input schema AST
```
Also, is this type guaranteed to be the same as the input type? If so, perhaps another opportunity for a generic?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -186,36 +230,35 @@ class RenameSchemaTypesVisitor(Visitor):
         }
```
obi1kenobi, line 230: Do we specifically need this to be a class attribute here? If not, can we move it to module-level and make it sit right adjacent to the `Union` definition that it duplicates?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -186,36 +230,35 @@ class RenameSchemaTypesVisitor(Visitor):
...
     )
 
-    def __init__(self, renamings, query_type, scalar_types):
+    def __init__(
+        self, renamings: Mapping[str, str], query_type: str, scalar_types: Set[str]
```
obi1kenobi, line 234:
```suggestion
        self, renamings: Mapping[str, str], query_type: str, scalar_types: AbstractSet[str]
```
https://docs.python.org/3/library/typing.html#typing.AbstractSet will allow both `set()` and `frozenset()`. This variant only allows `set()` specifically.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -312,3 +378,4 @@ def enter_field_definition(self, node, *args):
             else:  # Make copy of node with the changed name, return the copy
                 field_node_with_new_name = get_copy_of_node_with_new_name(node, new_field_name)
                 return field_node_with_new_name
+        return None
```
obi1kenobi, line 381:
```suggestion

        return None
```

obi1kenobi: If de-indenting two or more levels, I find it helps readability to add a blank line in between.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Nothing inherits from DocumentNode so the current annotation effectively does the same thing as a generic here

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: It's specifically used for `RenameSchemaTypesVisitor`, as it explicitly leaves out renaming the schema query type (i.e. `SchemaQuery`) and has `RenameQueryTypeFieldsVisitor` do that instead. Moving it to module-level could be confusing considering that we do actually modify the schema query type, just not in this particular class.

PR #852, opened 2020-06-22
Remove dataclasses backport

obi1kenobi:
We used to use `dataclasses==0.6` for the Pipfile because `dataclasses==0.7` (correctly) doesn't install on Python 3.7+ because there's no point in installing a backport for functionality that exists as a built-in, and yet we used the lockfile to install for all Pythons.

I think this is a better approach, where we use the lockfile to prime the environment in such a way that the last `pip install` will decide that a bunch of the dependencies are already installed, and will only install the Python version-specific packages in that step.


Pipfile
```diff
@@ -37,7 +37,6 @@ sphinx = ">=1.8,<2"
 [packages]  # Make sure to keep in sync with setup.py requirements.
 arrow = ">=0.15.0,<1"
 funcy = ">=1.7.3,<2"
-dataclasses = ">=0.6"
```
pmantica1, line 40: A con here is that if someone was using python3.6 to develop locally then they'll get import error. 

Technically, the Pipfile requires python 3.8, but I think that pipenv still works if you use an incorrect version of python. Also even though our `CONTRIBUTING.rst` information states that developers should use python3.8 for local development, our `TROUBLESHOOTING.rst` has documentation that makes it sound that you can use different python versions:
"# Please modify command to match the version of python you use for development."
"sudo apt-get install python3.6-dev"

If we are going to proceed with this PR, do you mind updating the `TROUBLESHOOTING.rst` and the readthedocs equivalent to make it clearer that developers are only supposed to use python3.8 for local development?

obi1kenobi: Updated all docs as suggested.

pmantica1: Thanks!

PR #853, opened 2020-06-22
Pin pylint to 2.4.x version

obi1kenobi:
Pylint 2.5.x seems to take an extremely long time to finish linting, and may even be hanging -- hard to tell since Travis cuts off the job as "hung" after 10 min with no output.

Pin pylint to 2.4.x while we investigate.


obi1kenobi:
I ran `pylint==2.5.3` for over two hours on the GraphQL compiler repo, and it still didn't terminate. My bet is on an infinite loop somewhere in the `pylint` code.

```
$ pylint --version
pylint 2.5.3
astroid 2.4.2
Python 3.8.5 (default, Jul 21 2020, 10:48:26) 
[Clang 11.0.3 (clang-1103.0.32.62)]

$ time pylint --rcfile=.pylintrc graphql_compiler/
^C
real    131m12.144s
user    131m9.018s
sys 0m2.527s
```

PR #855, opened 2020-06-23
Switch return value from None to IDLE

LWprogramming:
IDLE is a special return value specified in graphql-core's [visitor](https://github.com/graphql-python/graphql-core/blob/master/src/graphql/language/visitor.py#L34). Existing code returned None, but this would break when the special values get [redefined](https://github.com/graphql-python/graphql-core/issues/96).


LWprogramming:
Seems like something went wrong with the webhooks but the build is fine: https://travis-ci.org/github/kensho-technologies/graphql-compiler/builds/701290687

PR #858, opened 2020-06-23
Implement tests for type suppression

LWprogramming:
There are certain behaviors I haven't tested yet, such as suppressing enums and ensuring that attempting to suppress scalars doesn't actually remove the scalar from the schema, so I'll be adding tests for that there.


graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -183,6 +240,156 @@ def test_interface_rename(self):
             {"NewKid": "Kid", "NewCharacter": "Character"}, renamed_schema.reverse_name_map
         )
 
+    # def test_suppress_interface_implementation(self):
```
LWprogramming, line 243: commented out because not sure what to do with interfaces, see discussion in PR 859

PR #859, opened 2020-06-23
Implement type/union suppression

LWprogramming:
Update: I decided to cut scope for this PR and the other one with its tests to only focus on suppressing object types and unions to make this easier to read. 

Things not implemented here but coming soon:

* field suppression

* suppressing enums and interfaces-- makes more sense after implementing field suppression

* `renamings` error checking (e.g. if a value in `renamings` doesn't exist in the schema)

* looking into `reverse_name_map` optimization (as described below)

--

This PR implements:

* Suppressing object types, and unions

* Error-reporting for any illegal states that this can produce.

WIP: ~~suppressing enums and interfaces + various documentation. I suspect enums might already be taken care of, but I haven't confirmed it.~~ Also, basic linting, formatting, readability, better error messages, etc.

Tests are in PR 858 to keep the PRs from getting too big.

--

Questions:

* The code doesn't touch [user-defined scalars and enums](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema_transformation/rename_schema.py#L76), even if `renamings` includes it. What makes them different from types/ interfaces/ etc? Is it because we can't query for scalars and enums? I haven't been able to find anything that explicitly explains this.

* If we have a field of a type that we suppress, should we automatically suppress the field as well or raise a CascadingSuppressionError requiring the user explicitly suppress those fields as well (when we implement field suppression)? I implemented the latter because we chose to make things as explicit as possible but wonder if this might make it annoying if many fields are of a type that gets suppressed.

* [From here](https://github.com/kensho-technologies/graphql-compiler/pull/468#discussion_r313926425): "A concern was preserving an interface but not all of its implementations, since then a user could query for all instances of that interface and get back objects that are of types that don't exist in the schema." Did we ever decide on what to do?

--

Questions that don't affect end-user functionality

* [From here](https://github.com/kensho-technologies/graphql-compiler/pull/468#discussion_r313926425): what should we do if `renamings` attempts to rename or suppress a type/ interface/ etc. if that type doesn't appear in the schema. Should we simply ignore it (as we do now), or show a warning/ error? In addition, what if renaming attempts to do a 1-1 renaming from something to itself? Right now it doesn't change anything, but should we warn the user at least that this won't do anything (in case there was a typo)?

* What was the reasoning behind having [`reverse_name_map`](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema_transformation/rename_schema.py#L260) in `_rename_types` include non-renamed types? If it's not necessary anymore to store everything, I can refactor this to avoid problems when the schema gets very large.


obi1kenobi:
Great questions! Replies inline:

> Questions:
> 
> * The code doesn't touch [user-defined scalars and enums](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema_transformation/rename_schema.py#L76), even if `renamings` includes it. What makes them different from types/ interfaces/ etc? Is it because we can't query for scalars and enums? I haven't been able to find anything that explicitly explains this.

I'm not sure there's a particular good reason. The current code is a proof-of-concept-level prototype, and I wouldn't assume all the design decisions there are meant to be meaningful and cover all edge cases.

We should definitely not silently ignore `renamings` input. We should either error out if a user attempts to rename enum/scalar/input types, or actually implement that renaming. We don't have to pick the same decision for all of these special types (e.g. I'd error for `input` types for now at least), and we should be explicit about the combination we do and do not support.

> * If we have a field of a type that we suppress, should we automatically suppress the field as well or raise a CascadingSuppressionError requiring the user explicitly suppress those fields as well (when we implement field suppression)? I implemented the latter because we chose to make things as explicit as possible but wonder if this might make it annoying if many fields are of a type that gets suppressed.

Let's follow our overall strategy:
- Renaming types only renames types, and complains via errors if such renaming is illegal and would require doing more work first, such as field suppression.
- Offer a helper function that, for a given set of desired type renamings, calculates the exact "more work" needed and produces it in a form that is easy for a human to review and then feed into the functions that will apply the required schema changes such as the necessary field suppressions.

In a sense, we are designing a process that separates the "make diff" from the "apply diff" step:
- allows a human to specify type renamings
- get a reviewable and complete diff of the resulting changes to the schema as a whole
- review and sign off on it
- apply the prerequisite changes (field suppressions etc.)
- finally, apply the desired type renamings to get the final schema.

> * [From here](https://github.com/kensho-technologies/graphql-compiler/pull/468#discussion_r313926425): "A concern was preserving an interface but not all of its implementations, since then a user could query for all instances of that interface and get back objects that are of types that don't exist in the schema." Did we ever decide on what to do?

I've given this some thought, and the only thing that I can convince myself is safe is to require that all fields (both root schema query fields as well as edges) that point to that interface type are suppressed. The interface type can be kept purely to signal to a user that the remaining implementations share some common structure, but no query should be able to be constructed where a scope within that query is of that interface type while not also being a more-specific type (e.g. another interface that implements that one, or a type that implements that interface).

It seems to me that any approach other than this would have to deal with the "get back objects of types that don't exist in the schema" problem, and would require the compiler to do additional query rewriting such as adding `__typename @filter(op_name: "!=", value: ["$removed_type"])` clauses. While this is something that would definitely be possible to build, let's save it for a future version and implement the simpler approach first.

> * [From here](https://github.com/kensho-technologies/graphql-compiler/pull/468#discussion_r313926425): what should we do if `renamings` attempts to rename or suppress a type/ interface/ etc. if that type doesn't appear in the schema. Should we simply ignore it (as we do now), or show a warning/ error? In addition, what if renaming attempts to do a 1-1 renaming from something to itself? Right now it doesn't change anything, but should we warn the user at least that this won't do anything (in case there was a typo)?

Let's raise an error in both cases, and verify this with tests. These are both cases where the user may have either made a typo or may be operating under a mental model that doesn't match the schema the compiler is looking at, and we should let them know of this fact by raising an error.

Ideally, let's find _all_ such "invalid input" type problems with the specified renamings, and raise an error with all of them at once. If there are more than a couple such problems, it would be infuriating from a UX perspective to have to find and fix them one at a time. Let's aim for better UX than "infuriating" 😄 

> * What was the reasoning behind having [`reverse_name_map`](https://github.com/kensho-technologies/graphql-compiler/blob/master/graphql_compiler/schema_transformation/rename_schema.py#L260) in `_rename_types` include non-renamed types? If it's not necessary anymore to store everything, I can refactor this to avoid problems when the schema gets very large.

To be honest, I don't remember if there was a specific reason. I think we may have been attempting to err on the side of correctness rather than performance, and weren't worrying about the memory footprint. If it doesn't seem necessary, feel free to amend the code to not store everything, since we now know our schemas can be reasonably measured in GB of RAM 😄

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +117,148 @@ def rename_schema(
...
+                    union_member.name.value for union_member in union_type.types
+                ]
+            error_message_components.append(
+                "To fix this, you can suppress the union as well by adding `union_type: None` to "
+                "the `renamings` argument of `rename_schema`, for each value of `union_type` "
+                "described here. Note that adding suppressions may lead to other types, fields, "
+                "unions, etc. requiring suppression so you may need to iterate on this before "
+                "getting a legal schema."
+            )
+        raise CascadingSuppressionError("\n".join(error_message_components))
```
obi1kenobi, line 214: Consider, as a quality-of-life improvement (and perhaps in a subsequent PR), adding a way to programmatically read what went wrong that doesn't require parsing the error message text. When defining this exception, you get to specify what arguments its constructor takes, and what attributes it exposes — so you could in principle have a list of appropriately-typed values that describe what went wrong: for example, for each suppression, what additional steps must be taken to make it valid.

This would allow a pleasant UX workflow: attempt a renaming, observe it fail with constructive feedback, select which of the required changes you'd like to apply and for which you'd instead like to abandon the renaming, apply the prerequisite changes, then apply the renaming itself. The computer did almost all the work, and the user essentially only had to do the decision-making.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +117,148 @@ def rename_schema(
...
+        renamings: maps original field name to renamed name. If a name
+                   does not appear in the dict, it will be unchanged
+        query_type: name of the query type, e.g. 'RootSchemaQuery'
+
+    Raises:
+        - CascadingSuppressionError if a type suppression would require further suppressions
+        - NotImplementedError if renamings attempts to suppress an enum or an interface
+    """
+    _check_for_cascading_type_suppression(schema_ast, renamings, query_type)
+    _ensure_no_unsupported_suppression(schema_ast, renamings, query_type)
```
obi1kenobi, line 163: Perhaps also a good idea to ensure that the new names do not start with double underscores, since all such names are reserved for internal use by the GraphQL library and standard.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +117,148 @@ def rename_schema(
...
+            f"Type renamings mapped these schema interfaces to None: "
+            f"{visitor.unsupported_interface_suppressions}, attempting to suppress them. However, "
+            f"schema renaming has not implemented interface suppression yet."
+        )
+    if visitor.unsupported_interface_implementation_suppressions:
+        error_message_components.append(
+            f"Type renamings mapped these object types to None: "
+            f"{visitor.unsupported_interface_implementation_suppressions}, attempting to suppress "
+            f"them. Normally, this would be fine. However, these types each implement at least one "
+            f"interface and schema renaming has not implemented this particular suppression yet."
```
obi1kenobi, line 251: I'm confused about this error message. Am I reading it right, that types that implement any interface cannot be renamed? This is regardless of how the interface is used?

Consider writing a module-level docstring, perhaps, that explains (ideally with some small examples) which operations are and are not supported? I know we can't yet support renaming types that implement an interface when that interface is the exact type of a field, but I was surprised at this.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -173,6 +306,9 @@ def _rename_query_type_fields(
 
     Returns:
         modified version of the input schema AST
+
+    Raises:
+        - SchemaTransformError if renamings suppressed every type
```
obi1kenobi, line 311: Really love the thoroughness with these explicit `Raises` clauses! 💯

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -326,25 +471,27 @@ def enter(
...
 
 class RenameQueryTypeFieldsVisitor(Visitor):
-    def __init__(self, renamings: Mapping[str, str], query_type: str) -> None:
+    def __init__(self, renamings: Mapping[str, Optional[str]], query_type: str) -> None:
         """Create a visitor for renaming fields of the query type in a schema AST.
 
         Args:
-            renamings: maps original field name to renamed field name. Any name not in the dict will
-                       be unchanged
+            renamings: maps original field name to renamed field name or (for type suppression). Any
```
obi1kenobi, line 488: This line is a bit confusing and looks perhaps like an unintentional edit.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -326,25 +471,27 @@ def enter(
...
+            else:
+                # Name changed or suppressed, return new node, `visit` will make shallow copies
+                # along path
                 return renamed_node
         else:
             # All Node types should've been taken care of, this line should never be reached
             raise AssertionError('Unreachable code reached. Missed type: "{}"'.format(node_type))
 
 
 class RenameQueryTypeFieldsVisitor(Visitor):
```
obi1kenobi, line 483: Consider updating this name, since it no longer just renames and can now also remove.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -388,8 +541,170 @@ def enter_field_definition(
...
+                "self.current_type is None"
+            )
+        if self.current_type == field_type:
+            # Then node corresponds to a field belonging to type T that is also of type T.
+            # Therefore, we don't need to explicitly suppress the field as well and this should not
+            # raise errors.
+            return None
+        if self.current_type not in self.fields_to_suppress:
+            self.fields_to_suppress[self.current_type] = {}
+        self.fields_to_suppress[self.current_type][field_name] = field_type
```
obi1kenobi, line 622: You could also do `self.fields_to_suppress.setdefault(self.current_type, {})[field_name] = field_type` to do this as a one-liner. No particular preference for it one way or the other, this is fine.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -388,8 +541,170 @@ def enter_field_definition(
...
+        # Check if all the union members are suppressed.
+        for union_member in node.types:
+            union_member_type = get_ast_with_non_null_and_list_stripped(union_member).name.value
+            if self.renamings.get(union_member_type, union_member_type):
+                # Then at least one member of the union is not suppressed, so there is no cascading
+                # suppression error concern.
+                return None
+        if self.renamings.get(union_name):
+            # If the union is also suppressed, then nothing needs to happen here
+            return None
```
obi1kenobi, line 652: Is this intentional? This conditional looks like it's checking for the exact opposite of what the comment is saying.

Also, a small nit: it's somewhat nicer to be explicit and say we're checking for `None` in case there's ever a falsy value in there, like the empty string.

Which reminds me, do you ever check that the renamed names are non-empty, satisfy our naming rules (i.e. no illegal chars), and do not duplicate each other or other existing types?

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I looked at the helper function here and we can actually get away with just returning the helper's output (after a small change), so no need to actually come up with a good variable name here :)

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: The `check_type_name_is_valid` call does this, although the existing check only mentions one name at a time (as opposed to giving all illegal variable names at once). I can update this in a separate PR.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -67,22 +73,26 @@
 
 
 def rename_schema(
```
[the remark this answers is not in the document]
LWprogramming: For fields, I'll probably add to either an existing visitor or create a new one and call a function within here, rather than have two separate functions. I suppose we could then call it something more general like `manipulate_schema` but it might not be clear that we're allowing both suppression and renamings here.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I like that idea! Coming soon :)

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: For this PR, types that implement any interface cannot be suppressed (but [they can be renamed](https://github.com/kensho-technologies/graphql-compiler/blob/main/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py#L171)). I wanted to limit this PR's scope because of the tricky edge cases with interface renaming and there's already a good amount of code here. In a future PR I plan to allow this in some form.

Will document which operations are permitted!

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: fixed!

`check_type_name_is_valid` does check that the renamed names are valid and whether `renamings` duplicates anything, but checking against existing types only happens to a limited extent in `_rename_and_suppress_types`. I've noted all the improvements for error checking described here and plan to consolidate them all into `_validate_renamings` in a future PR just to keep this one at a manageable size.

LWprogramming:
Codecov failed because the tests are in a separate PR. However, when you add in the tests, my local codecov run yielded
`graphql_compiler/schema_transformation/rename_schema.py                   175      2    99%`
and overall: 
`TOTAL                                                                    8518    465    95%`
so it should be fine.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +130,149 @@ def rename_schema(
...
+    """Validate the renamings argument before attempting to rename the schema.
+
+    Check for fields with suppressed types or unions whose members were all suppressed. Also,
+    confirm renamings contains no enums, interfaces, or interface implementation suppressions
+    because that hasn't been implemented yet.
+
+    The input AST will not be modified.
+
+    Args:
+        schema_ast: schema that we're returning a modified version of
```
obi1kenobi, line 166: copy-pasta? We don't seem to be returning anything, let alone a modified AST.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +130,149 @@ def rename_schema(
...
+        renamings: maps original type name to renamed name or None (for type suppression). Any name
+                   not in the dict will be unchanged
+        query_type: name of the query type, e.g. 'RootSchemaQuery'
```
obi1kenobi, line 169: (for future PR) we should be able to load this from the AST, since the schema starts with
```
schema {
    query: RootSchemaQuery
    ...
}

...
```

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -67,29 +83,36 @@
...
 
     Returns:
-        RenamedSchemaDescriptor, a namedtuple that contains the AST of the renamed schema, and the
-        map of renamed type/field names to original names. Only renamed names will be included
-        in the map.
+        RenamedSchemaDescriptor containing the AST of the renamed schema, and the map of renamed
+        type/field names to original names. Only renamed names will be included in the map.
 
     Raises:
+        - CascadingSuppressionError if a type suppression would require further suppressions
```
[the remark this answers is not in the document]
LWprogramming: > Separately, it might be good to treat these "intermediate" error types as abstract, and always raise only errors whose types are not further refined.

Hmm, I don't quite understand this sentence-- what does it mean to "refine" a type here?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -1,6 +1,18 @@
...
+
+Currently you can perform 1-1 renaming of types, unions, enums, and interfaces. You can also
+suppress types that don't implement an interface and suppress unions.
+
+Operations that are not yet supported but will be implemented:
+Suppressions for fields, enums, interfaces, and types that implement interfaces.
+1-1 and 1-many renamings for fields and enum values.
+
+If you suppress all member types in a union, you must also suppress the union. If you suppress a
+type, you must also suppress any fields of that type (when field suppression is implemented).
```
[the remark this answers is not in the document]
LWprogramming: Hmm, I hadn't thought of this wording but I like it (I debated between this wording and "when you suppress type X, you must also suppress X-type fields" which both are a bit ambiguous). Will change and add test!

graphql_compiler/schema_transformation/rename_schema.py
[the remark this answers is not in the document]
obi1kenobi: All GraphQL compiler errors are subtypes of `GraphQLError`, but we never raise `GraphQLError` itself, instead opting for a subtype: any error the compiler raises passes `isinstance(e, GraphQLError)` but fails `type(e) == GraphQLError`.

I'm suggesting we could do the same for `SchemaTransformError`: have all exceptions raised from inside the `schema_transformation` module be derived from `SchemaTransformError` but always raise one of its subtypes instead of raising it directly. This is nice because then if you write `try ... except SchemaTransformError` you are explicitly stating that you intend to catch *all* such errors. Currently, since things can raise `SchemaTransformError` itself on its own, this is not true — there's no way to catch *just those errors* without also catching everything else, i.e. the intent behind that same `try ... except SchemaTransformError` would be unclear.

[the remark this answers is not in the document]
LWprogramming: Ah, got it. From our discussion a few weeks ago based on [this blog post](https://ericlippert.com/2008/09/10/vexing-exceptions/), at the time we decided to not make a new class since all the errors simply report an error message so a user wouldn't handle them differently. I do plan to update the errors so that the user can programmatically fix them, and at that point I think it makes sense to make a new class for this. For the time being, then, I'll just leave a note.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -1,6 +1,18 @@
 # Copyright 2019-present Kensho Technologies, LLC.
+"""Implement renaming and suppressing parts of the schema.
+
+Currently you can perform 1-1 renaming of types, unions, enums, and interfaces. You can also
+suppress types that don't implement an interface and suppress unions.
```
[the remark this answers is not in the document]
LWprogramming: This is correct- this particular PR implements suppression for types that don't implement interfaces and unions, whereas the 1-1 renaming implementation already existed for types, unions, enums, and interfaces. Will update the rest of the comment!

LWprogramming:
Noting here that, after [this PR](https://github.com/kensho-technologies/graphql-compiler/pull/877) gets merged, we'll be able to update some type hints and remove unnecessary code based on the changes in the latest GraphQL-core release.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +191,150 @@ def rename_schema(
...
+            f"renamed schema."
+        ]
+        if visitor.fields_to_suppress:
+            for object_type in visitor.fields_to_suppress:
+                error_message_components.append(f"Object type {object_type} contains: ")
+                error_message_components += [
+                    f"field {field} of suppressed type "
+                    f"{visitor.fields_to_suppress[object_type][field]}, "
+                    for field in visitor.fields_to_suppress[object_type]
+                ]
```
obi1kenobi, line 260: micro nit: `error_message_components.extend(<generator here instead of list comprehension>)` avoids materializing the intermediate list of error messages.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -107,26 +191,150 @@ def rename_schema(
...
+                    raise AssertionError(
+                        f"Union {union_name} must include one or more unique member types, but "
+                        f"{union_name}'s list of union types was None."
+                    )
+                error_message_components.append(
+                    f"Union type {union_type} has no non-suppressed members: "
+                )
+                error_message_components += [
+                    union_member.name.value for union_member in union_type.types
+                ]
```
obi1kenobi, line 282: same micro-nit as above

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -388,8 +619,169 @@ def enter_field_definition(
...
+        self,
+        node: FieldDefinitionNode,
+        key: Any,
+        parent: Any,
+        path: List[Any],
+        ancestors: List[Any],
+    ) -> None:
+        """Check that no type Y contains a field of type X, where X is suppressed."""
+        if self.current_type == self.query_type:
+            return IDLE
```
obi1kenobi, line 681: The combination of `-> None` and a `return` that takes a value to return is a bit strange. Does the new `graphql-core` keep `IDLE = None`, or will this need to change when we upgrade?

LWprogramming: This is technically correct, since the new version keeps [`IDLE = None`](https://github.com/graphql-python/graphql-core/commit/57c6e2a140c69f40d568d4325d3327beb948bf7c#diff-e2af0c27e1f7e27d2e12b2211e86fe7eR52) although in a slightly unexpected way, as `VisitorActionEnum` does not include `IDLE`. I could define a `TypeVar` to restrict returns to `type(IDLE)`?

obi1kenobi: Go ahead with whatever you think is best, I'm pretty agnostic. If `mypy` can figure out that `type(IDLE)` is the same thing as a `None` annotation, that sounds fine to me. If it can't, then this might cause further problems and it might not be worth it.

LWprogramming: Nope, it doesn't figure it out dynamically so we'll leave it as-is

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -54,42 +126,46 @@
...
-    appears in renamings will be renamed to the corresponding value. Any such names that do not
-    appear in renamings will be unchanged. Scalars, directives, enum values, and fields not
-    belonging to the root/query type will never be renamed.
+    appears in renamings will be renamed to the corresponding value if the value is not None. If the
+    value is None, it will be suppressed in the renamed schema and queries will not be able to
+    access it.
+
+    Any such names that do not appear in renamings will be unchanged.
+
+    Scalars and directives not belonging to the root/query type will never be renamed. In addition,
```
obi1kenobi, line 145: Confused about this sentence. It's a bit ambiguous: it can be parsed as "(scalars and directives) (not belonging to...)" or as "(scalars) and (directives not belonging to...)", and neither seems the clear meaning.

Also, I'm not sure what "belonging to the root/query type" means here. I don't know how scalars or directives can be seen as belonging to anything, and I'm not sure we'll ever want to rename directives.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -388,8 +602,161 @@ def enter_field_definition(
...
+
+        Args:
+            renamings: maps original type name to renamed name or None (for type suppression). Any
+                       name not in the dict will be unchanged
+            query_type: name of the query type (e.g. RootSchemaQuery)
+        """
+        self.renamings = renamings
+        self.query_type = query_type
+        self.current_type: Optional[str] = None
+        # Maps a type T to a dict which maps a field F belonging to T to the field's type T'
```
obi1kenobi, line 627: I'm not sure I understand this. The outer `str` is the name of a vertex type T. And then `field F belonging to T to the field's type T` maps each of T's fields to T's name? It seems like the values in the inner dict are redundant as described.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -388,8 +602,161 @@ def enter_field_definition(
...
+                # Suppress the type
+                return REMOVE
             else:  # Make copy of node with the changed name, return the copy
                 field_node_with_new_name = get_copy_of_node_with_new_name(node, new_field_name)
                 return field_node_with_new_name
 
         return IDLE
+
+
+class CascadingSuppressionCheckVisitor(Visitor):
```
obi1kenobi, line 615: Perhaps as a class docstring, explain how this visitor is intended to be used. Its API is currently undocumented — it seems it's something like "instantiate it, use it in a `visit()` call, and then look at some magic attributes".

Not only is the API not documented, but the "output" attributes aren't even defined and typed at class level: https://mypy.readthedocs.io/en/stable/class_basics.html#instance-and-class-attributes

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -388,8 +602,161 @@ def enter_field_definition(
...
+                # suppression error concern.
+                return IDLE
+        if self.renamings.get(union_name, union_name) is None:
+            # If the union is also suppressed, then nothing needs to happen here
+            return IDLE
+        self.union_types_to_suppress.append(node)
+
+
+class SuppressionNotImplementedVisitor(Visitor):
+    def __init__(self, renamings: Mapping[str, Optional[str]]) -> None:
```
obi1kenobi, line 710: Same comment as above with regard to the intended API of this object and its documentation.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -57,6 +57,20 @@ class InvalidCrossSchemaEdgeError(SchemaTransformError):
     """
 
 
+class CascadingSuppressionError(SchemaTransformError):
+    """Raised if existing suppressions would require further suppressions.
+
+    This may be raised during schema renaming if it suppresses all the fields of a type but not the
+    type itself, if it suppresses all the members of a union but not the union itself, or if it
+    suppresses a type X but there still exists a different type Y that has fields of type X.
```
obi1kenobi, line 65: Maybe use a bulleted list here? Might make this easier to read.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I re-read the original PR (#390) and I think I mis-parsed the original comment in the way you described in the first paragraph. When I read `Scalars, directives, enum values, and fields not belonging to the root/query type will never be renamed` in the context of everywhere else it mentions scalars, directives, or enum values, the correct reading in the original code is "(Scalars), (directives), (enum values), and (fields not belonging to the root/query type)". In this case, "field not belonging to the root/query type" should actually be "type that exists in the schema but not as a field in the object type `SchemaQueryType`".

I'll update this to only mention directives, since renaming/suppressing scalars types looks to be very similar to doing so with object types, which we've already implemented here.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Consider this part of a schema:
```
type Foo {
  ...
}
type Bar {
  baz: Foo
}
```
Then `F` is `baz`, `T` is `Bar`, and `T'` is Foo. I include `T'` so we can specify precisely which type `T'`'s suppression would cause a cascading error.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I've added a class docstring here and with the other visitor.

For typing at the class level vs. in-line, I believe we started a discussion a few weeks ago but I don't think we actually came to a full conclusion on it. At the time the concern was 

"How do we make `self.foo: str = "foo"` work but not `bar.foo: str = "foo"` for some random bar?"

although it turns out `mypy` does catch situations where these in-line type-hints happen in both cases where something gets type-hinted outside of the class.

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: I see. I was confused because I didn't realize `F` is in effect just a name while `T/T'` are types.

How about this alternative:
```
For type named T, and its field named F whose type has name V, this dict would be {"T": {"F": "V"}}
```
`T` and `T'` being named similar things is somewhat misleading since they don't have to be related types — any type names would work.

obi1kenobi:
The module docstring looks great btw! I think we'll be able to turn it into documentation on ReadTheDocs pretty easily given its present structure.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -1,6 +1,75 @@
...
+    type NewDog {
+        name: String
+    }
+    type Human {
+        pet: NewDog
+    }
+If "Dog" also appeared as a field in the schema's root type, it would be renamed to "NewDog" there
+as well.
+
+1-many renaming is an operation intended for fields in an object type and enum values only, in which
```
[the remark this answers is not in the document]
LWprogramming: Thanks for the catch! I've made a note of it.

PR #860, opened 2020-06-24
Make "main" the primary branch. Replace "master" everywhere possible.

obi1kenobi:
Updated Travis CI to consider `main` the primary branch that everything gets tested again. Replaced `master` terminology in comments, links, etc.

### DIFF
--- a/graphql_compiler/schema_generation/sqlalchemy/sqlalchemy_reflector.py
+++ b/graphql_compiler/schema_generation/sqlalchemy/sqlalchemy_reflector.py
@@ -16,8 +16,8 @@ def fast_sql_server_reflect(engine, metadata, schema, primary_key_selector=None)
     This function mutates the metadata object.
 
     Args:
-        engine: SQLAlchemy Engine, engine connected to a SQL server. It must connected to the master
-                database in the SQL Server.
+        engine: SQLAlchemy Engine, engine connected to a SQL server. It must connected to
+                the primary database in the SQL Server deployment, not a replica.
         metadata: MetaData object to reflect the metadata of the SQL Server to.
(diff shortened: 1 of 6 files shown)


graphql_compiler/schema_generation/sqlalchemy/sqlalchemy_reflector.py
```diff
@@ -16,8 +16,8 @@ def fast_sql_server_reflect(engine, metadata, schema, primary_key_selector=None)
     This function mutates the metadata object.
 
     Args:
-        engine: SQLAlchemy Engine, engine connected to a SQL server. It must connected to the master
-                database in the SQL Server.
+        engine: SQLAlchemy Engine, engine connected to a SQL server. It must connected to
+                the primary database in the SQL Server deployment, not a replica.
```
pmantica1, line 20: I don't think that this is quite right. I am not even sure if you can even have replicas of individual databases in a SQL server. 

"master" here refers to the core system database in the SQL server.

https://docs.microsoft.com/en-us/sql/relational-databases/databases/master-database?view=sql-server-ver15#:~:text=In%20SQL%20Server%2C%20system%20objects,initialization%20information%20for%20SQL%20Server.

The reason that this is a precondition is that only if you are within the context of the "master" database, i.e. you ran "USE master;",  can you run queries against all of the user databases.  (If you are within the context of  particular user database, then you can only run queries against tables in that database unless you switch the database context).

obi1kenobi: Ah I see, thanks. I'll include this link so that others don't make the same mistake.

PR #863, opened 2020-06-25
Upgrade to graphql-core 3.1.x.

obi1kenobi:
This allows us to remove three `xfail` test markers, since the underlying issues were resolved in `graphql-core`.

I also fixed a CONTRIBUTING.rst typo, where running the command it specified had the exact opposite effect from the one described.

In general, if we try to have less specific imports (`from graphql import print_schema` rather than `from graphql.utilities.schema_printer import print_schema`), these version upgrades will go more smoothly in the future. Nearly all the changes here were due to things moving around, breaking the specific imports but preserving the less-specific ones.

### DIFF
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -4,7 +4,7 @@ from collections import namedtuple
 from inspect import getmembers, isfunction
 from pprint import pformat
 import re
-from typing import Any, Dict, List, Optional, Set, Tuple, Union
+from typing import Dict, List, Optional, Set, Tuple, Union
 from unittest import TestCase
 
 from graphql import GraphQLList, parse
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -83,8 +83,8 @@ SCHEMA_TEXT = """
 
     directive @recurse(
         \"\"\"
-        Recurse up to this many times on this edge. A depth of 1 produces the current
-        vertex and its immediate neighbors along the given edge.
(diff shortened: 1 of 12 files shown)


graphql_compiler/tests/test_helpers.py
```diff
@@ -83,8 +83,8 @@
 
     directive @recurse(
         \"\"\"
-        Recurse up to this many times on this edge. A depth of 1 produces the current
-        vertex and its immediate neighbors along the given edge.
+        Recurse up to this many times on this edge. A depth of 1 produces the current \
+vertex and its immediate neighbors along the given edge.
```
pmantica1, line 87: I think that the failed tests might be due to this change. The linter errors are due to isort.

obi1kenobi: That's weird, my local tests refused to pass unless I made this change... 👀

LWprogramming: Just tried it locally and the test fails for me as well, but reverting it produces another failing test.

pmantica1: Maybe you need to delete/escape the whitespace before "Recurse" and after "edge." too?

obi1kenobi: The failing test was unrelated — it was simply producing an alternate but equally-valid order of definitions in the schema autogenerated from OrientDB. I added an order-independent schema definition comparison function and used it. It isn't perfect (let's not add any double newlines in docstrings 😄 ) but it'll be plenty good enough for tests.

Issue #862, opened 2020-06-24
Improve cost estimator with value-count statistics


bojanserafimov:
Value-count statistics reflect the popularity of each possible field (and edge endpoint) value in the database instance. The following commit sketches out the necessary changes to incorporate these statistics into the cost estimator, and raises design questions: https://github.com/kensho-technologies/graphql-compiler/commit/b2cb2e5a7ada061cdf9ac39b4cd8fd220f0539e1

PR #864, opened 2020-06-25
Use value count statistics for equality filters

bojanserafimov:
Addressing https://github.com/kensho-technologies/graphql-compiler/issues/862

Changes in this PR:
- Add `get_value_count` method to `Statistics` abstract class.
- Use `get_value_count` to improve estimation for `=` and `in_collection` filters.
- Add tests
- Fix unrelated tests that were in the way


graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1185,15 +1240,22 @@ def test_get_in_collection_filter_selectivity(self) -> None:
         self.assertEqual(expected_selectivity, selectivity)
 
         statistics_with_distinct_birthday_values_data = LocalStatistics(
-            dict(), distinct_birthday_values_data
+            dict(), distinct_field_values_counts=distinct_birthday_values_data
```
bojanserafimov, line 1243: Unrelated to this PR, but mypy found this issue so I had to fix the test.

.pylintrc
```diff
@@ -211,7 +211,6 @@ enable=abstract-class-instantiated,
     return-arg-in-generator,
     return-in-init,
     return-outside-function,
-    round-builtin,
```
bojanserafimov, line 214: https://stackoverflow.com/questions/59533912/what-is-the-alternative-to-built-in-round

The warning screams at any usage of `round` because the implementation changed from py2 to py3 and it might cause transition problems.

obi1kenobi: If you'd like to open a separate PR with just this change, I'd merge it right away.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -97,16 +115,39 @@ def get_field_quantiles(self, vertex_name, field_name):
...
-        class_counts,
-        vertex_edge_vertex_counts=None,
-        distinct_field_values_counts=None,
-        field_quantiles=None,
+        class_counts: Dict[str, int],
+        *,
+        vertex_edge_vertex_counts: Optional[Dict[Tuple[str, str, str], int]] = None,
+        distinct_field_values_counts: Optional[Dict[Tuple[str, str], int]] = None,
+        field_quantiles: Optional[Dict[Tuple[str, str], List[Any]]] = None,
+        sampling_summaries: Optional[Dict[str, SamplingSummary]] = None,
```
obi1kenobi, line 150: I think the need to fix the test below is because of this: the `*` means that `distinct_field_value_counts` is now kwarg-only, i.e. the test needed fixing. Mypy found it via static analysis, but this would have also failed at runtime.

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
obi1kenobi: See above, I think this is caused by your API change on `LocalStatistics` that I point out above. This is a fine change, and I like kwarg-only arguments, so don't change it back :)

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -97,16 +115,39 @@ def get_field_quantiles(self, vertex_name, field_name):
         """
         return None
 
+    def get_value_count(self, vertex_name: str, field_name: str, value: Any) -> Optional[int]:
```
obi1kenobi, line 118: Is `int` the right type hint here? Since we're estimating, do we always want integer-typed estimates? Is `1.47` not more useful than `1`?

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -1,10 +1,28 @@
...
 
 
+@dataclass
+class SamplingSummary:
+    """Results extracted from sampling a random rowset that are relevant for statistics."""
+
+    # The class this summary describes
+    class_name: str
+
+    # Mapping field_name -> field_value -> observed_count for all field names and observed
```
obi1kenobi, line 18:
```suggestion
    # Mapping field_name -> (field_value, observed_count) for all field names and observed
```

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -1,10 +1,28 @@
...
+    # field values on this class.
+    value_counts: Dict[str, Dict[Any, int]]
+
+    # The number_of_rows / number_of_samples ratio
+    sample_ratio: int
```
obi1kenobi, line 23: Maybe record the number of samples as well? I know it can be calculated by summing over the counts in `value_counts.values()` as well, but if there are a lot of them, it might be a more pleasant API to just have it ready.

graphql_compiler/cost_estimation/statistics.py
bojanserafimov: That would be `Dict[str, Tuple[Any, int]]`. It's a different thing. I can add parens if it's not implied: `field_name -> (field_value -> observed_count)`

graphql_compiler/cost_estimation/statistics.py
bojanserafimov: It is, but the smallest numbers we'll practically return are around `50`, and `51/50` is very close to `1`. I have no strong feelings either way, and I can use `float` if that's simpler to you.

graphql_compiler/cost_estimation/statistics.py
bojanserafimov: Mypy found if when I just added type hints, which reminded me to add `*`

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -163,44 +167,59 @@ def _get_query_interval_of_integer_inequality_filter(
...
+    """Calculate the selectivity of in_collection filter.
 
     Args:
         schema_info: QueryPlanningSchemaInfo
-        location_name: string, type of the location being filtered
-        filter_fields: tuple of str, listing all the fields being filtered over
+        location_name: type name of the location being filtered
+        filter_field: field affected by the filter
+        collection: the values the filter allows, if known. A value of None means the collection
+                    is not known at compile time.
```
bojanserafimov, line 183: Overview of changes in this particular function:
- Change `filter_fields` to `filter_field`. The function previously incorrectly assumed that if an index covers the `filter_fields`, then this filter is very selective. This is not the case since `filter_fields` are loosly defined, and can mean anything. In the case of `in_collection` and `=` filters, there's always just one filter field, so the wrong codepath was actually never called.
- Previously this function was a helper for `=` filters that was also used in estimating `in_collection` selectivity. We can't do that anymore since we dropped the assumption that the filter value doesn't matter. We need to look at all values inside the collection. So now we compute `=` filter selectivity using this helper by passing a single-element list for the `collection` argument.

graphql_compiler/cost_estimation/statistics.py
bojanserafimov: Hmm, could be useful to sanity-check `sample_ratio * number_of_samples ~= row_count`. I'd leave it for a future PR though, if you don't mind. There's many things to do for statistics validation, like canonicalizing values, and this can be done as part of that issue

graphql_compiler/cost_estimation/statistics.py
LWprogramming: ^ This is the only legal interpretation since the alternative interpretation (field_name -> field_value) -> observed_count would have a dict as a key which is not allowed because dicts are mutable (although you're free to add parens if you feel it helps readability! I think it probably does)

bojanserafimov: I'm used to `->` being right associative from Haskell/Idris, but there's no reason to bring those norms into a python codebase, so I'll add parens for clarity.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -163,44 +167,60 @@ def _get_query_interval_of_integer_inequality_filter(
...
-
-    Using the available unique indexes and/or the distinct_field_values_count statistic, this
-    function extracts the current location's selectivites, and then combines them, returning one
-    Selectivity object.
+def _estimate_filter_selectivity_of_in_collection(
+    schema_info: QueryPlanningSchemaInfo,
+    location_name: str,
+    filter_field: str,
+    collection: Optional[Set[Any]],
+):
```
LWprogramming, line 175:
```suggestion
) -> Selectivity:
```

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -163,44 +167,60 @@ def _get_query_interval_of_integer_inequality_filter(
...
+    )
 
-    result_selectivity = _combine_filter_selectivities(all_selectivities)
-    return result_selectivity
+    # Combine statistics to compute selectivity
+    if collection_value_counts is not None:
+        # TODO(bojanserafimov): If we have value_count statistics for the field, but we conclude
+        #                       that this is not one of the common values, we use the rule of 3
+        #                       (see statistics.get_value_count). The distinct_value_count stats
+        #                       could provide additional precision case if available.
```
LWprogramming, line 213: > additional precision case if available.

Possible extra word "case"? Or am I mis-parsing this phrase

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -436,8 +441,13 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
             if filter_info.op_name == "=":
                 # TODO(bojanserafimov): Check if the filter value is in the interval selected
                 #                       by the inequality filters.
-                selectivity = _estimate_filter_selectivity_of_equality(
-                    schema_info, location_name, filter_info.fields
+                collection = None
```
LWprogramming, line 444: This code seems almost identical to the code for `in_collection` filters right above. Is there something conceptually different about the way we compute selectivity for the two types of filters, or would it be easy to merge these two blocks of code into one?

(If the former, we can still combine the two loops through `filters_on_field` because they loop through the same thing)

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -128,15 +169,19 @@ class name) to count of edge instances of given class
                              element is a value greater than or equal to i/N of all present
                              values. The number N can be different for each entry. N has to be at
                              least 2 for every entry present in the dict.
-            TODO(bojanserafimov): Enforce a canonical representation for quantile values. Datetimes
-                                  should be in utc, decimals should have type float, etc.
+            sampling_summaries: dict, class_name: str -> SamplingSummary
```
LWprogramming, line 172: No need to describe types in the docstring since we have type hints now :)

graphql_compiler/cost_estimation/filter_selectivity_utils.py
LWprogramming: Also, unrelated but I can't comment on the function declaration itself on [line 345](https://github.com/kensho-technologies/graphql-compiler/pull/864/files?file-filters%5B%5D=.py#diff-5dfbb159b89d2baa7f0e0dd13732e415R345) but let's move the type hints from the docstring to the actual code since we're modifying this function.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -398,46 +428,21 @@ def get_selectivity_of_filters_at_vertex(schema_info, filter_infos, parameters,
...
+            if filter_info.op_name in ("=", "in_collection"):
+                filter_argument = get_only_element_from_collection(filter_info.args)
+                filter_field = get_only_element_from_collection(filter_info.fields)
+                collection = None
+                if is_runtime_parameter(filter_argument):
+                    # TODO(bojanserafimov): Check if the filter values are in the interval selected
+                    #                       by the inequality filters.
+                    collection = parameters[get_parameter_name(filter_argument)]
+                    if filter_info.op_name == "=":
+                        collection = [collection]
```
obi1kenobi, line 442: `mypy` is going to have things to say about the type of `collection` changing. I think this module's `mypy` rules are pretty lax right now, but it might be easier to fix it now rather than fixing it later when we try to dial up the rule strictness. Your call.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -1,10 +1,28 @@
...
 
 import six
 
 
+@dataclass
+class SamplingSummary:
+    """Results extracted from sampling a random rowset that are relevant for statistics."""
+
+    # The class this summary describes
+    class_name: str
```
obi1kenobi, line 16: Let's try to be consistent between `class/type/vertex` as much as possible. I'm not sure which one we've been using the most, but I feel like it's more than one.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
```diff
@@ -163,44 +167,60 @@ def _get_query_interval_of_integer_inequality_filter(
...
-        )
-
-        if statistics_result is not None:
-            # Assumption: all distinct field values are distributed evenly among vertex instances,
-            # so each distinct value occurs
-            # (# of current location vertex instances) / (# of distinct field values) times.
-            all_selectivities.append(
-                Selectivity(kind=FRACTIONAL_SELECTIVITY, value=1.0 / statistics_result)
-            )
+        num_entries = 1 if collection is None else len(collection)
```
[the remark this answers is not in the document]
bojanserafimov: I'd want to test it too, and at that point it's just dragging out this unrelated PR.

graphql_compiler/cost_estimation/filter_selectivity_utils.py
bojanserafimov: I'll come back when rules are dialed up. This section will require some casts/asserts too, so it's hard to blindly guess what passes.

graphql_compiler/cost_estimation/statistics.py
bojanserafimov: Yeah, `class/type/vertex/location_name`... it's not great. This is actually specific to a vertex so I'll rename to 'vertex'. Edge sampling summaries will be much different.

graphql_compiler/cost_estimation/statistics.py
obi1kenobi: I'd say that all other statistics are float-typed because we generally deal in estimates, so let's make this one `float` too if it's all the same to you.

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -97,16 +115,39 @@ def get_field_quantiles(self, vertex_name, field_name):
...
+        """Return the estimated number of times the given value appears in the database.
+
+        Args:
+            vertex_name: vertex on which the field is defined
+            field_name: field for which the value stands
+            value: value to be counted
+
+        Returns:
+            An estimate on how often this value currently appears in the given field, or None
+            if unkwnown.
```
obi1kenobi, line 128:
```suggestion
            if unknown.
```

graphql_compiler/cost_estimation/statistics.py
```diff
@@ -178,3 +225,25 @@ def get_field_quantiles(self, vertex_name, field_name):
...
+        vertex_sampling_summary = self._sampling_summaries.get(vertex_name)
+        if vertex_sampling_summary is None:
+            return None
+
+        field_sampled_value_counts = vertex_sampling_summary.value_counts.get(field_name)
+        if field_sampled_value_counts is None:
+            return None
+
+        sampled_value_count = field_sampled_value_counts.get(value)
+        if sampled_value_count:
```
obi1kenobi, line 240: Let's be explicit here. Is this a `is not None` or a `> 0` or both or something else entirely?

PR #865, opened 2020-06-26
Fix fold after recurse

bojanserafimov:
Using `@fold` inside `@recurse` is allowed, but our SQL backend didn't implement it, and didn't raise a good error. This PR implements this functionality.

Changes in this PR:
- Extend the `_find_columns_used_outside_folds` analysis to also reach into `FoldScopeLocation`s so that the recurse CTE can anticipate the upcoming fold and expose any columns needed for it
- Allow the fold subquery to be joined to a cte, by selecting the right cte output to serve as a primary key.
- Test

The test I added uncovered an unrelated recurse bug (see newly added TODO). I'll address this in a separate PR.

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -26,13 +26,13 @@ from .compiler_entities import BasicBlock
 from .compiler_frontend import IrAndMetadata
 from .expressions import ContextField, Expression
 from .helpers import (
-    get_vertex_path,
     BaseLocation,
     FoldPath,
     FoldScopeLocation,
     Location,
     QueryPath,
     get_edge_direction_and_name,
+    get_vertex_path,
 )
 from .metadata import LocationInfo
 
@@ -81,10 +81,10 @@ def _traverse_and_validate_blocks(ir: IrAndMetadata) -> Iterator[BasicBlock]:
         yield block
 
 
-def _find_columns_used_outside_folds(
+def _find_columns_used(
     sql_schema_info: SQLAlchemySchemaInfo, ir: IrAndMetadata
 ) -> Dict[VertexPath, Set[str]]:
-    """For each query path outside of a fold output, find which columns are used."""
+    """For each query path, find which of its columns would be used in the resulting query."""
     used_columns: Dict[VertexPath, Set[str]] = {}
 
     # Find filters used
@@ -96,7 +96,9 @@ def _find_columns_used_outside_folds(
     # Find foreign keys used
     for location, location_info in ir.query_metadata_table.registered_locations:
         for child_location in ir.query_metadata_table.get_child_locations(location):
-            edge_direction, edge_name = get_edge_direction_and_name(get_vertex_path(child_location)[-1])
+            edge_direction, edge_name = get_edge_direction_and_name(
+                get_vertex_path(child_location)[-1]
+            )
             vertex_field_name = f"{edge_direction}_{edge_name}"
             edge = sql_schema_info.join_descriptors[location_info.type.name][vertex_field_name]
             used_columns.setdefault(get_vertex_path(location), set()).add(edge.from_column)
@@ -105,7 +107,9 @@ def _find_columns_used_outside_folds(
             # A recurse implies an outgoing foreign key usage
             child_location_info = ir.query_metadata_table.get_location_info(child_location)
             if child_location_info.recursive_scopes_depth > location_info.recursive_scopes_depth:
-                used_columns.setdefault(get_vertex_path(child_location), set()).add(edge.from_column)
+                used_columns.setdefault(get_vertex_path(child_location), set()).add(
+                    edge.from_column
+                )
 
     # Find outputs used
     for _, output_info in ir.query_metadata_table.outputs:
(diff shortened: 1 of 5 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -1007,6 +1005,9 @@ def recurse(self, vertex_field: str, depth: int) -> None:
         # Instead of joining to the current _from_clause, we make this alias the _from_clause.
         # If the existing _from_clause had any information in it, then the _current_alias would
         # be a cte that contains all that information.
+        # TODO(bojanserafimov): If the existing _from_clause had no filters or traversals, but
+        #                       had an output, the output needs to be moved into the cte or
+        #                       we need to join to the _from_clause.
```
bojanserafimov, line 1010: https://github.com/kensho-technologies/graphql-compiler/issues/866

graphql_compiler/tests/test_compiler.py
```diff
@@ -5423,6 +5423,74 @@ def test_fold_on_many_to_one_edge(self):
...
+                anon_1 JOIN db_1.schema_1.[Animal] AS [Animal_2]
+                    ON anon_1.uuid = [Animal_2].parent
+            WHERE
+                anon_1.__cte_depth < 3
+        )
+        SELECT
+            [Animal_1].name AS animal_name,
+            folded_subquery_1.fold_output_name AS homes_list
+        FROM
+            db_1.schema_1.[Animal] AS [Animal_1],
```
bojanserafimov, line 5457: `[Animal_1]` should be joined to the rest, or `animal_name` should be picked from `anon_1`. See https://github.com/kensho-technologies/graphql-compiler/issues/866

Fix coming in separate PR.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -80,60 +81,54 @@ def _traverse_and_validate_blocks(ir: IrAndMetadata) -> Iterator[BasicBlock]:
         yield block
 
 
-def _find_columns_used_outside_folds(
+def _find_columns_used(
```
chewselene, line 84: nit/personal preference: I'd change this function name to `_find_used_columns`. Sound more readable to me for some reason.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -80,60 +81,54 @@ def _traverse_and_validate_blocks(ir: IrAndMetadata) -> Iterator[BasicBlock]:
...
     sql_schema_info: SQLAlchemySchemaInfo, ir: IrAndMetadata
 ) -> Dict[VertexPath, Set[str]]:
-    """For each query path outside of a fold output, find which columns are used."""
+    """For each query path, find which of its columns would be used in the resulting query."""
```
chewselene, line 87: nit: Not sure if the docstring properly tells the user that the "resulting query" is part of the IR. Maybe something like: `For each query path, find all columns used in the query described by the IR.`

graphql_compiler/tests/test_compiler.py
```diff
@@ -5423,6 +5423,74 @@ def test_fold_on_many_to_one_edge(self):
...
+                        WHERE
+                            anon_2.lives_in = [Location_1].uuid FOR XML PATH ('')),
+                   '') AS fold_output_name
+                FROM anon_1 AS anon_2
+            ) AS folded_subquery_1
+                ON anon_1.uuid = folded_subquery_1.uuid
+        """
+        expected_match = SKIP_TEST
+        expected_gremlin = SKIP_TEST
+        expected_cypher = SKIP_TEST
```
chewselene, line 5483: When do we decide to use `SKIP_TEST` vs `NotImplementedError`/emitted query? If this test is really only relevant for MSSQL, might be nice to include a note about why that is.

bojanserafimov: `SKIP_TEST`: Either not implemented yet, not tested yet, or no reason to test.
`NotImplementedError`: The desired behavior is to return `NotImplementedError`, and that's implemented.

I'll add a comment.

obi1kenobi: This is one of the things I'd love to eradicate in a future revamp of our test suite. "Not tested yet / no reason to test" situations shouldn't need explicit hand-written expected output, but if we have the test case inputs ready to go, we could definitely auto-generate a snapshot of the current behavior and alert if it ever changes. It would be a nice defense-in-depth strategy, but only worth it if we can make it easier to do.

Right now, if writing a test targeting something that's generally only interesting e.g. for a particular SQL flavor (e.g. `@fold` in MSSQL), we either have to write the expected test outcome for other SQLs and other query languages (tedious!) or `SKIP_TEST` to reduce the amount of effort a PR takes and not take advantage of the additional test opportunity. This is where an automatic snapshot test would be useful: low effort yet still helpful.

graphql_compiler/compiler/emit_sql.py
bojanserafimov: This function is a bit hard to spec. The IR only talks about fields, and the query builder interprets them as columns in complicated ways. This function is a bit ugly since it tries to guess what the query builder will need. I have two different fixes in mind:
a) Instead of using sqlalchemy selectables as the main building block for queries in this module, make a new mutable selectable class that keeps track of what columns are requested of it, and can be materialized into a sqlalchemy selectable that exposes only those columns. Then this function wouldn't need to exist.
b) Document in the spec of this function what is considered a "used column", and then make sure that:
  - all objects in `CompilationState._aliases` expose all the used columns
  - only used columns are selected from `CompilationState._aliases`

Solution (a) fixes code duplication and removes some source of bugs, but it's too complicated to implement and will likely introduce other kinds of problems. Solution (b) seems right, even though it doesn't solve the duplication problem. It just creates a chain of specs from definition of "used column" to usage of "used column" that it might make this more manageable.

I suspect properly specifying this function will reveal some bugs. For instance, it's a bit unclear whether meta fields produce "used columns", and whether using a meta field inside recurse would cause a problem. I'd do this in a separate PR though, let me know if you agree.

bojanserafimov: I can wrap part of this in a todo that explains what "the resulting query" is

bojanserafimov: Just wrote a clarifying todo

chewselene: Agreed that solution (a) seems too complicated. The todo helps clarify things. Thanks!

PR #867, opened 2020-06-29
Fix division by zero

bojanserafimov:
If our statistics think that there are `0` instances of a certain class, cost estimation in queries involving it divide by zero.

### DIFF
--- a/graphql_compiler/cost_estimation/cardinality_estimator.py
+++ b/graphql_compiler/cost_estimation/cardinality_estimator.py
@@ -229,6 +229,13 @@ def _estimate_edges_to_children_per_parent(
     # Count the number of parents, over which we assume the edges are uniformly distributed.
     parent_location_counts = schema_info.statistics.get_class_count(parent_name_from_location)
 
+    # Anticipate division by zero
+    if parent_location_counts == 0:
+        # This implies that edge_counts is also 0. The reason we don't assert that is that
+        # we can't expect statistics to be collected at the same time, so raising an error
+        # is too aggressive.
+        return 0
+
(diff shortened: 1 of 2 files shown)


graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -229,6 +229,13 @@ def _estimate_edges_to_children_per_parent(
     # Count the number of parents, over which we assume the edges are uniformly distributed.
     parent_location_counts = schema_info.statistics.get_class_count(parent_name_from_location)
 
+    # Anticipate division by zero
+    if parent_location_counts == 0:
+        # This implies that edge_counts is also 0. The reason we don't assert that is that
+        # we can't expect statistics to be collected at the same time, so raising an error
+        # is too aggressive.
+        return 0
```
chewselene, line 237: should this be returned as a float?

graphql_compiler/cost_estimation/cardinality_estimator.py
```diff
@@ -229,6 +229,13 @@ def _estimate_edges_to_children_per_parent(
     # Count the number of parents, over which we assume the edges are uniformly distributed.
     parent_location_counts = schema_info.statistics.get_class_count(parent_name_from_location)
 
+    # Anticipate division by zero
+    if parent_location_counts == 0:
+        # This implies that edge_counts is also 0. The reason we don't assert that is that
+        # we can't expect statistics to be collected at the same time, so raising an error
+        # is too aggressive.
```
chewselene, line 236: nit: sentence 2 is a hard to parse/long. Maybe something like `However, asserting that edge_counts is 0 is too aggressive because we can't expect [edge?] statistics to be collected at the same time.` instead?

PR #871, opened 2020-07-02
Type hint compiler/ir_lowering_match/between_lowering.py

obi1kenobi:


graphql_compiler/compiler/ir_lowering_match/between_lowering.py
```diff
@@ -37,54 +42,68 @@ def _extract_conjuction_elements_from_expression(expression):
...
                 new_expression = BinaryComposition(new_operator, expression.right, expression.left)
             else:
                 new_expression = expression
-            field_name = new_expression.left.field_name
-            field_type = new_expression.left.field_type
+
+            left_clause = new_expression.left
+            if not isinstance(left_clause, LocalField):
+                raise AssertionError(
+                    f"Expected left_clause to be a LocalField, but instead found {left_clause}. "
```
pmantica1, line 91: tiny nit: I wouldn't mention the "left_clause" variable name in the error message since we could easily rename the left_clause variable name and forget to update the error message here.

obi1kenobi: I'm not sure what a way to refer to this value that doesn't have "we could rename the variable" risk. The same is true of `new_expression.left` and of the value that `new_expression` comes from. Do you have a preference for a particular resolution?

pmantica1: Maybe "Expected a LocalField, but got {left_clause"} instead. This is a bug"?

obi1kenobi: To be honest, I'm not a huge fan — yes, that protects from the renaming risk, but at the cost of what feels to me like worse readability today. I think I'd prefer the variable to be named, raise a more informative error, and watch in code reviews for missing renamings, than be protected from renaming risk at the cost of less-informative errors.

PR #872, opened 2020-07-02
Type hint compiler/ir_lowering_match/ir_lowering.py

obi1kenobi:


graphql_compiler/compiler/ir_lowering_match/ir_lowering.py
```diff
@@ -8,27 +8,35 @@
...
 
 
 ##################################
 # Optimization / lowering passes #
 ##################################
 
 
-def rewrite_binary_composition_inside_ternary_conditional(ir_blocks):
+def rewrite_binary_composition_inside_ternary_conditional(
+    ir_blocks: List[BasicBlock],
```
pmantica1, line 35: I have another tiny nit. Sorry for these tiny nits, but I think that might be relevant for future prs. 

Why not make this a Sequence instead?

MichaelaShtilmanMinkin: Do we pass `ir_blocks` as something other than a list elsewhere?

pmantica1: That's a good point. I don't think we do. I would prefer type hints to be use the maximally general type. It makes the function more "powerful" and in this case, it would help make sure we weren't accidentally modifying the list. 

But I guess that this type hint is ok because it is consistent with the rest of the codebase. I'll approve the PR. Sorry for the troubles.

Meta-point:
Maybe we should follow the following convention when typing hinting a certain type:
1. If the type has never been type-hinted, then we should try to use the most general type.
2. If the type has been type-hinted before, we should either use the previously used type hint or change the existing type hint for a more general type.

obi1kenobi: In general, the guidelines are sound. A few things come to mind, however, that I think make them remain guidelines rather than strict rules:
- Tools that do runtime analysis of code (e.g. `monkeytype` and `pyannotate`) will always suggest the most-specific type, since they don't and can't know which of the properties of the object are important (e.g. whether `Sequence` is okay or whether modification is important and it has to be `List`). Between more typing and less typing across the codebase, we should always prefer more, so we should in general make that type-hint-addition process as smooth as possible.
- It's difficult to know in general for existing code whether it is allowed some capability but chooses not to use it (e.g. mutating the list), versus whether it isn't allowed that capability and should only know `Sequence` instead. So when type-hinting existing code, it's not always clear whether `List` or `Sequence` is appropriate, nor is it feasible to always find that out (per the above point).

So while these guidelines are good for newly-designed and newly-written code, and are in the general good spirit of "be liberal in what you accept and strict in what you provide," I think applying them to type-hinting existing code must be done with care and good judgment.

PR #873, opened 2020-07-05
Performance optimize tests

pmantica1:
I profiled our tests and it seems that a substantial amount of time was being spent building the test schema/schema infos for every test. So I wanted to check how much faster tests would run if we just built each schema/schema info once. Here are the results: 

Current test times:
37.06 - not-slow 
67.79 - total

New test times:
11.63 - not-slow 
35.82 - total 

This optimization does have the drawback that one test case can modify the data used by another test case. I still think it's worth it. I did want to get some other opinions though before spending any more time on this.

### DIFF
--- a/graphql_compiler/tests/test_helpers.py
+++ b/graphql_compiler/tests/test_helpers.py
@@ -41,7 +41,7 @@ from ..schema_generation.orientdb.utils import (
     ORIENTDB_SCHEMA_RECORDS_QUERY,
 )
 from ..schema_generation.schema_graph import SchemaGraph
-
+from functools import lru_cache
 
 # The strings which we will be comparing have newlines and spaces we'd like to get rid of,
 # so we can compare expected and produced emitted code irrespective of whitespace.
@@ -559,16 +559,16 @@ def compare_ignoring_whitespace(
     test_case.assertEqual(transform(expected), transform(received), msg=msg)
 
 
+_ast = parse(SCHEMA_TEXT)
+_schema = build_ast_schema(_ast)
 def get_schema() -> GraphQLSchema:
     """Get a schema object for testing."""
-    ast = parse(SCHEMA_TEXT)
-    schema = build_ast_schema(ast)
-    return schema
+    return _schema
 
 
 def get_type_equivalence_hints() -> TypeEquivalenceHintsType:
     """Get the default type_equivalence_hints used for testing."""
-    schema = get_schema()
+    schema = build_ast_schema(parse(SCHEMA_TEXT))
     type_equivalence_hints: Dict[
         Union[GraphQLInterfaceType, GraphQLObjectType], GraphQLUnionType
     ] = {}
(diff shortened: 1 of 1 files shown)


graphql_compiler/tests/test_helpers.py
```diff
@@ -559,6 +560,7 @@ def compare_ignoring_whitespace(
     test_case.assertEqual(transform(expected), transform(received), msg=msg)
 
 
+@lru_cache()
```
pmantica1, line 563: This is a horrible hack. It just minimizes the lines of code I had to change for my data exploration. If I were to do this properly, I would change the `get_schema` function for a constant.

graphql_compiler/tests/test_helpers.py
```diff
@@ -591,14 +593,15 @@ def get_type_equivalence_hints() -> TypeEquivalenceHintsType:
...
 def get_common_schema_info() -> CommonSchemaInfo:
     """Get the default CommonSchemaInfo used for testing."""
     return CommonSchemaInfo(get_schema(), get_type_equivalence_hints())
 
 
 def _get_schema_without_list_valued_property_fields() -> GraphQLSchema:
     """Get the default testing schema, skipping any list-valued property fields it has."""
-    schema = get_schema()
+    schema = build_ast_schema(parse(SCHEMA_TEXT))
```
pmantica1, line 604: This was necessary because this function modifies the schema stored in the `schema` variable.

pmantica1:
I profiled the code using pytest and did not find any other low-hanging performance optimization fruits other than this one.

obi1kenobi:
Out of curiosity, what's the performance difference between using the exact
same schema object for all tests, vs using deepcopy to make a new one?
Hopefully deepcopy is faster than parsing and building a schema from
scratch, and I'm wondering how much that is / if it's worth considering as
an alternative.

pmantica1:
It seems that you can't deepcopy a GraphQLSchema. See https://github.com/kensho-technologies/graphql-compiler/pull/874

pmantica1:
Another alternative that I tested was building the schema with `assume_valid=True`.

This reduced test times to:
24.24 seconds - not-slow 
59.63 seconds - total 

When doing this we have less performance benefits than we just use one copy and we lose test schema validation, which will probably help us pinpoint bug us in the future. So I still prefer the using one-copy approach. Thoughts?

obi1kenobi:
I think building a schema with validation once (as a test case on its own),
and then building it without repeated validation, is a good idea we should
definitely pursue.

For using a single schema object, I'm quite a bit more wary. It's not clear
to me that the extra 10s is worth the long-term risk on our test suite.

My recommendation would be to see what's up with the deepcopy approach, and
use that to help guide our ultimate decision. If it can be rectified and
deepcopy ends up being fast, that'd be ideal. If not, perhaps we can have
Travis stay maximally conservative and rebuild the schema, and have some
flag for local tests that allows tests to use a single schema copy. I
wouldn't add either of these until we have clarity on what's up with
deepcopy and (assuming the problem can be resolved) what the performance of
deepcopy would be.

pmantica1:
I am going to close this PR since obi1kenobi and I agreed that it is not worth it to keep working on this. (In other words, there are items higher on the priority list).

PR #875, opened 2020-07-06
Enable strict mypy enforcement with per-module exclusions.

obi1kenobi:
This is close to the strictest mypy configuration that currently passes. Honorable mentions to the `compiler.ir_lowering_common` and `post_processing` modules for passing all mypy checks in their entirety; all other modules have varying degrees of rule relaxations.

This setup should ensure that our typing effort does not significantly regress by virtue of people forgetting to add appropriate type hints. Over time, we should work to reduce and eventually eliminate all the rule relaxations, ideally on a module-by-module basis so that we can strengthen the checks as early as possible and limit the chances of a regression.

The setup currently also includes import exemptions for all modules that do not ship with baked-in type hints and are also lacking type stubs. Hopefully we'll also be able to reduce that list over time, though we have significantly less control in that area.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,3 @@ lines_after_imports = 2
 force_sort_within_sections = 1
 force_grid_wrap = 0
 combine_as_imports = True
-
-[mypy]
-ignore_missing_imports = True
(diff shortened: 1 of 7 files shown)


setup.cfg
```diff
@@ -34,6 +34,3 @@ lines_after_imports = 2
 force_sort_within_sections = 1
 force_grid_wrap = 0
 combine_as_imports = True
-
-[mypy]
-ignore_missing_imports = True
```
obi1kenobi, line 39: Moved to a dedicated `mypy.ini` file due to relatively large size.

PR #876, opened 2020-07-07
Generate the strictest possible mypy config.

obi1kenobi:
Auto-generated using `typing-copilot`: https://github.com/obi1kenobi/typing-copilot

Guaranteed to be very close to the strictest possible mypy config that still passes. The motivation is simple: the stricter the mypy config, the more likely we are to actually get closer to full type-hinting over time, and the less likely we are to suffer small typing regressions with each PR due to insufficiently strict enforcement. By setting the strictest possible config (and periodically updating as a newer, even more strict, config is available), our typing coverage can only remain constant or improve over time since those regressions will be caught and flagged by a failing mypy run.


pmantica1:
Looks good to me. Nice repo btw.

PR #877, opened 2020-07-07
Update lockfile to test with new GraphQL core release.

obi1kenobi:
Just a re-lock to pick up the new `graphql-core==3.1.2` release.

### DIFF
--- a/graphql_compiler/compiler/validation.py
+++ b/graphql_compiler/compiler/validation.py
@@ -55,8 +55,8 @@ def validate_schema_and_query_ast(schema, query_ast):
         ]
     )
 
-    # This directive is supported and ignored by the compiler, since it is meant as an indication
-    # to the user that a field should not be used.
+    # The following directives are supported and ignored by the compiler,
+    # since they are meant to communicate user-facing information.
     supported_default_directive = frozenset(
         [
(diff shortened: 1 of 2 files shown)


obi1kenobi:
Yeah, that's what I was worried about. The new `@specifiedBy` built-in directive breaks our extremely-conservative schema checking.

Fix coming soon.

graphql_compiler/compiler/validation.py
```diff
@@ -55,8 +55,8 @@ def validate_schema_and_query_ast(schema, query_ast):
         ]
     )
 
-    # This directive is supported and ignored by the compiler, since it is meant as an indication
-    # to the user that a field should not be used.
+    # The following directives are supported and ignored by the compiler,
+    # since they are meant to communicate user-facing information.
     supported_default_directive = frozenset(
```
pmantica1, line 60: nit: Missing an `s` after `supported_default_directive`. 

This is arguably an existing error so I'll approve the PR as is. If you want to fix it though in this PR feel free. I'll approve the PR immediately.

obi1kenobi: Good catch, thanks!

Issue #866, opened 2020-06-29
Incorrect sql output for recurse


bojanserafimov:
If there are no traversals, filters or coercions leading up to the recurse (therefore no need for a base-case cte), but there is an output field, the resulting sql output is incorrect, missing a join from the base where the output is selected, to the recursive cte.

It's not hard to fix this, just making an issue so I don't forget.

PR #880, opened 2020-07-14
Join recursive cte to base

bojanserafimov:
Currently we generate queries that have the following from clause `FROM anon_1, anon_2`, where `anon_1` is the query before a recurse, and `anon_2` is the recursive CTE. Since the two ctes are not joined, if outputs are selected from each, the result will be a cartesian product, which is incorrect.

There's two possible solutions:
a) Never select outputs from `anon_1`, since the same values can be accessed from `anon_2`
b) Join the ctes in the `FROM` clause

I went with option (b) since it makes it possible to implement recursion inside optional scope.

Note: When no outputs are selected from `anon_1`, there's no need to join, since `anon_1` can be excluded from the `FROM` clause completely. This optimization can be implemented later, and I'm not sure if it makes a difference in runtime.

Changes in this PR:
- Join the recursive CTE to the preceding CTE or alias
- Make sure this join does not violate `@optional` semantics if the `@recurse` is in optional scope
- Test
- Increase mypy strictness

This PR fixes https://github.com/kensho-technologies/graphql-compiler/issues/866


mypy.ini
```diff
@@ -9,7 +9,7 @@ disallow_untyped_calls = True
 disallow_incomplete_defs = True
 disallow_untyped_defs = True
 disallow_untyped_decorators = True
-warn_unused_ignores = False  # mypy-copilot: disabled as mypy only reports it if other checks pass
+warn_unused_ignores = True
```
bojanserafimov, line 12: @obi1kenobi I was getting the following warning here while running mypy:
`mypy.ini: [mypy]: warn_unused_ignores: Not a boolean: False  # mypy-copilot: disabled as mypy only reports it if other checks pass`

Seems to me like the `#` is not being parsed as a comment. The `#` comment syntax is not part of the basic spec, and only some implementations have it https://en.wikipedia.org/wiki/INI_file#Comments_2

In this case I just set it to true because it was an easily reviewable 5-line change that shouldn't add too much clutter. Let me know if you prefer a separate PR anyway.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -426,6 +426,92 @@ def test_recurse(self) -> None:
...
+            {
+                Animal {
+                    name @filter(op_name: "=", value: ["$starting_animal_name"])
+                    out_Animal_ParentOf {
+                        out_Animal_ParentOf @recurse(depth: 2){
+                            name @output(out_name: "descendant_name")
+                        }
+                    }
+                }
+            }""",
```
obi1kenobi, line 441: nit: Indent all of this in one step, to match the indentation of the starting `"""`.

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -426,6 +426,92 @@ def test_recurse(self) -> None:
...
                 ],
             ),
+            # Query 8: Skip depth 0
+            (
+                """
+            {
+                Animal {
+                    name @filter(op_name: "=", value: ["$starting_animal_name"])
+                    out_Animal_ParentOf {
+                        out_Animal_ParentOf @recurse(depth: 2){
```
obi1kenobi, line 436:
```suggestion
                        out_Animal_ParentOf @recurse(depth: 2) {
```

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -426,6 +426,92 @@ def test_recurse(self) -> None:
...
+            ),
+            # Query 9: Output child name
+            (
+                """
+            {
+                Animal {
+                    name @filter(op_name: "=", value: ["$starting_animal_name"])
+                    out_Animal_ParentOf {
+                        name @output(out_name: "child_name")
+                        out_Animal_ParentOf @recurse(depth: 2){
```
obi1kenobi, line 464:
```suggestion
                        out_Animal_ParentOf @recurse(depth: 2) {
```

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -426,6 +426,92 @@ def test_recurse(self) -> None:
...
+                            name @output(out_name: "descendant_name")
+                        }
+                    }
+                }
+            }""",
+                [
+                    {"child_name": "Animal 1", "descendant_name": "Animal 1"},  # depth 1 match
+                    {"child_name": "Animal 2", "descendant_name": "Animal 2"},  # depth 1 match
+                    {"child_name": "Animal 3", "descendant_name": "Animal 3"},  # depth 1 match
```
obi1kenobi, line 473: nit: off-by-one depth notation relative to the compiler's spec, these are depth-0 because the query says `@recurse(depth: 2)` i.e. the depths are 0 (no hops), 1 (one hop), and 2 (two hops), instead of 1-2-3 as the comments say here

graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
```diff
@@ -381,7 +381,7 @@ def test_eligible_fields(self) -> None:
 
     @pytest.mark.usefixtures("snapshot_orientdb_client")
     def test_distinct_result_set_estimates_with_revisit_counts(self):
-        schema_graph = generate_schema_graph(self.orientdb_client)  # type: ignore  # from fixture
+        schema_graph = generate_schema_graph(self.orientdb_client)
```
obi1kenobi, line 384: Nice, thank you!

mypy.ini
obi1kenobi: Yeah, I saw the same thing — I wish `mypy` would hard-fail instead of silently proceeding. Thanks for fixing, this is fine, though maybe update the comment on line 2 since this is no longer the same autogenerated file from `typing_copilot`.

graphql_compiler/tests/test_compiler.py
```diff
@@ -9662,10 +9677,14 @@ def test_simple_optional_recurse(self):
                 anon_1.[Animal_in_Animal_ParentOf__name] AS child_name,
                 anon_1.[Animal__name] AS name,
                 anon_2.name AS self_and_ancestor_name
-            FROM anon_1, anon_2
+            FROM
+                anon_1
+                LEFT OUTER JOIN anon_2
```
obi1kenobi, line 9682: I'm not sure why this is a `LEFT OUTER JOIN`. Isn't the `@optional` already handled by the `LEFT OUTER JOIN` on line 9651? As far as I can tell, because the recursive CTE always emits depth-0 recursions, this join should always have _something_ to join on, and `INNER JOIN` is therefore equivalent and probably semantically preferred.

bojanserafimov: Yeah actually, that makes sense. I'll run the integration test to double-check. This will simplify the implementation (no need for `self._came_from` shenanigans)

Going back to the PR description, I'd need to think more about solution (a) vs (b). I think I'll continue with (b), but it's good to know that (a) is still on the table.

obi1kenobi: If I understand correctly, I think joining the CTEs together is unavoidable because outputting from the recursive CTE vs the pre-recursion CTE might be semantically-different and corresponds to different input queries. Consider the following somewhat contrived but semantically-valid query:
```graphql
{
    Species {
        name @filter(op_name: "=", value: ["$dog"]) @output(out_name: "species")

        in_Animal_Species {
            name @output(out_name: "animal")

            out_Animal_ParentOf @recurse(depth: 1) {
                name @output(out_name: "self_or_child_animal")
            }
        }
    }
}
```

This is what I'd expect it to compile into for MSSQL:
```sql
WITH anon_1 AS (
    SELECT
        [Animal_1].name AS [Species_in_Animal_Species__name],
        [Animal_1].parent AS [Species_in_Animal_Species__parent],
        [Animal_1].uuid AS [Species_in_Animal_Species__uuid],
        [Species_1].name AS [Species__name]
    FROM
        db_1.schema_1.[Species] AS [Species_1]
        JOIN db_1.schema_1.[Animal] AS [Animal_1]
            ON [Species_1].uuid = [Animal_1].species
    WHERE
        [Species_1].name = :dog
),
anon_2(name, parent, uuid, __cte_key, __cte_depth) AS (
    SELECT
        anon_1.[Species_in_Animal_Species__name] AS name,
        anon_1.[Species_in_Animal_Species__parent] AS parent,
        anon_1.[Species_in_Animal_Species__uuid] AS uuid,
        anon_1.[Species_in_Animal_Species__uuid] AS __cte_key,
        0 AS __cte_depth
    FROM
        anon_1
    UNION ALL
    SELECT
        [Animal_2].name AS name,
        [Animal_2].parent AS parent,
        [Animal_2].uuid AS uuid,
        anon_2.__cte_key AS __cte_key,
        anon_2.__cte_depth + 1 AS __cte_depth
    FROM
        anon_2
        JOIN db_1.schema_1.[Animal] AS [Animal_2]
            ON anon_2.parent = [Animal_2].uuid
        WHERE anon_2.__cte_depth < 2
)
SELECT
    anon_1.[Species_in_Animal_Species__name] AS animal,
    anon_2.name AS self_or_child_animal,
    anon_1.[Species__name] AS species
FROM
    anon_1
    JOIN anon_2
        ON anon_1.[Species_in_Animal_Species__uuid] = anon_2.__cte_key
```

I don't see where we have the option to pick whether we output from `anon_1` or `anon_2`. Easy cases first:
- The value for `species` can only come from `anon_1`, hopefully that's self-evident since it has nothing to do with the recursion.
- The value for `self_or_child_animal` has to come from `anon_2` because `anon_1` doesn't know about it.

The only somewhat tricky case is the value for `animal`. I think we can rule out `anon_2.name` as the source for this value, because that's clearly the source for `self_or_child_animal` which has different semantics per the `@recurse`. `anon_2` doesn't output anything else on the basis of a `name` column (the explicitly-declared options are `name, parent, uuid, __cte_key, __cte_depth`), so I think this value has to come from `anon_1.[Species_in_Animal_Species__name]`.

So I've either entirely missed your point (and if so, I apologize), or it seems that the outputs from outside the `@recurse` block _must_ come from `anon_1` and the `JOIN` between the two CTEs is unavoidable.

bojanserafimov: `species` can come from `anon_2` if `anon_2` selects it in the base case, and copies it in the recursive step.

obi1kenobi: > `species` can come from `anon_2` if `anon_2` selects it in the base case, and copies it in the recursive step.

Sure — but that seems unnecessary. We'd just be carrying a known-constant value (from the perspective of any input row to `anon_2`) through the entire recursion without ever computing anything off of it. This would just add cost to the database computation, so I'd rather avoid it unless the code gets too gnarly.

> Back to the first comment in the chain: It's actually necessary to `LEFT JOIN`. On `@optional` miss, the preceding selectable will have a `NULL` value in the id column that joins to the recursive cte. So Even though the recursive CTE base case contains the whole of the preceding selectable, you can't join on `NULL`.

Ah yes, because of the ordering of steps (`anon_1` before `anon_2`) that's true. I forgot that the key isn't just coming from a `FROM Animal` clause and could therefore be `null`. Good catch!

graphql_compiler/compiler/emit_sql.py
```diff
@@ -743,7 +743,7 @@ def __init__(self, sql_schema_info: SQLAlchemySchemaInfo, ir: IrAndMetadata):
         self._relocate(ir.query_metadata_table.root_location)
 
         # Mapping aliases to the column used to join into them.
-        self._came_from: Dict[Alias, Column] = {}
+        self._came_from: Dict[Union[Alias, ColumnRouter], Column] = {}
```
bojanserafimov, line 746: The mention of `aliases` in the docstring is outdated. So is the entire `_aliases` field, since it doesn't only contain aliases anymore. I'm working on fixing this in the next PR, and I created an issue https://github.com/kensho-technologies/graphql-compiler/issues/882

graphql_compiler/compiler/emit_sql.py
```diff
@@ -913,10 +917,13 @@ def _wrap_into_cte(self) -> None:
         if self._recurse_needs_cte:
             self._current_alias = self.get_query(extra_outputs).cte(recursive=False)
             self._from_clause = self._current_alias
-
             self._filters = []  # The filters are already included in the cte
-            self._aliases = {
-                alias_key: ColumnRouter(
+
+            # After creating this CTE, accessing a field of any of the currently marked
+            # locations should redirect to the column of the cte that exposes that field.
```
obi1kenobi, line 923:
```suggestion
            # locations should redirect to the column of the CTE that exposes that field.
```

obi1kenobi: 2/3 uses of CTE in this block comment were capitalized, so I felt compelled to capitalize the last one :D

PR #881, opened 2020-07-14
Implement fast introspection execution module

lonerz:


graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,162 @@
...
+1.37 seconds. This module calls the same Type resolvers as graphql_sync to prevent
+code duplication with graphql-core.
+
+This module is fully dependent on the introspection query returned by `get_introspection_query`
+from graphql-core (with default parameters). Changes to that query will need to be
+reflected in this module. This module also assumes that the introspection query can
+be executed on the given schema.
+"""
+
+
```
obi1kenobi, line 14:
```suggestion

```

graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,162 @@
...
+__Schema = cast(GraphQLObjectType, introspection_types["__Schema"])
+__Directive = cast(GraphQLObjectType, introspection_types["__Directive"])
+__DirectiveLocation = cast(GraphQLEnumType, introspection_types["__DirectiveLocation"])
+__Type = cast(GraphQLObjectType, introspection_types["__Type"])
+__Field = cast(GraphQLObjectType, introspection_types["__Field"])
+__InputValue = cast(GraphQLObjectType, introspection_types["__InputValue"])
+__EnumValue = cast(GraphQLObjectType, introspection_types["__EnumValue"])
+__TypeKind = cast(GraphQLEnumType, introspection_types["__TypeKind"])
```
obi1kenobi, line 40: `pylint` is wrong here, `introspection_types` is subscriptable since it's of type `FrozenDict`, a subtype of `Dict`.
```suggestion
# pylint: disable=unsubscriptable-object
__Schema = cast(GraphQLObjectType, introspection_types["__Schema"])
__Directive = cast(GraphQLObjectType, introspection_types["__Directive"])
__DirectiveLocation = cast(GraphQLEnumType, introspection_types["__DirectiveLocation"])
__Type = cast(GraphQLObjectType, introspection_types["__Type"])
__Field = cast(GraphQLObjectType, introspection_types["__Field"])
__InputValue = cast(GraphQLObjectType, introspection_types["__InputValue"])
__EnumValue = cast(GraphQLObjectType, introspection_types["__EnumValue"])
__TypeKind = cast(GraphQLEnumType, introspection_types["__TypeKind"])
# pylint: enable=unsubscriptable-object
```
Feel free to open an issue in https://github.com/PyCQA/pylint about this, if there isn't one already — this is a pretty good reproducible example of a bug.

graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def _get_type_ref(type_: GraphQLType) -> Dict[str, Any]:
```
obi1kenobi, line 43: All these functions need docstrings that explain their purpose, otherwise this code is going to be very difficult to read and maintain.

cc @bojanserafimov @pmantica1 any idea why pydocstyle doesn't seem to mind the fact that docstrings are missing?

graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,162 @@
...
+        ],
+        "args": [
+            _get_input_value(arg) for arg in __Directive.fields["args"].resolve(directive, None)
+        ],
+    }
+
+
+def execute_fast_introspection_query(schema: GraphQLSchema) -> Dict[str, Any]:
+    """Compute the GraphQL introspection query."""
+
```
obi1kenobi, line 134: Avoid starting functions with blank lines after the docstring.
```suggestion
```

graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,162 @@
...
+            for location in __Directive.fields["locations"].resolve(directive, None)
+        ],
+        "args": [
+            _get_input_value(arg) for arg in __Directive.fields["args"].resolve(directive, None)
+        ],
+    }
+
+
+def execute_fast_introspection_query(schema: GraphQLSchema) -> Dict[str, Any]:
+    """Compute the GraphQL introspection query."""
```
obi1kenobi, line 133: This function is part of the user-facing public API of this module. As such, it needs quite a bit more detail in its docstring than this — otherwise, as a user, I probably wouldn't know what problem it solves for me, when it is appropriate to use it, and what functionality it replaces. That would mean it probably wouldn't get used, since users don't use features they don't understand or recognize the need for.

One especially important thing that we need to consider explicitly is this: how would a user know if _their_ introspection query is the same one that this module knows how to compute quickly? If the user is in-process and is therefore using `graphql-core` of the same version as the compiler, that's fine — but users don't have to do that. What if the user is using e.g. GraphiQL, which decides to send an introspection query to a server that is running this code? How would the server know if it can use this code to speed up the introspection or not? I don't think the current API solves this problem that users are very likely to run into, and I'd encourage you to revisit the end-to-end user experience and usage of this code in more detail.

graphql_compiler/fast_introspection.py
bojanserafimov: Seems like it's not configured to. http://www.pydocstyle.org/en/2.1.1/error_codes.html D102, D103 are missing.

Interesting that our `test_.*\.py` pydocstyle config has D102, therefore it's more strict than the non-test one.

bojanserafimov: The config seems correct. Another point of weirdness that might help: when I run pydocstyle locally as part of the lint script, it always fails.

bojanserafimov: Looking into it ..

bojanserafimov: http://www.pydocstyle.org/en/2.1.1/error_codes.html It seems like pydocstyle doesn't look at private functions

graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,300 @@
...
+}
+"""
+
+
+def remove_whitespace_from_query(query: str) -> str:
+    """Return an equivalent query with spaces and newline characters removed."""
+    return query.replace(" ", "").replace("\n", "")
+
+
+whitespace_free_introspection_query = remove_whitespace_from_query(_introspection_query)
```
obi1kenobi, line 141: nit: Thoughts on making these use underscore-prefixed names? I know tests need to import them and it's a bit iffy to import underscored names, but it makes it very explicit that these parts are not part of the public API and shouldn't be used from the outside (even from other places in the compiler other than this module's tests).

graphql_compiler/fast_introspection.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def try_fast_introspection(schema: GraphQLSchema, query: str) -> Optional[ExecutionResult]:
+    """Compute the GraphQL introspection query if query can be computed fastly.
+
+    This can be used in replacement of `graphql_sync`. It automatically checks if the query
+    is the expected introspection query in this module; if the query is not the expected
+    introspection query, this function returns None, meaning the query cannot be computed fastly
+    with this module. Otherwise, it will execute the introspection query fastly and return
+    the computed result in the form of a GraphQL ExecutionResult.
```
obi1kenobi, line 290: An (unfortunately poorly-documented) convention we do our best to stick to is that if a function requires a multi-line docstring, it should also always have explicit `Args / Returns` sections. For example: https://github.com/kensho-technologies/graphql-compiler/blob/main/graphql_compiler/compiler/compiler_frontend.py#L739

graphql_compiler/tests/test_fast_introspection.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_try_fast_introspection_none(self) -> None:
+        schema = get_schema()
+        self.assertEqual(try_fast_introspection(schema, "not the right query"), None)
+
+    def test_try_fast_introspection_equal_graphql_sync(self) -> None:
+        schema = get_schema()
+        execution_result = try_fast_introspection(schema, introspection_query)
+        self.assertIsNotNone(execution_result)
+        if execution_result:
```
obi1kenobi, line 31: This is implicitly `if bool(execution_result)` which is going to evaluate to `False` both if `execution_result == None` but also if somehow that object is falsy. That means that this test (and the one below) could pass if the calls return a falsy-but-non-None `ExecutionResult` value — perhaps one with no result rows, depending on the object's `__bool__` method.

I don't think this is intentional — I don't think that such an empty result is meant to be treated as a test success here. I'm assuming this was added because `mypy` didn't recognize that `assertIsNotNone()` will error on a `None` result? If so, let's cast the `Optional` away, or make this be `if execution_result is not None`.

graphql_compiler/fast_introspection.py
lonerz: Good suggestion, I think it's good to be explicit

graphql_compiler/tests/test_fast_introspection.py
lonerz: Yep, done to appease `mypy`. Added the explicit `is not None`.

PR #884, opened 2020-07-21
Schema rename collect conflict errors

LWprogramming:
Reporting all renaming conflicts at once instead of one at a time.

Specifically, the invariant we're preserving here is that no two types in the schema have the same name. There are two ways renaming might incorrectly let that happen: by renaming two types `Foo` and `Bar` both to the same name, or by renaming a type `Foo` to `Bar` when there already exists a type `Bar` in the schema that doesn't get renamed.

Previously, we had a `SchemaNameConflictError` which worked for both renaming and merging schemas. Now that we'd like to programmatically track the errors, I split them into `SchemaMergeNameConflictError` and `SchemaRenameNameConflictError`


graphql_compiler/schema_transformation/utils.py
```diff
@@ -39,16 +40,41 @@ class InvalidTypeNameError(SchemaTransformError):
...
+    """
+
+    name_conflicts: Dict[str, Set[str]]
+
+    def __init__(self, name_conflicts: Dict[str, Set[str]]) -> None:
+        """Record all renaming conflicts."""
+        super().__init__()
+        self.name_conflicts = name_conflicts
+
+    def __str__(self) -> str:
```
LWprogramming, line 65: So far all the exceptions I've seen get initialized with `ExceptionClassName(string_message_goes_here)`. When I searched it up, I found a number of different ways to modify the class (such as modifying `__str__` and/or `__repr__`) so that we get both a human-readable error message and the argument and it seems like this is the best way to do it.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -39,16 +40,54 @@ class InvalidTypeNameError(SchemaTransformError):
...
+        super().__init__()
+        self.name_conflicts = name_conflicts
+        self.renamed_to_scalar_conflicts = renamed_to_scalar_conflicts
+
+    def __str__(self) -> str:
+        """Explain renaming conflict and the fix."""
+        name_conflicts_message = ""
+        if self.name_conflicts != {}:
+            name_conflicts_message = (
+                f"Applying the renaming would produce a schema with multiple types with the same "
```
LWprogramming, line 73: I find the wording for these two messages a bit awkward, in particular these parts:

> Each entry here shows the type name that would appear in the renamed schema, along with all type names in the original schema that map to it

> it does not map a type name to a name that a different type will have in the resulting schema

> name already taken by a scalar

graphql_compiler/schema_transformation/utils.py
LWprogramming: The code coverage alert about `utils.py` is also because the `__str__` function doesn't get explicitly called in the tests.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -387,6 +387,10 @@ def _rename_and_suppress_types(
     """
     visitor = RenameSchemaTypesVisitor(renamings, query_type, scalars)
     renamed_schema_ast = visit(schema_ast, visitor)
+    if visitor.name_conflicts != {} or visitor.renamed_to_scalar_conflicts != {}:
```
obi1kenobi, line 390: Can these be `None`?

graphql_compiler/schema_transformation/utils.py
obi1kenobi: The code coverage alert is correct. Consider adding a test for this `__str__` since it isn't trivial.

For example, test coverage would probably have caught the fact that the function currently contains a subtle bug: the two error messages get concatenated without a space between them, resulting in a message like `[...] the resulting schema.Applying the renaming [...]`.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: D'oh, I forgot to change these after #885 and it slipped my notice when I was re-reading yesterday.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -477,6 +481,13 @@ class RenameSchemaTypesVisitor(Visitor):
             "UnionTypeDefinitionNode",
         }
     )
+    name_conflicts: Dict[str, Set[str]]  # Collects naming conflict errors. If renaming
+    # would result in multiple types being renamed to "Foo", name_conflicts will map "Foo" to an
+    # set containing the name of each such type.
+    renamed_to_scalar_conflicts: Dict[str, str]  # Collects attempts to rename a type with a name
+    # already used by a scalar. If renamings tries to rename a type named "Foo" to "Bar" and there
+    # already exists a scalar named "Bar" in the schema, then
+    # renamed_to_scalar_conflicts["Foo"] == "Bar"
```
chewselene, line 490: any reason to not include `reverse_name_map` up here rather than in the `__init__`? might be good to add typing/descriptions for the other class properties as well.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -543,17 +558,12 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
...
         ):
-            raise SchemaNameConflictError(
-                '"{}" and "{}" are both renamed to "{}"'.format(
-                    name_string, self.reverse_name_map[new_name_string], new_name_string
-                )
-            )
+            # At least two different types are mapped to the same new_name_string
+            if new_name_string not in self.name_conflicts:
+                self.name_conflicts[new_name_string] = {self.reverse_name_map[new_name_string]}
+            self.name_conflicts[new_name_string].add(name_string)
```
chewselene, line 564: I think this could be streamlined with `setdefault` https://www.geeksforgeeks.org/python-dictionary-setdefault-method/

graphql_compiler/schema_transformation/utils.py
chewselene: For the first,
`Received the following name conflicts: {self.name_conflicts}, where each entry maps a type in the renamed schema ("key") to the set of types in the original schema that would be renamed to "key".`
^ oof it is really tough to write this clearly, but this is my best attempt...

For the second, I think you can eliminate that part entirely because it is just rephrasing `ensure that no two types in the renamed schema have the same name` in a more confusing way. Instead,
`To fix this, modify the renamings argument of rename_schema to ensure that no two types in the renamed schema have the same name.`

I don't actually think the third sounds too bad, but maybe `used by` would sound better than `taken by`?

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Done for `reverse_name_map`! My rule for these is if an attribute gets used outside of the class, I make it a class attribute, so we can leave the others as-is.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: I think this came up at one point and neither Predrag or I had super strong opinions about it-- I find the existing code slightly easier to read, but that's just my personal preference

graphql_compiler/schema_transformation/utils.py
LWprogramming: Updated, how does this sound?

graphql_compiler/schema_transformation/utils.py
LWprogramming: Fixed the bug, although I'm not sure on how exactly to test `__str__`. We could look for an exact match for the hardcoded parts (i.e. ignore any non-deterministic issues when converting dicts and sets to strings), but that'd change as soon as the error message changes. At that point it seems like it'd be basically the same as testing that the code raises a `SchemaRenameNameConflictError` at all. 

Alternatively, we could test to see if attributes like `name_conflicts` show up in the error message (which wouldn't catch the bug you mentioned above, but would check that we're reporting the errors correctly). That could also get [messy](https://docs.python.org/3/library/ast.html#ast.literal_eval) especially if there are a lot of errors involved.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -477,6 +481,17 @@ class RenameSchemaTypesVisitor(Visitor):
...
+    # would result in multiple types being renamed to "Foo", name_conflicts will map "Foo" to an
+    # set containing the name of each such type.
+    renamed_to_scalar_conflicts: Dict[str, str]  # Collects attempts to rename a type with a name
+    # already used by a scalar. If renamings tries to rename a type named "Foo" to "Bar" and there
+    # already exists a scalar named "Bar" in the schema, then
+    # renamed_to_scalar_conflicts["Foo"] == "Bar"
+    reverse_name_map: Dict[str, str]  # From renamed type name to original type name
+    # reverse_name_map contains all non-suppressed types, including those that were unchanged. Must
+    # contain unchanged names to prevent renaming conflicts and raise SchemaNameConflictError when
+    # they arise.
```
obi1kenobi, line 494: These comments feel kind of awkward to read to me. Consider reformatting in a style like this:
```
# comment
# comment
attr: type

# comment
# comment
attr2: type
```

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: I find `setdefault()` slightly preferable but I wasn't going to block your PR over it 😄

graphql_compiler/schema_transformation/utils.py
obi1kenobi: Dicts are ordered (perhaps sets too?), so I don't think there should be non-determinism issues. I think it's fine to require the test to change if the error message changes.

Here's the underlying thing worth mentioning. Not writing a test for something is implicitly saying "this code isn't buggy, and the code itself is so obvious that debugging a broken test will be harder than just spotting the bug in the code itself." In this case, there was a bug due to some non-trivial logic, and the bug wasn't obvious but would have been easily caught by tests that exercise the various branches. So I think tests are warranted.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -39,16 +40,54 @@ class InvalidTypeNameError(SchemaTransformError):
...
+                f"to ensure that no two types in the renamed schema have the same name."
+            )
+        renamed_to_scalar_conflicts_message = ""
+        if self.renamed_to_scalar_conflicts:
+            renamed_to_scalar_conflicts_message = (
+                f"Applying the renaming would rename the following types to a name already used by "
+                f"a scalar: {self.renamed_to_scalar_conflicts}. To fix this, ensure that no type "
+                f"name is mapped to a scalar's name."
+            )
+        return " ".join(filter(None, [name_conflicts_message, renamed_to_scalar_conflicts_message]))
```
obi1kenobi, line 88: Is `" ".join([])` AKA empty string a legal output here? If not, let's prevent this error object from being constructed in such a way, and add an assertion against it here.

This is another thing that a test could have potentially uncovered.

graphql_compiler/schema_transformation/utils.py
LWprogramming: > Dicts are ordered (perhaps sets too?)

Did some digging and [dicts are ordered](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6) according to the spec for Python 3.7+ and as an implementation detail in the CPython implementation for 3.6, but not explicitly part of the spec. I'll implement the tests, but will this potentially cause issues later down the road?

obi1kenobi: Good catch. The question then becomes "are we more likely to support a non-CPython 3.6 version, or drop Python 3.6 first?"

I'd argue the answer is that we're more likely to drop Python 3.6 first.

In general, I don't love depending on implementation details like this, but in this case it seems like the cure would be worse than the disease. I'd argue the prudent choice here would be to get to "good" now vs getting to "perfect" later.

LWprogramming: Hmm, after some further experimentation and thinking I think this might've been a moot point because insertion order for the dicts and sets ([sets are still unordered anyways](https://stackoverflow.com/questions/45581901/are-sets-ordered-like-dicts-in-python3-6)) depends on the order in which the `GraphQL` visitor visits nodes at the same depth in the schema AST (e.g. one object type vs. the other), and I don't see any guarantees about that in the docs.

~~As a workaround, I'll split the string up and check the parts separately.~~ I'll split the two cases into two separate classes because my initial attempt at writing adequate tests for `__str__` when both cases are covered by `SchemaRenameNameConflictError` ended up requiring a lot of [stringly-typed](https://blog.codinghorror.com/new-programming-jargon/) silliness.

LWprogramming:
Going to split this into two PRs because this is doing two things at this point:

* treating custom scalars and built-in scalars differently

* collecting renaming conflict errors

Temporarily closing this PR, don't delete

PR #885, opened 2020-07-22
Move rename scalar error check

LWprogramming:
Previously, renamings would silently ignore any attempts to rename scalars (whether user-defined or built-in scalar types)

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -303,7 +303,7 @@ def _ensure_no_unsupported_scalar_operations(
     for scalar in scalars:
         if renamings.get(scalar, scalar) != scalar:
             # Then it either attempts to rename or suppress the scalar.
-            unsupported_scalar_operations[scalar] = renamings.get(scalar)
+            unsupported_scalar_operations[scalar] = renamings[scalar]
     if unsupported_scalar_operations != {}:
         raise NotImplementedError(
             f"Scalar renaming and suppression is not implemented yet, but renamings attempted to "
@@ -497,7 +497,7 @@ class RenameSchemaTypesVisitor(Visitor):
         # reverse_name_map contains all non-suppressed types, including those that were unchanged
         self.query_type = query_type
         self.scalar_types = frozenset(scalar_types)
-        
+
         # pylint produces a false positive-- see issue here:
         # https://github.com/PyCQA/pylint/issues/3743
         self.builtin_types = frozenset(specified_scalar_types.keys())  # pylint: disable=no-member
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -239,11 +239,11 @@ def _validate_renamings(
         - NotImplementedError if renamings attempts to suppress an enum, an interface, or a type
           implementing an interface
     """
-    _check_for_cascading_type_suppression(schema_ast, renamings, query_type)
-    _check_for_unsupported_operation(schema_ast, renamings, scalars)
+    _ensure_no_cascading_type_suppressions(schema_ast, renamings, query_type)
+    _ensure_no_unsupported_operations(schema_ast, renamings, scalars)
 
 
-def _check_for_cascading_type_suppression(
+def _ensure_no_cascading_type_suppressions(
     schema_ast: DocumentNode, renamings: Mapping[str, Optional[str]], query_type: str
 ) -> None:
     """Check for fields with suppressed types or unions whose members were all suppressed."""
@@ -287,15 +287,15 @@ def _check_for_cascading_type_suppression(
         raise CascadingSuppressionError("\n".join(error_message_components))
 
 
-def _check_for_unsupported_operation(
+def _ensure_no_unsupported_operations(
     schema_ast: DocumentNode, renamings: Mapping[str, Optional[str]], scalars: AbstractSet[str],
 ) -> None:
     """Check for unsupported renaming or suppression operations."""
-    _check_for_unsupported_scalar_operation(renamings, scalars)
-    _ensure_no_unsupported_suppression(schema_ast, renamings)
+    _ensure_no_unsupported_scalar_operations(renamings, scalars)
+    _ensure_no_unsupported_suppressions(schema_ast, renamings)
 
 
-def _check_for_unsupported_scalar_operation(
+def _ensure_no_unsupported_scalar_operations(
     renamings: Mapping[str, Optional[str]], scalars: AbstractSet[str],
 ) -> None:
     """Check for unsupported scalar operations."""
@@ -312,7 +312,7 @@ def _check_for_unsupported_scalar_operation(
         )
 
 
-def _ensure_no_unsupported_suppression(
+def _ensure_no_unsupported_suppressions(
     schema_ast: DocumentNode, renamings: Mapping[str, Optional[str]]
 ) -> None:
     """Confirm renamings contains no enums, interfaces, or interface implementation suppressions."""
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -280,6 +287,31 @@ def _check_for_cascading_type_suppression(
         raise CascadingSuppressionError("\n".join(error_message_components))
 
 
+def _check_for_unsupported_operation(
+    schema_ast: DocumentNode, renamings: Mapping[str, Optional[str]], scalars: AbstractSet[str],
+) -> None:
+    """Check for unsupported renaming/ suppression operations."""
```
obi1kenobi, line 293:
```suggestion
    """Check for unsupported renaming or suppression operations."""
```

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -226,14 +231,16 @@ def _validate_renamings(
...
+                 builtin scalars, excluding unused builtins
 
     Raises:
         - CascadingSuppressionError if a type suppression would require further suppressions
         - NotImplementedError if renamings attempts to suppress an enum, an interface, or a type
           implementing an interface
     """
     _check_for_cascading_type_suppression(schema_ast, renamings, query_type)
-    _ensure_no_unsupported_suppression(schema_ast, renamings)
+    _check_for_unsupported_operation(schema_ast, renamings, scalars)
```
obi1kenobi, line 243: Consider using a plural form, `operations`. Also, I kind of prefer `ensure` since it more clearly communicates the outcome if an unsupported operation is found. When I see `ensure` I definitely think "error will be raised if anything is found", when I see `check` I may very well think "it'll return a value telling me what it found".

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -280,6 +287,31 @@ def _check_for_cascading_type_suppression(
...
+) -> None:
+    """Check for unsupported scalar operations."""
+    unsupported_scalar_operations = {}  # Map scalars to value to be renamed.
+    for scalar in scalars:
+        if renamings.get(scalar, scalar) != scalar:
+            # Then it either attempts to rename or suppress the scalar.
+            unsupported_scalar_operations[scalar] = renamings.get(scalar)
+    if unsupported_scalar_operations != {}:
+        raise NotImplementedError(
+            f"Scalar renaming/ suppression is not implemented yet, but renamings attempted to "
```
obi1kenobi, line 309:
```suggestion
            f"Scalar renaming and suppression is not implemented yet, but renamings attempted to "
```

LWprogramming:
> Can you confirm that we don't rename / suppress / otherwise do anything to change the built-in scalars?

The code checks for that [here](https://github.com/kensho-technologies/graphql-compiler/pull/885/files#diff-3b859f46e97b0af834db0b3b7de08658R549)

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -465,7 +497,9 @@ def __init__(
         # reverse_name_map contains all non-suppressed types, including those that were unchanged
         self.query_type = query_type
         self.scalar_types = frozenset(scalar_types)
-        self.builtin_types = frozenset({"String", "Int", "Float", "Boolean", "ID"})
+        self.builtin_types = frozenset(specified_scalar_types.keys())  # pylint: disable=E1101
```
obi1kenobi, line 500: Please use the full name of the erroneous pass rather than the error code, because error codes are opaque and it's impossible to easily tell what is being suppressed.

If the full name is too long to fit on one line, use a `disable` on the line before, followed by an explicit `enable` on the next line.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -786,6 +768,14 @@ def test_field_in_list_still_depends_on_suppressed_type(self):
...
         class PrefixNewDict(object):
             def get(self, key, default=None):
+                if key == "Date" or key in specified_scalar_types:  # pylint: disable=E1135
+                    # pylint produces a false positive when checking for membership for
+                    # specified_scalar_types-- see issue here:
+                    # https://github.com/PyCQA/pylint/issues/3743
+                    # specified_scalar_types is a FrozenDict whose keys are built-in scalar type
+                    # names (e.g. "String")
+                    return key  # Making an exception for Date and all built-in scalar types because
+                    # they are scalars and renaming and suppressing them hasn't been implemented yet
```
obi1kenobi, line 778: Don't we have other custom scalars in addition to `Date`? I'm not sure why `Date` in particular is being special-cased, it would be good to explicitly mention that.

Also, I find this comment hard to read. How about this formatting instead:
```suggestion
                    # names (e.g. "String")
                    #
                    # Making an exception for Date and all built-in scalar types because
                    # they are scalars and renaming and suppressing them hasn't been implemented yet
                    return key  
```
or perhaps even better, code explanation first, pylint suppression comment and description after?

LWprogramming: `Date` is the only scalar used in the schema-- however, after I wrote this code, I found a better way to do this that doesn't require any pylint suppressions or hard-coded values.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -280,7 +287,32 @@ def _check_for_cascading_type_suppression(
...
+
+def _ensure_no_unsupported_scalar_operations(
+    renamings: Mapping[str, Optional[str]], scalars: AbstractSet[str],
+) -> None:
+    """Check for unsupported scalar operations."""
+    unsupported_scalar_operations = {}  # Map scalars to value to be renamed.
+    for scalar in scalars:
+        if renamings.get(scalar, scalar) != scalar:
+            # Then it either attempts to rename or suppress the scalar.
+            unsupported_scalar_operations[scalar] = renamings.get(scalar)
```
obi1kenobi, line 306: Why `.get()` here, and not `renamings[scalar]`?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -280,7 +287,32 @@ def _check_for_cascading_type_suppression(
...
+    if unsupported_scalar_operations != {}:
```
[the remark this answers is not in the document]
LWprogramming: Hmm, I remember seeing something about explicitly checking for `foo is not None` instead of `if foo` at one point-- what makes this one different? I'm seeing some contradictory answers online.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -280,7 +287,32 @@ def _check_for_cascading_type_suppression(
...
+
+
+def _ensure_no_unsupported_scalar_operations(
+    renamings: Mapping[str, Optional[str]], scalars: AbstractSet[str],
+) -> None:
+    """Check for unsupported scalar operations."""
+    unsupported_scalar_operations = {}  # Map scalars to value to be renamed.
+    for scalar in scalars:
+        if renamings.get(scalar, scalar) != scalar:
+            # Then it either attempts to rename or suppress the scalar.
```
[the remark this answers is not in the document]
LWprogramming: The intent here is that `renamings` attempts to do _something_ with the scalar, maybe renaming or maybe suppression. This is the case iff `renamings.get(scalar, scalar)` returns something that is _not_ `scalar`

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: That's odd, last time I tried that I got some sort of error about how the brackets only work for dicts and not for `Mapping` but upon re-running it now I don't see a problem, so I'll change it.

graphql_compiler/schema_transformation/rename_schema.py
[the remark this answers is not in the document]
obi1kenobi: Yeah, I figured that's what's happening. I'd just rephrase the comment to clarify it, since it's unusual for a `Then` (especially a capitalized one at the start of a sentence) to be matched up with the `if` in the code. It definitely threw me off for a bit when I first read it.

PR #886, opened 2020-07-27
Add typing_extensions as a conditional dependency.

obi1kenobi:
In compiler/ir_lowering_common/common.py we import `from typing_extensions import TypedDict`. However, `TypedDict` exists in the standard `typing` library as of Python 3.8: https://docs.python.org/3/library/typing.html#typing.TypedDict

Before that version, it has to be used from `typing_extensions`, but `typing_extensions` was not declared as a package dependency in such Python versions.

To resolve, adding a conditional import to the top-level typedefs.py file, and a conditional installation of typing_extensions for Python versions prior to 3.8.

### DIFF
--- a/graphql_compiler/typedefs.py
+++ b/graphql_compiler/typedefs.py
@@ -4,10 +4,15 @@ from typing import Union
 from graphql import GraphQLList, GraphQLNonNull, GraphQLScalarType
 
 
+# The below code is an import shim for TypedDict: we don't want to conditionally import it from
+# every file that needs it. Instead, we conditionally import it here and then import from this file
+# in every other location where this is needed.
+#
+# Hence, the "unused import" warnings on TypedDict here are false-positives.
 try:
-    from typing import TypedDict
+    from typing import TypedDict  # noqa  # pylint: disable=unused-import
 except ImportError:  # TypedDict was only added to typing in Python 3.8
-    from typing_extensions import TypedDict
+    from typing_extensions import TypedDict  # noqa  # pylint: disable=unused-import
 
 
 # The compiler's supported GraphQL types for query arguments. The GraphQL type of a query argument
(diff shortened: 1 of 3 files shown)


graphql_compiler/typedefs.py
```diff
@@ -4,6 +4,12 @@
 from graphql import GraphQLList, GraphQLNonNull, GraphQLScalarType
 
 
+try:
+    from typing import TypedDict
+except ImportError:  # TypedDict was only added to typing in Python 3.8
+    from typing_extensions import TypedDict
```
MichaelaShtilmanMinkin, line 10: Some some reason flake8 isn't recognizing this as being used, even though we import it.

obi1kenobi: The `flake8` and `pylint` import use analyses run on a per-file basis, they don't expect that we're importing this here to be available for importing from other files.

graphql_compiler/typedefs.py
```diff
@@ -4,6 +4,17 @@
...
 
+# The below code is an import shim for TypedDict: we don't want to conditionally import it from
+# every file that needs it. Instead, we conditionally import it here and then import from this file
+# in every other location where this is needed.
+#
+# Hence, the "unused import" warnings on TypedDict here are false-positives.
+try:
+    from typing import TypedDict  # noqa  # pylint: disable=unused-import
+except ImportError:  # TypedDict was only added to typing in Python 3.8
+    from typing_extensions import TypedDict  # noqa  # pylint: disable=unused-import
```
MichaelaShtilmanMinkin, line 15: Looks like this is a `no fix` sort of situation btw: https://github.com/python/mypy/issues/1107

obi1kenobi: Thankfully, I think `mypy` doesn't mind in this case. Another positive in this situation is that `TypedDict` has the same signature no matter where it is imported from.

PR #887, opened 2020-07-27
Fix recurse duplication bug

bojanserafimov:
Exposed and fix a duplication bug caused by `@recurse`.

Changes:
- Added integration test that used to fail
- Fixed implementation so test passes
- Fixed compiler tests to match output
- Optimization: Removed semi join when there's no base CTE (@obi1kenobi please double-check the correctness here)

Future PR:
- Since the new recurse implementation is simpler, see if any outputs we used to expose are now unused

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1022,6 +1022,13 @@ class CompilationState(object):
             + [base_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
         )
         if self._recurse_needs_cte:
+            # Optimization: Only compute the recursion for the valid starting points -- ones that
+            # will not be discarded when the recursive cte is joined to the rest of the query.
+            # The SQL database could do this predicate pushdown itself, but most implementations
+            # don't.
+            # This optimization is absolutely necessary for queries with strong filters before
+            # the recursion. It improves performance by a factor of the filter selectivity,
+            # which can go up to 1 million times in common queries.
             base = base.where(
                 base_alias.c[primary_key].in_(sqlalchemy.select([previous_alias.c[primary_key]]))
             )
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1026,7 +1026,7 @@ class CompilationState(object):
             # will not be discarded when the recursive cte is joined to the rest of the query.
             # The SQL database could do this predicate pushdown itself, but most implementations
             # don't.
-            # This optimization is absolutely necessary for queries with strong filters before
+            # This optimization is absolutely necessary for queries with highly selective filters before
             # the recursion. It improves performance by a factor of the filter selectivity,
             # which can go up to 1 million times in common queries.
             base = base.where(
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1028,7 +1028,7 @@ class CompilationState(object):
             # don't.
             # This optimization is absolutely necessary for queries with highly selective filters before
             # the recursion. It improves performance by a factor of the filter selectivity,
-            # which can go up to 1 million times in common queries.
+            # which can be 1 million times or more in common queries.
             base = base.where(
                 base_alias.c[primary_key].in_(sqlalchemy.select([previous_alias.c[primary_key]]))
             )
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -133,7 +133,7 @@ def _find_used_columns(
 
     # Columns used in the base case of CTE recursions should be made available from parent scope
     # TODO(bojanserafimov): Some of these are no longer needed, since we don't select from
-    #                       the base cete, but only semijoin to its primary key now.
+    #                       the base cte, but only semijoin to its primary key now.
     for location, _ in ir.query_metadata_table.registered_locations:
         for recurse_info in ir.query_metadata_table.get_recurse_infos(location):
             traversal = f"{recurse_info.edge_direction}_{recurse_info.edge_name}"
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1017,16 +1017,15 @@ class CompilationState(object):
 
         # The base of the recursive CTE selects all needed columns and sets the depth to 0
         base_alias = self._current_alias.alias()
-        base = (
-            sqlalchemy.select(
-                [base_alias.c[col].label(col) for col in used_columns]
-                + [base_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
-            )
-            .where(
+        base = sqlalchemy.select(
+            [base_alias.c[col].label(col) for col in used_columns]
+            + [base_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
+        )
+        if self._recurse_needs_cte:
+            base = base.where(
                 base_alias.c[primary_key].in_(sqlalchemy.select([previous_alias.c[primary_key]]))
             )
-            .cte(recursive=True)
-        )
+        base = base.cte(recursive=True)
 
         # The recursive step selects all needed columns, increments the depth, and joins to the base
         step = self._current_alias.alias()
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1027,7 +1027,7 @@ class CompilationState(object):
             # The SQL database could do this predicate pushdown itself, but most implementations
             # don't.
             # This optimization is absolutely necessary for queries with highly selective filters before
-            # the recursion. It improves performance by a factor of the filter selectivity,
+            # the recursion. It improves performance by a factor of (1 / filter selectivity),
             # which can be 1 million times or more in common queries.
             base = base.where(
                 base_alias.c[primary_key].in_(sqlalchemy.select([previous_alias.c[primary_key]]))
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1014,10 +1014,13 @@ class CompilationState(object):
         used_columns = sorted(self._used_columns[self._current_location.query_path])
 
         # The base of the recursive CTE selects all needed columns and sets the depth to 0
+        base_alias = self._current_alias.alias()
         base = sqlalchemy.select(
-            [previous_alias.c[col].label(col) for col in used_columns]
-            + [previous_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
-        ).cte(recursive=True)
+            [base_alias.c[col].label(col) for col in used_columns]
+            + [base_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
+        ).where(base_alias.c[primary_key].in_(
+            sqlalchemy.select([previous_alias.c[primary_key]])
+        )).cte(recursive=True)
 
         # The recursive step selects all needed columns, increments the depth, and joins to the base
         step = self._current_alias.alias()
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -132,6 +132,8 @@ def _find_used_columns(
         used_columns.setdefault(query_path, set()).add(tag_info.location.field)
 
     # Columns used in the base case of CTE recursions should be made available from parent scope
+    # TODO(bojanserafimov): Some of these are no longer needed, since we don't select from
+    #                       the base cete, but only semijoin to its primary key now.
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -555,6 +555,129 @@ def test_recurse(self) -> None:
...
+                    {"animal": "Animal 3", "ancestor": "Animal 1"},  # parent
+                    {"animal": "Animal 3", "ancestor": "Animal 1"},  # grandparent
+                    {"animal": "Animal 4", "ancestor": "Animal 4"},  # self
+                    {"animal": "Animal 4", "ancestor": "Animal 3"},  # parent
+                    {"animal": "Animal 4", "ancestor": "Animal 1"},  # grandparent
+                ],
+            ),
+            # Query 4: Unfold recursion to omit self
+            #
+            #  XXX: This output is incorrect. I'm just showing the current output. Results
```
obi1kenobi, line 630: This might be better split up off into its own test suite, with each test query as a separate test function, and this last one as an `xfail` test (with the correct input) until we can fix it.

bojanserafimov: The point of the first 3 tests is to run before the fourth. I'm testing recurse results in relation to non-recurse results:
- If the data changes and test 4 fails, likely test 3 will fail as well, pointing to the issue instead of suggesting recurse might be broken.
- If the data changes and test 4 passes, but is no longer a good test for duplication, test 3 will likely fail.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1014,10 +1014,13 @@ def recurse(self, vertex_field: str, depth: int) -> None:
         used_columns = sorted(self._used_columns[self._current_location.query_path])
 
         # The base of the recursive CTE selects all needed columns and sets the depth to 0
```
obi1kenobi, line 1016: This comment might need expanding / updating, once you've gotten this working.

obi1kenobi:
Here are my notes from debugging this, and how I figured out the working alternative that @bojanserafimov is implementing. The overall approach I used to get to the solution is to break down the query into composable parts and solve them separately. That allowed me to figure out that our recursive CTE was incorrectly defined (or, equivalently, was applying an illegal optimization — depending on how you want to look at it).

Here’s the query in the failing test:
```graphql
{
    Animal {
        name @output(out_name: "animal")
        in_Animal_ParentOf {
            in_Animal_ParentOf @recurse(depth: 1) {
                name @output(out_name: "ancestor")
            }
        }
    }
}
```

Here’s my sketch of the data graph we’re querying over (top left/center) and the query’s graph representation (top right):
image

In the bottom part, you see the two components of the query graph, semantically equivalent to the following two queries (with all outputs stripped out and replaced with `<...>` since they are irrelevant at the moment).

The first component is equivalent to the following query:
```graphql
{
    Animal {
        <...>
        in_Animal_ParentOf {
            <...>
        }
    }
}
```

The second component is equivalent to the following query:
```
{
    Animal {
        in_Animal_ParentOf @recurse(depth: 1) {
            <...>
        }
    }
}
```

Adjacent to the query graph diagrams of the components, you see the matching result sets we’d get if we executed just that component of the query. (“Result set” is a well-defined term in our [language specification](https://graphql-compiler.readthedocs.io/en/latest/language_specification/definitions.html) but I don’t think the definition is surprising, just wanted to link to it in case you’d like to read more.)

GraphQL compiler queries satisfy the composition property. (All the following stuff is explicitly defined in the spec linked above as well.) That means that when we split up the original query into two components, we can solve each component separately and then combine the answers to compute the answer to the original query. Remember that when splitting the original query, we actually duplicated the “middle” vertex in the original query graph — it appears in both query components. Therefore, to get the result sets that answer the original query, we construct them from the set of pairs `(A, B)` where `A` is a result set for the first component, `B` is a result set for the second component, and `A` and `B` have the same vertex in the “duplicated” position. This is equivalent to a SQL `INNER JOIN` i.e. what happens when you just write `JOIN`. So far, so good.

For the query being tested, the compiler was producing the following SQL:
```sql
WITH anon_2 AS (
    SELECT 
        [Animal_1].name AS [Animal__name], 
        [Animal_1].parent AS [Animal__parent], 
        [Animal_2].name AS [Animal_in_Animal_ParentOf__name], 
        [Animal_2].parent AS [Animal_in_Animal_ParentOf__parent], 
        [Animal_2].uuid AS [Animal_in_Animal_ParentOf__uuid] 
    FROM 
        db_1.schema_1.[Animal] AS [Animal_1] 
    JOIN 
        db_1.schema_1.[Animal] AS [Animal_2] ON [Animal_1].parent = [Animal_2].uuid
), 
anon_1(name, parent, uuid, __cte_key, __cte_depth) AS (
    SELECT 
        anon_2.[Animal_in_Animal_ParentOf__name] AS name, 
        anon_2.[Animal_in_Animal_ParentOf__parent] AS parent, 
        anon_2.[Animal_in_Animal_ParentOf__uuid] AS uuid, 
        anon_2.[Animal_in_Animal_ParentOf__uuid] AS __cte_key, 
        0 AS __cte_depth 
    FROM 
        anon_2 
    UNION ALL 
    SELECT 
        [Animal_3].name AS name, 
        [Animal_3].parent AS parent, 
        [Animal_3].uuid AS uuid, 
        anon_1.__cte_key AS __cte_key, 
        anon_1.__cte_depth + 1 AS __cte_depth 
    FROM 
        anon_1 
    JOIN 
        db_1.schema_1.[Animal] AS [Animal_3] ON anon_1.parent = [Animal_3].uuid 
    WHERE
        anon_1.__cte_depth < 1
)
SELECT 
    anon_1.name AS ancestor, 
    anon_2.[Animal__name] AS animal 
FROM 
    anon_2 
JOIN 
    anon_1 ON anon_2.[Animal_in_Animal_ParentOf__uuid] = anon_1.__cte_key
```

Here, `anon_2` is an exact match for the first query component: to confirm, just replace the bottom select with `SELECT * FROM anon_2` and you’ll see the same results produced as I wrote down on paper. However, doing the same for `anon_1` unexpectedly produces different results! Aha — a smoking gun!

Let’s dig into `anon_1`. Notice that our diagram for the second query component (bottom left in above photo) visually says “start from Animal and recurse to depth 1”, but `anon_1` says `FROM anon_2` followed by the (correct) recursion. That’s strange! Changing this to `FROM db_1.schema_1.[Animal]` produces correct results — so we’re on the right track.

That change, however, undoes an optimization we badly want: the SQL database is now likely to join all of the Animal table to itself, before discarding down to only the rows we want. But imagine if we were only interested in the family tree of a single animal: there's no point in computing *every* animal's family tree and then discarding all but the one we care about. Unfortunately, since CTEs are optimization barriers in most flavors of SQL, this is precisely what would happen with this setup — and what the original incorrect optimization was trying to avoid.

To make the query more efficient, we’d like to limit the vertices where the recursion is allowed to start to only the vertices that are known to be in the result sets produced by `anon_2`. The recursive CTE in `anon_1` consists of two `SELECT` parts: the base case (the first `SELECT`), and the recursive step (the `UNION ALL SELECT`). To apply this optimization, we add a `WHERE` clause to the base case that only selects vertices that were present in `anon_2`, a so-called [“semijoin”](https://sqlperformance.com/2018/02/sql-plan/row-goals-part-2-semi-joins) operation: `WHERE uuid IN (SELECT anon_2.[Animal_in_Animal_ParentOf__uuid] FROM anon_2)`, where `uuid` is the CTE key (selected as `__cte_key`).

This completes our journey: our query is now complete, correct, and has the same optimization we intended it to have before. Remember that this semijoin-based approach is just one way to apply this optimization, and there are many others. We now have something that works, but I expect we will continue to revisit this over time, and we may find another way that is consistently faster than the semijoin. This is not only fine but highlights the power of GraphQL compiler — our users simply write `@recurse` and don’t care about the underlying approach the compiler chose for their query! In other words, they get all the benefits (faster queries over time as we improve the compiler) with none of the downsides (they don't need to rewrite their query, and they don’t even notice we upgraded the compiler outside of seeing performance improvements).

graphql_compiler/compiler/emit_sql.py
bojanserafimov: Explained a few lines below. See if the comment needs more clarity/context.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -132,6 +132,8 @@ def _find_used_columns(
         used_columns.setdefault(query_path, set()).add(tag_info.location.field)
 
     # Columns used in the base case of CTE recursions should be made available from parent scope
+    # TODO(bojanserafimov): Some of these are no longer needed, since we don't select from
+    #                       the base cete, but only semijoin to its primary key now.
```
LWprogramming, line 136:
```suggestion
    #                       the base cte, but only semijoin to its primary key now.
```

obi1kenobi: Maybe capitalize `CTE` as well?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1014,10 +1016,23 @@ def recurse(self, vertex_field: str, depth: int) -> None:
...
-        ).cte(recursive=True)
+            [base_alias.c[col].label(col) for col in used_columns]
+            + [base_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
+        )
+        if self._recurse_needs_cte:
+            # Optimization: Only compute the recursion for the valid starting points -- ones that
+            # will not be discarded when the recursive cte is joined to the rest of the query.
+            # The SQL database could do this predicate pushdown itself, but most implementations
+            # don't.
+            # This optimization is absolutely necessary for queries with strong filters before
```
obi1kenobi, line 1029:
```suggestion
            # This optimization is absolutely necessary for queries with highly selective filters before
```
and this might force re-wrapping of lines to fit within 100 chars, sorry 😬

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1014,10 +1016,23 @@ def recurse(self, vertex_field: str, depth: int) -> None:
...
+            # the recursion. It improves performance by a factor of the filter selectivity,
+            # which can go up to 1 million times in common queries.
```
obi1kenobi, line 1031:
```suggestion
            # which can be 1 million times or more in common queries.
```

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1014,10 +1016,23 @@ def recurse(self, vertex_field: str, depth: int) -> None:
...
+            [base_alias.c[col].label(col) for col in used_columns]
+            + [base_alias.c[primary_key].label(CTE_KEY_NAME), literal_0.label(CTE_DEPTH_NAME),]
+        )
+        if self._recurse_needs_cte:
+            # Optimization: Only compute the recursion for the valid starting points -- ones that
+            # will not be discarded when the recursive cte is joined to the rest of the query.
+            # The SQL database could do this predicate pushdown itself, but most implementations
+            # don't.
+            # This optimization is absolutely necessary for queries with strong filters before
+            # the recursion. It improves performance by a factor of the filter selectivity,
```
obi1kenobi, line 1030:
```suggestion
            # the recursion. It improves performance by a factor of (1 / filter selectivity),
```
Selectivity is a fraction of 1, so presumably the improvement (speedup) is a multiple of inverse selectivity?

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -555,6 +555,112 @@ def test_recurse(self) -> None:
...
+
+        # TODO(bojanserafimov): Only testing in MSSQL because none of our backends agree on recurse
+        #                       semantics when multiple paths to the same output are inolved:
+        #                       - Our Match backend would represent each result once, even though it
+        #                         was reached multiple times by different paths.
+        #                       - Our SQL backend would duplicate the output row once for each path
+        #                       - Our Neo4j backend would find all different paths that use each
+        #                         edge at most once, and duplicate the result for each one.
+        for graphql_query, expected_results in queries:
+            self.assertResultsEqual(graphql_query, parameters, test_backend.MSSQL, expected_results)
```
obi1kenobi, line 662: Can we check this in Postgres as well? I expect our CTE setup should behave equivalently across SQL implementations, but N=2 will make me a lot more confident in that statement than N=1 😄

obi1kenobi:
A few minor nits, otherwise looks great. I'll release a new dev version as soon as this is merged in.

graphql_compiler/tests/integration_tests/test_backends_integration.py
bojanserafimov: Postgres is not configured for integration testing, and it's not trivial to do so (different db + schema hierarchy)

graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -555,6 +555,112 @@ def test_recurse(self) -> None:
         for graphql_query, expected_results in queries:
             self.assertResultsEqual(graphql_query, parameters, test_backend.MSSQL, expected_results)
 
+    @integration_fixtures
+    def test_recurse_duplication_regression(self) -> None:
```
obi1kenobi, line 559: Since this is quite a tricky test case, maybe add a docstring on the test function, or expand the comment on Query 4 to explain what's so special about it? Perhaps ideally also link to this PR, so we can easily link it back to the bug that we resolved and we are protecting against?

PR #888, opened 2020-08-05
Fix unintended type heterogeneity in validate_schema_and_query_ast()

obi1kenobi:
The return type of the `validate_schema_and_query_ast()` function was unintentionally `List[Union[GraphQLError, str]]`, but was treated by callers as `List[str]`. This PR fixes the type to be `List[str]` as seemingly originally intended.

After adding types, I used `typing-copilot` to tighten the `mypy.ini` file as follows:
```
tc init --overwrite
```
Then manually correct the `warn_unused_ignores` line since inline comments are not supported in `mypy.ini` and I haven't released a patched version of `typing-copilot` to correct the auto-generation yet.


graphql_compiler/compiler/validation.py
```diff
@@ -20,7 +23,7 @@ def validate_schema_and_query_ast(schema, query_ast):
     Returns:
         list containing schema and/or query validation errors
     """
-    core_graphql_errors = validate(schema, query_ast)
+    core_graphql_errors = [str(error) for error in validate(schema, query_ast)]
```
pmantica1, line 26: It makes more sense to me to be consistent with the GraphQL core next library and have the `validate` function return a list of `GraphQLError` objects. Thoughts?

mypy.ini
```diff
@@ -9,9 +8,10 @@ disallow_untyped_calls = True
 disallow_incomplete_defs = True
 disallow_untyped_defs = True
 disallow_untyped_decorators = True
-warn_unused_ignores = True
 ignore_missing_imports = False
 
+# mypy-copilot: disabled as mypy only reports it if other checks pass
```
pmantica1, line 13: I am confused as to why this warrants setting `ward_unused_ignores=False`. For me this change means that mypy doesn't raise an error if all other checks pass except this one. I don't understand why we want that. Do you mind making the comment above a bit clearer?

obi1kenobi: I'd argue we don't necessarily care that much whether the setting is on or off, because unused ignores aren't really errors per se. They are mypy saying "this code type checks just fine, there's no need for a `type: ignore` here." That means that this setting being on or off isn't particularly important, and we can get essentially all the benefit by just flipping it on every few months and cleaning up what it finds.

`False` is just what `typing-copilot` generates at the moment. I can flip it to `True`, we'll just have to keep flipping it back until `typing-copilot` is able to check for it not being necessary.

graphql_compiler/compiler/validation.py
obi1kenobi: This is a type-hinting PR, so I tried to keep functionality the same. I'd prefer to keep type hinting separate from functionality changes, otherwise the type hinting process becomes super tedious.

We have working, well-tested code in the codebase, even though not all parts are designed in the way we'd design them if we started from scratch today. But if we stop and fix everything that we'd like to redesign *while* we're type-hinting the codebase, we'll never finish type-hinting because there will always be something to fix and redesign.

Let's keep our eyes on the ball here: we're trying to change how people query their data systems. Whether this internal function returns strings, or returns error objects which immediately get stringified by the caller, is completely immaterial for that goal.

mypy.ini
pmantica1: Resolved offline: This comment is autogenerated by the typing copilot utility. I would prefer if the my.ini file was 100% autogenerated so I am ok with keeping this comment as is even though it could be written better.

PR #890, opened 2020-08-06
Use graphql.GraphQLError as base exception and remove mirror exception class. 

pmantica1:
It is bug prone to have a mirror `GraphQLError` class in the graphql compiler library.

### DIFF
--- a/graphql_compiler/query_formatting/common.py
+++ b/graphql_compiler/query_formatting/common.py
@@ -92,7 +92,7 @@ def validate_argument_type(name: str, expected_type: QueryArgumentGraphQLType, v
             try:
                 decimal.Decimal(value)
             except decimal.InvalidOperation as e:
-                raise GraphQLInvalidArgumentError(e)
+                raise GraphQLInvalidArgumentError(str(e))
     elif is_same_type(GraphQLDate, stripped_type):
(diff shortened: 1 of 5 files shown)


graphql_compiler/query_formatting/common.py
```diff
@@ -92,22 +92,22 @@ def validate_argument_type(name: str, expected_type: QueryArgumentGraphQLType, v
             try:
                 decimal.Decimal(value)
             except decimal.InvalidOperation as e:
-                raise GraphQLInvalidArgumentError(e)
+                raise GraphQLInvalidArgumentError(str(e))
```
pmantica1, line 95: The `GraphQLError` class takes in a `message` str parameter.

obi1kenobi:
I really don't think this is worth it. The added coupling is going to be more trouble than it's worth, and the effort we've gone to (you to put this PR together and refactor all the code, and me to review and write this) is already outweighing any benefit we might have drawn from this.

At most, we might want to rename `GraphQLError` into `GraphQLCompilerError` though even there I'd argue the opportunity cost of doing something else with that time massively outweighs any benefit we'd draw from it. We have much bigger fish to fry in this project, let's focus on those.

PR #891, opened 2020-08-06
Fix meta field insertion into existing schemas.

obi1kenobi:
In my rush to fix meta fields being broken a long time ago, I sloppily didn't add test coverage for the `insert_meta_fields_into_existing_schema()` function. As a result, our recent GraphQL core library version bump broke this function silently, without any test catching it.

I discovered this thanks to mypy: I added type hints, and mypy immediately pointed out the use of non-existent attributes on the underlying type.

To fix:
- Updated the GraphQLSchema object's use to refer to the new attributes, in line with the most recent GraphQL core library version.
- Add a unit test to cover the buggy function.
- Add type hints to the entire source and test files where the problematic function was located and tested, as penance.
- Tighten the mypy.ini file using `typing-copilot` so that type hints are required in this file (and other checks able to be tightened for other files are indeed tightened) to prevent future regressions.

### DIFF
--- a/graphql_compiler/schema/__init__.py
+++ b/graphql_compiler/schema/__init__.py
@@ -4,7 +4,7 @@ from datetime import date, datetime
 from decimal import Decimal
 from hashlib import sha256
 from itertools import chain
-from typing import Any, FrozenSet
+from typing import Any, FrozenSet, Iterable
 
 import arrow
 from graphql import (
@@ -267,7 +267,7 @@ INBOUND_EDGE_FIELD_PREFIX = "in_"
 VERTEX_FIELD_PREFIXES = frozenset({OUTBOUND_EDGE_FIELD_PREFIX, INBOUND_EDGE_FIELD_PREFIX})
 
 
-def is_vertex_field_name(field_name):
+def is_vertex_field_name(field_name: str) -> bool:
     """Return True if the field's name indicates it is a non-root vertex field."""
     # N.B.: A vertex field is a field whose type is a vertex type. This is what edges are.
     return field_name.startswith(OUTBOUND_EDGE_FIELD_PREFIX) or field_name.startswith(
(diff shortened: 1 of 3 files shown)


mypy.ini
```diff
@@ -76,9 +74,6 @@ check_untyped_defs = False
 disallow_untyped_calls = False
 disallow_untyped_defs = False
 
-[mypy-graphql_compiler.compiler.helpers.*]
-disallow_untyped_calls = False
```
obi1kenobi, line 80: This is perhaps my favorite part about `typing-copilot` — it catches and tightens settings I never would have thought about doing by hand. In this case, it noticed that `graphql_compiler.compiler.helpers` was calling a previously-untyped function that is now typed.

Since this suppression is no longer necessary, it is removed. That module is guaranteed to now and forever more only call fully-typed code.

graphql_compiler/schema/__init__.py
```diff
@@ -296,7 +296,7 @@ def _serialize_date(value):
     return value.isoformat()
 
 
-def _parse_date_value(value):
+def _parse_date_value(value: str) -> date:
```
pmantica1, line 299: nit: I would use `Any` instead of `str` for two reasons:
1. To be consistent with the `_parse_datetime_value` function and all of the GraphQL core next parsing functions. (These GraphQL core type annotation are also probably in turn derived from the GraphQL js library). 
2.  The input here is likely going to be a JSON value and the type of this value might be an union of the possible JSON types. In that case, mypy might raise an error unless we use a `cast`.

pmantica1: The third and most important reason is that I would like to avoid future discussions over the parsing function type annotations and seems like the best way to do that is to be consistent from the start.

obi1kenobi: Yeah, you're right — the proper type should be `Any`.
https://github.com/graphql-python/graphql-core/blob/master/src/graphql/type/scalars.py#L100
https://github.com/graphql-python/graphql-core/blob/master/src/graphql/type/scalars.py#L64

PR #893, opened 2020-08-07
Add type hints to schema/schema_info.py

obi1kenobi:


graphql_compiler/schema/schema_info.py
```diff
@@ -248,11 +248,16 @@ def _create_sql_schema_info(
...
 
 def make_sqlalchemy_schema_info(
-    schema, type_equivalence_hints, dialect, vertex_name_to_table, join_descriptors, validate=True
-):
+    schema: GraphQLSchema,
+    type_equivalence_hints: TypeEquivalenceHintsType,
+    dialect: Dialect,
+    vertex_name_to_table: Dict[str, sqlalchemy.Table],
+    join_descriptors: Dict[str, Dict[str, DirectJoinDescriptor]],
+    validate: bool = True,
```
chewselene, line 256: The `Args` section specifies that this is `Optional`. 

Also some of the `Args` descriptions are just types so it might be nice to put descriptions there.

obi1kenobi: > The `Args` section specifies that this is `Optional`.

I blame Python's unfortunate naming collision between "optional" and "nullable" :) The argument is optional in the sense that it has a default value (`True`) and is therefore allowed to be not provided, and is not nullable `Optional[bool]`. I can lowercase `Optional` -> `optional`, but I'm not sure what else we can do to make this more clear.

> Also some of the `Args` descriptions are just types so it might be nice to put descriptions there.

Normally I'd agree, but this is an interesting and somewhat unusual situation, and I'm not sure. This function is essentially a "double-check args, and then instantiate `SQLAlchemySchemaInfo`" kind of function. Since all the args except `validate` go directly to `SQLAlchemySchemaInfo`, the docstring says `See the documentation of SQLAlchemySchemaInfo for more detailed documentation of the args.` since that's what ultimately governs their use.

We can copy-paste the descriptions of the arguments from `SQLAlchemySchemaInfo` here, but if we ever change the argument meanings or descriptions, I'm not convinced we'd remember to propagate the changes here as well. We can leave them blank, forcing people to check `SQLAlchemySchemaInfo` for their detailed descriptions. Or we can try to add a simplified, more generic description here, and risk that people might miss the fact that `SQLAlchemySchemaInfo` has more info on the arguments.

Modulo tiny things like `Optional -> optional`, I'm really not sure how to meaningfully improve the situation here. Any ideas?

obi1kenobi: Here's my attempt, let me know what you think: #894

chewselene: lgtm, thanks for updating!

Issue #897, opened 2020-08-11
Improve test coverage of query pagination code for queries with type coercions


obi1kenobi:
While adding types to the query pagination code, I realized that a good chunk of the code appears to implicitly assume that queries do not contain type coercions: for example, there was a lot of use of `ast.name.value`, which is only defined for `FieldNode` and doesn't work for `InlineFragmentNode` AST types.

This situation emerged because our SQL backend doesn't currently support type coercions (because SQL in general doesn't do table inheritance), and because we prototyped the query pagination code on SQL.

I am doing my best to fix the code as best I can, but I'm not sure I can get everything. We need more test coverage to prove that query pagination works for all queries, including ones with type coercions.

bojanserafimov:
Worth checking why `test_query_pagination.py:test_with_compiler_tests` doesn't catch those bugs. It runs pagination against all defined compiler test inputs, checking that no errors are raised.

bojanserafimov:
Ah, the current tests pass because we always paginate at the root right now.

PR #898, opened 2020-08-11
Improve type checking in query pagination and AST manipulation code.

obi1kenobi:
Includes type-level fixes for handling `InlineFragmentNode` AST values in query pagination code, a couple of new helper functions, and type hints for the AST manipulation module.

Based on the tests passing and my careful refactoring, I am fairly confident that I haven't made pagination any worse. However, I opened #897 to verify that query pagination in the presence of type coercions (signaled by `InlineFragmentNode` AST values) actually works.

This is quite a tricky PR, so I'd like to have @bojanserafimov review it before we merge, in addition to anyone else that might want to check it out.

### DIFF
--- a/graphql_compiler/ast_manipulation.py
+++ b/graphql_compiler/ast_manipulation.py
@@ -1,41 +1,53 @@
 # Copyright 2019-present Kensho Technologies, LLC.
+from typing import Any, Optional, Type, Union
+
 from graphql.error import GraphQLSyntaxError
 from graphql.language.ast import (
     DocumentNode,
+    FieldNode,
     InlineFragmentNode,
     ListTypeNode,
+    NamedTypeNode,
+    Node,
     NonNullTypeNode,
     OperationDefinitionNode,
     OperationType,
+    SelectionNode,
+    TypeNode,
 )
 from graphql.language.parser import parse
+from graphql.pyutils import FrozenList
 
 from .exceptions import GraphQLParsingError
 
 
-def get_ast_field_name(ast):
+def get_ast_field_name(ast: FieldNode) -> str:
     """Return the field name for the given AST node."""
     return ast.name.value
 
 
-def get_ast_field_name_or_none(ast):
+def get_ast_field_name_or_none(ast: Union[FieldNode, InlineFragmentNode]) -> Optional[str]:
     """Return the field name for the AST node, or None if the AST is an InlineFragment."""
     if isinstance(ast, InlineFragmentNode):
         return None
     return get_ast_field_name(ast)
 
 
-def get_human_friendly_ast_field_name(ast):
+def get_human_friendly_ast_field_name(ast: Node) -> str:
     """Return a human-friendly name for the AST node, suitable for error messages."""
     if isinstance(ast, InlineFragmentNode):
         return "type coercion to {}".format(ast.type_condition)
     elif isinstance(ast, OperationDefinitionNode):
         return "{} operation definition".format(ast.operation)
+    elif isinstance(ast, FieldNode):
+        return get_ast_field_name(ast)
+    else:
+        # Fall back to Node's __repr__() method.
+        # If we need more information for a specific type, we can add another branch in the if-elif.
+        return repr(ast)
 
-    return get_ast_field_name(ast)
 
-
-def safe_parse_graphql(graphql_string):
+def safe_parse_graphql(graphql_string: str) -> DocumentNode:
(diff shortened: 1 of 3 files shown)


graphql_compiler/ast_manipulation.py
```diff
@@ -45,7 +57,9 @@ def safe_parse_graphql(graphql_string):
     return ast
 
 
-def get_only_query_definition(document_ast, desired_error_type):
+def get_only_query_definition(
+    document_ast: DocumentNode, desired_error_type: Type[Exception]
+) -> OperationDefinitionNode:
     """Assert that the Document AST contains only a single definition for a query, and return it."""
     if not isinstance(document_ast, DocumentNode) or not document_ast.definitions:
```
LWprogramming, line 64: Now that there's a type hint, do we still need to perform the `isinstance` check?

obi1kenobi: Unfortunately, yes — until all uses of this function are typed to something stricter than `Any`, the `isinstance` check is still meaningful and can affect code execution.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -241,12 +249,16 @@ def _add_pagination_filter_recursively(
...
-    if not isinstance(node_ast, (FieldNode, InlineFragmentNode, OperationDefinitionNode)):
-        raise AssertionError(
-            f'Input AST is of type "{type(node_ast).__name__}", which should not be a selection.'
-        )
-
     if len(query_path) == 0:
+        if isinstance(node_ast, OperationDefinitionNode):
+            # N.B.: We cannot use assert_selection_is_a_field_or_inline_fragment_node() here, since
+            #       that function's return type annotation is a Union type that isn't compatible
+            #       with the generic type signature of _add_pagination_filter_at_node().
```
bojanserafimov, line 256: Interesting how `assert_selection_is_a_field_or_inline_fragment_node` works with unions: "We used to know x is of type A, but we ran some assertions and now we're not sure if it's A or B"

Can't think of a cleaner solution.

Btw, we use `N.B.` and `NOTE` in the codebase. No need to standardize, but letting you know that `NOTE` gets highlighted by many text editors, just like `TODO` is.

graphql_compiler/ast_manipulation.py
```diff
@@ -96,22 +119,68 @@ def get_only_selection_from_ast(ast, desired_error_type):
...
+
+    if not isinstance(ast, NamedTypeNode):
+        raise AssertionError(
+            f"Expected the AST value to be a NamedTypeNode, but unexpectedly instead found: {ast}"
+        )
+
+    return ast
+
+
+def assert_selection_is_a_field_node(ast: Any) -> FieldNode:
```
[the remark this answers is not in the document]
obi1kenobi: I like the idea — perhaps `checked_cast` as an alternative name, to make the answer to "how is this safer" a bit more explicit?

graphql_compiler/query_pagination/query_parameterizer.py
obi1kenobi: If `checked_cast()` works out, we might not need this note anymore. 🤞

graphql_compiler/ast_manipulation.py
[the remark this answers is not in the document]
obi1kenobi: `checked_cast()` doesn't work with `Union` because `Union` is not a `Type`, so `Type[T]` doesn't match.

For bounded-length tuples, this is achievable:
```python
def checked_cast_to_union2(target_types: Tuple[Type[T_A], Type[T_B]], value: Any) -> Union[T_A, T_B]:
    if not isinstance(value, target_types):
        raise AssertionError()

    return value
```

I don't know of a way to make `target_types: Tuple[Type, ...]` work and connect the types it contains to the types inside the `Union` result.

Given that we're somewhat unlikely to have unions of more than like 3 or so things, I'm fine with adding `checked_cast_to_union2()` and in the future potentially adding a `checked_cast_to_union3()` function that takes a 3-tuple as necessary.

What do you think?

graphql_compiler/query_pagination/query_parameterizer.py
obi1kenobi: I wrote a `checked_cast_to_union2()` function but it turns out that even it can't do the trick with generics. I updated the note with my new observations, let me know if you want to talk about it.

obi1kenobi:
@bojanserafimov I'd love a re-review here when you get the chance.

PR #899, opened 2020-08-11
Transfer custom scalar code from PR #884

LWprogramming:
Originally the code treated non-scalar renaming to a name that a scalar already uses as a special case. This case was treated differently from, for instance, 2 object types renamed to the same thing.

For example, renaming Foo -> USER_DEFINED_SCALAR when there's already a scalar named USER_DEFINED_SCALAR gets treated differently from renaming Foo and Bar both to Baz.

However, since it's (eventually) going to be fine to rename/ suppress scalars, it makes more sense to not draw that distinction but rather draw the line between "this renaming would work if not for this other type or renaming resulting in conflict" vs. "this renaming tries to rename Dog to String which is never allowed because String is a built-in scalar type." This PR reflects the change.

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -545,20 +545,29 @@ class RenameSchemaTypesVisitor(Visitor):
             return REMOVE
         check_type_name_is_valid(new_name_string)
 
+        # Either new_name_string is the name of a type (that may or may not have been
+        # renamed) in reverse_name_map or new_name_string is the name of a custom scalar.
+        # Either way, we want the original name of the type that is would be named
+        # new_name_string in the renamed schema.
         if (
             new_name_string in self.reverse_name_map
             and self.reverse_name_map[new_name_string] != name_string
         ) or new_name_string in self.custom_scalar_names:
-            # Either new_name_string is the name of a type (that may or may not have been
-            # renamed) in reverse_name_map or new_name_string is the name of a custom scalar.
-            # Either way, we want the original name of the type that is would be named
-            # new_name_string in the renamed schema.
-            other_name_string_renamed_to_new_name_string = self.reverse_name_map.get(
+            # This type of conflict arises when two types with different names in the original
+            # schema have the same name in the new schema.
+            # If neither type is a custom scalar, the two type names in the original schema are
+            # name_string and self.reverse_name_map[new_name_string].
+            # If one of the types is a custom scalar, it won't appear in self.reverse_name_map
+            # because custom scalar renaming has not been implemented yet. Since this renaming has
+            # not been implemented yet, we know for sure that the scalar's name is new_name_string.
+            # So, in either case, the two types in the original schema will be named name_string and
+            # conflictingly_renamed_type_name.
+            conflictingly_renamed_type_name = self.reverse_name_map.get(
                 new_name_string, new_name_string
             )
             raise SchemaNameConflictError(
                 '"{}" and "{}" are both renamed to "{}"'.format(
-                    name_string, other_name_string_renamed_to_new_name_string, new_name_string
+                    name_string, conflictingly_renamed_type_name, new_name_string
                 )
             )
         if new_name_string in builtin_scalar_type_names:
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -84,7 +84,6 @@ from graphql import (
     ObjectTypeDefinitionNode,
     UnionTypeDefinitionNode,
     build_ast_schema,
-    specified_scalar_types,
 )
 from graphql.language.visitor import IDLE, REMOVE, Visitor, VisitorAction, visit
 import six
@@ -94,11 +93,12 @@ from .utils import (
     CascadingSuppressionError,
     SchemaNameConflictError,
     SchemaTransformError,
+    builtin_scalar_types,
     check_ast_schema_is_valid,
     check_type_name_is_valid,
     get_copy_of_node_with_new_name,
+    get_custom_scalar_names,
     get_query_type_name,
-    get_scalar_names,
 )
 
 
@@ -187,13 +187,13 @@ def rename_schema(
 
     schema = build_ast_schema(schema_ast)
     query_type = get_query_type_name(schema)
-    scalars = get_scalar_names(schema)
+    custom_scalar_types = get_custom_scalar_names(schema)
 
-    _validate_renamings(schema_ast, renamings, query_type, scalars)
+    _validate_renamings(schema_ast, renamings, query_type, custom_scalar_types)
 
     # Rename types, interfaces, enums, unions and suppress types, unions
     schema_ast, reverse_name_map = _rename_and_suppress_types(
-        schema_ast, renamings, query_type, scalars
+        schema_ast, renamings, query_type, custom_scalar_types
     )
     reverse_name_map_changed_names_only = {
         renamed_name: original_name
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -542,15 +548,22 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
         if (
             new_name_string in self.reverse_name_map
             and self.reverse_name_map[new_name_string] != name_string
-        ):
+        ) or new_name_string in self.custom_scalar_types:
+            # Either new_name_string is the name of a type (that may or may not have been
+            # renamed) in reverse_name_map or new_name_string is the name of a custom scalar.
+            # Either way, we want the original name of the type that is would be named
+            # new_name_string in the renamed schema.
+            other_name_string_renamed_to_new_name_string = self.reverse_name_map.get(
```
LWprogramming, line 556: not a big fan of this variable name

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -542,15 +548,22 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
...
+                new_name_string, new_name_string
+            )
             raise SchemaNameConflictError(
                 '"{}" and "{}" are both renamed to "{}"'.format(
-                    name_string, self.reverse_name_map[new_name_string], new_name_string
+                    name_string, other_name_string_renamed_to_new_name_string, new_name_string
                 )
             )
-        if new_name_string in self.scalar_types or new_name_string in self.builtin_types:
+        if new_name_string in self.custom_scalar_types or new_name_string in builtin_scalar_types:
```
chewselene, line 564: seems that the `new_name_string in self.custom_scalar_types` part of the if statement repeated in both this block and the one above? keeping it down here might solve your awkward variable name you mentioned in the comment above as well.

LWprogramming: Oops! The custom scalar check should only be in the if statement above, since that one checks for renaming conflicts for renamable/ suppressable types, will fix.

That doesn't quite solve the awkward variable name, though-- see follow up comment there.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: The idea behind this error is when the original schema has two different types `Foo` and `Bar` that end up having the same name after calling `rename_schema`. In this case, `name_string` gets renamed to `new_name_string` when there's some other type (maybe a custom scalar type, maybe some other type) that we know will have the name `new_name_string` in the resulting schema, resulting in conflict. `other_name_string_renamed_to_new_name_string` is supposed to be the name of this other type.

obi1kenobi: It seems like the `new_name_string` is the desired name for the type? If so, let's name it something like `desired_type_name`?

In general, I'm not a huge fan of the `_string` suffix or reverse Hungarian notation as a whole: https://en.wikipedia.org/wiki/Hungarian_notation

Hungarian notation is sometimes useful! For example: For example: https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/

These days, Hungarian notation's benefits can be replicated by `NewType()` in Python and similar tricks in other languages, and pure "this is a string" Hungarian (as opposed to "this int is a height, and not a width") is in general a code smell.

So I'd advise against using the `_string` suffix and instead focusing on the purpose of each variable in the algorithm you are constructing, then picking an appropriate name. Sometimes, `new` is good enough, or the best we can come up with. But frequently, we can do better.

In this case, `other_name_string_renamed_to_new_name_string` is simply `conflictingly_renamed_type_name`.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -542,15 +548,22 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
         if (
             new_name_string in self.reverse_name_map
             and self.reverse_name_map[new_name_string] != name_string
-        ):
+        ) or new_name_string in self.custom_scalar_types:
+            # Either new_name_string is the name of a type (that may or may not have been
+            # renamed) in reverse_name_map or new_name_string is the name of a custom scalar.
+            # Either way, we want the original name of the type that is would be named
+            # new_name_string in the renamed schema.
```
obi1kenobi, line 555: Can you add one line here about why `.get()` is the correct call on the next line? I'm sure working through the complex logic in the conditional, one can convince themselves that is the case, but it's nice to optimize the code for readability and help the reader out with such logical leaps.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -542,15 +548,22 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
         if (
             new_name_string in self.reverse_name_map
             and self.reverse_name_map[new_name_string] != name_string
```
obi1kenobi, line 550: Technically in the scope of the PR, since you're modifying this conditional — this can be more succinctly and more clearly stated as `self.reverse_name_map.get(new_name_string, None) != name_string`. Since an easy alternative like this is available, not taking it tricks the reader into thinking something subtle and unusual is going on — I had to double-check, for example, that the same `self.reverse_name_map` dict was used in both clauses.

I'd much prefer the `.get()` approach here as a result.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -148,7 +152,7 @@ def get_scalar_names(schema):
     scalars = {
```
obi1kenobi, line 152: Consider renaming this variable name, `scalars` is only telling part of the story now.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: This is actually a bit trickier than it looks on the surface, since we're checking if there already is a mapping for `new_name_string` that doesn't correspond to our current `name_string`, i.e. there's a conflict. The code that makes this work is `self.reverse_name_map.get(new_name_string, name_string) != name_string`, i.e. the default is `name_string` instead of `None`. I don't have a strong preference since I've seen this code a whole bunch of times, but would you say one seems more understandable than the other?

obi1kenobi: I'd definitely prefer the code one-liner + some explanation in English, over the current two-clauses-on-two-lines approach. I think it'd be much more readable.

obi1kenobi: Especially because it's trickier than it looks on the surface :D

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -528,38 +530,73 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
...
             # Suppress the type
             return REMOVE
-        check_type_name_is_valid(new_name_string)
-
+        check_type_name_is_valid(desired_type_name)
+
+        # Renaming conflict arises when two types with different names in the original schema have
+        # the same name in the new schema. There are two ways to produce this conflict:
+        # * when neither type is a custom scalar
+        # * when one type is a custom scalar and the other is not
```
chewselene, line 551: nit: i would tab and/or replace the `*` with numbers. visually the `# *` just looks kind of odd to me.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -528,38 +530,73 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
...
+        # If one type in the conflict is a custom scalar, then the conflict arises from
+        # attempting to rename a non-scalar type (from type_name to desired_type_name) when
+        # there already exists a custom scalar type named desired_type_name. Custom scalar
+        # renaming has not been implemented yet so we know for sure that the scalar's name (in
+        # both the original and new schema) is desired_type_name.
+        #
+        # It's also possible to produce a similar conflict where one type is not a custom scalar but
+        # rather a built-in scalar, e.g. String. That case is handled separately because renaming a
+        # type to a built-in scalar will never be allowed in any schema, whereas the conflicts here
+        # arise from conflicting with the existence of other types in the schema.
```
chewselene, line 568: This giant comment is super helpful for understanding the code! Thanks for adding 😄

PR #901, opened 2020-08-12
Deduplicate conversions between query string / AST with parameters.

obi1kenobi:
We were converting between `QueryStringWithParameters` and `ASTWithParameters` separately in several places, with identical code. In this PR, I add classmethods on both of those dataclasses that should be the only way to convert between the two, and use them to replace all the prior independent implementations.

I intend to do more refactoring in these areas, just keeping my PRs small and easy to review.


graphql_compiler/global_utils.py
```diff
@@ -17,13 +20,27 @@ class PropertyPath(NamedTuple):
...
 class QueryStringWithParameters:
     """A query string and parameters that validate against the query."""
 
     query_string: str
     parameters: Dict[str, Any]
 
+    @classmethod
+    def from_ast_with_parameters(
+        cls: Type[QueryStringWithParametersT], ast_with_params: "ASTWithParameters"
+    ) -> QueryStringWithParametersT:
```
pmantica1, line 39: Can you not use a string here instead, i.e. "QueryStringWithParameters"? The main reason I prefer a string here is for future consistency; I would argue that other developers are likely to use a string here since this what is suggested by the top StackOverlow post when you search for "return type of a method is the same as class python".

https://stackoverflow.com/questions/33533148/how-do-i-specify-that-the-return-type-of-a-method-is-the-same-as-the-class-itsel

graphql_compiler/global_utils.py
```diff
@@ -32,6 +49,14 @@ class ASTWithParameters:
     query_ast: DocumentNode
     parameters: Dict[str, Any]
 
+    @classmethod
+    def from_query_string_with_parameters(
```
pmantica1, line 53: How about we make a `to_query_string_with_parameters` instance method in this class and a `to_ast_with_parameters` method in the `QueryStringWithParameters`? That way the conversion would be more concise and we are less likely to go over the 100 character limit when typing it out.

graphql_compiler/cost_estimation/analysis.py
```diff
@@ -486,8 +485,21 @@ def pagination_capacities(self) -> Dict[PropertyPath, int]:
 
 
 def analyze_query_string(
-    schema_info: QueryPlanningSchemaInfo, query: QueryStringWithParameters
+    schema_info: QueryPlanningSchemaInfo, query_and_params: QueryStringWithParameters
```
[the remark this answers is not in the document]
obi1kenobi: Done.

graphql_compiler/global_utils.py
obi1kenobi: The two options on the table are the following:
```python
# option 1 -- the current code
QueryStringWithParametersT = TypeVar(
    "QueryStringWithParametersT", bound="QueryStringWithParameters"
)
@classmethod
def from_ast_with_parameters(
    cls: Type[QueryStringWithParametersT], ast_with_params: "ASTWithParameters"
) -> QueryStringWithParametersT:
    pass

# option 2 -- your suggestion
@classmethod
def from_ast_with_parameters(
    cls, ast_with_params: "ASTWithParameters"
) -> "QueryStringWithParameters":
    pass
```

Despite the similarity, they say very different things. The former says:
- "Given a type, which is a subtype of `QueryStringWithParameters`, return an instance of that type".

The latter says:
- "Given a type, return an instance of `QueryStringWithParameters`. The input type is not (and likely, cannot safely be) used anywhere in the type definition or implementation.

I think the former is preferable, because it continues to be safe in the presence of subclasses. Also, see below.

graphql_compiler/global_utils.py
obi1kenobi: With a `to_` instance method, both the "source" and the "destination" types are pinned down at definition time (source type by the `self` type, destination by the return type). This means the caller at the call site has no freedom to pick the destination type other than by looking for an appropriate `to_` method. If such a method doesn't exist, the caller is just out of luck.

With a `from_` classmethod, the call site controls the target type, because the call site does `Foo.from_x()` and `Foo` becomes the `cls` argument. This not only gives us more flexibility and convenience, but also means that we get more functionality out of fewer from-to function pairs.

PR #905, opened 2020-08-14
Replace "cached-property" with the built-in and the backport.

### LINKED ISSUE #904: Remove cached-property dependency
I think we should remove our dependency on `cached-property`, for a few reasons:
- We use a very minimal piece of functionality we can easily replicate and improve upon ourselves.
- It isn't type-hinted, and the open issue for it is over a year old with no activity: https://github.com/pydanny/cached-property/issues/172
- The lack of type hints means that we have to always suppress `mypy`'s `disallow_untyped_decorators` rule. It also means that `@cached_property` properties return type `Any`, which makes `mypy` even less useful.
- `@cached_property` doesn't inherit from `@property`, causing a number of other type issues. Here's the tracking issue for it, which has also been inactive in many years: https://github.com/pydanny/cached-property/issues/26

obi1kenobi:
The `cached_property` decorator is available as of Python 3.8, but up to this point we were using a 3rd party library for all such functionality, to avoid having different code in different Python versions. Unfortunately, this 3rd party library has its own issues described in #904, so I think it's time to replace it.

Migrating to the built-in allowed mypy to catch a type error that was previously hidden by missing type hints, so I also fix that type error and clean up the surrounding documentation and errors.

Two considerations I'd recommend we apply going forward:
- Prefer to rely on built-in functionality and explicit backports, rather than 3rd party libraries with the same functionality.
- Strongly consider whether we need a particular dependency before adding it, especially if it lacks type hints, or its functionality is trivially small and simple, and can be easily recreated locally in strongly-typed fashion and in a few lines of code.

Resolves #904.

### DIFF
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -9,6 +9,13 @@ import six
 from .ast_manipulation import safe_parse_graphql
 
 
+# Imported from here, to avoid spreading the conditional import everywhere.
+try:
+    from functools import cached_property  # noqa
+except ImportError:
+    from backports.cached_property import cached_property  # noqa
+
+
 # A path starting with a vertex and continuing with edges from that vertex
 VertexPath = Tuple[str, ...]
 
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -13,7 +13,7 @@ from .ast_manipulation import safe_parse_graphql
 try:
     from functools import cached_property  # noqa
 except ImportError:
-    from backports.cached_property import cached_property  # noqa
+    from backports.cached_property import cached_property  # noqa  # type: ignore[no-redef]
 
 
 # A path starting with a vertex and continuing with edges from that vertex
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -13,7 +13,7 @@ from .ast_manipulation import safe_parse_graphql
 try:
     from functools import cached_property  # noqa
 except ImportError:
-    from backports.cached_property import cached_property  # noqa  # type: ignore[no-redef]
+    from backports.cached_property import cached_property  # type: ignore[no-redef] # noqa
 
 
 # A path starting with a vertex and continuing with edges from that vertex
--- a/graphql_compiler/global_utils.py
+++ b/graphql_compiler/global_utils.py
@@ -11,9 +11,9 @@ from .ast_manipulation import safe_parse_graphql
 
 # Imported from here, to avoid spreading the conditional import everywhere.
 try:
-    from functools import cached_property  # noqa
+    from functools import cached_property  # noqa  # pylint: disable=unused-import
 except ImportError:
-    from backports.cached_property import cached_property  # type: ignore[no-redef] # noqa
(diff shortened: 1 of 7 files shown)


graphql_compiler/query_pagination/__init__.py
```diff
@@ -10,22 +10,22 @@
...
-        int, estimated number of pages if the query were executed.
-
-    Raises:
-        ValueError if page_size is below 1.
+        int, estimated number of pages of the desired size into which the supplied query
+        should be split up
     """
     if page_size < 1:
-        raise ValueError(
+        raise AssertionError(
```
obi1kenobi, line 28: This is a trivial invariant for the caller to enforce. `ValueError` made it the callee's problem, `AssertionError` makes it the caller's problem, which I believe is better since the caller is better positioned to produce a meaningful error message if this were to ever legitimately occur.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -64,12 +64,9 @@ def paginate_query_ast(
               - page_and_remainder.page_size == page_size
             - Tuple of PaginationAdvisory objects that communicate what can be done to improve
               pagination
-
-    Raises:
-        ValueError if page_size is below 1.
     """
     if page_size < 1:
-        raise ValueError(
+        raise AssertionError(
```
obi1kenobi, line 69: Same reasoning as above.

mypy.ini
```diff
@@ -399,9 +398,6 @@ disallow_untyped_calls = False
 [mypy-arrow.*]
 ignore_missing_imports = True
 
-[mypy-cached_property.*]
-ignore_missing_imports = True
-
```
obi1kenobi, line 404: Thanks to `typing-copilot` here. I remembered to remove the `untyped_decorators` check above, but I completely forgot this one.

setup.py
```diff
@@ -57,15 +57,17 @@ def find_long_description() -> str:
     install_requires=[  # Make sure to keep in sync with Pipfile requirements.
         "arrow>=0.15.0,<1",
         "funcy>=1.7.3,<2",
-        "graphql-core>=3.1,<3.2",
+        "graphql-core>=3.1.2,<3.2",
```
obi1kenobi, line 60: Just getting this in line with our `Pipfile` — since this is a patch release, I don't think it's a big deal.

graphql_compiler/global_utils.py
```diff
@@ -9,6 +9,13 @@
 from .ast_manipulation import safe_parse_graphql
 
 
+# Imported from here, to avoid spreading the conditional import everywhere.
+try:
+    from functools import cached_property  # noqa
+except ImportError:
+    from backports.cached_property import cached_property  # type: ignore[no-redef] # noqa
```
obi1kenobi, line 16: See https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker if you want to learn more about `# type: ignore[no-redef]` — this is something we should get used to using everywhere since it disables only a particular type of mypy check.

pmantica1:
Looks good to me. Thanks for the Github comments. They were pretty useful in helping me understand your code. Just fix the Pipfile.lock file and I'll approve.

pmantica1:
The bot is putting up a fight...

PR #909, opened 2020-08-18
Remove unnecessary value list in null-related filters

LWprogramming:
`value: []` used to be required but then we relaxed the @filter definition to allow it to not be present.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1263,7 +1263,7 @@ Example Use
     {
         Animal {
             name @output(out_name: "animal_name")
-            color @filter(op_name: "is_null", value: [])
+            color @filter(op_name: "is_null")
         }
     }
 
@@ -1274,7 +1274,7 @@ Constraints and Rules
 ^^^^^^^^^^^^^^^^^^^^^
 
 -  Must be applied on a property field.
--  :code:`value` must be empty.
+-  :code:`value` must not appear in the filter (shown here) or be an empty list.
 
 is\_not\_null
 ~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


LWprogramming:
Does any change need to happen with read-the-docs documentation as well?

README.rst
```diff
@@ -1274,7 +1274,7 @@ Constraints and Rules
 ^^^^^^^^^^^^^^^^^^^^^
 
 -  Must be applied on a property field.
--  :code:`value` must be empty.
+-  :code:`value` must not appear in the filter (shown here) or be an empty list.
```
obi1kenobi, line 1277:
```suggestion
-  :code:`value` must either not appear in the filter (shown in the example) or be an empty list.
```
Similar change to the entry below.

JulianGoetz: Should we remove `value` from the examples as well? Seems superfluous.

obi1kenobi: It seems like it is removed — see changes above an below in the `Example use` sections. Are there other examples you were thinking of?

obi1kenobi:
> Does any change need to happen with read-the-docs documentation as well?

I think this content also exists in `docs/source/language_specification/query_directives.rst`, so probably yes.

LWprogramming:
The [CI](https://travis-ci.org/github/kensho-technologies/graphql-compiler/builds/719310700) passed but the webhook is broken, so feel free to override!

LWprogramming:
d'oh-- fixed!

PR #910, opened 2020-08-18
Update pylintrc with new checks. Fix code as necessary.

obi1kenobi:
I enabled a few more Pylint checks that:
- seemed useful
- caught real problems, or suggest that lint they catch is indicative of a real problem
- do not cause excessive CPU time to be consumed (looking at you, `useless-return` check)

I also did my best to fix the errors that Pylint finds in this new configuration, though a few of them are in docstrings in code that I'm not familiar with — I'd appreciate suggestions on how to make them even better.


graphql_compiler/schema_generation/schema_graph.py
```diff
@@ -364,10 +364,16 @@ def __repr__(self):
...
 class VertexType(SchemaElement):
+
+    # The parent class' __init__ method is marked abstract, and must be overridden.
+    # See the parent class for more details.
+    # pylint: disable=useless-super-delegation
     def __init__(self, class_name, abstract, properties, class_fields):
         """See base class __init__ method."""
         super(VertexType, self).__init__(class_name, abstract, properties, class_fields)
 
+    # pylint: enable=useless-super-delegation
```
pmantica1, line 375: I think that these can actually be removed. Only the methods marked `@abstractmethod` actually have to be overriden.

obi1kenobi: I thought the same thing at first. See lines 368-369 above.
```
    # The parent class' __init__ method is marked abstract, and must be overridden.
    # See the parent class for more details.
```

PR #912, opened 2020-08-21
Collect InvalidTypeNameErrors instead of reporting one at a time

LWprogramming:


graphql_compiler/schema_transformation/utils.py
```diff
@@ -200,25 +200,20 @@ def check_schema_identifier_is_valid(identifier: str) -> None:
...
-    if not isinstance(name, str):
-        raise InvalidTypeNameError('Name "{}" is not a string.'.format(name))
-    if not re_name.match(name):
-        raise InvalidTypeNameError('"{}" is not a valid GraphQL name.'.format(name))
-    if name.startswith("__"):
-        raise InvalidTypeNameError(
-            '"{}" starts with two underscores, which is reserved for '
-            "GraphQL internal use and is not allowed.".format(name)
-        )
+    return bool(re_name.match(name)) and not name.startswith("__")
```
obi1kenobi, line 216: It seems like this would make our error messages less good — we capture the correct outcome, but don't capture the _reason_ behind it. The "two leading underscores" rule is particularly non-obvious: how many users have actually read the GraphQL specification before writing queries? I'd bet the number is not high.

LWprogramming: That's fair, although both error messages [here](https://github.com/kensho-technologies/graphql-compiler/pull/912/files#diff-3b859f46e97b0af834db0b3b7de08658R381) and [here](https://github.com/kensho-technologies/graphql-compiler/pull/912/files#diff-78fe4ecd19bb243ccd07437313dd81b4R404) do mention it. I considered splitting this into multiple functions to make it more explicit what's being checked, but then anyone who wants to validate a GraphQL type name would have to make sure to call all the functions in order to validate properly.

obi1kenobi (approved):

PR #913, opened 2020-08-25
Make some analysis fields private

bojanserafimov:
Exploring ways to add `QueryPlanningAnalysis` to the public api. Not intending to merge this PR, just making inline comments to start discussion.

We can't just make `QueryPlanningAnalysis` public as is, because then any function signature change would need a compiler major version bump if we were to respect semver. Fields like `pagination_capactiy` are not stable yet, so that's a risky move.

So I've made most fields private, only leaving public the fields that are already public in some way, or we intend to make public for sure. This created uses of private fields outside of this class. I address each problem separately with inline comments.

I argue that the `QueryPlanningAnalysis` class is still effective, even after half its fields are marked private. The main goal is to avoid recomputation of analysis that many of our passes require, and that's still possible. The class is even more effective now that we can avoid recomputation even among public library api calls.

It's worth noting that now plumbing of analysis becomes a bit more tedious: A function either:
- Uses only public analysis passes, or
- Is one of the analysis passes

@obi1kenobi @MichaelaShtilmanMinkin thoughts?

### DIFF
--- a/graphql_compiler/query_pagination/__init__.py
+++ b/graphql_compiler/query_pagination/__init__.py
@@ -79,10 +79,10 @@ def paginate_query_ast(
 
     # See if we can and should split the query
     num_pages = 1
-    if query_analysis.classes_with_missing_counts:
+    if query_analysis._classes_with_missing_counts:
         advisories += tuple(
             MissingClassCount(class_name)
-            for class_name in query_analysis.classes_with_missing_counts
+            for class_name in query_analysis._classes_with_missing_counts
         )
     else:
         result_size = query_analysis.cardinality_estimate
--- a/graphql_compiler/query_pagination/pagination_planning.py
+++ b/graphql_compiler/query_pagination/pagination_planning.py
@@ -136,7 +136,7 @@ def get_pagination_plan(
     # Get the pagination capacity
     vertex_path = (root_node,)
     property_path = PropertyPath(vertex_path, pagination_field)
-    capacity = query_analysis.pagination_capacities.get(property_path)
+    capacity = query_analysis._pagination_capacities.get(property_path)
     # If the pagination capacity is None, then there must be no quantiles for this property.
     if capacity is None:
         ideal_min_num_quantiles_per_page = 5
--- a/graphql_compiler/query_pagination/query_parameterizer.py
+++ b/graphql_compiler/query_pagination/query_parameterizer.py
@@ -86,7 +86,7 @@ def _is_new_filter_stronger(
     Returns:
         whether the old filter can be removed with no change in query meaning.
     """
-    vertex_type = query_analysis.types[property_path.vertex_path].name
+    vertex_type = query_analysis._types[property_path.vertex_path].name
     new_int_value = convert_field_value_to_int(
         query_analysis.schema_info, vertex_type, property_path.field_name, new_filter_value,
     )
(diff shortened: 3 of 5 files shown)


graphql_compiler/query_pagination/pagination_planning.py
```diff
@@ -136,7 +136,7 @@ def get_pagination_plan(
     # Get the pagination capacity
     vertex_path = (root_node,)
     property_path = PropertyPath(vertex_path, pagination_field)
-    capacity = query_analysis.pagination_capacities.get(property_path)
+    capacity = query_analysis._pagination_capacities.get(property_path)
```
bojanserafimov, line 139: This usage of private field outside of the class can be removed by making pagination_planning one of the analysis passes, just as we did with cost estimation.

graphql_compiler/query_pagination/__init__.py
```diff
@@ -79,10 +79,10 @@ def paginate_query_ast(
 
     # See if we can and should split the query
     num_pages = 1
-    if query_analysis.classes_with_missing_counts:
+    if query_analysis._classes_with_missing_counts:
         advisories += tuple(
             MissingClassCount(class_name)
-            for class_name in query_analysis.classes_with_missing_counts
+            for class_name in query_analysis._classes_with_missing_counts
```
bojanserafimov, line 85: This code block can be a helper method called `get_desired_number_of_pages(desired_page_size: int, ...)`, and that method can be part of the `QueryPlanningAnalysis` class, and can freely access the private field `_classes_with_missing_counts`.

obi1kenobi:
This doesn't seem right to me. It's strange that `analyze_query_string()` returns an object capable of performing many analyses, but some are prohibited from being read from the outside.

Why not define a superclass of `QueryPlanningAnalysis` that contains just the parts you intend to make part of the public API, and make public-facing returns be of that supertype while still returning `QueryPlanningAnalysis` objects? Then `mypy` will enforce which methods are used from the outside.

graphql_compiler/query_pagination/query_parameterizer.py
```diff
@@ -86,7 +86,7 @@ def _is_new_filter_stronger(
     Returns:
         whether the old filter can be removed with no change in query meaning.
     """
-    vertex_type = query_analysis.types[property_path.vertex_path].name
+    vertex_type = query_analysis._types[property_path.vertex_path].name
```
bojanserafimov, line 89: We could make `types` public.

Reminder how it's different than the type information in the `metadata_table`: It doesn't care about revisits.

bojanserafimov:
> This doesn't seem right to me. It's strange that `analyze_query_string()` returns an object capable of performing many analyses, but some are prohibited from being read from the outside.
> 
> Why not define a superclass of `QueryPlanningAnalysis` that contains just the parts you intend to make part of the public API, and make public-facing returns be of that supertype while still returning `QueryPlanningAnalysis` objects? Then `mypy` will enforce which methods are used from the outside.

Let's call that supertype `PublicAnalysis` for now. If `analyze_query_string` is annotated to return `PublicAnalysis`, but actually returns `QueryPlanningAnalysis`, mypy will accept that. But mypy will not allow you to pass this into `get_pagination_plan`, which requires a `QueryPlanningAnalysis`.

PR #914, opened 2020-08-25
Type hint and convert namedtuple to dataclass in merge_schemas.py

chewselene:
The PR adds type hints and converts namedtuples to dataclasses. Also does a bit of documentation cleanup.

### DIFF
--- a/graphql_compiler/schema_transformation/merge_schemas.py
+++ b/graphql_compiler/schema_transformation/merge_schemas.py
@@ -1,11 +1,33 @@
 # Copyright 2019-present Kensho Technologies, LLC.
-from collections import namedtuple
+from collections import OrderedDict  # pylint: disable=unused-import
 from copy import deepcopy
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Set, Tuple, Union
 
 from graphql import build_ast_schema
-from graphql.language import ast as ast_types
+from graphql.language.ast import (
+    ArgumentNode,
+    DefinitionNode,
+    DirectiveDefinitionNode,
+    DirectiveNode,
+    DocumentNode,
+    EnumTypeDefinitionNode,
+    FieldDefinitionNode,
+    InterfaceTypeDefinitionNode,
+    ListTypeNode,
+    NamedTypeNode,
+    NameNode,
+    ObjectTypeDefinitionNode,
+    OperationType,
+    OperationTypeDefinitionNode,
+    ScalarTypeDefinitionNode,
+    SchemaDefinitionNode,
+    StringValueNode,
+    UnionTypeDefinitionNode,
+)
 from graphql.language.printer import print_ast
 from graphql.pyutils import FrozenList
+from graphql.type import GraphQLInterfaceType, GraphQLObjectType, GraphQLSchema, GraphQLUnionType
 import six
 
 from ..ast_manipulation import get_ast_with_non_null_stripped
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,40 +42,58 @@
...
-    "MergedSchemaDescriptor",
-    (
-        "schema_ast",  # Document, AST representing the merged schema
-        "schema",  # GraphQLSchema, representing the same schema as schema_ast
-        "type_name_to_schema_id",
-        # Dict[str, str], mapping type name to the id of its schema, includes Interface, Object,
-        # Union, and Enum types
-    ),
-)
+@dataclass
```
obi1kenobi, line 45: Consider adding `frozen=True` for each of these, since `frozen=False` is the default for dataclasses whereas the original `namedtuple` is immutable.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,40 +42,58 @@
...
+    # Describing the inbound field of the edge.
+    inbound_field_reference: FieldReference
 
-def merge_schemas(schema_id_to_ast, cross_schema_edges, type_equivalence_hints=None):
+    # Whether the edge is bidirectional.
+    out_edge_only: bool
+
+
+def merge_schemas(
+    schema_id_to_ast: "OrderedDict[str, DocumentNode]",
```
obi1kenobi, line 91: Interesting — is `OrderedDict` not generic by default? Might be good to leave a short comment here, just to save the reader of the code some time. I also wasn't able to find a clear and definitive answer in the docs.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -20,40 +42,58 @@
...
+    cross_schema_edges: List[CrossSchemaEdgeDescriptor],
+    type_equivalence_hints: Optional[
+        Dict[Union[GraphQLInterfaceType, GraphQLObjectType], GraphQLUnionType]
```
obi1kenobi, line 94: `from ..schema.typedefs import TypeEquivalenceHintsType` might be useful here, it's just shorthand for the complex `Dict[...]` type signature.

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -764,22 +818,25 @@ def _scalars_match(scalar_name1, scalar_name2):
...
 
 def _add_edge_field(
-    source_type_node, sink_type_name, source_field_name, sink_field_name, edge_name, direction
-):
+    source_type_node: Union[InterfaceTypeDefinitionNode, ObjectTypeDefinitionNode],
+    sink_type_name: str,
+    source_field_name: str,
+    sink_field_name: str,
+    edge_name: str,
+    direction: str,
```
pmantica1, line 826: Probably a good idea to use `typing.LITERAL['out', 'in']` here for the direction.

graphql_compiler/schema_transformation/merge_schemas.py
chewselene: I got this pylint error `E1136: Value 'OrderedDict' is unsubscriptable (unsubscriptable-object)`. I'll add a note about it here.

LWprogramming: A note that this is only a problem in Python 3.6 according to [this](https://github.com/python/mypy/issues/6904#issuecomment-496207426)

graphql_compiler/schema_transformation/merge_schemas.py
```diff
@@ -373,33 +418,38 @@ def _process_scalar_definition(scalar, existing_scalars, type_name_to_schema_id,
...
-        generic_type: Any of EnumTypeDefinitionNode, InterfaceTypeDefinitionNode,
-                      ObjectTypeDefinitionNode, or UnionTypeDefinitionNode, an AST node
-                      representing the definition of a type.
-        schema_id: str, the identifier of the schema that this type came from.
-        existing_scalars: Set[str], set of names of all existing scalars.
-        type_name_to_schema_id: Dict[str, str], mapping names of types to the identifier of the
-                                schema that they came from.
-        merged_schema_ast: DocumentNode, AST representing a schema.
+        generic_type: AST node representing the definition of a type.
+        schema_id: the identifier of the schema that this type came from.
```
LWprogramming, line 443:
```suggestion
        schema_id: identifier of the schema that this type came from.
```
Nit: removing article

LWprogramming: Also shows up in a few places elsewhere, I think ctrl + f ": the" covers all of them

graphql_compiler/schema_transformation/merge_schemas.py
obi1kenobi: This would be a nice to have if nothing prevents it from existing, but I'm happy to wait for a future PR as well.

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
```diff
@@ -554,7 +554,13 @@ def test_invalid_identifiers(self):
             )
         with self.assertRaises(ValueError):
             merge_schemas(
-                OrderedDict([(42, parse(ISS.basic_schema)), ("enum", parse(ISS.enum_schema)),]), [],
+                OrderedDict(
+                    [
+                        (42, parse(ISS.basic_schema)),  # type: ignore
```
obi1kenobi, line 559: Why the `type: ignore`? Can we resolve with a cast instead?

graphql_compiler/schema_transformation/merge_schemas.py
chewselene: @pmantica1 discussed this a bit offline. Agreed, that it would be nice, but it is a bit trickier than adding the Literal type hint. We pass `OUT/INBOUND_EDGE_DIRECTION` to the function, which are constants with type `str`. This causes a mypy error. Possibly would be good to make `OUT/INBOUND_EDGE_DIRECTION` Literals as well, but seems outside the scope of this PR.

graphql_compiler/tests/schema_transformation_tests/test_merge_schemas.py
chewselene: Casting `42` to `str`? This test ensures that you can't pass an `int` as the identifier. Since the type hint specifies that the identifier must be a `str`, we have to ignore the mypy error in order to make sure that this raises a `ValueError`. Let me know if I'm misunderstanding you!

obi1kenobi: Oooh I see. No worries, this is fine.

graphql_compiler/schema_transformation/merge_schemas.py
obi1kenobi: Fair enough!

PR #915, opened 2020-08-27
Implement query on schema adapter

lonerz:


graphql_compiler/tests/test_schema_adapter.py
```diff
@@ -0,0 +1,124 @@
...
+                "property_name": "uuid",
+                "property_description": None,
+                "is_property_deprecated": False,
+                "property_type": "ID",
+            },
+        ]
+
+        _ensure_query_produces_expected_output(self, query, args, expected_results)
+
+    # def test_vertex_type_with_no_properties(self) -> None:
```
lonerz, line 108: Currently, there are no types in the test schema that don't have properties, but can be added easily -- just ran out of time

obi1kenobi (approved):

PR #916, opened 2020-09-01
Type hint schema_transformation/utils.py

LWprogramming:
After this PR gets merged, I'd be happy to try out typing-copilot before integrating the work in #912 with this, since #912 is currently a draft because some of the code there gets away with stuff it shouldn't (due to a lack of type hints).

### DIFF
--- a/graphql_compiler/schema_transformation/utils.py
+++ b/graphql_compiler/schema_transformation/utils.py
@@ -6,7 +6,8 @@ from typing import Dict, Set, FrozenSet, Optional, List, Union, Any, Type, TypeV
 from graphql import build_ast_schema, specified_scalar_types, GraphQLSchema
 from graphql.language.ast import FieldNode, InlineFragmentNode, NameNode, Node, \
     ObjectTypeDefinitionNode, FieldDefinitionNode, DocumentNode, DirectiveDefinitionNode, \
-    DirectiveNode, SelectionSetNode, FragmentSpreadNode
+    DirectiveNode, SelectionSetNode, FragmentSpreadNode, EnumTypeDefinitionNode, \
+    InterfaceTypeDefinitionNode, NamedTypeNode, UnionTypeDefinitionNode, ScalarTypeDefinitionNode
 from graphql.language.visitor import Visitor, visit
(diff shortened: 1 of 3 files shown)


obi1kenobi:
> After this PR gets merged, I'd be happy to try out typing-copilot before integrating the work in #912 with this, since #912 is currently a draft because some of the code there gets away with stuff it shouldn't (due to a lack of type hints).

You should try out `typing-copilot` with this PR :) Given that you've added type hints to an entire file, odds are that the `mypy.ini` file can be tightened relative to its current state. We want to keep the `mypy.ini` as tight as possible because that's the only thing that prevents type regressions — right now, if someone else posts a PR after yours that adds new untyped code to the same file you just type-hinted, nothing will stop that from being merged.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -119,15 +135,49 @@ class CascadingSuppressionError(SchemaTransformError):
...
+# For the same reason as with RenameTypes, these types have to be written out explicitly instead of
+# relying on allowed_types in get_copy_of_node_with_new_name.
+# Unlike RenameTypes, RenameNodes also includes fields because it's used in the function
+# get_copy_of_node_with_new_name which rename_query depends on to rename the root field in a query.
+# Meanwhile, RenameTypes applies only for rename_schema and field renaming in the schema is not
+# implemented yet.
+RenameNodes = Union[
+    RenameTypes, FieldNode, FieldDefinitionNode,
+]
+RenameNodesT = TypeVar("RenameNodesT", bound="RenameNodes")
```
obi1kenobi, line 177: Any reason to use `"RenameNodes"` here as a forward reference even though `RenameNodes` is defined on the previous line?

obi1kenobi, line 177:
```suggestion
RenameNodesT = TypeVar("RenameNodesT", bound=RenameNodes)
```

obi1kenobi: See below

LWprogramming: Ah, gotcha-- will fix

PR #917, opened 2020-09-09
Add comment explaining renamings only apply to types existing in original schema

LWprogramming:


obi1kenobi (commented):
Should it be an error to have attempted renamings that don't actually rename anything? I haven't had a chance to think this through all the way, so it's totally possible this is intentional and necessary. I'm just generally wary of and not a fan of silently ignoring user input.

LWprogramming:
> Should it be an error to have attempted renamings that don't actually rename anything? I haven't had a chance to think this through all the way, so it's totally possible this is intentional and necessary. I'm just generally wary of and not a fan of silently ignoring user input.

I considered raising errors for this and for no-op renames like `{"Foo": "Foo"}`, but since `renamings` is a `Mapping` and not a `Dict` there isn't a way to check for these. I also considered adding in these checks for just `Dict` in particular, but that wouldn't work because there could be other `Mapping`s that aren't `Dict`s where these no-op renamings exist.

obi1kenobi:
> > Should it be an error to have attempted renamings that don't actually rename anything? I haven't had a chance to think this through all the way, so it's totally possible this is intentional and necessary. I'm just generally wary of and not a fan of silently ignoring user input.
> 
> I considered raising errors for this and for no-op renames like `{"Foo": "Foo"}`, but since `renamings` is a `Mapping` and not a `Dict` there isn't a way to check for these. I also considered adding in these checks for just `Dict` in particular, but that wouldn't work because there could be other `Mapping`s that aren't `Dict`s where these no-op renamings exist.

Good point. Is it possible/reasonable to check if `renamings` is iterable, and apply the checks then? Again, I haven't thought this through all the way, and it's possible it's not a good idea.

LWprogramming:
> > > Should it be an error to have attempted renamings that don't actually rename anything? I haven't had a chance to think this through all the way, so it's totally possible this is intentional and necessary. I'm just generally wary of and not a fan of silently ignoring user input.
> > 
> > 
> > I considered raising errors for this and for no-op renames like `{"Foo": "Foo"}`, but since `renamings` is a `Mapping` and not a `Dict` there isn't a way to check for these. I also considered adding in these checks for just `Dict` in particular, but that wouldn't work because there could be other `Mapping`s that aren't `Dict`s where these no-op renamings exist.
> 
> Good point. Is it possible/reasonable to check if `renamings` is iterable, and apply the checks then? Again, I haven't thought this through all the way, and it's possible it's not a good idea.

That could work and it wouldn't take too long to take to implement, but this could result in counterintuitive situations, e.g. `renamings` is an object whose `get` method looks like

```
def get(name):
    if name == "Foo":
        return "Bar"
    return name
```

which is valid but if the schema doesn't contain a type named `Foo` then it'll do nothing silently, while a dict doing the same thing would raise errors.

obi1kenobi:
> That could work and it wouldn't take too long to take to implement, but this could result in counterintuitive situations, e.g. `renamings` is an object whose `get` method looks like
> 
> ```
> def get(name):
>     if name == "Foo":
>         return "Bar"
>     return name
> ```
> 
> which is valid but if the schema doesn't contain a type named `Foo` then it'll do nothing silently, while a dict doing the same thing would raise errors. I lean slightly against but if it turns out that most renamings use iterable mappings, then I'd be fine with checking for this.

To me, unused renamings feel kind of like lint. We don't want it to accumulate if we can prevent it, but we shouldn't check arbitrary getters for lint because that is simply infeasible. The common case would be the user passing a dict or similar iterable mapping type, and as a result should offer the best protections against users unintentionally not following good practices. I'd consider the callable-style mappings to be an advanced feature where "the training wheels come off" and at which point the user takes on the responsibility for their correctness and reasonableness.

To fully drive home the point, we can't prevent the user from making a getter like this, either:
```python
# DO NOT EVER DO THIS
def get(name):
     if name == "Foo":
         return "Bar_" + str(uuid4()).replace("-", "")
     return name
```
We simply can't catch and prevent arbitrary code doing inadvisable things. I don't think we should use that as an excuse to not provide a good user experience in the cases where we **can** catch and prevent issues early.

LWprogramming:
> > That could work and it wouldn't take too long to take to implement, but this could result in counterintuitive situations, e.g. `renamings` is an object whose `get` method looks like
> > ```
> > def get(name):
> >     if name == "Foo":
> >         return "Bar"
> >     return name
> > ```
> > 
> > 
> > which is valid but if the schema doesn't contain a type named `Foo` then it'll do nothing silently, while a dict doing the same thing would raise errors. I lean slightly against but if it turns out that most renamings use iterable mappings, then I'd be fine with checking for this.
> 
> To me, unused renamings feel kind of like lint. We don't want it to accumulate if we can prevent it, but we shouldn't check arbitrary getters for lint because that is simply infeasible. The common case would be the user passing a dict or similar iterable mapping type, and as a result should offer the best protections against users unintentionally not following good practices. I'd consider the callable-style mappings to be an advanced feature where "the training wheels come off" and at which point the user takes on the responsibility for their correctness and reasonableness.
> 
> To fully drive home the point, we can't prevent the user from making a getter like this, either:
> 
> ```python
> # DO NOT EVER DO THIS
> def get(name):
>      if name == "Foo":
>          return "Bar_" + str(uuid4()).replace("-", "")
>      return name
> ```
> 
> We simply can't catch and prevent arbitrary code doing inadvisable things. I don't think we should use that as an excuse to not provide a good user experience in the cases where we **can** catch and prevent issues early.

Makes sense to me. I notice that the [cyclic renaming test case](https://github.com/kensho-technologies/graphql-compiler/blob/main/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py#L269) actually covers this edge case neatly. If we had an incorrect implementation that renames all types named `Human` to `Droid`, and only then renames all types named `Droid` to `Dog` (etc), it would fail on that test.

Given this, in this PR I'll just modify the docstring to mention the edge case and check for the unused renamings in a separate PR (which turns out to be a little trickier than I expected at first but on the bright side, my draft of the solution has already automatically caught about 3 or 4 test cases in the codebase that had the no-op renamings) :)

obi1kenobi (approved):

obi1kenobi:
Can you rename the PR before merging it?

PR #920, opened 2020-09-14
Add test for rename_schema only in type namespace

LWprogramming:


obi1kenobi (approved):
Doesn't Lilo get a type? Just Stitch? 😄

PR #922, opened 2020-09-16
Prevent noop renames when renamings is iterable

LWprogramming:
Covers both cases

1. when renamings maps a name to itself

2. when renamings contains an entry not in the schema


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -77,8 +77,21 @@
...
+  words, a string type_name may be in renamings only if there exists a type in the schema named
+  type_name (since otherwise that entry would not affect any type in the schema). Also, if type_name
+  is in renamings, then renamings[type_name] != type_name (since applying the renaming would not
+  change the type named type_name).
+  Note that these rules apply only if renamings is iterable. If renamings is a non-iterable
+  object that just implements a get() method, there's no way to enforce these rules since there's
+  no way to keep arbitrary code from doing valid but inadvisable things. These rules exist
+  specifically for iterables because that's the most common usage for rename_schema.
 """
+# Note that importing from collections.abc instead of collections is due to a pylint issue that goes
```
LWprogramming, line 91: Note: black auto-formatted this comment to put it above all the imports and not just the relevant import

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -186,6 +261,34 @@ def test_original_unmodified_suppress(self):
...
+        self.assertTrue(check_no_op_renaming_error_message({"Dinosaur"}, set(), e.exception))
+
+    def test_rename_legal_noop_unused_renaming(self):
+        # Unlike with test_rename_illegal_noop_unused_renaming, here the renaming is not
+        # iterable. As a result, this renaming is inadvisable but it is technically legal to
+        # write a renaming like this since the intended "Dinosaur" -> "NewDinosaur" mapping is
+        # unused and will silently do nothing when applied to the given schema.
+        class RenameMapping(object):
+            def __init__(self):
+                pass
```
obi1kenobi, line 276: You don't need this function.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -186,6 +261,34 @@ def test_original_unmodified_suppress(self):
...
+
+            def get(self, key, default=None):
```
obi1kenobi, line 278: Need type hints here, and probably a docstring. I wonder why the linter isn't complaining? Could you look into that?

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -186,6 +261,34 @@ def test_original_unmodified_suppress(self):
...
+        with self.assertRaises(NoOpRenamingError) as e:
+            rename_schema(parse(ISS.basic_schema), {"Dinosaur": "NewDinosaur"})
+        self.assertTrue(check_no_op_renaming_error_message({"Dinosaur"}, set(), e.exception))
+
+    def test_rename_legal_noop_unused_renaming(self):
+        # Unlike with test_rename_illegal_noop_unused_renaming, here the renaming is not
+        # iterable. As a result, this renaming is inadvisable but it is technically legal to
+        # write a renaming like this since the intended "Dinosaur" -> "NewDinosaur" mapping is
+        # unused and will silently do nothing when applied to the given schema.
+        class RenameMapping(object):
```
obi1kenobi, line 274:
```suggestion
        class RenameMapping:
```
Since we're not in Python 2 land anymore...

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -233,6 +336,38 @@ def test_multiple_type_suppress(self):
...
+        # write a renaming like this since the intended "Dinosaur" -> None mapping is unused and
+        # will silently do nothing when applied to the given schema.
+        class SuppressMapping(object):
+            def __init__(self):
+                pass
+
+            def get(self, key, default=None):
+                if key == "Dinosaur":
+                    return None
+                return key
```
obi1kenobi, line 356: Same comments as above.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -21,6 +22,38 @@
...
+
+    Args:
+        error_message_component: error message part to validate
+        expected_prefix: explanation of why the error was raised
+        expected_data_structure: description of which parts of the schema caused the error
+
+    Returns:
+        True iff the error message correctly consists of the prefix followed by the expected data
+        structure.
+    """
```
obi1kenobi, line 45: To be honest, this feels like way overkill and also like testing code by writing more (and more complex) code. Testing error messages by relying on a correct implementation of this function, including uses of `literal_eval` etc., seems problematic.

If non-determinism is the source of the problem, let's tackle it directly — output the items in sorted order in the error message. That will allow you to write out the actual error message strings in the tests, which would make it much easier for reviewers and readers of the code to review the content of the error message and evaluate its usefulness.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -608,7 +651,15 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
             self.renamed_to_builtin_scalar_conflicts[type_name] = desired_type_name
 
         self.reverse_name_map[desired_type_name] = type_name
+        # renamings is performing a 1:1 renaming so if renamings is iterable, record that this entry
+        # in renamings is used to ensure all renamings are used.
+        if isinstance(self.renamings, Iterable) and type_name in self.renamings:
```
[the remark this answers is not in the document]
LWprogramming: `self.renamed_types` is meant to contain the entries from the `renamings` mapping that are actually used (i.e. if `renamings` maps `A` to `B` _and_ `A` is in the schema, then `A` is to appear in `self.renamed_types`) and only meant to be used when `renamings` is iterable, to check that all the renamings were used. 

Upon re-reading this code, I found an easier way to do this (since we already have `self.reverse_name_map`, which isn't sufficient on its own because it doesn't contain suppressed types, but has all the renames) that also helps with the error message problems

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
LWprogramming: Turns out the mypy.ini has `disallow_untyped_defs = False` for the tests and uncommenting that uncovered roughly 1 complaint per test. For now, I'll just type-hint this and separately open a PR type-hinting the rest

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -21,6 +22,38 @@
...
+    try:
+        if literal_eval(error_message_component[len(expected_prefix) :]) != expected_data_structure:
+            return False
+    except SyntaxError:
+        # In case it's syntactically invalid
+        return False
+    return True
+
+
 def check_rename_conflict_error_message(
```
LWprogramming, line 57: Having revised the data structures to check for no-op renames, I removed the big method checking that string for the `NoOpRenamingError`. Unfortunately, since name conflicts have a more complicated structure (i.e. it's not just "renamings contains these keys which do nothing"), and since it's easier to generate many name conflicts accidentally (as opposed to manually creating a renamings dict with many unused entries), I'm not sure about just sorting everything in the error message. Thoughts?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -408,6 +417,26 @@ def _rename_and_suppress_types(
...
+        # If renamings is iterable, then every renaming must be used and no renaming can map a
+        # name to itself
+        for type_name in visitor.suppressed_types:
+            if type_name not in renamings:
+                raise AssertionError(
+                    f"suppressed_types should be a subset of the set of keys in renamings, but "
+                    f"found {type_name} in suppressed_types that is not a key in renamings. This "
+                    f"is a bug."
+                )
+        renamed_types = set(
```
obi1kenobi, line 430: nit: Use a set comprehension instead of `set()` with a generator inside.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -408,6 +417,26 @@ def _rename_and_suppress_types(
...
+            visitor.reverse_name_map[type_name]
+            for type_name in visitor.reverse_name_map
+            if type_name != visitor.reverse_name_map[type_name]
+        )
+        no_op_renames: Set[str] = (
+            set(key for key in renamings) - renamed_types - set(visitor.suppressed_types)
```
obi1kenobi, line 436: The first clause here is actually `set(renamings)`.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
obi1kenobi: I'm not sure what the current structure is that is more complicated. I agree that there could be many name conflicts — the canonical solution to that is to check the list length and if too large, only output a few elements (~7ish or so) from the head of the sorted list, and a few (maybe ~3) from the tail. This keeps the error message small. Then, you can add an instance attribute to the error type that contains the full list of errors, allowing callers to see and recover the full list even though it's not in the error message.

LWprogramming: Ah, gotcha-- to clarify, this is the order of events that happened:

* PR #884 gets merged (for `SchemaRenameNameConflictError`, which has a complicated error message because the data structures are involved are complicated).

* This PR comes out (for `NoOpRenamingError`, which also had a complicated error message that's constructed in a similar way as `SchemaRenameNameConflictError`).

* Since the two error types have similarly constructed error messages, I decide to abstract out some of it into this `validate_error_message_component` function.

* I find a way to simplify the data structure for `NoOpRenamingError` so only `SchemaRenameNameConflictError` tests use `validate_error_message_component` function now.

In the process of writing this out I rubber-duck-debugged myself into realizing that this means any code involving `validate_error_message_component` no longer has to do with this PR's purpose since `NoOpRenamingError` tests don't rely on complicated string checks anymore, so I'll revert those changes and split into a separate PR as appropriate.

obi1kenobi (approved):

PR #924, opened 2020-09-18
Annotate functions in tests

LWprogramming:
Type-hinting PR described in #922 

Since `Mapping` is actually always iterable, even though what we really need for `renamings` is simply an object that implements the `get` method described in the code, I modified the code so that it permits non-iterable renamings to work.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -24,3 +24,6 @@ exclude_lines =
 
     # Don't complain if non-runnable code isn't run:
     if __name__ == .__main__.:
+
+    # Don't complain if ellipsis never gets executed
+    grep -r "^[ ]*\.\.\.$"
--- a/.coveragerc
+++ b/.coveragerc
@@ -26,4 +26,4 @@ exclude_lines =
     if __name__ == .__main__.:
 
     # Don't complain if ellipsis never gets executed
-    grep -r "^[ ]*\.\.\.$"
+    ^[ ]*\.\.\.$
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -79,7 +79,8 @@ Renaming constraints:
   being renamed or suppressed.
 """
 from collections import namedtuple
-from typing import AbstractSet, Any, Dict, List, Optional, Protocol, Set, Tuple, Union, cast
+import sys
+from typing import AbstractSet, Any, Dict, List, Optional, Set, Tuple, Union, cast
 
 from graphql import (
     DocumentNode,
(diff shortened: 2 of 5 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -79,7 +79,7 @@
   being renamed or suppressed.
 """
 from collections import namedtuple
-from typing import AbstractSet, Any, Dict, List, Mapping, Optional, Set, Tuple, Union, cast
+from typing import AbstractSet, Any, Dict, List, Optional, Protocol, Set, Tuple, Union, cast
```
obi1kenobi, line 82: `Protocol` is new in 3.8: https://docs.python.org/3/library/typing.html#typing.Protocol

This is why the 3.6 and 3.7 builds are failing. Fortunately, `typing_extensions` has a backport available, so just add to the conditional import in the top-level `typedefs.py` file that I tweaked to add `Literal` earlier.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -111,6 +112,16 @@
 )
 
 
+# We prefer the explicit sys.version_info check instead of the more common try-except ImportError
+# approach, because at the moment mypy seems to have an easier time with the sys.version_info check:
+# https://github.com/python/mypy/issues/1393
+if sys.version_info[:2] >= (3, 8):
```
obi1kenobi, line 118: Add this to the top-level `typedefs.py` file instead, so we don't have to do repeated conditional imports wherever we require `Protocol`.

.coveragerc
```diff
@@ -24,3 +24,6 @@ exclude_lines =
 
     # Don't complain if non-runnable code isn't run:
     if __name__ == .__main__.:
+
+    # Don't complain if ellipsis never gets executed
+    grep -r "^[ ]*\.\.\.$"
```
obi1kenobi, line 29: Confused about the `grep -r` part, is this just a bad paste?

LWprogramming: Oops! Yes, will fix that along with the typedef thing

graphql_compiler/typedefs.py
```diff
@@ -16,9 +16,13 @@
 # Hence, the "unused import" warnings here are false-positives.
 if sys.version_info[:2] >= (3, 8):
     # TypedDict and Literal were only added to typing in Python 3.8
-    from typing import Literal, TypedDict  # noqa  # pylint: disable=unused-import
+    from typing import Literal, TypedDict, Protocol  # noqa  # pylint: disable=unused-import
```
obi1kenobi, line 19: Maybe update line 8 to not reference specific imports anymore, since it's not just `Literal` and `TypedDict` anymore?

obi1kenobi: See line 8.

LWprogramming: Ah, didn't see the 1 before the 8

obi1kenobi (approved):

PR #925, opened 2020-09-21
Type hint exported compiler frontend functions

obi1kenobi:
Add minimal type hints for `compiler_frontend.py` and fix broken type hints elsewhere. Then, constrict `mypy.ini` to the strictest configuration that passes.

### DIFF
--- a/graphql_compiler/compiler/compiler_frontend.py
+++ b/graphql_compiler/compiler/compiler_frontend.py
@@ -1102,10 +1102,9 @@ def ast_to_ir(
     """Convert the given GraphQL AST object into compiler IR, using the given schema object.
 
     Args:
-        schema: GraphQL schema object, created using the GraphQL library
-        ast: AST object to be compiled to IR
-        type_equivalence_hints: optional dict of GraphQL interface or type -> GraphQL union.
-                                Used as a workaround for GraphQL's lack of support for
+        schema: schema object created using the GraphQL library for which the query must be valid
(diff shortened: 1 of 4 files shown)


chewselene (dismissed):
looks good!

graphql_compiler/compiler/compiler_frontend.py
```diff
@@ -1091,7 +1094,11 @@ def _compile_output_step(query_metadata_table):
...
 
-def ast_to_ir(schema, ast, type_equivalence_hints=None):
+def ast_to_ir(
+    schema: GraphQLSchema,
+    ast: DocumentNode,
+    type_equivalence_hints: Optional[TypeEquivalenceHintsType] = None,
+) -> IrAndMetadata:
     """Convert the given GraphQL AST object into compiler IR, using the given schema object.
 
     Args:
```
LWprogramming, line 1104: We can drop the types in the docstring since we have type hints now (same for `graphql_to_ir` below)

LWprogramming:
Looks like the failing lint is just due to updating `black` earlier, besides that and dropping the types from the function headers it looks good to me

graphql_compiler/compiler/compiler_frontend.py
obi1kenobi: Thanks, good call! Will fix shortly.

LWprogramming (approved):

PR #926, opened 2020-09-22
Add Literal to typing import shim, change to mypy-friendly approach.

obi1kenobi:
Without the `sys.version_info` check, this used to require a `type: ignore` due to the following problem:
https://twitter.com/PredragGruevski/status/1308085206285529089?s=20

Mypy tracking issue for the import error approach is here: https://github.com/python/mypy/issues/1393

For our needs, `sys.version_info` is an acceptable solution for Python import shims.

### DIFF
--- a/graphql_compiler/typedefs.py
+++ b/graphql_compiler/typedefs.py
@@ -6,7 +6,7 @@ from graphql import GraphQLList, GraphQLNonNull, GraphQLScalarType
 
 
 # The below code is an import shim for TypedDict and Literal: we don't want to conditionally import
-# then from every file that needs them. Instead, we conditionally import them here and then import
+# them from every file that needs them. Instead, we conditionally import them here and then import
 # from this file in every other location where they are needed.
 #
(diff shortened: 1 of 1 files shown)


graphql_compiler/typedefs.py
```diff
@@ -1,18 +1,24 @@
...
 from typing import Union
 
 from graphql import GraphQLList, GraphQLNonNull, GraphQLScalarType
 
 
-# The below code is an import shim for TypedDict: we don't want to conditionally import it from
-# every file that needs it. Instead, we conditionally import it here and then import from this file
-# in every other location where this is needed.
+# The below code is an import shim for TypedDict and Literal: we don't want to conditionally import
+# then from every file that needs them. Instead, we conditionally import them here and then import
```
LWprogramming, line 9:
```suggestion
# them from every file that needs them. Instead, we conditionally import them here and then import
```

obi1kenobi: Thanks!

LWprogramming (approved):

PR #931, opened 2020-09-24
Add functionality overview to interpreter module docstring.

obi1kenobi:
Intended to be a basic intro to the purpose and structure of the interpreter module. We could perhaps reuse some or all of this in the ReadTheDocs description of the interpreter — thoughts?


bojanserafimov:
Looks good to me. You forgot the copyright header.

Once we're cofortable with having this in public docs the description of the compiler can change from "library that simplifies database querying and exploration" to "a schema indirection layer for arbitrary data sources" or something similar.

obi1kenobi:
I suppose technically the diff coverage _is_ 0% because nothing ever imports that file with no executable code in it. But also, as a vacuous truth, technically all executable code in the module I added _is_ executed in tests, so maybe the diff coverage is 100% instead 🤔

graphql_compiler/interpreter/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+proceed by compiling them to another query language -- no such target query language exists.
+Instead, the queries need to be executed using an *interpreter*: a sophisticated piece of code
+that executes queries incrementally in a series of steps, such as "fetch the value of this field"
+or "filter out this data point if its value is less than 5."
+
+Some parts of the interpreter (e.g. "fetch the value of this field") obviously need to be aware of
+the schema and the underlying data source. Other parts (e.g. "filter out this data point") are
+schema-agnostic -- they work in the same way regardless of the schema and data source. This library
+provides efficient implementations of all schema-agnostic interpreter components. All schema-aware
+logic is abstracted into the straightforward API of the InterpreterAdapter class, which should be
```
branen, line 21:
```suggestion
logic is abstracted into the straightforward, four-method API of the InterpreterAdapter class, which should be
```
Although you mention the compactness & simplicity of the API footprint below, I think it's worth mentioning here, too, as an enticement to the prospective user to keep reading.  (It is EXTREMELY enticing!)

branen (changes requested):
Looks good!  I suggested two small changes meant to make it more alluring to prospective users.

branen (approved):

PR #932, opened 2020-09-24
Add ImmutableStack dataclass implementation.

obi1kenobi:
This is a simple internal data structure of the interpreter, it's an implementation detail of some of the types that form the basic public-facing API of the module.

### DIFF
--- a/graphql_compiler/interpreter/immutable_stack.py
+++ b/graphql_compiler/interpreter/immutable_stack.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from typing import Any, Optional, Tuple
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, init=False)
 class ImmutableStack:
     """An immutable stack of arbitrary (heterogeneously-typed) values.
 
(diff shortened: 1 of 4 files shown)


graphql_compiler/interpreter/immutable_stack.py
```diff
@@ -0,0 +1,35 @@
...
+    """An immutable stack of arbitrary (heterogeneously-typed) values.
+
+    Specifically designed for cheap structural sharing, in order to avoid deep copies or
+    bugs caused by mutations of shared data.
+    """
+
+    # The following attributes are considered visible and safe for direct external use.
+    value: Any  # The value contained within this stack node.
+    depth: int  # The number of stack nodes contained in the tail of the stack.
+    tail: Optional["ImmutableStack"]  # The node that represents the rest of the stack, if any.
```
bojanserafimov, line 17: You can leave out `depth` from the constructor to avoid invalid stacks like `ImutableStack(5, 8, None)`.

obi1kenobi: Yeah good point 👍

obi1kenobi: My initial thought was that nobody should be instantiating `ImmutableStack` directly using the constructor, but hidden constructors aren't really a thing in Python...

bojanserafimov (approved):

PR #933, opened 2020-09-24
Eliminate remaining documentation duplication

pmantica1:
In this diff I eliminate the remaining documentation duplication. I effectively:

1. Copy to the troubleshooting documentation to the bottom of the CONTRIBUTING.rst file and delete the 
   TROUBLESHOOTING.rst file. This documentation is only about troubleshooting issues when setting up your local 
   environment. So I think that it's best to have it as a section of the CONTRIBUTING.rst file. (We already had it set up this way in
   ReadTheDocs). 

2. Replace the ReadTheDocs code of conduct, contributing and changelog documentation with an include directive that 
    "includes" the top-level directory CODE_OF_CONDUCT.rst, CONTRIBUTING.rst and CHANGELOG.rst files. This way the 
    documentation is readily accessible from both ReadTheDocs and github. I copied this trick from the ReadTheDocs github 
    repository.

I verified by using git diff that the existing ReadTheDocs documentation matched the top level github documentation. So I am not deleting any documentation here.

 

 

-


CONTRIBUTING.rst
```diff
@@ -85,7 +85,7 @@ teardown. These can be optionally skipped during development by running:
 
    pytest -m 'not slow'
 
-If you run into any issues, please consult the TROUBLESHOOTING.md file.
+If you run into any issues, please consult the `troubleshooting guide <#troubleshooting-guide>`__.
```
pmantica1, line 88: Verified that this link worked.

CONTRIBUTING.rst
```diff
@@ -97,8 +97,8 @@ Code of Conduct
 ---------------
 
 This project adheres to the Contributor Covenant `code of
-conduct <CODE_OF_CONDUCT.rst>`__. By participating, you are expected to
-uphold this code. Please report unacceptable behavior at
+conduct <https://graphql-compiler.readthedocs.io/en/latest/about/code_of_conduct.html>`__. By
```
pmantica1, line 100: I changed this link because it won't work from ReadTheDocs as is. I verified that it works.

bojanserafimov (approved):

PR #939, opened 2020-09-28
Add InterpreterAdapter class with thorough class docstring.

obi1kenobi:
The `typedefs.py` file inside the interpreter library has a ton of subtleties and a decent amount of complexity, so I'm chunking it up across several PRs for easier review.

This PR contains just the most bare-bones type definitions required to describe the `InterpreterAdapter` class, together with a class-level docstring that goes into some depth. In the interest of making reviews faster and smaller, I chose to open the PR without detailed function docstrings for `InterpreterAdapter`'s four methods. I was going to add them in a separate PR later, though if you'd prefer I can add them here instead.

I also did not document or implement the `DataContext` type, instead choosing to leave that for a future PR as well.


obi1kenobi:
There's no coverage because there's no functionality yet, all I've added are types.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+    The set of hints (and the information each hint provides) could grow in the future. Currently,
+    the following hints are offered:
+    - runtime_arg_hints: the values of any runtime arguments provided to the query for use in
+      filtering operations (recall queries' "$foo" filter parameter syntax).
+    - used_property_hints: the property names within the scope relevant to the called function that
+      the query will eventually need, e.g. for filtering on, or to output as the final result.
+    - filter_hints: information about the filters applied within the scope relevant to the called
+      function, such as "which filtering operation is being performed?" and "with which arguments?"
+    - neighbor_hints: information about the edges originating from the scope relevant to the called
+      function that the query will eventually need to expand.
```
branen, line 136: What does "the scope relevant to the called function" mean?  Also, is the scope relevant to the called function, or to the particular invocation of the function?

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+
+    More details on these hints, and suggestions for their use, can be found in the API methods
+    docstrings available below.
```
branen, line 139:
```suggestion
    More details on these hints, and suggestions for their use, can be found in the methods'
    docstrings, available below.
```

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+        **hints: Any,
+    ) -> Iterable[Tuple[DataContext[DataToken], Iterable[DataToken]]]:
+        """Produce the neighbors along a given edge for each of an iterable of input DataTokens."""
+        # TODO(predrag): Add more docs in an upcoming PR.
+        #
+        # If using a generator instead of a list for the Iterable[DataToken] part,
+        # be careful -- generators are not closures! Make sure any state you pull into
+        # the generator from the outside does not change, or that bug will be hard to find.
+        # Remember: it's always safer to use a function to produce the generator, since
+        # that will explicitly preserve all the external values passed into it.
```
branen, line 192:
```suggestion
        # If using a generator or a mutable data type for the Iterable[DataToken] part,
        # be careful! Make sure any state it depends upon
        # does not change, or that bug will be hard to find.
```
I think this admonition is important enough to warrant keeping it short and sweet.  I don't know that we necessarily need to make additional points about closures and functions, but if we do, I think they might be made more profitably in short code examples than in prose.

branen (changes requested):
Nice work!  I like the structure of it, and it reads very clearly.

I suggested several small edits, for clarity and copy, and have one large question about "the scope relevant to the called function."

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+    - an instance of a custom class which has *some* of the values of the vertex properties, and
+      has sufficient information to look up the rest of them if they are ever requested.
+
+    The best choice of DataToken type is dependent on the specific use case, e.g. whether the data
+    is already available in Python memory, or is on a local disk, or is a network hop away.
+
+    Implementers are free to choose any DataToken type and the interpreter code will happily use it.
+    However, for the sake of easier debugging and testing using the built-in functionality in
+    this library, it is desirable to make DataToken be a deep-copyable type that implements
+    equality beyond a simple referential equality check.
```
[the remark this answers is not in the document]
obi1kenobi: I'd say "work best" instead of "only work", what do you think?

The debugging functionality makes deep copies because `DataToken` might be mutable, but in principle this can be avoided if the `DataToken` isn't deep-copyable. Deep-copying prevents referential equality from being that useful.

Debugging traces can always be read by a human, but if semantic equality isn't available then those traces might not be meaningfully comparable to each other — `==` might produce `False` even for semantically-identical traces, due to referential equality not matching.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+    - get_tokens_of_type() produces an iterable of DataTokens of the type specified by its argument.
+      The calling function will wrap the DataTokens into a bookkeeping object called a DataContext,
+      where a particular token is currently active and specified in the "current_token" attribute.
+    - For an iterable of such DataContexts, project_property() can be used to get the value
+      of one of the properties on the vertex type represented by the currently active DataToken
+      in each DataContext; project_property() therefore returns an iterable of
+      tuples (data_context, value).
+    - project_neighbors() is similar: for an iterable of DataContexts and a specific edge name,
+      it returns an iterable (data_context, iterable_of_neighbor_tokens) where each result
+      contains the DataTokens of the neighboring vertices along that edge for the vertex whose
```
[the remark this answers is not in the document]
obi1kenobi: I did struggle with this sentence. There are too many clauses already for my taste, but I didn't find a good way to preserve clarity and unambiguousness while making it simpler.

The issue with just saying "vertices that neighbor the vertex ..." is that there may be many vertices that are neighbors but along _a different edge_ than the one specified. Those don't count here, and shouldn't get returned.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+    - Interpreters perform lazy evaluation by default: if exactly 3 query results are requested,
+      then only the minimal data necessary to produce *exactly 3 rows' worth* of outputs is loaded.
+    - When computing a particular result, data loading for output fields is deferred
+      until *after* all filtering operations have been completed, to minimize data loads.
+    - Data caching is easy to implement within this API -- simply have
+      your API function's implementation consult a cache before performing the requested operation.
+    - Batch-loading of data can be performed by simply advancing the input generator multiple times,
+      then operating on an entire batch of input data before producing corresponding outputs:
+
+        def project_property(
```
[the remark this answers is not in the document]
obi1kenobi: Yay! 🙌  I added it because I found through talking to people that this pattern seems to be in the "obvious in retrospect" category, and it's a very easy and powerful way to make queries faster since most APIs support some flavor of batch operation.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+      DataToken is currently active in that DataContext.
+    - can_coerce_to_type() is used to check whether a DataToken corresponding to one vertex type
+      can be safely converted into one representing a different vertex type. Given an iterable of
+      DataContexts and the name of the type to which the conversion is attempted, it produces
+      an iterable of tuples (data_context, can_coerce), where can_coerce is a boolean.
+
+    ## Performance and optimization opportunities
+
+    The design of the API, including its generator-style operation, enables a variety of
+    optimizations to either happen automatically or be available with minimal additional work.
```
chewselene, line 93:
```suggestion
    The design of the API, including its generator-style operation, enables a variety of
    optimizations, available either automatically or with minimal additional work.
```

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+    A few simple examples:
+    - Interpreters perform lazy evaluation by default: if exactly 3 query results are requested,
+      then only the minimal data necessary to produce *exactly 3 results' worth* of outputs is loaded.
```
chewselene, line 96:
```suggestion
    - Interpreters perform lazy evaluation by default: if 3 query results are requested,
      only the minimal data necessary to produce *exactly 3* results worth of outputs is loaded.
```
I think this makes the emphasis a bit more effective. Not positive about the apostrophe, but I don't think it is needed?

chewselene (commented):
Looking good! Excited to get this in the compiler 🎉

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+    described by your chosen schema. For example, if building a subclass of InterpreterAdapter
+    called MyAdapter with dict as the DataToken type, MyAdapter should be defined as follows:
+
+        class MyAdapter(InterpreterAdapter[dict]):
+            ...
+
+    Here are a few common examples of DataToken types in practice:
+    - a dict containing the type name of the vertex and the values of all its properties;
+    - a dataclass containing the type name of the vertex, and the collection name and primary key
+      of the database entry using which the property values can be looked up, or
```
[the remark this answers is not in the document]
branen: I also got tripped up here.  How about something like this?

```suggestion
    - a dataclass containing the type name of the vertex, and a collection name and primary key
      that can be used to retrieve its property values from a database, or
```

graphql_compiler/interpreter/typedefs.py
branen: :+1: to everything, including dropping the apostrophe.

graphql_compiler/interpreter/typedefs.py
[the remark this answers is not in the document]
branen: :+1: to "work best".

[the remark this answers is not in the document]
branen: (In retrospect, it probably would have been more correct from an "intelligible comment history" viewpoint to have created a new one--I'll try to remember this for next time.)

graphql_compiler/interpreter/typedefs.py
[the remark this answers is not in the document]
branen: Right, got it.  I'll make a new suggestion, but it'll need to span two lines, so I'm not sure I can do it in this thread.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -0,0 +1,208 @@
...
+      The calling function will wrap the DataTokens into a bookkeeping object called a DataContext,
+      where a particular token is currently active and specified in the "current_token" attribute.
+    - For an iterable of such DataContexts, project_property() can be used to get the value
+      of one of the properties on the vertex type represented by the currently active DataToken
+      in each DataContext; project_property() therefore returns an iterable of
+      tuples (data_context, value).
+    - project_neighbors() is similar: for an iterable of DataContexts and a specific edge name,
+      it returns an iterable (data_context, iterable_of_neighbor_tokens) where each result
+      contains the DataTokens of the neighboring vertices along that edge for the vertex whose
+      DataToken is currently active in that DataContext.
```
branen, line 84:
```suggestion
    - project_neighbors() is similar: for an iterable of DataContexts and a specific edge name,
      it returns an iterable (data_context, iterable_of_neighbor_tokens) where
      iterable_of_neighbor_tokens yields a DataToken for each vertex that can be reached by
      following the specified edge from data_context's vertex. 
```

branen: Here's a second shot at the `project_neighbors` description.  It's tricky!

branen (changes requested):
Here's a second shot at the `project_neighbors` description.

graphql_compiler/interpreter/typedefs.py
obi1kenobi: I like where this is going. I might tweak it minimally, just to make it clear that "available automatically" means "they are always in place" as opposed to "you have to enable them, even though you don't have to implement them."

graphql_compiler/interpreter/typedefs.py
obi1kenobi: I rewrote this part to explicitly define which scope is in play for each function, let me know what you think.

branen (approved):
I'm so excited for this to land!

PR #942, opened 2020-10-01
Interpreter: Pass hints into project_property calls

bojanserafimov:


graphql_compiler/interpreter/expression_ops/immediate_expression_handlers.py
```diff
@@ -41,6 +41,9 @@ def evaluate_context_field(
     location = expression.location.at_vertex()
     field_name = expression.location.field
 
+    # TODO(bojanserafimov): Memoize hints for each location.
+    hints = construct_hints_for_location(query_metadata_table, query_arguments, location)
```
obi1kenobi, line 45: Can you add a similar thing around line 29?

In the output-generation stage, `project_property` calls are made while resolving `ContextField` expressions so this works for the particular test case. However, when filtering, `project_property` calls come from `LocalField` expressions (because the property is in the local scope) so this code path won't get used. Even more confusingly for the end user, if using a tagged parameter for a filter, the `project_property` for the tagged value will have hints because it too will be a `ContextField`.

obi1kenobi (approved):
This is fine, and if you'd like you can merge it as-is. But you can implement something significantly nicer with only a bit more work 😄

graphql_compiler/interpreter/expression_ops/immediate_expression_handlers.py
```diff
@@ -23,10 +23,13 @@ def evaluate_local_field(
     expression: LocalField,
     data_contexts: Iterable[DataContext],
 ) -> Iterator[Tuple[DataContext, Any]]:
-    # TODO(predrag): Add hints here.
+    location = expression.location.at_vertex()
```
obi1kenobi, line 26:
```suggestion
    vertex_location = expression.location.at_vertex()
```

graphql_compiler/interpreter/expression_ops/immediate_expression_handlers.py
```diff
@@ -23,10 +23,13 @@ def evaluate_local_field(
...
+
+    # TODO(bojanserafimov): Memoize hints for each location.
+    hints = construct_hints_for_location(query_metadata_table, query_arguments, location)
```
obi1kenobi, line 29:
```suggestion
    hints = construct_hints_for_location(query_metadata_table, query_arguments, vertex_location)
```

obi1kenobi (approved):
This looks good. For bonus points, add a trace that does a simple filter on one field and outputs another.

graphql_compiler/interpreter/expression_ops/immediate_expression_handlers.py
```diff
@@ -23,10 +23,13 @@ def evaluate_local_field(
     expression: LocalField,
     data_contexts: Iterable[DataContext],
 ) -> Iterator[Tuple[DataContext, Any]]:
-    # TODO(predrag): Add hints here.
+    vertex_location = expression.location.at_vertex()
```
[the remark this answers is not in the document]
obi1kenobi: I knew there was a reason I didn't add hints to `project_property` the first time around 😄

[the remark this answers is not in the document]
bojanserafimov: We could just convert local fields to context fields as an IR pass, but I think that's problematic, since the context consists of locations already processed.

The alternative is to plumb the current location in here through the dispatcher, and everywhere else that's required to make that work.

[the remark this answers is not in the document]
obi1kenobi: It's also illegal to do that conversion — `LocalField` expressions show up _before_ the location is declared in a `MarkLocation` block, so they'd be forward-referencing a still-undeclared location value.

We probably need to do the current location plumbing.

graphql_compiler/tests/interpreter_tests/test_interpreter_behavior.py
```diff
@@ -237,3 +240,27 @@ def test_single_generator_pull_grabs_only_one_result_from_adapter(self) -> None:
...
+        list(result_gen)  # drain the iterator
+
+        trace = adapter.recorder.get_trace()
+
+        project_property_calls = [
+            operation
+            for operation in trace.operations
+            if operation.kind == "call" and operation.name == "project_property"
+        ]
+        self.assertEqual(3, len(project_property_calls))  # TODO(bojanserafimov): Why 3?
```
bojanserafimov, line 266: @obi1kenobi There's 10 animals in the dataset, not sure how this completes with 3 calls to `project_property`

obi1kenobi: Ah, you've stumbled upon precisely the part that makes the interpreter difficult to implement, but once implemented makes the API pleasant and powerful.

Look at the `field_name` values of the `project_property` calls. They are going to be `name`, `color`, and `name`, almost certainly in that order. Here's what's happening:
- First call (`name`): this is to get the value of the `name @filter(...)` part.
- Second call (`color`): this is to get a value for the `value: ["%color"]` tagged parameter inside the filter.
- Third call (`name`): this is to perform `name @output(...)`.

You correctly point out that there are 10 animals, and it would be a valid execution of the interpreter if it called `project_property` for each used property for each animal (30 calls in total). But it clearly doesn't do that.

To see why, recall the interpreter's desirable optimality ("anything you can do, I can do at least as well") property. This property doesn't hold if we can show an optimization that is prevented by the interpreter API's design, or by the interpreter library's implementation.

Making anything more than the 3 calls I described above does prevent a very desirable optimization: user-specified batching. Imagine `name` for animals 1-4 is fetched in one `project_property` call, and for animals 5-10 in another `project_property` call. This prevents the user from batching 10 (or 8, or all, or `random.randint()`) animals at a time — no implementation of the `InterpreterAdapter` API is able to batch over input data it is unable to receive as input, since the `data_contexts` input generator will run out.

So optimality requires that *all* the relevant `data_contexts` are delivered in a single call to `project_property` (or any of the other two functions that receive generators as input). That means that the set of function *calls* in the trace is purely a function of the input query, and does not depend on the data.

Taking a step back, *this* is why the interpreter is so challenging to build: because it has to dynamically construct a chain of generators in *just the right way* that fulfills the target query without knowing how the 4-method `InterpreterAdapter` API is going to be implemented, and by extension guaranteeing correct results while not knowing the exact order of operations between those generators beyond "data is created before it is consumed."

graphql_compiler/interpreter/expression_ops/dispatch.py
```diff
@@ -30,7 +31,8 @@ def evaluate_expression(
     adapter: InterpreterAdapter[DataToken],
     query_metadata_table: QueryMetadataTable,
     query_arguments: Dict[str, Any],
-    current_type_name: str,
+    current_location: Optional[Location],
+    current_type_name: str,  # TODO remove this?
```
obi1kenobi, line 35: This should be safe to remove now. The only place where the current type name would not match the location's type is during `can_coerce_to_type` where the current type name is the pre-coercion type (the `coerced_from_type` attribute instead of the `type` attribute on `LocationInfo`), but no expressions can exist during `can_coerce_to_type` and the type coercion happens before any filtering / outputting.

Feel free to change this in the next PR if you'd prefer not to grow this one any more.

graphql_compiler/interpreter/expression_ops/immediate_expression_handlers.py
```diff
@@ -19,28 +20,34 @@ def evaluate_local_field(
...
     query_metadata_table: QueryMetadataTable,
     query_arguments: Dict[str, Any],
+    current_location: Optional[Location],
     current_type_name: str,
     expression: LocalField,
     data_contexts: Iterable[DataContext],
 ) -> Iterator[Tuple[DataContext, Any]]:
-    # TODO(predrag): Add hints here.
+    # TODO(bojanserafimov): Memoize hints for each location.
+    hints = construct_hints_for_location(query_metadata_table, query_arguments, current_location)
```
obi1kenobi, line 29: This is not a legal call, `current_location` is `Optional[Location]` but is being passed to a function that expects `Location`.

`current_location` is only `None` in the global scope, which cannot have `LocalField` expressions, so you can assert that `current_location` is not `None` here.

graphql_compiler/interpreter/expression_ops/dispatch.py
bojanserafimov: Sure. Wrote a clearer todo

graphql_compiler/interpreter/expression_ops/immediate_expression_handlers.py
```diff
@@ -25,6 +25,10 @@ def evaluate_local_field(
     expression: LocalField,
     data_contexts: Iterable[DataContext],
 ) -> Iterator[Tuple[DataContext, Any]]:
+    if current_location is None:
+        raise AssertionError(f"Received None for current_location, which indicates we are "
+                             f"in the global scope, but there can't be any local fields there.")
```
obi1kenobi, line 30:
```suggestion
        raise AssertionError(
            f"Unexpectedly attempting to evaluate a LocalField while in the global scope "
            f"(current_location is None). This is a bug: {expression}"
        )
```

bojanserafimov: Good fix, thanks. With the way I wrote it, a user would have no idea if this is a bug or an obscure user error.

obi1kenobi (approved):
Thanks for the fix!

PR #945, opened 2020-10-06
Add debug and test helpers for interpreters.

obi1kenobi:
These are used in tests.

Adding the test code requires merging much more code first, so for the moment they'll show up as uncovered code. You can see how they are used in the tests on the `interpreted_mode_v3` branch, starting from this line onward: https://github.com/kensho-technologies/graphql-compiler/blob/interpreted_mode_v3/graphql_compiler/tests/interpreter_tests/test_interpreter_behavior.py#L138

### DIFF
--- a/graphql_compiler/interpreter/debugging.py
+++ b/graphql_compiler/interpreter/debugging.py
@@ -144,8 +144,8 @@ class TraceRecorder(Generic[DataToken]):
 
         Yields:
             for each piece of data yielded by the input iterable, yields a tuple of:
-            - the operation's unique identifier for the yielded data
-            - the input iterable's yielded data itself
+            - the operation's unique identifier for the yielded data, and
+            - the input iterable's yielded data itself.
         """
         for item in iterable:
             item_uid = len(self._operation_log)
--- a/graphql_compiler/interpreter/debugging.py
+++ b/graphql_compiler/interpreter/debugging.py
@@ -42,10 +42,19 @@ def _unzip_and_yield_second(iterable: Iterable[Tuple[T, U]]) -> Iterable[U]:
 class AdapterOperation:
     """The record of an action performed on or by an InterpreterAdapter function as part of a trace.
 
+    Actions that can be recorded like this include function calls, function returns, function yields
+    (including yields that contain nested generators, whose yields will also be recorded),
+    and yields of generators that were passed as inputs to function calls. The kind of action being
+    recorded is reflected in the kind attribute, and carries a value suitable to the kind of action
+    in the data attribute.
+
     For example, calling get_tokens_of_type("Foo") would produce an AdapterOperation
     with kind "call", with name "get_tokens_of_type", with unique ID and parent unique IDs
-    appropriate for the trace, and with data (("Foo",), {}) as the tuple of
-    the positional arguments tuple and kwargs dict.
+    appropriate for the trace, and with data (("Foo",), {}) representing a (positional_args, kwargs)
+    tuple for the function call. For function returns, the data attribute of AdapterOperation
+    holds the value returned by the function. For yields of input or output generators, the content
+    of the data attribute is determined by the specific circumstances of the function named in
+    the name attribute -- e.g. whether the function outputs a generator of generators, etc.
 
     The parent_uid field allows us to track dependencies across different operations. For example,
     data yielded by a generator-style function will be recorded in AdapterOperations
--- a/graphql_compiler/interpreter/debugging.py
+++ b/graphql_compiler/interpreter/debugging.py
@@ -42,14 +42,14 @@ def _unzip_and_yield_second(iterable: Iterable[Tuple[T, U]]) -> Iterable[U]:
 class AdapterOperation:
     """The record of an action performed to or by an InterpreterAdapter function as part of a trace.
 
-    For example, calling get_tokens_of_type("Foo") would be an AdapterOperation with kind "call",
+    For example, calling get_tokens_of_type("Foo") would produce an AdapterOperation with kind "call",
     with name "get_tokens_of_type", with unique ID and parent unique IDs appropriate for the trace,
     and with data (("Foo",), {}) as the tuple of the positional arguments tuple and kwargs dict.
 
     The parent_uid field allows us to track dependencies across different operations. For example,
-    data yielded by a call to a generator-style function will be recorded in AdapterOperations
-    with kind "yield" whose parent_uid points to the AdapterOperation of the particular invocation
-    of the function that is yielding that data.
+    data yielded by a generator-style function will be recorded in AdapterOperations
+    of kind "yield" whose parent_uids point to the AdapterOperation of the particular invocation
+    of the function that is generating that data.
     """
 
     kind: Literal["call", "yield", "return"]
@@ -61,7 +61,7 @@ class AdapterOperation:
 
 @dataclass(frozen=True)
 class RecordedTrace(Generic[DataToken]):
-    """A complete immutable recording of the execution of a query via an InterpreterAdapter.
+    """A complete, immutable recording of the execution of a query via an InterpreterAdapter.
 
     Includes a linearized sequence of all operations performed during the execution of the query:
     - adapter function calls, yields, and returns;
@@ -73,10 +73,10 @@ class RecordedTrace(Generic[DataToken]):
     The recording is sufficiently detailed to be used in at least the following use cases:
     - to examine, either manually or through automated tests, a query's execution flow;
     - to "impersonate" an InterpreterAdapter implementation for a given query, allowing bugs in
-      the interpreter library code to be isolated and reproduced without an InterpreterAdapter;
-    - to "impersonate" the interpreter library code toward an InterpreterAdapter implementation,
+      the interpreter code to be isolated and reproduced without an InterpreterAdapter;
+    - to "impersonate" the interpreter code toward an InterpreterAdapter implementation,
       allowing adapter implementers to test and examine their implementations in isolation from
-      the rest of the interpreter library code.
+      the rest of the interpreter code.
     """
 
     DEFAULT_ROOT_UID: ClassVar[int] = -1
--- a/graphql_compiler/interpreter/debugging.py
+++ b/graphql_compiler/interpreter/debugging.py
@@ -1,0 +1,196 @@
+from copy import deepcopy
+from dataclasses import dataclass, field
+from typing import (
+    Any,
+    ClassVar,
+    Dict,
+    Generic,
+    Iterable,
+    List,
+    Sequence,
+    Tuple,
+    TypeVar,
+)
+
+from ..typedefs import Literal
+from .typedefs import DataToken
+
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def _enumerate_starting_at(
+    sequence: Sequence[T],
+    first_index: int,
+) -> Iterable[Tuple[int, T]]:
+    """Enumerate only a portion of a sequence, without slicing the sequence.
+
+    Equivalent to enumerate(sequence[first_index:], start=first_index). The built-in enumerate()
+    always iterates the entire sequence and only changes the counter's value based on
+    the start parameter. In this function, we skip elements with indices in [0, first_index)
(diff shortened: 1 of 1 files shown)


graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,186 @@
...
+
+        Args:
+            operation_name: name to associate with this operation; usually the name of the function
+                            whose iterable output is being recorded.
+            parent_uid: unique identifier of the parent operation of this operation.
+            iterable: iterable to record
+
+        Yields:
+            for each piece of data yielded by the input iterable, yields a tuple of the
+            operation's unique identifier for that data together with the data itself
```
chewselene, line 131: sentence is a bit wordy. solution might be to do `yields a tuple of:` and then a bulleted list with the elements of the tuple. comes up in `record_compound_iterable` as well.

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,186 @@
...
+
+def _unzip_and_yield_second(iterable: Iterable[Tuple[T, U]]) -> Iterable[U]:
+    """Unpack an iterable of tuples and only yield the second element of each tuple."""
+    for _, second in iterable:
+        yield second
+
+
+@dataclass(frozen=True)
+class AdapterOperation:
+    """The record of an action performed to or by an InterpreterAdapter function as part of a trace.
```
chewselene, line 43: what does `an action performed to a function` mean?

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,196 @@
...
+
+    For example, calling get_tokens_of_type("Foo") would be an AdapterOperation with kind "call",
+    with name "get_tokens_of_type", with unique ID and parent unique IDs appropriate for the trace,
+    and with data (("Foo",), {}) as the tuple of the positional arguments tuple and kwargs dict.
+
+    The parent_uid field allows us to track dependencies across different operations. For example,
+    data yielded by a call to a generator-style function will be recorded in AdapterOperations
+    with kind "yield" whose parent_uid points to the AdapterOperation of the particular invocation
+    of the function that is yielding that data.
```
branen, line 52:
```suggestion
    The parent_uid field allows us to track dependencies across different operations. For example,
    data yielded by a generator-style function will be recorded in AdapterOperations
    of kind "yield" whose parent_uids point to the AdapterOperation of the particular invocation
    of the function that is generating that data.
```
This edit consists of two nits:
- Disentangle the invocation of a generator-style function to produce a generator and the invocation of the generator to yield the data
- Match the cardinalities of "AdapterOperations" and "parent_uid"

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,196 @@
...
+    kind: Literal["call", "yield", "return"]
+    name: str
+    uid: int
+    parent_uid: int
+    data: Any
+
+
+@dataclass(frozen=True)
+class RecordedTrace(Generic[DataToken]):
+    """A complete immutable recording of the execution of a query via an InterpreterAdapter.
```
branen, line 64:
```suggestion
    """A complete, immutable recording of the execution of a query via an InterpreterAdapter.
```

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,196 @@
...
+
+    Includes a linearized sequence of all operations performed during the execution of the query:
+    - adapter function calls, yields, and returns;
+    - actuation of all input and output iterables.
```
branen, line 68:
```suggestion
    Includes a linear sequence of all operations performed during the execution of the query.
```
- "Linearized" seems to suggest a transformation, but the sequence is actually just all the operations that occurred in their actual execution order on our single thread of execution, right?

- Isn't the "actuation of iterables" covered by the "adapter function yields"?

I don't think we need to enumerate all the operation types here, since we just covered them in the prior class's docstring and since the type annotations make it very clear where to look for additional information.

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,196 @@
...
+    Traces may record complete executions of queries (i.e. ones where the generators are run
+    to completion), and may also be used to record partial executions (e.g. "fetch only 3 results").
+
+    The recording is sufficiently detailed to be used in at least the following use cases:
+    - to examine, either manually or through automated tests, a query's execution flow;
+    - to "impersonate" an InterpreterAdapter implementation for a given query, allowing bugs in
+      the interpreter library code to be isolated and reproduced without an InterpreterAdapter;
+    - to "impersonate" the interpreter library code toward an InterpreterAdapter implementation,
+      allowing adapter implementers to test and examine their implementations in isolation from
+      the rest of the interpreter library code.
```
branen, line 79:
```suggestion
    The recording is sufficiently detailed to be used in at least the following use cases:
    - to examine, either manually or through automated tests, a query's execution flow;
    - to "impersonate" an InterpreterAdapter implementation for a given query, allowing bugs in
      the interpreter code to be isolated and reproduced without an InterpreterAdapter;
    - to "impersonate" the interpreter code toward an InterpreterAdapter implementation,
      allowing adapter implementers to test and examine their implementations in isolation from
      the rest of the interpreter code.
```
Since we've settled on the interpreter/adapter dichotomy, I think it's worth dropping "library" for brevity & clarity.  I think "library" connotes "something that I call" to a lot of developers, and its use might confuse people who are responsible for writing adapters but who may not be responsible for running the service in which their adapters are used.

branen (changes requested):
Nice work!  I made a few small suggestions.  I suspect I missed a few commits, so I'll circle around to those momentarily.

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,196 @@
...
+        first_index: the first index of the sequence to produce in the resulting iterable
+
+    Yields:
+        tuples (index, element_at_that_index_in_sequence), analogous to the built-in enumerate()
+    """
+    for index in range(first_index, len(sequence)):
+        yield index, sequence[index]
+
+
+def _unzip_and_yield_second(iterable: Iterable[Tuple[T, U]]) -> Iterable[U]:
```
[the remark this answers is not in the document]
obi1kenobi: There are 7 more call sites in this file, in code I chose not to include in this PR for the sake of keeping the size manageable. It's a very useful helper around `record_iterable()` and `record_compound_iterable()` calls, most of which aren't in this PR but are available on the `interpreted_mode_v3` branch: https://github.com/kensho-technologies/graphql-compiler/blob/interpreted_mode_v3/graphql_compiler/interpreter/debugging.py#L137

graphql_compiler/interpreter/debugging.py
obi1kenobi: Yeah this is not my favorite phrasing either. The functions being recorded aren't "plain" functions but generators (AKA coroutines), which may be resumed by calling their `send()` method (in Python, functions can have methods — weird, right?)

This is arguably an action performed _on_ the function, rather than _by_ the function. I probably should have used "on" instead of "to" but I doubt the bad choice of preposition is the crux of the issue.

I had a really hard time figuring out something concise enough to be the top line of this docstring. Any ideas?

graphql_compiler/interpreter/debugging.py
obi1kenobi: It sort of is a transformation, in a sense. The trace represents one possible linearization of the execution, but there are many others that would also be valid and could be produced with a slightly different adapter implementation. "Linearized" here is (perhaps non-obviously) referring to the property of [linearizability](https://en.wikipedia.org/wiki/Linearizability) — the trace is a concrete linearization of the observed execution, a "proof by construction" that the interpreter satisfies the linearizability property. Linearizability isn't that important at the moment (the interpreter executes synchronously on a single thread) but becomes more interesting if/when we add `async` support.

For "actuation of iterables", you're right that it is partially covered by "adapter function yields", but it's not entirely the same — the `data_contexts` input argument to adapter functions is also actuated, but in that case the adapter function is _yielded to_. This operation is also recorded in the trace, and I felt it was worth mentioning here since the previous class' docstring wasn't explicit about that since it doesn't distinguish between the two kinds of yields.

This is all tricky, and I recognize that the current description doesn't do a good job of communicating what needs to be communicated. Let me know what you think about the above, and let's work together to find a better way to describe all this.

obi1kenobi:
Will integrate the feedback into the code in the morning, just wanted to post a few comment replies to give you the option to read them while I work on the code changes.

graphql_compiler/interpreter/debugging.py
obi1kenobi: I'm adding another short paragraph here explaining how input iterable arguments are recorded, which should hopefully clarify this a bit. Let me know if that did the trick, or if I should do something else as well.

graphql_compiler/interpreter/debugging.py
obi1kenobi: Tweaked as you suggested, let me know if this is better.

graphql_compiler/interpreter/debugging.py
obi1kenobi: Changed the "actuation of iterables" to make the two bullet points have no overlap between each other.

graphql_compiler/interpreter/debugging.py
```diff
@@ -0,0 +1,195 @@
...
+    an iterable are consumed, those values are recorded in their own AdapterOperation records
+    with kind "yield", parent unique ID matching the unique ID of the "call" record, and
+    name matching the placeholder name in the "call" record.
+    """
+
+    kind: Literal["call", "yield", "return"]
+    name: str
+    uid: int
+    parent_uid: int
+    data: Any
```
chewselene, line 66: This sentence: `data (("Foo",), {}) as the tuple of the positional arguments tuple and kwargs dict.` makes `data` sound like something specific so being type `Any` confused me. Looks it varies based on the `kind`? I think more explanation is need for what `data` is expected to be.

graphql_compiler/interpreter/debugging.py
chewselene: The `on` definitely helps! I don't have a suggestion for how to make this better so I'd say it is fine :)

graphql_compiler/interpreter/debugging.py
obi1kenobi: I added some more information about what `data` is expected to be. Unfortunately, when recording a `kind="yield"` operation, the `data` value can be a variety of different things that are specific to the use case and recorder class/function being used. I tried to clarify that as best I could while still trying to keep the docstring concise, let me know what you think.

graphql_compiler/interpreter/debugging.py
```diff
@@ -172,8 +179,9 @@ def record_compound_iterable(
             compound_iterable: iterable to record
 
         Yields:
-            for each piece of data yielded by the input compound_iterable, yields a tuple of the
-            operation's unique identifier for that data together with the data itself
+            for each piece of data yielded by the input compound_iterable, yields a tuple of:
+            - the operation's unique identifier for the yielded data
+            - the input compound_iterable's yielded data itself
```
branen, line 184:
```suggestion
            - the operation's unique identifier for the yielded data, and
            - the input compound_iterable's yielded data itself.
```

graphql_compiler/interpreter/debugging.py
```diff
@@ -137,8 +143,9 @@ def record_iterable(
             iterable: iterable to record
 
         Yields:
-            for each piece of data yielded by the input iterable, yields a tuple of the
-            operation's unique identifier for that data together with the data itself
+            for each piece of data yielded by the input iterable, yields a tuple of:
+            - the operation's unique identifier for the yielded data
+            - the input iterable's yielded data itself
```
branen, line 148:
```suggestion
            - the operation's unique identifier for the yielded data, and
            - the input iterable's yielded data itself.
```

branen (dismissed):
Looks good!  I suggested a couple of edits that I think improve readability slightly, but I don't think they're strictly necessary, and if they don't make it in, I will not be sad.

branen (approved):
(appeasing GitHub)

PR #947, opened 2020-10-07
Add docstring for project_property().

obi1kenobi:


graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,73 @@ def project_property(
...
+                    bar @output(out_name: "bar_value")
+                }
+            }
+        Once the interpreter has used the get_tokens_of_type() function to obtain
+        an iterable of DataTokens for the Foo type, it will need to get the value of
+        the "bar" property on those tokens. This is where project_property() comes in.
+        The interpreter would call project_property() with arguments current_type_name = "Foo"
+        and field_name = "bar" along with an iterable of DataContext values, each of which
+        contains an optional DataToken as its current_token attribute. These DataContext
+        objects simply allow the interpreter to keep track of "which data came from where";
```
chewselene, line 191: Consider including this bit about what `DataContexts` are earlier in the doc string. I think the difference between `DataContext` and `DataToken` is not immediately obvious without this and the reader might be confused until they get to this point.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,73 @@ def project_property(
...
+
+                    # Remember to always yield the DataContext alongside the produced value
+                    yield data_context, property_value
+
+        Args:
+            data_contexts: iterable of DataContext objects which specify the DataTokens whose
+                           property data needs to be loaded
+            current_type_name: name of the vertex type whose property is being loaded. Guaranteed
+                               to be the name of a type defined in the schema being queried.
+            field_name: name of the property whose data to load. Guaranteed to refer to a property
```
chewselene, line 222: I think the tense changes here in a vaguely confusing way. Maybe `name of the property whose data is to be loaded` or `name of the property whose data needs to be loaded` instead?

graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,73 @@ def project_property(
...
+                        that is valid for the supplied current_type_name based on the schema.
+            runtime_arg_hints: names and values of any runtime arguments provided to the query
+                               for use in filtering operations (e.g. "$foo").
```
chewselene, line 225: Commented this on the `get_tokens_of_type` docstring PR as well, but maybe `baz` for the stand-in runtime arg since `Foo` and `bar` are already used in this docstring.

graphql_compiler/interpreter/typedefs.py
LWprogramming: Should be [coming soon](https://github.com/kensho-technologies/graphql-compiler/blob/936e47970eff81c2a6eeaead44fdf4d8d3c1a599/graphql_compiler/interpreter/typedefs.py#L23)

obi1kenobi: Upon reading this one more time, I think I agree with @chewselene. While the reader could refer to the (not-yet-written) docs for `DataContext`, it's entirely possible that they miss the fact that `DataToken` and `DataContext` are not the same thing, and get confused. If we can make this part stand better on its own, we should. Let me give it another shot.

obi1kenobi: Tweaked this, let me know what you think!

graphql_compiler/interpreter/typedefs.py
branen: Can "whose property data needs to be loaded", "whose property is being loaded", and "whose data to load" all be changed to roughly the same form?

graphql_compiler/interpreter/typedefs.py
branen:
```suggestion
        objects allow the interpreter to keep track of "which data came from where";
```

graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,73 @@ def project_property(
...
+                }
+            }
+        Once the interpreter has used the get_tokens_of_type() function to obtain
+        an iterable of DataTokens for the Foo type, it will need to get the value of
+        the "bar" property on those tokens. This is where project_property() comes in.
+        The interpreter would call project_property() with arguments current_type_name = "Foo"
+        and field_name = "bar" along with an iterable of DataContext values, each of which
+        contains an optional DataToken as its current_token attribute. These DataContext
+        objects simply allow the interpreter to keep track of "which data came from where";
+        only the DataToken value within the current_token attribute is relevant to this function.
```
branen:
```suggestion
        only the DataToken value bound to the current_token attribute is relevant to this function.
```
"Within" seems to imply an interstitial container--"bound to" or "referred to by" is probably clearer.

branen (changes requested):

graphql_compiler/interpreter/typedefs.py
```diff
@@ -176,22 +176,27 @@ def project_property(
...
+
         For example, consider the following query:
             {
                 Foo {
                     bar @output(out_name: "bar_value")
                 }
             }
+
         Once the interpreter has used the get_tokens_of_type() function to obtain
-        an iterable of DataTokens for the Foo type, it will need to get the value of
```
branen: YES!!!!!  This is so much clearer!

graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,84 @@ def project_property(
...
+                Foo {
+                    bar @output(out_name: "bar_value")
+                }
+            }
+
+        Once the interpreter has used the get_tokens_of_type() function to obtain
+        an iterable of DataTokens for the Foo type, it will automatically wrap each of them in
+        a "bookkeeping" object called DataContext. These DataContext objects allow
+        the interpreter to keep track of "which data came from where"; only the DataToken value
+        within their current_token attribute is relevant to the InterpreterAdapter API.
```
branen, line 191:
```suggestion
        a "bookkeeping" object called DataContext. These DataContext objects allow
        the interpreter to keep track of "which data came from where"; only the DataToken value
        bound to each current_token attribute is relevant to the InterpreterAdapter API.
```

graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,84 @@ def project_property(
...
+                    else:
+                        property_value = < load the field_name property's value for current_token >
+
+                    # Remember to always yield the DataContext alongside the produced value
+                    yield data_context, property_value
+
+        Args:
+            data_contexts: iterable of DataContext objects which specify the DataTokens whose
+                           property data needs to be loaded
+            current_type_name: name of the vertex type whose property is being loaded. Guaranteed
```
branen, line 235:
```suggestion
            current_type_name: name of the vertex type whose property needs to be loaded. Guaranteed
```
Unless there's a reason to use different phrasing here, making it concordant with the former and subsequent args' descriptions will reduce an opportunity for confusion.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -172,8 +172,84 @@ def project_property(
...
+                               to be the name of a type defined in the schema being queried.
+            field_name: name of the property whose data needs to be loaded. Guaranteed to refer
+                        either to a property that is defined in the supplied current_type_name
+                        in the schema, or to the "__typename" meta field that is valid for all
```
branen, line 239: Nice!

branen (changes requested):
I really like the paragraphs you reworded in response to Selene's feedback!  I suggested a couple of minor edits for clarity.

branen (commented):
:+1:

graphql_compiler/interpreter/typedefs.py
obi1kenobi: Changed to match in several places.

graphql_compiler/interpreter/typedefs.py
obi1kenobi: Good catch!

chewselene (approved):
I think this reads really clearly 🙌

LWprogramming:
Ah, hm, it seems like the requested changes are still showing up, blocking merge-- I'll dismiss that since that was resolved

PR #949, opened 2020-10-09
Add docstring for can_coerce_to_type().

obi1kenobi:
Using the same sort of general format as we established in the other API methods.

### DIFF
--- a/graphql_compiler/interpreter/typedefs.py
+++ b/graphql_compiler/interpreter/typedefs.py
@@ -307,8 +307,8 @@ class InterpreterAdapter(Generic[DataToken], metaclass=ABCMeta):
 
         The interpreter uses can_coerce_to_type() for this purpose: it calls this function with
         an iterable of DataContexts, with current_type_name set to "Bar" (the schema-implied type
-        of the query's scope) and with coerce_to_type_name set to "BarImpl" as the type to which
-        coercion is being attempted. For each DataContext in the input iterable, this function
+        of the query's scope) and with coerce_to_type_name set to "BarImpl" (the type to which
+        coercion is being attempted). For each DataContext in the input iterable, this function
         yields a tuple containing the context itself together with a bool set to True if
         the coercion to the new type could be completed.
 
--- a/graphql_compiler/interpreter/typedefs.py
+++ b/graphql_compiler/interpreter/typedefs.py
@@ -272,5 +272,103 @@ class InterpreterAdapter(Generic[DataToken], metaclass=ABCMeta):
         neighbor_hints: Optional[Collection[Tuple[EdgeInfo, NeighborHint]]] = None,
         **hints: Any,
     ) -> Iterable[Tuple[DataContext[DataToken], bool]]:
-        """Determine if each of an iterable of input DataTokens can be coerced to another type."""
-        # TODO(predrag): Add more docs in an upcoming PR.
+        """Determine if each of an iterable of input DataTokens can be coerced to another type.
+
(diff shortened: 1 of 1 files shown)


graphql_compiler/interpreter/typedefs.py
```diff
@@ -272,5 +272,103 @@ def can_coerce_to_type(
...
+                          such as "which filtering operations are being performed?"
+                          and "with which arguments?"
+            neighbor_hints: information about the edges of the vertices being processed
+                            that the query will eventually need to expand.
+            **hints: catch-all kwarg field making the function's signature forward-compatible with
+                     future revisions of this library that add more hints.
+
+        Yields:
+            tuples (data_context, can_coerce), with a DataContext together with whether
+            the current_token referenced by that context can be coerced to the specified type.
```
chewselene, line 371:
```suggestion
            tuples (data_context, can_coerce), containing a DataContext and a corresponding boolean indicating
            whether its current_token can be coerced to the specified type.
```
nit: the `with`/`with` got me a little tripped up.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -272,5 +272,103 @@ def can_coerce_to_type(
...
+        When resolving the out_Foo_Bar edge in the query using project_neighbors(), the interpreter
+        receives DataTokens that (per the schema) are instances of the Bar interface type. However,
+        the query's subsequent "... on BarImpl" type coercion clause requires that the interpreter
+        discard any neighboring vertices that are not instances of BarImpl, a subtype of
+        the Bar interface type.
+
+        The interpreter uses can_coerce_to_type() for this purpose: it calls this function with
+        an iterable of DataContexts, with current_type_name set to "Bar" (the schema-implied type
+        of the query's scope) and with coerce_to_type_name set to "BarImpl" as the type to which
+        coercion is being attempted. For each DataContext in the input iterable, this function
```
chewselene, line 311:
```suggestion
        of the query's scope) and with coerce_to_type_name set to "BarImpl" (the type to which
        coercion is being attempted). For each DataContext in the input iterable, this function
```

graphql_compiler/interpreter/typedefs.py
```diff
@@ -272,5 +272,103 @@ def can_coerce_to_type(
...
+        yields a tuple containing the context itself together with a bool set to True if
```
chewselene, line 312:
```suggestion
        yields a tuple containing the context itself with a bool set to True if
```

chewselene (commented):
Some super small nits, but overall I think this looks good!

graphql_compiler/interpreter/typedefs.py
obi1kenobi: I initially mis-parsed this as "context itself with a bool" being one clause. How about this instead:
`yields a tuple containing the context itself and a bool set to True ...`

graphql_compiler/interpreter/typedefs.py
obi1kenobi: I like this! One tweak I'd suggest is working around that `its` in `its current_token` since it is not obvious that it refers to `DataContext` and not `corresponding boolean`. Maybe if I change it to `indicating whether the DataContext's current_token` instead? Let me know what you think.

graphql_compiler/interpreter/typedefs.py
chewselene: 👍  yep, i like that better!

chewselene (approved):

PR #951, opened 2020-10-12
Fix pagination capacity bug

bojanserafimov:
Tests were interpretting the `Animal_ParentOf` edge in the wrong direction, which caused a bug in pagination capacity calculation not to be detected.

### DIFF
--- a/graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
+++ b/graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
@@ -8,7 +8,7 @@ from ...cost_estimation.filter_selectivity_utils import Selectivity
 from ...cost_estimation.interval import Interval
 from ...cost_estimation.statistics import LocalStatistics
 from ...global_utils import QueryStringWithParameters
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/snapshot_tests/test_cost_estimation_analysis.py
```diff
@@ -412,15 +412,51 @@ def test_distinct_result_set_estimates_with_revisit_counts(self):
...
+        # estimate at the root is 1. However, since that animal has at most one parent,
+        # the distinct result set estimate at that location is 1 as well, even though
+        # there are no explicit filters on it.
         query = QueryStringWithParameters(
             """{
             Animal {
                 name @filter(op_name: "=", value: ["$animal_name"])
                 in_Animal_ParentOf @optional {
-                    name @output(out_name: "child_name")
+                    name @output(out_name: "parent_name")
```
bojanserafimov, line 431: You can see that the test was using `in_Animal_ParentOf`, which is the opposite of `ParentOf`, meaning `Parent`, but calling the destination a child. That was the source of an incorrect test. I recommend looking at the raw file while reviewing, github didn't do a good job diffing.

PR #953, opened 2020-10-15
Integration tests for Postgres

chewselene:
Turns on integration test for the Postgres backend. Uncovers a [bug](https://github.com/kensho-technologies/graphql-compiler/blob/main/graphql_compiler/tests/integration_tests/test_backends_integration.py#L779) in fold that will need to be addressed before landing.


graphql_compiler/tests/integration_tests/test_backends_integration.py
```diff
@@ -150,7 +151,7 @@ def compile_and_run_query(
         if backend_name in SQL_BACKENDS:
             engine = cls.sql_backend_name_to_engine[backend_name]  # type: ignore
             results, output_metadata = compile_and_run_sql_query(
-                cls.sql_schema_info, graphql_query, parameters, engine  # type: ignore
+                cls.sql_schema_info[backend_name], graphql_query, parameters, engine  # type: ignore
```
obi1kenobi, line 154: Is there something we can do about the `type: ignore` here?

graphql_compiler/tests/test_data_tools/data_tool.py
```diff
@@ -283,33 +284,72 @@ def init_sql_integration_test_backends():
...
+                engine.execution_options(isolation_level="AUTOCOMMIT").execute(
+                    CreateSchema("schema_2")
+                )
+                engine.execution_options(isolation_level="AUTOCOMMIT").execute(
+                    CreateSchema("schema_3")
+                )
+                engine.execution_options(isolation_level="AUTOCOMMIT").execute(
+                    CreateSchema("schema_4")
+                )
+
```
obi1kenobi, line 316:
```suggestion
```

graphql_compiler/tests/integration_tests/test_backends_integration.py
chewselene: The error here is `error: "Type[IntegrationTests]" has no attribute "sql_schema_info". We set up per backend attributes in `conftest.py`. Throughout this file we `# type: ignore` on all instances where we're accessing a backend specific attribute to get pylint to pass. Is there a better way to be doing this?

obi1kenobi: Ah okay, so it's a pytest fixture issue. We don't yet have a better way to do this, other than maybe making the `# type: ignore` specific to the error code we expect here (currently it ignores _all_ errors on this line). Not worth fixing in this PR, probably.

obi1kenobi (approved):

PR #955, opened 2020-10-15
Add docstring for project_neighbors().

obi1kenobi:
Let me know if I adequately explained "generators are not closures" and the associated implications for how this function should be implemented to avoid bugs.

### DIFF
--- a/graphql_compiler/interpreter/typedefs.py
+++ b/graphql_compiler/interpreter/typedefs.py
@@ -398,35 +398,79 @@ class InterpreterAdapter(Generic[DataToken], metaclass=ABCMeta):
 
         ## Common bug to avoid in your implementation
 
-        Importantly, when making the "neighbors" generator in the code example above, we called
-        a function that returns a generator instead of using a generator expression such as:
-            # BUGGY CODE FOLLOWS
-            neighbors = (
-                _your_function_that_makes_a_data_token(neighbor_id)
-                for neighbor_id in _your_function_that_gets_neighbor_ids(
-                    current_type_name, edge_info, current_token
-                )
-            )
-            # END OF BUGGY CODE
-        The above code is the victim of an unfortunate and difficult-to-find bug similar to
-        a race condition in multithreaded code:
-        - The current_token value is set to a new value in each iteration of the function's
-          outer "for data_context in data_contexts" loop.
-        - The generator in the buggy snippet refers to current_token -- but instead of referring
-          to "whatever is the value of current_token at the time the generator is defined,"
-          it instead refers to "whatever value current_token points to at each generator iteration".
-        - Since the outer loop's iterations and the buggy generator's iterations may occur in
-          an arbitrarily-iterleaved order, the buggy generator's current_token value changes
-          unpredictably during its execution therefore causing the bug.
-        - Since Python generators are not started until their first result is requested, the above
-          buggy snippet can't even guarantee that the first use of current_token has the intended
-          value -- the outer loop might have iterated before the generator was first invoked.
+        In the previous code example, note that we called a module-scoped function,
+        `_your_function_that_gets_neighbors_for_a_given_token`, instead of one defined in the scope
+        of `project_neighbors`.  Because Python evaluates references to variables in outer scopes
+        at the time a function or generator is invoked--not at the time it's defined--, it's very
+        easy to introduce subtle race conditions when defining generator factories in a nested
+        scope.
 
-        This bug is avoided (as in the working example code) by calling a function and supplying
-        the value of current_token to it. This creates a new Python scope (a so-called "closure")
-        that remembers the value current_token had when the closure was created. A succinct,
-        internet-search-friendly way to describe the cause of this bug is to say that in Python,
-        generators are not closures.
+        Because generators may be evaluated in arbitrary order, these bugs can appear only
+        intermittently and can be very difficult to troubleshoot.  Always defining generator
+        factories in the module scope is one reliable way to avoid this problem.
+
+        In this example code, we use a for-loop to yield several generators from a generator
+        factory.  Notice that we don't pass any arguments to the generator factory--the value its
+        generators yield come from its enclosing scope.
+
+        >>> def yield_generators():
+        ...     for target in range(1, 4):
+        ...         def _generator_factory():
+        ...             while True:
+        ...                 # refers to `target` in the enclosing scope
+        ...                 yield target
+        ...         yield _generator_factory()
+        ...
+        >>> gens = yield_generators()
+        >>> one = next(gens)
+        >>> next(one)
+        1
+        >>> two = next(gens)
+        >>> next(two)
+        2
+        >>> next(one)  # We expect 1, but get 2
+        2
+        >>> three = next(gens)
+        >>> next(three)
+        3
+        >>> next(two)  # We expect 2, but get 3
+        3
+        >>> next(one)  # We expect 1, got 2, and now get 3
+        3
+
+        Although we have three distinct generators, they're all yielding the same `target` from
+        `yield_generator`'s scope, which is also the same `target` that the for-loop advances with
+        each iteration.
+
+        If we define `_generator_factory` in the scope of the module, then we can't refer
+        inadvertently to shared state in an enclosing scope, which saves us from this bug.
+
+        >>> def _generator_factory(target):
+        ...     while True:
+        ...         # refers to the argument `target`, which exists
+        ...         # only in the local scope
+        ...         yield target
+        ...
+        >>> def yield_generators():
+        ...     for target in range(1, 4):
+        ...         yield _generator_factory(target)
+        ...
+        >>> gens = yield_generators()
+        >>> one = next(gens)
+        >>> next(one)
+        1
+        >>> two = next(gens)
+        >>> next(two)
+        2
+        >>> next(one)
+        1
+        >>> three = next(gens)
+        >>> next(three)
+        3
+        >>> next(two)
+        2
+        >>> next(one)
+        1
 
         ## Hints supplied to this function refer to neighboring vertices
 
--- a/graphql_compiler/interpreter/typedefs.py
+++ b/graphql_compiler/interpreter/typedefs.py
@@ -298,6 +298,8 @@ class InterpreterAdapter(Generic[DataToken], metaclass=ABCMeta):
                            property data needs to be loaded
             current_type_name: name of the vertex type whose property needs to be loaded. Guaranteed
                                to be the name of a type defined in the schema being queried.
+                               (current_type_name names the concrete type of DataToken contained in
+                               the DataContext containers yielded by the data_context iterator.)
             field_name: name of the property whose data needs to be loaded. Guaranteed to refer
                         either to a property that is defined in the supplied current_type_name
                         in the schema, or to the "__typename" meta field that is valid for all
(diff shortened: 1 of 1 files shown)


graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+            # END OF BUGGY CODE
+        The above code is the victim of an unfortunate and difficult-to-find bug similar to
+        a race condition in multithreaded code:
+        - The current_token value is set to a new value in each iteration of the function's
+          outer "for data_context in data_contexts" loop.
+        - The generator in the buggy snippet refers to current_token -- but instead of referring
+          to "whatever is the value of current_token at the time the generator is defined,"
+          it instead refers to "whatever value current_token points to at each generator iteration".
+        - Since the outer loop's iterations and the buggy generator's iterations may occur in
+          an arbitrarily-iterleaved order, the buggy generator's current_token value changes
```
chewselene, line 419:
```suggestion
          an arbitrarily-interleaved order, the buggy generator's current_token value changes
```

chewselene (commented):
Tons of information clearly conveyed through this doc string - great job! The `common bug` section is especially nice; the bulleted list really helps break it down. There were a few areas I found confusing so I added some suggestions for clarity.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+                Foo {
+                    out_Foo_Bar {
+                        name @output(out_name: "name")
+                    }
+                }
+            }
+        The project_neighbors() call corresponding to the out_Foo_Bar edge traversal would contain
+        used_property_hints=frozenset({"name"}). This is because project_neighbors() here would
+        be called DataContexts whose current_token would point to Foo vertices, and the neighboring
+        vertices (ones along the outbound Foo_Bar edge) have their "name" property queried.
```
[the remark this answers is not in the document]
branen: I like the rewording!  To improve on the last sentence, how about this?
```
This is because used_property_hints correspond to neighboring vertices,
and the neighboring Bar vertices (along the outbound Foo_Bar edge)
are being queried for their "name" property.
```

graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+
+        Args:
+            data_contexts: iterable of DataContext objects which specify the DataTokens whose
+                           neighboring DataTokens need to be loaded.
+            current_type_name: name of the vertex type whose neighbors need to be loaded. Guaranteed
```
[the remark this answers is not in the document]
branen: FWIW, I prefer the original language because of the explicit relationship to neighbors.  Maybe we could add something about the relationship to DataTokens to the end of the following sentence?

branen (changes requested):
Nice work!  I made a few minor suggestions and a placeholder for one larger one--I think the "Common bug to avoid in your implementation" could use some reworking.  I have some ideas and will write them up in a separate review.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+          unpredictably during its execution therefore causing the bug.
+        - Since Python generators are not started until their first result is requested, the above
+          buggy snippet can't even guarantee that the first use of current_token has the intended
+          value -- the outer loop might have iterated before the generator was first invoked.
+
+        This bug is avoided (as in the working example code) by calling a function and supplying
+        the value of current_token to it. This creates a new Python scope (a so-called "closure")
+        that remembers the value current_token had when the closure was created. A succinct,
+        internet-search-friendly way to describe the cause of this bug is to say that in Python,
+        generators are not closures.
```
branen, line 429:
```suggestion
        ## Common bug to avoid in your implementation
        
        In the previous code example, note that we called a module-scoped
        function, `_your_function_that_gets_neighbors_for_a_given_token`,
        instead of one defined in the scope of `project_neighbors`.  Because
        Python evaluates references to variables in outer scopes at the time a
        function or generator is invoked--not at the time it's defined--, it's
        very easy to introduce subtle race conditions when defining generator
        factories in a nested scope.
        
        Because generators may be evaluated in arbitrary order, these bugs can
        appear only intermittently and can be very difficult to troubleshoot.
        Always defining generator factories in the module scope is one
        reliable way to avoid this problem.
        
        In this example code, we use a for-loop to yield several generators from
        a generator factory.  Notice that we don't pass any arguments to the
        generator factory--the value its generators yield come from its
        enclosing scope.
        
        >>> def yield_generators():
        ...     for target in range(1, 4):
        ...         def _generator_factory():
        ...             while True:
        ...                 # refers to `target` in the enclosing scope
        ...                 yield target
        ...         yield _generator_factory()
        ...
        >>> gens = yield_generators()
        >>> one = next(gens)
        >>> next(one)
        1
        >>> two = next(gens)
        >>> next(two)
        2
        >>> next(one)  # We expect 1, but get 2
        2
        >>> three = next(gens)
        >>> next(three)
        3
        >>> next(two)  # We expect 2, but get 3
        3
        >>> next(one)  # We expect 1, got 2, and now get 3
        3
        
        Although we have three distinct generators, they're all yielding the
        same `target` from `yield_generator`'s scope, which is also the same
        `target` that the for-loop advances with each iteration.
        
        If we define `_generator_factory` in the scope of the module, then we
        can't refer inadvertently to shared state in an enclosing scope, which
        saves us from this bug.
        
        >>> def _generator_factory(target):
        ...     while True:
        ...         # refers to the argument `target`, which exists
        ...         # only in the local scope
        ...         yield target
        ...
        >>> def yield_generators():
        ...     for target in range(1, 4):
        ...         yield _generator_factory(target)
        ...
        >>> gens = yield_generators()
        >>> one = next(gens)
        >>> next(one)
        1
        >>> two = next(gens)
        >>> next(two)
        2
        >>> next(one)
        1
        >>> three = next(gens)
        >>> next(three)
        3
        >>> next(two)
        2
        >>> next(one)
        1
```

In the long run, I think we should drop the example code from the docstring, end the warning after the second paragraph, and include a RTD link for readers who'd like a more in-depth explanation (with example code) than can reasonably fit here.  I'd love to write that documentation, since I think it could be useful well beyond this project, have some neat pedagogical ideas I'd like to try out, and am pretty excited about it.

I've kept the code here short and simple, so as to avoid occluding the cause of the bug or dressing it up in unneeded complexity.  I think it'd be fine to include some example buggy and corrected `project_neighbors` implementations on a future RTD page, but think simplicity is better here.  (I considered foregoing the example code altogether, but on balance, I think it's probably worth keeping for now in its minimal form.)

I've also tried to avoid terminology that might be misleading for some readers, in favor of terminology that refers more specifically to the language rules that underlie this class of bug.

branen (changes requested):
Finally!  Here are my suggested changes for the "Common bug to avoid" documentation.

graphql_compiler/interpreter/typedefs.py
LWprogramming: I personally like the code example in this new version! If we do include a link to RTD at some point, would the explanation there mention closures? I'm thinking it would be nice to have that internet-search-friendly terms there since that's where the more detailed explanation would go.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+            data_contexts: iterable of DataContext objects which specify the DataTokens whose
+                           neighboring DataTokens need to be loaded.
+            current_type_name: name of the vertex type whose neighbors need to be loaded. Guaranteed
+                               to be the name of a type defined in the schema being queried.
+            edge_info: direction and name of the edge along which neighboring vertices need to be
+                       loaded. For example, in the query example above, this argument would be set
+                       to ("out", "Foo_Bar").
+            runtime_arg_hints: names and values of any runtime arguments provided to the query
+                               for use in filtering operations (e.g. "$arg_name").
+            used_property_hints: the property names of the neighboring vertices being loaded that
```
LWprogramming, line 459:
```suggestion
            used_property_hints: property names of the neighboring vertices being loaded that are
```

Nit: argument descriptions don't have leading articles. Multi-line suggestions don't seem to be working for me but this suggestion would also mean removing the leading word "are" from the line below.

graphql_compiler/interpreter/typedefs.py
branen: I'm not opposed to considering SEO for some future document, so long as it doesn't diminish the document's pedagogical value.

LWprogramming: Ah, I was primarily thinking about internet search in the other direction, as in if someone comes across the documentation, wonders if there are technical terms to search up that describe what we see here-- something like what currently is here:

`A succinct, internet-search-friendly way to describe the cause of this bug is to say that in Python, generators are not closures.`

branen: Ah, gotcha.  In general, I'm not a fan of this approach, since the result set for any phrase will vary by search engine, client IP address, locale, cookies, and time, and we run the risk of sending users in unhelpful or misleading directions.  I think it's better for us and better for our readers if we link to specific resources.

graphql_compiler/interpreter/typedefs.py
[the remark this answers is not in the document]
branen: Queued for commit with my suggested change.

graphql_compiler/interpreter/typedefs.py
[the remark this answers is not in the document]
branen: Kept the original language and staged for commit with this parenthetical:

> (current_type_name names the concrete type of DataToken contained in the DataContext containers yielded by the data_context iterator.)

graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+        a "bookkeeping" object called DataContext. These DataContext objects allow
+        the interpreter to keep track of "which data came from where"; only the DataToken value
+        bound to each current_token attribute is relevant to the InterpreterAdapter API.
+
+        Having obtained an iterable of DataTokens and converted it to an iterable of DataContexts,
+        the interpreter needs to find the neighbors along the outbound Foo_Bar edge for each of
+        those DataTokens. To do so, the intepreter calls project_neighbors() with the iterable
+        of DataContexts, setting current_type_name = "Foo" and edge_info = ("out", "Foo_Bar"),
+        requesting an iterable of DataTokens representing the neighboring vertices for each
+        DataContext with its corresponding current_token. If the DataContext's current_token
```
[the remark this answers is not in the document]
branen: Staged for commit.

graphql_compiler/interpreter/typedefs.py
branen: Staged for commit.

graphql_compiler/interpreter/typedefs.py
branen: Staged for commit.

graphql_compiler/interpreter/typedefs.py
```diff
@@ -338,12 +338,140 @@ def project_neighbors(
...
+        that remembers the value current_token had when the closure was created. A succinct,
+        internet-search-friendly way to describe the cause of this bug is to say that in Python,
+        generators are not closures.
+
+        ## Hints supplied to this function refer to neighboring vertices
+
+        Hint kwargs such as used_property_hints, filter_hints, and neighbor_hints in this function
+        describe the desired structure of the *neighboring* vertices that this function produces,
+        instead of to the vertices supplied via the data_contexts input. For example, consider
+        the following query:
```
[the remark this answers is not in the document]
branen: Staged for commit.

graphql_compiler/interpreter/typedefs.py
branen: Staged for commit.

branen:
Ah, shoot, Leon also made a suggestion, and I forgot to include his Co-authored-by in the commit message.  We should fix that in the squash.

```
-            used_property_hints: the property names of the neighboring vertices being loaded that
+            used_property_hints: property names of the neighboring vertices being loaded that
```

branen (approved):

branen:
Barring objection, I'll merge this (with credit to Leon, as above) Thursday morning.

branen:
Squash-merged in eec75c09b1fd0b2a89cc221fa7c7ab78c3b28810.

PR #957, opened 2020-10-16
Add ability to filter then traverse inside a fold for SQL backend

chewselene:
We previously only allowed filtering at the output location when inside a fold in the SQL backend, but did not have an appropriate error message if a user tried to filter at a different location within a fold. This PR adds functionality and tests for queries that involve a fold, filter, traversal, and output (in that order). See example query in `test_input_data`.


obi1kenobi (commented):
Looks great modulo the one formatting nitpick.

PR #958, opened 2020-10-21
Use LEFT OUTER JOIN for Postgres folds

### LINKED ISSUE #954: Postgres back produces incorrect results for empty fold lists
Turning on integrations tests in #953 uncovered a bug. This [db-fiddle](https://www.db-fiddle.com/f/pwwFd8dy46TagLaLjb9RN/0) is modeled after the current output of this [integration test](https://github.com/kensho-technologies/graphql-compiler/blob/main/graphql_compiler/tests/integration_tests/test_backends_integration.py#L764). For `Animal 4`, we expect to see the result `{child_names: []}`, but instead get no results for `Animal 4`.

chewselene:
Postgres and MSSQL require different types of `JOIN` for joining the folded subquery. 

For Postgres, a `LEFT OUTER JOIN` is needed so that every row in the outer table has at least 1 result returned. Some of these results can be `null` which will then be coalesced to `[]` in the top `SELECT` statement.

For MSSQL, a `JOIN` (inner join) is needed because every row in the outer table is guaranteed to have exactly 1 result in the folded subquery. Coalescing to `''` (the empty array equivalent for MSSQL) occurs within the folded subquery itself.

The PR updates the sqlalchemy join function to use the appropriate value for `isouter` depending on dialect.

Resolves #954

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1158,6 +1158,15 @@ class CompilationState(object):
                 f"Attempted to unfold while the _current_alias was None during fold {self}."
             )
         outer_vertex_primary_key_name = self._get_current_primary_key_name("@fold")
+
+        # Postgres uses a LEFT OUTER JOIN and coalesces nulls to an empty array in the top SELECT.
+        if isinstance(self._sql_schema_info.dialect, PGDialect):
+            is_left_outer_join = True
+        # MSSQL folded subquery returns exactly 1 folded result for each row in the outer table
+        # so should use an INNER JOIN.
+        else:
+            is_left_outer_join = False
+
         self._from_clause = sqlalchemy.join(
             self._from_clause,
             fold_subquery_alias,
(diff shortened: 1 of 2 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -1158,14 +1158,23 @@ def unfold(self) -> None:
...
             )
         outer_vertex_primary_key_name = self._get_current_primary_key_name("@fold")
+
+        # Postgres uses a LEFT OUTER JOIN and coalesces nulls to an empty array in the top SELECT.
+        if isinstance(self._sql_schema_info.dialect, PGDialect):
+            is_left_outer_join = True
+        # MSSQL folded subquery returns exactly 1 folded result for each row in the outer table
+        # so should use an INNER JOIN.
+        else:
+            is_left_outer_join = False
```
obi1kenobi, line 1168: I think most other SQL flavors are likely to be similar to PostgreSQL and not to MSSQL here, so I'd either make the conditional be `if-elif-else raise AssertionError()` or make the `else` case be PostgreSQL + everything else, with MSSQL the "special" if case.

chewselene: Good point! I went with a `NotImplementedError` because that's what we have in other places where we're checking the dialect.

obi1kenobi (approved):

PR #959, opened 2020-10-22
Field renaming [WIP]

LWprogramming:
I'd advise _against_ trying to read this, especially not from start to finish. While getting field renaming functionality to its current state, I noticed ways to break this down into smaller pieces, so I'd recommend only reading those pieces (which will be marked as non-draft).

----

This PR does two things:

* ensure my existing work on field renamings doesn't exist only on my computer

* let me break the PR down into parts more easily-- I find Github's PR diff viewer to be especially helpful for larger PRs, so I'm putting this out as a draft

----

It contains:

* All the test cases I've come up with so far, going through quite a number of edge cases.

* The implementation itself

* Updated docs, comments, etc

* A number of TODOs marked `TODO(Leon)` for things that I'd be open to feedback on, and I'll be describing them in more detail in the smaller PRs coming soon.


LWprogramming:
PRs #970 and 971 have been merged, so this PR is no longer needed.

PR #960, opened 2020-10-23
Fix recurse bug

bojanserafimov:
There was a bug in the code that chooses which columns to select from the base CTE of a recursion that caused the compiler to raise `KeyError`.

Line 43 (see red code) uses the `edge` variable, which is is garbage from another scope. The solution is to remove the line, and the entire block surrounding it (see TODO in the red code).

I don't add any regression tests since this is a very silly thing to regression test. Finding a query that fails is hard, but if I find one I might add it since it doesn't hurt. I've manually verified that the query that caused this problem now works.

To prevent more bugs in this code, the best thing to do is explained in the TODO at the top of the function. In fact, the implementation of that TODO would have prevented this bug.

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -1158,6 +1158,20 @@ class CompilationState(object):
                 f"Attempted to unfold while the _current_alias was None during fold {self}."
             )
         outer_vertex_primary_key_name = self._get_current_primary_key_name("@fold")
+
+        # Postgres uses a LEFT OUTER JOIN and coalesces nulls to an empty array in the top SELECT.
+        if isinstance(self._sql_schema_info.dialect, PGDialect):
+            is_left_outer_join = True
+        # MSSQL folded subquery returns exactly 1 folded result for each row in the outer table
+        # so should use an INNER JOIN.
+        elif isinstance(self._sql_schema_info.dialect, MSDialect):
+            is_left_outer_join = False
+        else:
+            raise NotImplementedError(
+                "Fold only supported for MSSQL and PostgreSQL, "
+                f"dialect set to {self._sql_schema_info.dialect.name}."
+            )
+
         self._from_clause = sqlalchemy.join(
             self._from_clause,
             fold_subquery_alias,
@@ -1165,7 +1179,7 @@ class CompilationState(object):
                 self._current_alias.c[outer_vertex_primary_key_name]
                 == fold_subquery_alias.c[outer_vertex_primary_key_name]
             ),
-            isouter=False,
+            isouter=is_left_outer_join,
         )
 
         # 5. Clear the fold from the compilation state.
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -117,6 +117,9 @@ def _find_used_columns(
             # A recurse implies an outgoing foreign key usage
             child_location_info = ir.query_metadata_table.get_location_info(child_location)
             if child_location_info.recursive_scopes_depth > location_info.recursive_scopes_depth:
+                used_columns.setdefault(get_vertex_path(location), set()).add(
+                    edge.to_column
+                )
                 used_columns.setdefault(get_vertex_path(child_location), set()).add(
                     edge.from_column
                 )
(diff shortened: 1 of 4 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -117,6 +117,9 @@ def _find_used_columns(
             # A recurse implies an outgoing foreign key usage
             child_location_info = ir.query_metadata_table.get_location_info(child_location)
             if child_location_info.recursive_scopes_depth > location_info.recursive_scopes_depth:
+                used_columns.setdefault(get_vertex_path(location), set()).add(
+                    edge.to_column
+                )
```
bojanserafimov, line 122: The PR is still draft because I want to spend more time thinking if this is sufficient. Maybe I should explicitly add the primary key as a used column?

obi1kenobi (changes requested):
How difficult would it be to find a repro test case that we can put in the repo for this?

I recognize we have a real bug that we need to fix, but this code is really tricky and therefore we're likely to have more bugs here — especially until we get a chance to address that large TODO. The "one-off, no tests, no meaningful test coverage" here feels like it will cause even more problems down the line than it'll solve right now.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -105,6 +105,14 @@ def _find_used_columns(
 
     # Find foreign keys used
     for location, location_info in ir.query_metadata_table.registered_locations:
+
+        primary_key = None
+        primary_keys = (
+            sql_schema_info.vertex_name_to_table[location_info.type.name].alias().primary_key
+        )
+        if len(primary_keys) > 1:
+            primary_key = str(primary_keys[0].name)
```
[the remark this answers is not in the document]
bojanserafimov: It's wrong, supposed to be `== 1`. It's hard to find a test case for the bug, but we should have a test case that covers this branch at least. Will add one.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -103,8 +103,15 @@ def _find_used_columns(
...
 
-    # Find foreign keys used
+    # Find columns used to emit edges
     for location, location_info in ir.query_metadata_table.registered_locations:
+        primary_key = None
+        primary_keys = (
+            sql_schema_info.vertex_name_to_table[location_info.type.name].alias().primary_key
+        )
+        if len(primary_keys) == 1:
+            primary_key = str(primary_keys[0].name)
```
obi1kenobi, line 113: Is fall-through on the other cases expected / correct behavior, or is this just a hack?

Maybe add `elif len(primary_keys) > 1: raise NotImplementedError()` and perhaps raise `AssertionError` for unreachable code in the `else`? Let's be explicit about what we expect here, it's not at all obvious to me why `== 1` is a special case and we already had a couple of bugs around this code so that's making me feel extra uncomfortable with the current setup.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -46,6 +46,36 @@
...
+            f"This information is required to emit a {directive_name} directive."
+        )
+    if len(primary_keys) > 1:
+        raise NotImplementedError(
+            f"The table for vertex {self._current_classname} has a composite primary key "
+            f"{primary_keys}. The SQL backend does not support "
+            f"{directive_name} on tables with composite primary keys."
+        )
+
+    return str(primary_keys[0].name)
```
bojanserafimov, line 76: this code is copied from below

graphql_compiler/compiler/emit_sql.py
```diff
@@ -46,6 +46,36 @@
...
+                        primary key to exist. Used in error messages only.
+
+    Returns:
+        name of the single-column primary key
+    """
+    primary_keys = alias.primary_key
+
+    if not primary_keys:
+        raise AssertionError(
+            f"The table for vertex {self._current_classname} has no primary key specified. "
```
bojanserafimov, line 66: oops, self is out of scope

graphql_compiler/compiler/emit_sql.py
```diff
@@ -46,6 +46,38 @@
...
+def _get_primary_key_name(alias: Alias, vertex_type_name: str, directive_name: str) -> str:
+    """Return the name of the single-column primary key for the alias.
+
+    If there is no single-column primary key for this alias, an error is raised.
+
+    Args:
+        alias: the sqlalchemy object with primary key information
+        vertex_type_name: The vertex name that represents the alias. Used for error messages
+                          only.
+        directive_name: name of the directive that requires for the single-column
```
obi1kenobi, line 58:
```suggestion
        directive_name: name of the directive that requires the single-column
```

obi1kenobi (commented):
Looks good other than the one small nit.

obi1kenobi (approved):

PR #961, opened 2020-10-23
Copyright line

LWprogramming:
Noticed this come up during my lint, although it didn't make the linting actually fail.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -199,7 +199,7 @@ fi
 if  [[
         (
             ("$run_fast_linters" == 1) && (
-                ("copyright_line_check_exit_code" != "0") ||
+                ("$copyright_line_check_exit_code" != "0") ||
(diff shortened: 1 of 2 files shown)


scripts/lint.sh
```diff
@@ -199,6 +199,7 @@ fi
 if  [[
         (
             ("$run_fast_linters" == 1) && (
+                ("copyright_line_check_exit_code" != "0") ||
```
obi1kenobi, line 202:
```suggestion
                ("$copyright_line_check_exit_code" != "0") ||
```

obi1kenobi: Pretty sure this is why the tests failed.

obi1kenobi (commented):
Nice catch! I think there's a small bug in the `lint.sh` script but otherwise this is great.

obi1kenobi (approved):

PR #962, opened 2020-10-23
Refactor renamings argument

LWprogramming:
We'll want to rename fields soon #959, so it should be clear that the existing `renamings` refer to renamings for types.


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -56,6 +49,13 @@
 its fields) removed. If "Dog" also appeared as a field in the schema's root type, it would be
 removed there as well.
 
+For both 1-1 or 1-many renaming, renamings only apply to types, fields, and enum values that exist
```
LWprogramming, line 52: This gets moved because right now the comment's structure goes 
- 1-1 renaming description
- 1-many renaming description
- This paragraph
- 1-0 renaming description
- other information including implementation details

This PR moves the paragraph to go with existing implementation details, especially since #959 introduces more of these details.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -76,14 +76,15 @@
...
-    - A string type_name may be in renamings only if there exists a type in the schema named
-      type_name (since otherwise that entry would not affect any type in the schema).
-    - A string type_name is in renamings, then renamings[type_name] != type_name (since applying the
-      renaming would not change the type named type_name).
-  If not iterable, then these no-op rules don't apply.
+  being renamed or suppressed. The same rules apply for fields that belong to the same type, since
+  they share a namespace as well.
+- There exist special rules for iterable renamings (e.g. if renamings are represented as a
+  dictionary)-- specifically, no-op renamings are not allowed.
+  - If type_renamings argument is iterable:
```
obi1kenobi, line 83:
```suggestion
- When the renaming argument is an iterable value (e.g., a dict), a few additional rules apply:
```
Then de-indent the below bullet points. This says the same thing more succinctly and with one fewer level of indentation.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -137,19 +138,21 @@
...
+    schema_ast: DocumentNode, type_renamings: TypeRenamingMapping
+) -> RenamedSchemaDescriptor:
+    """Create a RenamedSchemaDescriptor; rename/suppress types and root type fields.
 
-    Any type, interface, enum, or fields of the root type/query type whose name
-    appears in renamings will be renamed to the corresponding value if the value is not None. If the
-    value is None, it will be suppressed in the renamed schema and queries will not be able to
-    access it.
+    Any object type, interface type, enum type, or field of the root type/query type has a name. Let
+    the name be called type_name. If type_renamings.get(type_name, _) is not None, the type or field
```
obi1kenobi, line 153: Confused about the `_` in `type_renamings.get(type_name, _)`. Is that supposed to be `None` instead?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -530,49 +533,55 @@ class RenameSchemaTypesVisitor(Visitor):
...
+    # consist only of alphanumeric characters and underscores, do not start with a number, and do
+    # not start with two underscores), invalid_type_names will map "Foo" to the invalid type name.
     invalid_type_names: Dict[str, str]
-    # Collects the type names for types that get suppressed. If renaming would suppress a type named
-    # "Foo", renamed_types will contain "Foo".
-    suppressed_types: Set[str]
+    # Collects the type names for types that get suppressed. If type_renamings would suppress a type
+    # named "Foo", suppressed_type_names will contain "Foo".
+    suppressed_type_names: Set[str]
+    # reverse_field_name_map maps type name to a dict, which in turn maps the name of a field in the
```
obi1kenobi, line 555: This seems new, and doesn't seem like it refers to any particular attribute here.

For readability and to make errors like this easier to spot, consider using new lines to visually separate the comments from the code above them (the code not related to them).

graphql_compiler/schema_transformation/utils.py
```diff
@@ -46,16 +46,62 @@ class SchemaStructureError(SchemaTransformError):
...
+    start with a numeric character (as required by GraphQL), and doesn't start with double
+    underscores as such type names are reserved for GraphQL internal use.
+
+    Not to be confused with InvalidTypeNameError, which is used for schemas that aren't being
+    renamed because this type of error includes information from the renaming process that
+    InvalidTypeNameError doesn't.
+    """
+
+    # TODO(Leon): Would like feedback on this-- is there a better way to do this than have two types
+    # of invalid name errors? See docstring for why I split them up.
```
obi1kenobi, line 74: I actually didn't get what the docstring was saying. Can you maybe give an example to clarify the distinction? "includes information that the other error doesn't" is not the most clear statement.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: The idea here is that either the type named `type_name` gets suppressed or not, and whether or not it gets suppressed doesn't depend on the default return value but rather only on `type_name`.

`type_renamings.get(type_name, type_name)` or `type_renamings.get(type_name, None)` would both give a correct sentence here, despite the former leaving types alone by default and the latter suppressing them by default; in either case, if the return isn't None, then the type/ query type field gets renamed to whatever gets returned.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: This additional layer is actually part of the `renamings` -> `type_renamings` refactor, since when we get field renamings it'd be helpful to distinguish between `type_renamings`, field renamings for a particular type (which could be a `Dict[str, str]`), and field renamings in general (which could be, among other things, a `Dict[str, FieldRenamingsForParticularType]`)

graphql_compiler/schema_transformation/utils.py
LWprogramming: `InvalidTypeNameError` used to have double duty: one when checking that a given schema has no invalid names (which is something that happens in code unrelated to schema renaming), and one to check that schema renaming doesn't produce a schema with invalid names.

My goal here was to make a clear error if schema renaming involves invalid names, which means tracking the problem renamings (perhaps a dict mapping valid names in the original schema to the invalid names in the renamed schema). The problem is, this sort of bookkeeping doesn't make sense when the code raises `InvalidTypeNameError` when just checking a given schema unrelated to schema renaming specifically.

This extra information means that we have two errors that do different things despite having similar names.

pmantica1: I don't think we should add the `SchemaRenameInvalidNameError` error class, specially as part of this PR, for the following reasons:
1. This PR renames `renamings` to `type_renamings` so that we can also add a `field_renamings` 
     field or something akin to that in the future. The addition of this class seems completely unrelated to that goal 
     and PRs should be aiming to do one thing and one thing only.  
2. The addition of this error class is only arguably an improvement to the code and I think we have more 
     important stuff to work on. I understand the goal you are trying to achieve here, but there is a downside to 
     having an excessive amount of error classes, the `InvalidTypeNameError` seemed perfectly fine to me and 
     the distinction between `SchemaRenameInvalidNameError` and `InvalidTypeNameError` is confusing. So 
     even though you might see this as an improvement, it is really not obvious to me that it is.

graphql_compiler/schema_transformation/rename_schema.py
pmantica1: I agree with Leon's choice here.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Ok, makes sense to me-- one clarification:

* This sentence (which occurs in the Github reply) is true: `If type_renamings.get(type_name, None) is not None, then the type or root query field will be renamed to whatever is returned`

* This sentence (which occurs in the comment in the code) is also true: `Any such names that do not appear in renamings will be unchanged`

* The part that I worded confusingly is my original Github reply, because it brings up a scenario that doesn't occur in the code (default `type_renamings` return value being `None`) in order to show that any default return value would make the comment true, even if that return value is `None`. I'll change it back, since it's misleading

graphql_compiler/schema_transformation/utils.py
```diff
@@ -46,16 +46,62 @@ class SchemaStructureError(SchemaTransformError):
...
+            "must not start with double underscores."
+        )
+        sorted_invalid_type_names = sorted(self.invalid_type_names.items())
+        invalid_type_names_message = (
+            f"The following is a list of tuples that describes what needs to be fixed for type "
+            f"renamings. Each tuple is of the form (original_name, invalid_new_name) where "
+            f"original_name is the name in the original schema and invalid_new_name is what "
+            f"original_name would be renamed to: {sorted_invalid_type_names}"
+        )
+        return "\n".join(filter(None, [explanation, invalid_type_names_message]))
```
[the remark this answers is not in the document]
LWprogramming: This is also an artifact from splitting code off the large field renamings draft PR in #959-- when field renamings come into play, we'll also have an invalid names argument for fields, which means that there could be situations where one of the invalid names variables ends up being `None` (e.g. when if the only invalid names are type names, not field names). The filter is for that purpose-- but since right now only type renamings are supported in this PR, it doesn't make immediate sense. Will remove

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -1056,7 +1119,7 @@ def test_field_in_list_still_depends_on_suppressed_type(self) -> None:
             rename_schema(parse(ISS.list_schema), {"Height": None})
 
     def test_rename_using_dict_like_prefixer_class(self) -> None:
-        class PrefixNewDict(RenamingMapping):
+        class PrefixNewDict:
```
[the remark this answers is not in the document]
LWprogramming: Keeping it is technically valid, although removing it shows that `type_renamings` doesn't have to inherit from `TypeRenamingMapping` but only simply implements the `get` method defined in `TypeRenamingMapping`.

LWprogramming:
Ok, it seems I've accidentally done something weird with Git that I'd rather not debug at the moment since it's getting late-- going to close this and open a fresh PR with the changes I want since that'll be faster, and figure out what happened here tomorrow.

PR #965, opened 2020-11-17
Implement and test query printing and parameter binding

bojanserafimov:
Implement query printing and parameter binding for mssql and postgresql. Add unit and integration tests.


graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -64,15 +66,32 @@ def print_sqlalchemy_query_string(query, dialect):
...
+    if isinstance(dialect, supported_dialects):
+        class BindparamCompiler(dialect.statement_compiler):
+            def visit_bindparam(self, bindparam, **kwargs):
+                # A bound parameter with name param is represented as ":param". However,
+                # if the parameter is expanding (list-valued) it is represented as
+                # "([EXPANDING_param])" by default. This is an internal sqlalchemy
+                # representation that is not understood by databases, so we explicitly
+                # make sure to print it as ":param".
+                bindparam.expanding = False
+                return f":{bindparam.key}"
```
bojanserafimov, line 79: This is not copy-paste. Notice the explicit string formatting here.

Calling the super method only works sometimes with postgres, and I can't figure out the pattern of when it works and when it doesn't. When it doesn't, it prints `BINDPARAM_{key}` instead of just `:{key}`

graphql_compiler/tests/integration_tests/integration_test_helpers.py
```diff
@@ -88,10 +90,15 @@ def compile_and_run_sql_query(
     engine: Engine,
 ) -> Tuple[List[Dict[str, Any]], Dict[str, OutputMetadata]]:
     """Compile and run a SQL query against the SQL engine, return result and output metadata."""
-    compilation_result = graphql_to_sql(sql_schema_info, graphql_query, parameters)
+    compilation_result = compile_graphql_to_sql(sql_schema_info, graphql_query)
```
bojanserafimov, line 93: I'm considering implementing the top-level `graphql_to_sql` function like this (print the query, wrap it in `text`, then attach parameters) so that we don't have 2 different execution paths to test.

obi1kenobi: Are there any downsides we should be worried about?

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -64,15 +66,31 @@ def print_sqlalchemy_query_string(query, dialect):
...
+        raise AssertionError(f"Unsupported dialect {dialect}. "
+                             f"Only {supported_dialects} are supported.")
+
+
+def bind_parameters_to_query_string(query, parameters):
+    bound_parameters = [
+        sqlalchemy.bindparam(
+            parameter_name,
+            value=parameter_value,
+            expanding=isinstance(parameter_value, (list, tuple)),
```
obi1kenobi, line 91: Is it feasible to read this information from the parameter metadata instead? The metadata is the canonical place for this info, and I'd hate to see us add support for some new collection type (e.g. `np.array()`) as a parameter input type only to create the wrong `bindparam` here. It feels like it might be painful bug to track down.

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -64,15 +66,31 @@ def print_sqlalchemy_query_string(query, dialect):
...
+            def visit_bindparam(self, bindparam, **kwargs):
+                # A bound parameter with name param is represented as ":param". However,
+                # if the parameter is expanding (list-valued) it is represented as
+                # "([EXPANDING_param])" by default. This is an internal sqlalchemy
+                # representation that is not understood by databases, so we explicitly
+                # make sure to print it as ":param".
+                return f":{bindparam.key}"
+
+        return str(BindparamCompiler(dialect, query).process(query))
+    else:
```
obi1kenobi, line 81: nit: would you mind putting the shorter clause first, for the sake of readability?

graphql_compiler/compiler/sqlalchemy_extensions.py
obi1kenobi: Should we be worried about injection through `bindparam.key` here?

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -64,15 +66,31 @@ def print_sqlalchemy_query_string(query, dialect):
     Returns:
         string that can be ran using sqlalchemy.sql.text(result)
     """
+    supported_dialects = (PGDialect_psycopg2, MSDialect_pyodbc)
+    if isinstance(dialect, supported_dialects):
+        class BindparamCompiler(dialect.statement_compiler):
```
obi1kenobi, line 71: Since `dialect` here is of a dynamically-specified type, would you mind type-hinting this file (perhaps in a separate PR that merges first) and then adding type hints to all these functions? It's possible there's no other way to do this, but I'd rather hit mypy issues when this code is in your brain's cache than sometime down the road when nobody remembers how and why this works :)

graphql_compiler/tests/integration_tests/integration_test_helpers.py
bojanserafimov: The performance downside is negligible.

I'm not aware of any semantic differences, but if there are any, we'd rather know sooner rather than later.

One other difference is that the resulting query would be more opaque, since it wouldn't be a sqlalchemy selectable, but a text query.

graphql_compiler/compiler/sqlalchemy_extensions.py
bojanserafimov: Good catch. I'll figure out something.

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -75,4 +102,18 @@ def visit_bindparam(self, bindparam, **kwargs):
...
+    bound_parameters = [
+        sqlalchemy.bindparam(
+            parameter_name,
+            value=parameter_value,
+            expanding=isinstance(parameter_value, (list, tuple)),
+        )
+        for parameter_name, parameter_value in parameters.items()
+    ]
+
+    return sqlalchemy.text(query).bindparams(*bound_parameters)
```
bojanserafimov, line 119: This function literally parses the query with a regex looking for ":<key>" and then matches the parameters to the keys found. I don't fully understand it, but I hope it's just a validation step, and not a find-and-replace strategy.

graphql_compiler/compiler/sqlalchemy_extensions.py
bojanserafimov: The behaviour of `np.array()` should be the same. I get the sentiment though, and it would be good. I'd like to resolve the `graphql_to_sql` top-level-function comment first to avoid plumbing twice.

graphql_compiler/tests/integration_tests/integration_test_helpers.py
obi1kenobi: I'm not really worried about performance. I'm worried about two things: the likelihood of running into bugs (security or correctness) in the insertion of arguments into a `text()` representation of the query, and the loss of flexibility with the generated query. When we return `Selectable`, the user is able to take that and use it in many different ways: add clauses (e.g. `LIMIT` or `ORDER BY`), plug it in elsewhere without immediately executing the query (e.g. use it to create a new vertex backed by the `Selectable`) etc.

Not having two different execution paths certainly simplifies _our_ lives as maintainers. But I'm not convinced it's the right thing to do from a user-facing perspective. So I'm inclined to leave the flexibility in and take the slight maintainability hit here.

What do you think?

bojanserafimov: No strong opinions. Leaving it as is, and testing both execution paths against each other.

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -54,18 +62,38 @@ def not_contains_operator(collection, element):
...
-def print_sqlalchemy_query_string(query, dialect):
+def print_sqlalchemy_query_string(
+    query: Select, dialect: Union[PGDialect_psycopg2, MSDialect_pyodbc]
+) -> str:
     """Return a string form of the parameterized query.
 
     Args:
         query: sqlalchemy.sql.selectable.Select
-        dialect: sqlalchemy.engine.interfaces.Dialect
+        dialect: a sqlalchemy.engine.interfaces.Dialect. Currently only postgres
```
chewselene, line 72: I'd get rid of the types here since the function is already type hinted.

graphql_compiler/compiler/sqlalchemy_extensions.py
```diff
@@ -75,4 +102,25 @@ def visit_bindparam(self, bindparam, **kwargs):
...
 
-    return str(BindparamCompiler(dialect, query).process(query))
+    return str(BindparamCompiler(printing_dialect, query).process(query))
+
+
+def bind_parameters_to_query_string(
+    query: str, input_metadata: Dict[str, GraphQLType], parameters: Dict[str, Any]
+) -> TextClause:
+    """Assign values to query parameters."""
+
```
obi1kenobi, line 112:
```suggestion
```

obi1kenobi (commented):
Seems fine to me, with the one suggestion. Will let @chewselene take another look and merge if she agrees.

chewselene (approved):

PR #966, opened 2020-11-17
Apply changes from PR #962

LWprogramming:
See comment here:
https://github.com/kensho-technologies/graphql-compiler/pull/962#issuecomment-729261585

---

Brief summary of the changes:

*  Refactor `renamings` argument to `type_renamings` since we'll want to rename fields soon and we'll want to keep the fields separate.

* Remove the redundant error message testing if these checks are very similar to each other.

A design decision (written here for reference): Although I planned to create a new error class for invalid names that arise during schema renaming in #962, I've decided to keep all invalid name errors as a single class rather than splitting it off as originally intended because my original extra-class reasoning wasn't quite right. When I originally made a separate class, I envisioned that the `SchemaRenameInvalidNameError` object would contain data structures describing what parts of the renaming were invalid. I'd planned to keep these data structures to eventually allow for some sort of (semi)automated error correction by being able to inspect the problematic renamings programmatically instead of just outputting them directly in an error message string (see `CascadingSuppressionError` for an example of this in action).

Unfortunately, unlike something like cascading suppressions, there isn't really a natural automatic thing to do for invalid names.  Whereas for cascading suppressions, the natural thing to do is to simply continue suppressing unreachable things until the schema is valid again, there isn't a really good way to guess at what an invalid GraphQL name _should_ be. As a result, the extra data structures wouldn't help, so keeping everything in one class is fine.

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -40,13 +40,6 @@ data.
         secondname: String
     }
 
-For both 1-1 or 1-many renaming, renamings only apply to types, fields, and enum values that exist
-in the original schema. For example, if a schema contains a type named "Foo" and a type named "Bar"
-but not a type named "Baz" and renamings maps "Foo" to "Bar" and "Bar" to "Baz", then the renamed
-schema will contain a type named "Bar" (corresponding to the original schema's type "Foo") and a
-type named "Baz" (corresponding to the original schema's type "Bar"), instead of containing two
-types both named "Baz".
-
 Suppressing part of the schema removes it altogether. For instance, given the following part of a
 schema:
     type Dog {
@@ -55,6 +48,13 @@ schema:
 suppressing "Dog" would produce an otherwise-identical schema but with that type (and therefore all
 its fields) removed. If "Dog" also appeared as a field in the schema's root type, it would be
 removed there as well.
+
+For both 1-1 or 1-many renaming, renamings only apply to types, fields, and enum values that exist
+in the original schema. For example, if a schema contains a type named "Foo" and a type named "Bar"
+but not a type named "Baz" and type_renamings maps "Foo" to "Bar" and "Bar" to "Baz", then the
+renamed schema will contain a type named "Bar" (corresponding to the original schema's type "Foo")
+and a type named "Baz" (corresponding to the original schema's type "Bar"), instead of containing
+two types both named "Baz".
 
 Operations that are already supported:
 - 1-1 renaming of object types, unions, enums, and interfaces.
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -56,6 +49,13 @@
 its fields) removed. If "Dog" also appeared as a field in the schema's root type, it would be
 removed there as well.
 
+For both 1-1 or 1-many renaming, renamings only apply to types, fields, and enum values that exist
```
LWprogramming, line 52: Comment from 962: 

This gets moved because right now the comment's structure goes 
- 1-1 renaming description
- 1-many renaming description
- This paragraph
- 1-0 renaming description
- other information including implementation details

This PR moves the paragraph to go with existing implementation details, especially since #959 introduces more of these details.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -76,14 +76,15 @@
...
-    - A string type_name may be in renamings only if there exists a type in the schema named
-      type_name (since otherwise that entry would not affect any type in the schema).
-    - A string type_name is in renamings, then renamings[type_name] != type_name (since applying the
-      renaming would not change the type named type_name).
-  If not iterable, then these no-op rules don't apply.
+  being renamed or suppressed. The same rules apply for fields that belong to the same type, since
+  they share a namespace as well.
+- There exist special rules for iterable renamings (e.g. if renamings are represented as a
+  dictionary)-- specifically, no-op renamings are not allowed.
+  - If type_renamings argument is iterable:
```
LWprogramming, line 83: Comment from 962:

This additional layer of indentation is actually part of the `renamings` -> `type_renamings` refactor, since when we get field renamings it'd be helpful to distinguish between `type_renamings`, field renamings for a particular type (which could be a `Dict[str, str]`), and field renamings in general (which could be, among other things, a `Dict[str, FieldRenamingsForParticularType]`)

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -137,19 +138,21 @@
...
+    schema_ast: DocumentNode, type_renamings: TypeRenamingMapping
+) -> RenamedSchemaDescriptor:
+    """Create a RenamedSchemaDescriptor; rename/suppress types and root type fields.
 
-    Any type, interface, enum, or fields of the root type/query type whose name
-    appears in renamings will be renamed to the corresponding value if the value is not None. If the
-    value is None, it will be suppressed in the renamed schema and queries will not be able to
-    access it.
+    Any object type, interface type, enum type, or field of the root type/query type has a name. Let
+    the name be called type_name. If type_renamings.get(type_name, type_name) is not None, the type
```
LWprogramming, line 153: 962 comment: 

Needed to revise this `type_renamings` comment because the original comment mentions the idea of a name being "in" `type_renamings`, which only applies if `type_renamings` is iterable. If, instead, we have an object that simply implements a get method that maps strings to strings, it doesn't make sense to speak of anything being "in" the object. 

There was some discussion previously about whether the original paragraph was clearer versus this new explanation, but I find that the new one is more accurate so I kept this change here.

pmantica1 (approved):

pmantica1 (approved):

PR #967, opened 2020-11-19
Refactor names

LWprogramming:
Small refactoring for code that was specific to types, not fields, but wasn't part of the `renamings`-> `type_renamings` transition in #966. After this PR, #959 (which contains the field_renaming code) will be very close to being ready for review-- I have some remaining nits to iron out there, but it's getting close!

### DIFF
--- a/graphql_compiler/schema_transformation/utils.py
+++ b/graphql_compiler/schema_transformation/utils.py
@@ -254,18 +254,18 @@ def check_schema_identifier_is_valid(identifier: str) -> None:
         )
 
 
-def is_valid_unreserved_name(name: str) -> bool:
-    """Check if input is a valid, unreserved GraphQL name.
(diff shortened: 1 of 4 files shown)


graphql_compiler/schema_transformation/utils.py
```diff
@@ -254,18 +254,18 @@ def check_schema_identifier_is_valid(identifier: str) -> None:
         )
 
 
-def type_name_is_valid(name: str) -> bool:
-    """Check if input is a valid, nonreserved GraphQL type name.
+def is_valid_unreserved_name(name: str) -> bool:
+    """Check if input is a valid, unreserved GraphQL name.
```
chewselene, line 258: minor preference: `nonreserved` (maybe with a hyphen?) rather than `unreserved` since unreserved already has a different meaning in english.

pmantica1 (approved):

PR #970, opened 2020-11-24
Field renaming implementation

LWprogramming:
The first half of field renamings as outlined in draft PR #959. This PR is the implementation for field renamings, while the following PR (#971) contains all the tests specific to field renaming, so that the very large amount of code in #959 gets split into more manageable pieces.

* `rename_schema.py` contains the code to actually rename the fields, as well as keep track of invalid states that may arise from renaming to report in errors.

* `utils.py` contains some additions to the error handling for field renamings, using the data structures created in `rename_schema.py`.

* Almost every change in the tests in this PR are just to specify no field renamings in existing tests (in other words, test semantics stay the same in this PR).


graphql_compiler/schema_transformation/utils.py
```diff
@@ -157,27 +202,93 @@ class NoOpRenamingError(SchemaTransformError):
       the schema named type_name
     * type_renamings is iterable and maps a string type_name to itself, i.e.
       type_renamings[type_name] == type_name
+    * There exists an object type T named type_name in the schema such that
+      field_renamings[type_name] is both iterable and contains a string field_name but there doesn't
+      exist a field named field_name belonging to T in the schema.
+    * field_renamings is iterable and contains a string [type_name] but there doesn't exist an
```
chewselene, line 208: why is `type_name` in brackets here?

graphql_compiler/schema_transformation/utils.py
```diff
@@ -157,27 +202,93 @@ class NoOpRenamingError(SchemaTransformError):
...
+        no_op_nonexistent_type_field_renames: Set[str],
+    ) -> None:
         """Record all no-op renamings."""
-        if not no_op_type_renames:
+        # TODO(Leon): Would like feedback on this-- I notice a pattern where I create error message
+        # objects that must have at least one of these parameters as non-empty. Is there a better
+        # way to ensure this than manually writing this verification (which needs to change as the
+        # parameters change)? This also comes up when deciding whether or not to raise these errors
+        # as well-- for instance, checking if any of a number of error-collecting data structures
+        # are nonempty, and raising an error if any are.
```
chewselene, line 231: only makes it slightly better, but maybe using `not any([...])` where `[...]` is the list of things to check?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -291,12 +383,19 @@ def _ensure_no_cascading_type_suppressions(
...
-                "etc. requiring suppression so you may need to iterate on this before getting a "
-                "legal schema."
+                "A schema containing a union with no members is invalid. To fix this, suppress the "
+                "unions using the type_renamings argument of rename_schema."
             )
+        if visitor.types_to_suppress:
+            error_message_components.append(
+                f"The following types have no non-suppressed fields, which is invalid: "
+                f"{sorted(visitor.types_to_suppress)}. To fix this, suppress the types using the "
+                f"type_renamings argument of rename_schema."
```
pmantica1, line 393: Have in mind that there might be an UI built on top of this function. So python arguments might not be a thing for the user. I think that error messages can be improved in the future and this is fine enough for now. So just have this in mind.

pmantica1 (commented):
First round of comments. (I didn't have a chance to fully review the PR in one sitting).

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -129,6 +165,10 @@
         "schema",  # GraphQLSchema, representing the same schema as schema_ast
         "reverse_name_map",  # Dict[str, str], renamed type/query type field name to original name
         # reverse_name_map only contains names that were changed
+        "reverse_field_name_map",  # Dict[str, Dict[str, str]], mappings of the form
+        # type_name: reverse_field_name_mapping, where type_name is the name of a type in the
+        # original schema and reverse_field_name_mapping is a dict mapping renamed field names to
+        # original name for the given type. Contains only field names that were changed
```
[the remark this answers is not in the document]
LWprogramming: Going to slightly modify the wording on here to specify it's an field name to field name mapping, so no field nodes. Otherwise looks good to me!

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -144,18 +184,45 @@ def get(self, key: str, default: Optional[str]) -> Optional[str]:
         ...
 
 
+# Unfortunately this can't be a nested class because mypy has a bug and it's safer to make
+# the classes slightly less organized in exchange for being able to use mypy here.
+# https://github.com/python/mypy/issues/6393
+class FieldRenamingsForParticularType(Protocol):
+    def get(self, field_name: str, default: Set[str]) -> Set[str]:
+        """Define mapping for renaming fields belonging to a particular type."""
+        # TODO(Leon): I'd like this to work with dicts, but that would require the default argument
```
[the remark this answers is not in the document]
LWprogramming: Dictionaries are the most straightforward way to write a single type renaming and it's how most of the tests perform renamings, so I'd like to do the same with field renamings-- specifically, allowing `field_renamings` to be of the type `Dict[str, Dict[str, Set[str]]]` which allows for creating renamings with only data structures already built into Python.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -144,18 +184,45 @@ def get(self, key: str, default: Optional[str]) -> Optional[str]:
...
+        #  and the return to be Optional[Set[str]]. However, it's natural to represent suppressions
+        #  as the empty set, so actually changing the type hint here to be Optional would be
+        #  misleading.
+        ...
```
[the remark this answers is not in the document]
LWprogramming: Here, they do the same thing, but Pedro pointed out that we can drop them entirely so I've removed them.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -167,13 +234,16 @@ def rename_schema(
         type_renamings: maps original type name to renamed name or None (for type suppression). A
                         type named "Foo" will be unchanged iff type_renamings does not map "Foo" to
                         anything, i.e. type_renamings.get("Foo", "Foo") returns "Foo"
+        field_renamings: maps type names to the renamings for its fields. The renamings map field
+                         names belonging to the type to a set of field names for the renamed
+                         schema
```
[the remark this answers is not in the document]
LWprogramming: Hmm, I can see why it'd be a bit tricky. The original comment avoids using the word "dictionary" because the renamings could also just be an object implementing the `get` method. I also tried separating the type from its name `type_name`, so I'm not quite sure how to approach changing this one on either front without losing accuracy.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -278,9 +371,8 @@ def _ensure_no_cascading_type_suppressions(
                     )
                 )
             error_message_components.append(
-                "A schema containing a field that is of a nonexistent type is invalid. When field "
-                "suppression is supported, you can fix this problem by suppressing the fields "
-                "shown above."
+                "A schema containing a field that is of a nonexistent type is invalid. To fix "
+                "this, suppress the fields using the field_renamings argument of rename_schema."
```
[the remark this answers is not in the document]
LWprogramming: Yep! The description is just above but not shown in the diff because it wasn't changed-- I'll change this error message slightly to clarify which fields should be changed.

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Hmm, I don't quite follow-- what arguments are we talking about, and why might they be unavailable for UIs?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -567,9 +719,47 @@ class RenameSchemaTypesVisitor(Visitor):
...
+
+    # Collects no-op renamings for fields, mapping the type name that contains the field to the set
+    # of field names for which field_renamings contained no-op renamings. Applies only to types for
+    # which field_renamings[type_name] is iterable. If field_renamings would rename a field named
+    # "foo" to "foo" (in a type named "Bar"), or if it attempts to rename a field named "foo" (in a
+    # type named "Bar") when such a field does not exist, no_op_field_renamings will map "Bar" to a
+    # set containing "foo".
+    no_op_field_renamings: DefaultDict[str, Set[str]]
+
+    # Collects type names for each object type that has field renamings and had them applied. After
```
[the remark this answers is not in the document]
LWprogramming: A concrete example: 
* Let the schema have a single type `Foo`
* Let `field_renamings` be a dict with field renamings for `Foo` and `Bar`

Then `types_with_field_renamings_processed` should contain the string `Foo` because its field renamings were applied. Then, when detecting no-op renamings, we have `set(field_renamings) - visitor.types_with_field_renamings_processed` which is `{"Bar"}`.

graphql_compiler/schema_transformation/rename_schema.py
pmantica1: Imagine someone built a graphical UI on top of this function. The UI would ultimately use this function underneath to rename the schema and report any errors. When the user sees `some_python_argument` in the error message they might not be know what the error message refers to since the UI from the perspective is purely graphical. So `some_python_argument` is not even a thing from their perspective. Here an example of `some_python_argument` would be `type_renamings`. In this case, I think that the user of the graphical UI would still be able to get what's wrong based on the error message. I think that this holds true for most of your error messages, so I don't think that it's worth spending time on this. But I just wanted to point this out so that you have this in mind in the future.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -141,21 +181,41 @@
 class TypeRenamingMapping(Protocol):
     def get(self, key: str, default: Optional[str]) -> Optional[str]:
         """Define mapping for type_renamings object."""
-        ...
+
+
+# Unfortunately this can't be a nested class because mypy has a bug and it's safer to make
+# the classes slightly less organized in exchange for being able to use mypy here.
+# https://github.com/python/mypy/issues/6393
+class FieldRenamingsForType(Protocol):
```
LWprogramming, line 189: I'd like this to work with dicts, but that would require the default argument and the return to be Optional[Set[str]]. However, it's natural to represent suppressions as the empty set, so actually changing the type hint here to be Optional would be misleading.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -567,9 +717,44 @@ class RenameSchemaTypesVisitor(Visitor):
...
+    # (valid, non-reserved names consist only of alphanumeric characters and underscores, do not
+    # start with a number, and do not start with two underscores), invalid_field_names will map
+    # "Bar" to a dict that maps "foo" to the invalid field name.
+    invalid_field_names: DefaultDict[str, Dict[str, str]]
+
+    # Collects naming conflict errors involving fields. If field_renamings would rename multiple
+    # fields (in a type named "Bar" in the original schema) to "foo", field_name_conflicts will map
+    # "Bar" to a dict that maps "foo" to a set containing the names of the fields in the original
+    # schema that would be renamed to "foo".
+    field_name_conflicts: Dict[str, Dict[str, Set[str]]]
```
LWprogramming, line 752: Would welcome suggestions on how to better organize the data structures in here, since this visitor class has picked up quite a few fields just because of the various types
of errors that might arise during renaming and needing to keep track of it all.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -673,13 +867,82 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
...
+            node_with_new_name = get_copy_of_node_with_new_name(
+                fields_renamed_node, desired_type_name
+            )
             return node_with_new_name
 
+    def _rename_fields(self, node: ObjectTypeDefinitionNode) -> ObjectTypeDefinitionNode:
+        """Rename node's fields, if applicable and return new node with updated fields."""
+        type_name = node.name.value
+        current_type_field_renamings = self.field_renamings.get(type_name, None)
+        if current_type_field_renamings is None:
```
LWprogramming, line 892: Since current_type_field_renamings need not be a dict, I allow for it to be None if we don't want to do any field renamings for the current type. However, that means it'd technically be valid for it to be an empty dict and I'm not sure how to best check for this, especially since it's not the falsiness of current_type_field_renamings that we're looking for but rather whether it contains any field renamings.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -157,27 +197,81 @@ class NoOpRenamingError(SchemaTransformError):
...
-    def __init__(self, no_op_type_renames: Set[str]) -> None:
+    def __init__(
+        self,
+        no_op_type_renames: Set[str],
+        no_op_field_renames: Dict[str, Set[str]],
+        no_op_nonexistent_type_field_renames: Set[str],
+    ) -> None:
         """Record all no-op renamings."""
-        if not no_op_type_renames:
+        if not any([no_op_type_renames, no_op_field_renames, no_op_nonexistent_type_field_renames]):
```
LWprogramming, line 221: Would like feedback on this-- I notice a pattern where I create error message objects that must have at least one of these parameters as non-empty. Is there a better way to ensure this than manually writing this verification (which needs to change as the parameters change)? This also comes up when deciding whether or not to raise these errors as well-- for instance, checking if any of a number of error-collecting data structures are nonempty, and raising an error if any are.

graphql_compiler/schema_transformation/utils.py
```diff
@@ -70,22 +70,26 @@ class SchemaRenameNameConflictError(SchemaTransformError):
...
 
     def __init__(
         self,
         type_name_conflicts: Dict[str, Set[str]],
         renamed_to_builtin_scalar_conflicts: Dict[str, str],
+        field_name_conflicts: Dict[str, Dict[str, Set[str]]],
     ) -> None:
         """Record all renaming conflicts."""
-        if not type_name_conflicts and not renamed_to_builtin_scalar_conflicts:
+        if not any(
```
LWprogramming, line 82: same question as here: https://github.com/kensho-technologies/graphql-compiler/pull/970/files#r532948039

LWprogramming:
Quick summary of everything that's happened:
* Addressed comments (didn't find all of them at first because I think I started editing first so had to do several passes-- should be good now!)
* Moved TODO comments from the code to Github comments at Pedro's suggestion so we don't miss them.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -80,14 +102,28 @@
   they share a namespace as well.
 - There exist special rules for iterable renamings (e.g. if renamings are represented as a
   dictionary)-- specifically, no-op renamings are not allowed.
-  - If type_renamings argument is iterable:
+  - If type_renamings is iterable:
     - A string type_name may be in type_renamings only if there exists a type in the original schema
       named type_name (since otherwise that entry would not affect any type in the schema).
-    - If string type_name is in type_renamings, then type_renamings[type_name] != type_name (since
-      if they were the same, then applying the renaming would not change the type named type_name).
+    - If string type_name is in type_renamings, then
```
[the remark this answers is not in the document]
pmantica1: if type_renamings is iterable, then I think we always expect `type_renamings.get(type_name, type_name) != type_name`,

pmantica1:
I am pretty convinced at this point that we should use input dicts for the type renamings instead of the current protocol classes for the following reasons:
1. I don't see any substantial benefits of using the protocol types. Based on our offline discussion, there are two benefits that you 
    had in mind: 
        a. Lessened memory usage. Based on experiments, I ran offline I found the that memory used by renaming input dicts is 
            diminutive compared to the memory used by the `GraphQLSchema` and schema ast objects.  See attached screenshots 
            illustrating this. One of us the screenshots is over a false schema and the other is over an actual schema that we use at 
            Kensho. (The later screenshot is more minimal just so that I don't accidentally share any proprietary code).
        b. It grants us the ability to specify renaming rules that work without knowing the input schema or that work for a general set 
            of schemas. This seem like a benefit, but it really isn't. One of the inputs to `rename_schema` is the `schema_ast` so we 
            have access to the schema at the point that we are calling the function. Also, there is nothing stopping us from writing a 
            function that produces materialized renamings given "renaming rules" and an a schema. So we still should be able to 
            write rules that work for many schemas.
2. I've had much more difficulty understanding the code in this PR than I thought I would because of the input protocol types. For 
    instance, the protocol types don't implement `__contains__` and `__getitem__` operators so we can't just write:
 ```
if type_name in type_renamings: 
    desired_type_name = type_renamings[type_name]
    if desired_type_name == type_name:
         # I know that this is not right name for no-op errors and that you raise a single error mentioning all no-op issues,
         # but I hope you get the point.
         raise NoOpError()
``` 
   Instead we have to write 
```
desired_type_name = self.type_renamings.get(type_name, type_name)
```
   and have the no-op renaming check elsewhere, which only happens if the input type is iterable and makes things even more 
   complex.
3. Many of the comments I've left are related to issues with documenting the protocol types and their usage, re-highlighting that 
    it's not that easy to reason about them.
4. The distinction between how we treat iterable and non-iterable types is a bit hacky. For instance, we assume that the iterator 
     values are the "keys". But it would be completely valid for the`type_renamings` class to return an iterator over key-value 
     pairs when we try to iterate over it. It would also be valid for it to return an empty iterator. 

I would also like some thoughts from @obi1kenobi since he originally reviewed this protocol code.

Screenshot from 2020-12-02 15-23-50
Screenshot from 2020-12-02 15-16-27

LWprogramming:
For convenience, some links to prior discussion of topics here:

* On whether no-op renamings should be allowed: https://github.com/kensho-technologies/graphql-compiler/pull/917

* On needing to make a `reverse_name_map` data structure which at some point needs to include all the changes anyways in order to prevent name conflicts: https://github.com/kensho-technologies/graphql-compiler/pull/390#discussion_r301722935. Before writing the actual code I wasn't quite sure if we'd need to include everything or not, but now it seems there's no getting out of it.

Definitely am also interested in hearing @ obi1kenobi's opinion on design decisions around Protocols vs. just using dictionaries for everything at this point! Especially pconsidering that in principle it's possible to write a function mapping any (suitable renamings object, schema) pair to a dictionary containing all the renamings needed.

obi1kenobi:
@pmantica1 makes some compelling points, I'm convinced. If you both think it's preferable to go with plain `Mapping` types, and it's easy to refactor in that direction, let's go for it!

pmantica1 (approved):

PR #971, opened 2020-11-24
Field renaming tests

LWprogramming:
The tests that go along with the changes outlined in PR #970.

### DIFF
--- a/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
+++ b/graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
@@ -321,6 +321,354 @@ class TestRenameSchema(unittest.TestCase):
             {"Dog": "Droid", "Human": "Dog", "Droid": "Human"}, renamed_schema.reverse_name_map
         )
 
+    def test_field_rename(self) -> None:
+        renamed_schema = rename_schema(parse(ISS.basic_schema), {}, {"Human": {"id": {"new_id"}}})
+        renamed_schema_string = dedent(
+            """\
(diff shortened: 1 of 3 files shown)


graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
```diff
@@ -289,6 +289,175 @@ def test_cyclic_rename(self) -> None:
...
+            """\
+            schema {
+              query: SchemaQuery
+            }
+
+            directive @stitch(source_field: String!, sink_field: String!) on FIELD_DEFINITION
+
+            type Human {
+              id: String
+              new_id: String
```
pmantica1, line 331: Shouldn't this lead to a name conflict error because there are two fields named "new_id"?

pmantica1 (commented):
I think that you should also add tests for when all of type's fields are deleted and for field suppression over interfaces, which is not yet implemented.

graphql_compiler/tests/schema_transformation_tests/test_rename_schema.py
LWprogramming: Nope, there's just one `new_id`-- this is a separate test from the previous test, `test_field_rename`

pmantica1: Yep you are right.

pmantica1:
Can you also add a test where two fields are renamed to the same value?

pmantica1 (approved):

Issue #969, opened 2020-11-23
Improve Codecov settings to take advantage of "after_n_builds" and (if possible) flags


obi1kenobi:
Courtesy of @jerrode:

1 thing I'd definitely recommend to deal with your notifications issue: `after_n_builds` in the .yml so Codecov knows to wait to notify until everything is uploaded

https://docs.codecov.io/docs/codecovyml-reference#codecovnotifyafter_n_builds

As far as running separate unit and integration tests, have you checked out our `flags` feature where you can tag separate builds? For example, tagging unit and integration tests separately? 

https://docs.codecov.io/docs/flags

Lastly, if you are only uploading partial runs of your CI suite, you should check out Carryforward Flags so we can intelligently pick up only the new files uploaded: https://blog.codecov.io/2020/04/21/new-test-only-what-you-change-with-carryforward-flags/

_Originally posted by @jerrode in https://github.com/kensho-technologies/graphql-compiler/issues/806#issuecomment-730669829_

jerrode:
Good luck! Let me know if I can help.

obi1kenobi:
> Good luck! Let me know if I can help.

Much appreciated, thank you!

nida624:
components/tikv_util/src/topn.rs

PR #972, opened 2020-11-25
Prevent early, spurious Codecov reports

LWprogramming:
Fixes the problem in issue #969 where Codecov will generate reports before all the jobs have completed by having it only report coverage results after all lint and test jobs have completed.

I also looked into using flags as described to separate reports for different test-related jobs, but it seems like that wouldn't work-- the way to divide tests seems to be [slow vs. not-slow](https://github.com/kensho-technologies/graphql-compiler/blob/main/.github/workflows/tests_and_lint.yml#L64) rather than unit vs. integration tests. The problem with this approach is that tests with the decorator `@pytest.mark.slow` are in multiple places, including but not limited to `graphql_compiler/tests/integration_tests`, and could change whenever a new test is added, so I'm not sure that trying to do something with flags makes sense here.

I'm also a bit confused about how choosing paths for flags would work here anyways, since the [documentation example](https://docs.codecov.io/docs/flags#flags-example) looks like a monorepo where each module is isolated from the others, but the tests in this repo often use the same code paths in different ways (e.g. the not-slow compiler tests and the slow integration tests both must use the same code path to generate an IR).


LWprogramming:
A note: I received an email from Codecov before all 14 of the jobs passed, saying that coverage got worse-- not sure if that's the old settings or if this doesn't actually fix the problem

obi1kenobi:
> A note: I received an email from Codecov before all 14 of the jobs passed, saying that coverage got worse-- not sure if that's the old settings or if this doesn't actually fix the problem

Doesn't seem to have fixed it — the email reported `Merging #972 (d206f92) into main (1a7c7c4) will decrease coverage by 8.80%.` whereas the current state is `decrease coverage by 0.00%`.

LWprogramming:
Figured out what the problem was-- it turns out that `after_n_builds` is a parameter for both `comment` and `notify`, but the latter is for Slack/ IRC/ etc notifications whereas the emails are generated from the codecov bot comment on the PR itself. Now the email shows up after the first 8 builds have completed-- unfortunately, at the moment I haven't seen an easy way to distinguish between builds for tests (which affect coverage) and builds for linting (which don't). I think it'd be a bit easier to understand if `after_n_builds` just depends on the number of test builds required, so I'm going to look into [flags](https://docs.codecov.io/docs/flags) to see if there's an easy way to specify which 8 builds need to finish before comment, but if there's no easy way to do it then I'll just change the number to 14 to have codecov wait for everything.

LWprogramming:
Summary of changes up to this point: 

- There seems to be no easy way to separate test and lint builds, so I've decided for simplicity to just make the codecov bot wait until all builds are complete.

- 8 test + 6 lint + 1 for codecov (patch) + 1 for codecov (project) come out to 16 builds, so I expected the bot to send an email after all the builds are done. However, this is not what happened-- according to the email timestamp, it was sent at 2:40:52 PM EST, but the following builds completed after this time: pylint, typing-copilot, 3.6 not-slow tests, 3.6 slow tests, 3.7 slow tests, 3.8 slow tests, 3.9 slow tests. That seems like progress, since codecov is no longer leaving comments after the very first build is done, but it's still too early, sending an update while only about half the builds are done.

LWprogramming:
Tagging @jerrode if there's anything we missed here about using `after_n_builds`?

jerrode:
Hey @LWprogramming, thanks for this very clear summary and sorry for delay on response. Tapping in my colleague @thomasrockhu, our dev relations lead, to see if canhelp out

codecov.yml
```diff
@@ -5,3 +5,5 @@ coverage:
         target: auto
         threshold: 0.03%
         base: auto
+  comment:
+    after_n_builds: 16 # Prevent early, spurious Codecov reports before all tests finish: https://github.com/kensho-technologies/graphql-compiler/pull/806#issuecomment-730622647. 16 is calculated from the number of jobs to run, which is specified in the .github/workflows/tests_and_lint.yml file: 6 lint jobs (1 job per combination of python-version and lint-flags) and 8 test jobs (1 job per combination of python-version and markers), plus 1 for codecov patch build, plus 1 for the codecov project build.
```
thomasrockhu, line 9: `comment` is actually a top-level section in the `yaml`.

```suggestion
comment:
  after_n_builds: 16 # Prevent early, spurious Codecov reports before all tests finish: https://github.com/kensho-technologies/graphql-compiler/pull/806#issuecomment-730622647. 16 is calculated from the number of jobs to run, which is specified in the .github/workflows/tests_and_lint.yml file: 6 lint jobs (1 job per combination of python-version and lint-flags) and 8 test jobs (1 job per combination of python-version and markers), plus 1 for codecov patch build, plus 1 for the codecov project build.
```

thomasrockhu:
Thanks @jerrode, looks like the `after_n_builds` isn't catching because the `comment` section is improperly indented.

thomasrockhu:
I would also note that it should probably be 14, not 16. Builds are considered uploads to Codecov, not statuses.

LWprogramming:
Confirmed, this works! All tests and linting jobs finished by 4:46:45 and the email came out at 4:47:09 EST. Thanks so much!

thomasrockhu:
Woo hoo! Glad it worked for you @LWprogramming

obi1kenobi (approved):

PR #973, opened 2020-11-30
Add py.typed file to export type hints

justinaustin:


pmantica1:
I don't know much about the `package_data` argument but are we sure that `""` is a valid package name.

The following stackoverflow post seems to say otherwise:
https://stackoverflow.com/questions/29036937/how-can-i-include-package-data-without-a-manifest-in-file

justinaustin:
`""` is what we use in internal kensho packages (and it seems to work), but I didn't realize the meaning behind the keys. I can switch to the package name if you all prefer!

pmantica1:
According to the link below, "The empty string stands for the root package." 
https://docs.python.org/3/distutils/examples.html#pure-python-distribution-by-package 

So both `{"graphql_compiler": "py.typed"}` and `{"": "py.typed"}` should both be semantically equal and lead to successful type hinting. I tested locally that both led to successful type hinting. I prefer `{"": py.typed}` since it is less repetitive.

To check that I didn't make any blunders when verifying that type hinting worked correctly, I checked that type hinting didn't work if we remove the `package_data=...` block or if we passed in dummy values like `{"asdasd": ""py.typed"}`. For those curious, a possibly verification blunder here would be to run `pip install -e .` instead of `pip wheel . && pip install pip install graphql_compiler-1.11.0-py2.py3-none-any.whl". The first only makes something like a "symlink" to the `graphql_compiler` package in your local directory. So the py.typed file would always be included regardless of what you put in `package_data`. The later actually packages the distribution package into a binary, (a wheel).

One thing that kind of confused me here initially is that in python "package" refers to both "import packages" and "distribution packages". The distribution package is "graphql-compiler". An import package is just a directory with an `__init__.py` file. The "packages" that are referred to in the `package_data` documentation are the "import packages" that are discovered by the `find_packages` function, e.g. graphql_compiler, graphql_compiler.schema, graphql_compiler.interpreter, etc. 

https://packaging.python.org/guides/distributing-packages-using-setuptools/#package-data
https://packaging.python.org/glossary/#term-Distribution-Package
https://packaging.python.org/glossary/#term-Import-Package

pmantica1 (approved):

PR #979, opened 2020-12-09
Prefer fields with inequality filters for pagination

bojanserafimov:
If fields A and B are correlated, and there's a strong inequality filter on A, then paging on B is a bad idea. Most of the pages will come back empty.

In this PR, we detect this scenario and choose A as a pagination field. To do that we need to relax the user-provided pagination field preference just a little bit -- enough to allow flexibility, but not enough to lose control.

This PR is an example of why we don't want to publish the pagination API yet. It's gonna change like this many times until we get it right.

Since there's lots of noise in the tests in this PR, I've singled out the one test that actually tests something with a comment.

### DIFF
--- a/graphql_compiler/schema/schema_info.py
+++ b/graphql_compiler/schema/schema_info.py
@@ -392,7 +392,7 @@ class QueryPlanningSchemaInfo:
     # that are eligible pagination on that vertex in a general context. Some of the provided
     # properties might be ineligible for pagination in certain queries, for instance if the
     # query has a "=" filter on the field and it's uniquely indexed. The order of properties
-    # within this list is used as a final (least significant) step in choosing which one
+    # within this sequence is used as a final (least significant) step in choosing which one
     # to use when multiple options are available. The pagination key list for a vertex can
     # be omitted, making the entire vertex ineligible for pagination.
     pagination_keys: Mapping[str, Sequence[str]]
--- a/graphql_compiler/schema/schema_info.py
+++ b/graphql_compiler/schema/schema_info.py
@@ -4,7 +4,7 @@ from collections import namedtuple
 from dataclasses import dataclass, field
 from enum import Enum, Flag, auto, unique
 from functools import partial
-from typing import Dict, Optional, List
+from typing import Dict, Optional, Tuple
 
 from graphql.type import GraphQLSchema
 from graphql.type.definition import GraphQLInterfaceType, GraphQLObjectType
@@ -388,14 +388,14 @@ class QueryPlanningSchemaInfo:
     # A Statistics object giving statistical information about all objects in the schema.
     statistics: Statistics
 
-    # Dict mapping vertex names in the graphql schema to the non-empty list of property names
+    # Dict mapping vertex names in the graphql schema to the non-empty tuple of property names
     # that are eligible pagination on that vertex in a general context. Some of the provided
     # properties might be ineligible for pagination in certain queries, for instance if the
     # query has a "=" filter on the field and it's uniquely indexed. The order of properties
     # within this list is used as a final (least significant) step in choosing which one
     # to use when multiple options are available. The pagination key list for a vertex can
     # be omitted, making the entire vertex ineligible for pagination.
-    pagination_keys: Dict[str, List[str]]
+    pagination_keys: Dict[str, Tuple[str, ...]]
 
     # Dict mapping vertex names in the graphql schema to a dict mapping property names that
     # are known to contain uniformly distributed uuid values to the ordering method used for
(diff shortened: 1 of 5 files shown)


graphql_compiler/tests/snapshot_tests/test_query_pagination.py
```diff
@@ -669,8 +689,12 @@ def test_pagination_datetime(self):
...
         schema_graph = generate_schema_graph(self.orientdb_client)
         graphql_schema, type_equivalence_hints = get_graphql_schema_from_schema_graph(schema_graph)
-        pagination_keys = {vertex_name: "uuid" for vertex_name in schema_graph.vertex_class_names}
-        pagination_keys["Event"] = "event_date"  # Force pagination on datetime field
+        pagination_keys = {
+            vertex_name: ("uuid",) for vertex_name in schema_graph.vertex_class_names
+        }
+        # We allow pagination on uuid as well and leave it to the pagination planner to decide to
+        # paginate on event_date to prevent empty pages if the two fields are correlated.
+        pagination_keys["Event"] = ("uuid", "event_date")
```
bojanserafimov, line 697: this is the test, everything else is just refactor

graphql_compiler/schema/schema_info.py
```diff
@@ -388,16 +388,14 @@ class QueryPlanningSchemaInfo:
...
-    #                       schema graph, due to the process of type overrides that happens
-    #                       during schema generation.
-    pagination_keys: Dict[str, str]
+    # Mapping vertex names in the graphql schema to the non-empty sequence of property names
+    # that are eligible pagination on that vertex in a general context. Some of the provided
+    # properties might be ineligible for pagination in certain queries, for instance if the
+    # query has a "=" filter on the field and it's uniquely indexed. The order of properties
+    # within this list is used as a final (least significant) step in choosing which one
+    # to use when multiple options are available. The pagination key list for a vertex can
+    # be omitted, making the entire vertex ineligible for pagination.
```
obi1kenobi, line 397: When you say "omitted" do you mean that the list is empty? If so, let's say that explicitly because it wasn't quite obvious to me.

graphql_compiler/schema/schema_info.py
```diff
@@ -388,16 +388,14 @@ class QueryPlanningSchemaInfo:
...
-    #
-    # NOTE(bojanserafimov): The type of this property might be different in the
-    #                       schema graph, due to the process of type overrides that happens
-    #                       during schema generation.
-    pagination_keys: Dict[str, str]
+    # Mapping vertex names in the graphql schema to the non-empty sequence of property names
+    # that are eligible pagination on that vertex in a general context. Some of the provided
+    # properties might be ineligible for pagination in certain queries, for instance if the
+    # query has a "=" filter on the field and it's uniquely indexed. The order of properties
+    # within this list is used as a final (least significant) step in choosing which one
```
obi1kenobi, line 395:
```suggestion
    # within this sequence is used as a final (least significant) step in choosing which one
```

graphql_compiler/schema/schema_info.py
```diff
@@ -388,16 +388,14 @@ class QueryPlanningSchemaInfo:
...
+    # to use when multiple options are available. The pagination key list for a vertex can
```
obi1kenobi, line 396:
```suggestion
    # to use when multiple options are available. The pagination key sequence for a vertex can
```

obi1kenobi (commented):
Looks good. Minor nits.

graphql_compiler/schema/schema_info.py
bojanserafimov: It says above that the sequences are non-empty. I can repeat here if you think it adds value

obi1kenobi: Oh okay so omitted means "not present in this dict"? In that case, how about the phrasing below:

graphql_compiler/schema/schema_info.py
```diff
@@ -388,16 +388,14 @@ class QueryPlanningSchemaInfo:
...
-    #                       schema graph, due to the process of type overrides that happens
-    #                       during schema generation.
-    pagination_keys: Dict[str, str]
+    # Mapping vertex names in the graphql schema to the non-empty sequence of property names
+    # that are eligible pagination on that vertex in a general context. Some of the provided
+    # properties might be ineligible for pagination in certain queries, for instance if the
+    # query has a "=" filter on the field and it's uniquely indexed. The order of properties
+    # within this sequence is used as a final (least significant) step in choosing which one
+    # to use when multiple options are available. The pagination key sequence for a vertex can
+    # be omitted, making the entire vertex ineligible for pagination.
```
obi1kenobi, line 397:
```suggestion
    # to use when multiple options are available.
    #
    # If a vertex name is omitted from this dict, the entire vertex is ineligible for pagination.
```

obi1kenobi (approved):

PR #982, opened 2020-12-10
Query renaming field renaming

LWprogramming:
Implements query renaming now that field renaming has been implemented.

### DIFF
--- a/graphql_compiler/schema_transformation/rename_query.py
+++ b/graphql_compiler/schema_transformation/rename_query.py
@@ -31,14 +31,15 @@ def rename_query(
     any occurrences of type Y in the input query ast will be renamed to X.
 
     All type names (including ones in type coercions), as well as root vertex fields (fields
-    of the query type) will be renamed. No other field names will be renamed.
+    of the query type) will be renamed.
 
     Args:
         ast: represents a query
         renamed_schema_descriptor: namedtuple including the attribute reverse_name_map, which maps
-                                   the new, renamed names of types to their original names. This
-                                   function will revert these renamed types in the query ast back to
-                                   their original names
+                                   the new, renamed names of types to their original names, and
+                                   reverse_field_name_map which has a similar role for fields. This
+                                   function will revert these renamed types/ fields in the query ast
+                                   back to their original names
 
     Returns:
         New AST representing the renamed query
@@ -91,7 +92,7 @@ class RenameQueryVisitor(Visitor):
         type_renamings: Dict[str, str],
         field_renamings: Dict[str, Dict[str, str]],
     ) -> None:
-        """Create a visitor for renaming types and root vertex fields in a query AST.
+        """Create a visitor for renaming types and fields in a query AST.
 
         Args:
             schema: The renamed schema that the original query is written against
--- a/graphql_compiler/schema_transformation/rename_query.py
+++ b/graphql_compiler/schema_transformation/rename_query.py
@@ -24,14 +24,13 @@ from .utils import RenameQueryNodeTypesT, get_copy_of_node_with_new_name
 def rename_query(
     ast: DocumentNode, renamed_schema_descriptor: RenamedSchemaDescriptor
 ) -> DocumentNode:
-    """Translate names of types using reverse_name_map of the input RenamedSchemaDescriptor.
+    """Translate names of types/ fields using reverse_name_map of the input RenamedSchemaDescriptor.
 
-    The direction in which types and fields are renamed is opposite of the process that
-    produced the renamed schema descriptor. If a type X was renamed to Y in the schema, then
-    any occurrences of type Y in the input query AST will be renamed to X.
+    The direction in which types/ fields are renamed is opposite of the process that
+    produced the renamed schema descriptor. If a type/ field X was renamed to Y in the schema, then
+    any occurrences of Y in the input query AST will be renamed to X.
 
-    All type names (including ones in type coercions), as well as root vertex fields (fields
-    of the query type) will be renamed.
+    All type names (including ones in type coercions) and field names will be renamed.
 
     Args:
         ast: represents a query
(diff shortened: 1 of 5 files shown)


graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -131,18 +169,56 @@ def leave_selection_set(
...
+                    f"Current type {current_type_name}'s ast_node field should be an "
+                    f"ObjectTypeDefinitionNode. However, the actual type was "
+                    f"{type(current_type.ast_node)}. This is a bug."
+                )
+            current_type_fields = current_type.ast_node.fields
+            for field_node in current_type_fields:
+                # Unfortunately, fields is a list instead of some other datastructure so we actually
+                # have to loop through them all.
+                if field_node.name.value == node.name.value:
+                    field_type_name = get_ast_with_non_null_and_list_stripped(
```
LWprogramming, line 209: Linting fails here because this function is untyped. I tried adding type hints so that it takes a `TypeNode` argument (based on how this function is used) and returns a `Node`, but it'd involve some more issues where the code assumes the `Node` objects have a `name` field so I wanted to hold off on this in case there was a cleaner way to do this.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -27,14 +31,15 @@ def rename_query(
...
 
     Args:
         ast: represents a query
         renamed_schema_descriptor: namedtuple including the attribute reverse_name_map, which maps
-                                   the new, renamed names of types to their original names. This
-                                   function will revert these renamed types in the query ast back to
-                                   their original names
+                                   the new, renamed names of types to their original names, and
+                                   reverse_field_name_map which has a similar role for fields. This
+                                   function will revert these renamed types/ fields in the query ast
```
obi1kenobi, line 41: nit: AST is an acronym, so let's try to consistently capitalize it here and elsewhere, and especially in end-user-facing messages and docstrings. This function is going to be part of the public API, so the quality bar should be particularly high.
```suggestion
                                   function will revert these renamed types/ fields in the query AST
```

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -27,14 +31,15 @@ def rename_query(
     any occurrences of type Y in the input query ast will be renamed to X.
 
     All type names (including ones in type coercions), as well as root vertex fields (fields
```
obi1kenobi, line 33: This docstring mentions quite a bit of stuff about types, but based on the `Args` section it seems like the function will also rename fields? It's quite surprising to read the entire docstring up to the `Args` section and not see a mention of field renaming (other than root vertex fields, which are types). Even the one-liner at the top of the docstring only mentions types.

Consider touching up the rest of this docstring if the function's contract is indeed changing to include field renaming as well.

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -131,18 +169,56 @@ def leave_selection_set(
...
-            if renamed_node is node:  # Name unchanged, continue traversal
-                return None
-            else:  # Name changed, return new node, `visit` will make shallow copies along path
-                return renamed_node
+            self.current_type_name.append(node.name.value)
+        else:
+            # Entered a regular field and we want to find its type
+            current_type_name = self.current_type_name[-1]
+            current_type = self.schema.get_type(current_type_name)
+            if not current_type:
```
obi1kenobi, line 185: Do we expect this to be `None` if this branch gets hit? If so, I have a mild preference for the more explicit `is None` check because otherwise we have to know what could make a type falsy. `if not current_type` does `bool(current_type)` under the hood, which usually does the right thing, but if someone defines a `__bool__` method then anything goes. With `is None`, I know exactly what the code does with no wiggle room.

I realize this is the opposite of the Google style guide, so if you decide you want to keep it this way, that's fine too.
```suggestion
            if current_type is None:
```

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -131,18 +169,56 @@ def leave_selection_set(
...
+                raise AssertionError(
+                    f"Current type is {current_type_name} which doesn't exist in the schema. This "
+                    f"is a bug."
+                )
+            if not current_type.ast_node:
+                raise AssertionError(
+                    f"Current type {current_type_name} should have non-null field ast_node, which "
+                    f"contains information such as the current type's fields. However, ast_node "
+                    f"was None. This is a bug."
```
obi1kenobi, line 194: Here's a practical example of why I suggested the change above. The error message says that `ast_node` was `None` — but we don't actually know that for sure. All we know is that it was falsy, which means it was _probably_ `None`. I've had way too many debugging odysseys over "trivial" differences like this so I'd advise going for the explicit `None` check again:
```suggestion
            if current_type.ast_node is None:
                raise AssertionError(
                    f"Current type {current_type_name} should have non-null field ast_node, which "
                    f"contains information such as the current type's fields. However, ast_node "
                    f"was None. This is a bug."
```

graphql_compiler/schema_transformation/rename_query.py
obi1kenobi: Yeah, it's not great and it's a tricky thing to fix. In the interest of speed, let's allow (only) this file to call untyped code by modifying the mypy.ini. Let me know if you'd like a hand with that.

LWprogramming: This ended up being a bigger change than expected but that's just because `typing_copilot` merged a lot of rules into one now that we're allowing untyped calls again

graphql_compiler/schema_transformation/rename_query.py
```diff
@@ -24,21 +24,20 @@
...
     ast: DocumentNode, renamed_schema_descriptor: RenamedSchemaDescriptor
 ) -> DocumentNode:
-    """Translate names of types using reverse_name_map of the input RenamedSchemaDescriptor.
+    """Translate names of types/ fields using reverse_name_map of the input RenamedSchemaDescriptor.
 
-    The direction in which types and fields are renamed is opposite of the process that
-    produced the renamed schema descriptor. If a type X was renamed to Y in the schema, then
-    any occurrences of type Y in the input query ast will be renamed to X.
+    The direction in which types/ fields are renamed is opposite of the process that
+    produced the renamed schema descriptor. If a type/ field X was renamed to Y in the schema, then
```
obi1kenobi, line 30: Is `type/ field` the proper way to use spacing around `/`? I'm no expert on orthography but this looks strange to me.

obi1kenobi (approved):
One comment where I'm honestly not sure about the answer. If you're confident that it's correct, feel free to merge.

obi1kenobi (approved):

PR #984, opened 2020-12-15
Use inclusive language [removing sanity check]

chewselene:
Remove ableist language.

### DIFF
--- a/graphql_compiler/compiler/ir_validation.py
+++ b/graphql_compiler/compiler/ir_validation.py
@@ -23,13 +23,13 @@ from .ir_lowering_common.common import extract_folds_from_ir_blocks
 from .metadata import QueryMetadataTable
 
 
-def sanity_check_ir_blocks_from_frontend(
+def validate_ir_blocks_from_frontend(
     ir_blocks: List[BasicBlock], query_metadata_table: QueryMetadataTable
 ) -> None:
     """Assert that IR blocks originating from the frontend do not have nonsensical structure.
 
     Args:
-        ir_blocks: list of BasicBlocks representing the IR to sanity-check
+        ir_blocks: list of BasicBlocks representing the IR to validate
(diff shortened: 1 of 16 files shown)


graphql_compiler/compiler/ir_validation.py
```diff
@@ -255,7 +255,7 @@ def _sanity_check_every_location_is_marked(ir_blocks: List[BasicBlock]) -> None:
             mark_location_blocks_count = 0
 
 
-def _sanity_check_coerce_type_outside_of_fold(ir_blocks: List[BasicBlock]):
+def _validate_coerce_type_outside_of_fold(ir_blocks: List[BasicBlock]):
```
bojanserafimov, line 258:
```suggestion
def _validate_coerce_type_outside_of_fold(ir_blocks: List[BasicBlock]) -> None:
```

branen (commented):
Instead of `self_consistency_check` and `self_consistency_checks`, what about `assert` and `assertions`?  They describe what's actually happening, are unambiguous to the reader, and are much easier to read, type, and say.

chewselene:
> Instead of `self_consistency_check` and `self_consistency_checks`, what about `assert` and `assertions`? They describe what's actually happening, are unambiguous to the reader, and are much easier to read, type, and say.

I originally had this as validate/validation! gonna loop in @obi1kenobi who said:
` would it be okay to use self-consistency check instead of validate when possible for this? there’s a subtle distinction, but compiler IR can be locally valid (all real blocks with valid arguments etc.) but self-inconsistent (invariants that span multiple blocks are not upheld) so for example, if there’s a Traverse without a matching Backtrack, or a Traverse to a location that is never marked with MarkLocation`

I think assert/assertion likely has a similar issue to validate/validation, but Branen also pointed out that all these functions do raise assertion errors so maybe it is appropriate?

obi1kenobi (approved):
Looks fine to me!

PR #986, opened 2021-01-05
Validate arguments before interpreting query.

obi1kenobi:
We didn't use to validate that all the expected arguments are present before starting to interpret queries. Now we do :)


obi1kenobi:
Ugh I forgot that the `interpreted_mode_v3` branch doesn't lint cleanly at the moment 😢  Still need to put in the work to clean that up and merge it into `main`, let's not block this PR for the sins of the parent branch.

The `pytest` runs succeeded though!

chewselene (approved):
Looks good to me!

PR #988, opened 2021-01-06
Type hint split query

chewselene:
Adds type hints to `graphql_compiler/schema_transformation/split_query.py`. Also converts `namedtuple` to `dataclass`.


graphql_compiler/schema_transformation/split_query.py
```diff
@@ -116,42 +214,67 @@ def split_query(query_ast, merged_schema_descriptor):
...
+        Dictionary mapping (type name, vertex field name) to (source field name, sink field name)
+        used in the @stitch directive, for each cross schema edge.
     """
     edge_to_stitch_fields = {}
     for type_definition in merged_schema_descriptor.schema_ast.definitions:
         if isinstance(type_definition, (ObjectTypeDefinitionNode, InterfaceTypeDefinitionNode)):
             for field_definition in type_definition.fields:
                 stitch_directive = try_get_ast_by_name_and_type(
-                    field_definition.directives, "stitch", DirectiveNode
+                    list(field_definition.directives), "stitch", DirectiveNode
```
chewselene, line 239: I added this `list()` here and in other places to get rid of mypy errors like `Argument 1 to "try_get_ast_by_name_and_type" has incompatible type "List[FieldNode]"; expected "Optional[List[Node]]"`. Not sure if this is the best way to go about it. I used these docs: https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance. In some cases, I tried adding explicit type annotations to variables, but was still seeing errors so tried this method instead. Let me know if there's a better way!

obi1kenobi: Would it be feasible to change the type hint of `try_get_ast_by_name_and_type` from `Optional[List[Node]]` to `Optional[Sequence[Node]]` which I believe will be covariant? I think that function doesn't actually modify the list, so it shouldn't need to know that the type is a list and not a sequence.

Using abstract `Sequence/AbstractSet/Mapping` instead of concrete `List/Set/Dict` when possible is something we should probably try to get better at over time.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -564,41 +739,45 @@ def _get_child_query_node_and_out_name(ast, child_type_name, child_field_name, n
     )
     # Get new child_selections by replacing or adding in new property field
     child_property_fields_map, child_vertex_fields = _split_selections_property_and_vertex(
-        child_selections
+        list(child_selections)
```
obi1kenobi, line 742: Assuming this is the same kind of issue? Can we solve it by typing the callee with `Sequence` instead?

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -218,36 +343,38 @@ def _split_query_one_level(
...
     Returns:
-        Field, InlineFragment, or OperationDefinition, the AST with which to replace the input
-        AST in the selections that contain it
+        The AST with which to replace the input AST in the selections that contain it.
     """
+    if ast.selection_set is None:
+        raise AssertionError("AST's selection_set cannot be None.")
     type_info.enter(ast.selection_set)
-    selections = ast.selection_set.selections
+    selections = list(ast.selection_set.selections)
```
obi1kenobi, line 377: Is this `list()` feasible to avoid as well? It's not terribly expensive on its own, but it doesn't seem like we are mutating `selections` itself, so hopefully it might not be so bad? In the worst case, we might even be able to get away with a cast to the necessary type on the next line instead of a copy of the list.

graphql_compiler/schema_transformation/split_query.py
```diff
@@ -491,23 +637,32 @@ def _split_vertex_fields_intra_and_cross_schema(
     return intra_schema_fields, cross_schema_fields
 
 
-def _get_selections_from_property_and_vertex_fields(property_fields_map, vertex_fields):
+def _get_selections_from_property_and_vertex_fields(
+    # OrderedDict is unsubscriptable (pylint E1136)
+    property_fields_map: "OrderedDict[str, FieldNode]",
+    vertex_fields: List[FieldNode],
+) -> List[FieldNode]:
```
obi1kenobi, line 644: Could this return `List[SelectionNode]` and maybe take `OrderedDict[str, SelectionNode]` and `Sequence[SelectionNode]`?

If we can't change the input types easily, could we add a runtime `isinstance` check on all the things we're adding to the `selections` list to make sure they are all `SelectionNode`? That way the input types can stay the same, the return type can become what we need it to be downstream, and we only pay a small runtime cost that doesn't increase memory pressure.

obi1kenobi (commented):
A couple of remaining minor things. If they aren't easy to solve, this is probably alright to merge as-is.

obi1kenobi (approved):

PR #989, opened 2021-01-06
Eliminate absolute imports of our own package.

obi1kenobi:
A small quality of life improvement: replace all absolute imports of our own package's submodules with relative imports of the same.

Unless there's a good reason, we should prefer to import our own submodules via relative imports rather than absolute ones. In some of the cases in this PR, we had separate absolute and relative imports of the same submodule, which makes the code more difficult to read and understand.

Currently we don't have a mechanism for enforcing that this type of lint doesn't come back, but I'm hoping that something based on https://twitter.com/PredragGruevski/status/1344725604520878080 might be able to help in the long run :)


chewselene (approved):
Looks good to me!

PR #990, opened 2021-01-07
Type hint make query plan

chewselene:
Adds type hints and converts namedtuples to dataclasses in `schema_transformation/make_query_plan.py`.

### DIFF
--- a/graphql_compiler/schema_transformation/make_query_plan.py
+++ b/graphql_compiler/schema_transformation/make_query_plan.py
@@ -44,7 +44,7 @@ class SubQueryPlan:
 
 @dataclass(frozen=True)
 class OutputJoinDescriptor:
-    """Description of what outputs should be join and how."""
+    """Description of what outputs should be joined and how."""
 
     # (parent output name, child output name)
     # May be expanded to have more attributes, e.g. is_optional, describing how the join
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_transformation/make_query_plan.py
```diff
@@ -221,18 +246,23 @@ def _add_filter_at_field_with_output(ast, field_out_name, input_filter_name):
...
 
 
-def _is_output_directive_with_name(directive, out_name):
+def _is_output_directive_with_name(directive: DirectiveNode, out_name: str) -> bool:
     """Return whether or not the input is an @output directive with the desired out_name."""
     if not isinstance(directive, DirectiveNode):
         raise AssertionError('Input "{}" is not a directive.'.format(directive))
+    # The directive argument must be a string since output directives must have a
+    # non-null string `out_name`.
+    directive_out_name_value_node = directive.arguments[0].value
```
obi1kenobi, line 255: This refactor is unfortunately not equivalent to the prior code. In the original code, `directive.name.value == OutputDirective.name` is guaranteed to evaluate first (before `directive.arguments[0].value`), and since by this point we expect that the query has passed validation against the schema, that means that `directive.arguments[0]` actually exists i.e. the directive has arguments.

Unfortunately, it's not safe to flip this order around: not all directives have arguments, so `directive.arguments[0]` might not exist and may cause an unexpected exception.

Options are to embed the `isinstance` check inside the chain of `and`-ed expressions, or check the name explicitly and have another early-return before this argument `isinstance` check. To prevent this bug from being reintroduced in the future, it may also be good to add an explicit assertion that `directive.arguments` is not `None` or an empty list like `if not directive.arguments: raise AssertionError()`.

obi1kenobi (changes requested):
Other than the one minor bugfix, looks good!

obi1kenobi (approved):

PR #991, opened 2021-01-07
Allow microseconds in datetimes and allow identity-function parsing.

obi1kenobi:
Given that we are working toward supporting interpreter-based query execution, rather than just purely compilation-based ones, we should do our best to align our expectations of incoming data to the capabilities and use cases of interpreters.

Up to this point, we've had a couple of limitations that this PR helps us move away from:
- The GraphQL `Date` and `DateTime` types' `parse()` functions always expected string inputs, because that's what was reasonable to get from a client "over the wire." However, in the case of interpreters, it's much more likely that _the client_ is themselves running the interpreter, and therefore the argument validation functions (which depend on calling `parse()` on GraphQL type objects) may in fact be given an already-constructed Python date or datetime object. In such cases, `parse()` should simply return the input as-is, after validating that it is in fact a supported value (e.g. that the datetime is timezone-naive, the same requirement which exists for string-based inputs).
- The GraphQL `DateTime` type had precision down to seconds rather than microseconds. While seconds are convenient in some cases (e.g. for use with OrientDB), Python `datetime` objects and datetime objects in many databases (e.g. PostgreSQL) go down to microseconds. Limiting precision to seconds for the sake of OrientDB was therefore disadvantaging interpreters and databases like PostgreSQL that could support higher precision.

In this PR, we allow microsecond-level precision for GraphQL `DateTime` query arguments, and we allow the `parse()` functions for `Date` and `DateTime` to accept the kinds of objects as input that they would themselves return, simply returning them as-is.


graphql_compiler/schema/__init__.py
```diff
@@ -340,36 +340,47 @@ def _serialize_date(value: Any) -> str:
 
 def _parse_date_value(value: Any) -> date:
     """Deserialize a Date object from its proper ISO-8601 representation."""
-    return arrow.get(value, "YYYY-MM-DD").date()
+    if type(value) == date:
+        # We prefer exact type equality instead of isinstance() because datetime objects
+        # are subclasses of date but are not interchangeable for dates for our purposes.
+        return value
+    elif isinstance(value, str):
+        return arrow.get(value, "YYYY-MM-DD").date()
```
eyurtsev, line 348: Any need to assert that value follows ISO-8601 prior to parsing / truncating?

```In [7]: arrow.get('2010-01-01 2020-01-01 boom boom', 'YYYY-MM-DD')                   
Out[7]: <Arrow [2010-01-01T00:00:00+00:00]>
```

branen, line 348: It seems incongruous to me that we disallow subtypes of `date` to prevent implicit loss of precision, but then implicitly drop precision ourselves when presented with string-specified date-times.

```python
>>> arrow.get('0001-02-03 04:05:06', "YYYY-MM-DD").date()
datetime.date(1, 2, 3)
```

I think there are reasonable arguments to be made in favor of either behavior, and I'd rather we choose one and stick with it.  (If we decide to stick with disallowing implicit loss of precision, then we should start emitting warnings for the string case now and switch to raising exceptions in the 2.0 release.)

graphql_compiler/schema/__init__.py
```diff
@@ -340,36 +340,47 @@ def _serialize_date(value: Any) -> str:
...
+    if isinstance(value, datetime) and value.tzinfo is None:
         return value.isoformat()
     else:
         raise ValueError(
-            f"Expected a timezone naive datetime object. Got {value} of type {type(value)} instead."
+            f"Expected a timezone-naive datetime object. Got {value} of type {type(value)} instead."
         )
 
 
 def _parse_datetime_value(value: Any) -> datetime:
```
branen, line 369:
```suggestion
def _parse_datetime_value(value: Union[datetime, str]) -> datetime:
```

graphql_compiler/schema/__init__.py
```diff
@@ -389,10 +400,11 @@ def _parse_datetime_value(value: Any) -> datetime:
...
     name="DateTime",
     description=(
-        "The `DateTime` scalar type represents timezone-naive second-accuracy timestamps."
+        "The `DateTime` scalar type represents timezone-naive timestamps with microsecond accuracy."
         "Values are serialized following the ISO-8601 datetime format specification, "
-        'for example "2017-03-21T12:34:56". All of these fields must be included, '
-        "including the seconds, and the format followed exactly, or the behavior is undefined."
+        'for example "2017-03-21T12:34:56.012345" or "2017-03-21T12:34:56". All fields down to '
+        "and including seconds must be included, while fractional seconds are optional."
+        "If this format is not followed, the behavior is undefined."
```
branen, line 407: Is there a compelling reason to leave this behavior undefined?  Strict adherence to ISO-8601 isn't particularly onerous, and I'd much rather we just reject deviations outright than interpret them in undefined ways.

graphql_compiler/deserialization.py
```diff
@@ -1,5 +1,6 @@
 # Copyright 2020-present Kensho Technologies, LLC.
 """Convert values to their underlying GraphQLType."""
+from datetime import date, datetime
 from types import MappingProxyType
 from typing import Any, Callable, Mapping, Tuple, Type
```
branen, line 5:
```suggestion
from typing import Any, Callable, Mapping, Tuple, Type, Union
```

(See my suggestion for `_parse_datetime_value`.)

branen (changes requested):

graphql_compiler/schema/__init__.py
```diff
@@ -340,36 +340,47 @@ def _serialize_date(value: Any) -> str:
...
+    elif isinstance(value, str):
+        return arrow.get(value, "YYYY-MM-DD").date()
+    else:
+        raise ValueError(
+            f"Expected a date object or its ISO-8601 string representation. "
+            f"Got {value} of type {type(value)} instead."
+        )
 
 
 def _serialize_datetime(value: Any) -> str:
```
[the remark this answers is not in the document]
obi1kenobi: Unfortunately we have to follow the upstream contract, since this function is going to get assigned to a GraphQL type object's `serialize()` slot. The upstream contract is that both the `parse()` and the `serialize()` functions accept `Any`:
https://github.com/graphql-python/graphql-core/blob/master/src/graphql/type/scalars.py#L105

Hence the runtime type checks.

graphql_compiler/schema/__init__.py
obi1kenobi: I think this is not allowed for the same reason as above — it does not follow the upstream contract, which is that `parse()` and `serialize()` on GraphQL type objects (which both this function and the function above are used as) take `Any` arguments.
https://github.com/graphql-python/graphql-core/blob/master/src/graphql/type/scalars.py#L105

graphql_compiler/schema/__init__.py
obi1kenobi: This is a good point. I think we should not allow implicit loss of precision, and we're pretty committed to only releasing 2.0-series versions at this point (the `main` branch already has breaking changes), so I'd like to go straight to raising exceptions for loss of precision in the interest of time and simplicity.

obi1kenobi: This is a good point. Maybe I should just use the Python builtins for this case, since there are many fewer subtleties in the date case than in the datetime case.

Unfortunately, `arrow` in general is really permissive with its inputs, and I don't immediately know of a good alternative for ISO 8601 datetime parsing. If you have one you like, I'm happy to eliminate `arrow` as a dependency altogether and switch to it entirely.

obi1kenobi: Maybe this one?
https://pypi.org/project/ciso8601/

Also @branen what do you think? Is it worth ripping out `arrow` entirely as a dependency and switching to `ciso8601` instead?

graphql_compiler/schema/__init__.py
obi1kenobi: Honestly, largely because I don't trust date/datetime parsing libraries to not get "too creative" in how they interpret dates. If we specify a particular behavior, then we are on the hook for any parsing bugs or inconsistencies in those libraries that happen _outside of_ ISO 8601 string cases.

For example, the `parse()` function on the GraphQL `DateTime` type can be called with any input data, including `int` and `bytes` and `float` and almost-but-not-quite ISO 8601, and while we have some best-effort tests for many of these cases, I don't want to overpromise nor have to own the API and test quality of the underlying libraries for anything outside of the "known to work" ISO 8601 seconds-included, fractionals-optional format we declare acceptable and extensively tested.

graphql_compiler/schema/__init__.py
branen: @obi1kenobi, I'd like to give this some more thought before I express an opinion about `arrow` vs `ciso8601` vs anything else, but I do want to note this, which is disappointing, given its purported support for ISO-8601:

```python
>>> arrow.get('2020-W02')
<Arrow [2020-01-01T00:00:00+00:00]>
```

(It should be 2020-01-08.)

branen: I read through `ciso8601` code and poked at it a little, and it looks great.  I definitely think we should use it in place of `arrow`.

That said, since we only support a subset ISO-8601, it'd be nice to phrase the error messages we use in a way that reflect this.  `ciso8601` implements something pretty close to RFC 3339, so I'll suggest some deltas that reference it.

branen: Sounds good!  (I agree on disallowing implicit loss of precision.)

graphql_compiler/schema/__init__.py
branen: I understand your argument, but I'm not convinced that a note in the docs is better than implementing our own strict parser--an important part of the compiler's value proposition is that it eliminates entire classes of weird surprises, so saying "don't do this or you might get weird surprises!" feels jarringly out of place to me.

eyurtsev: I would suggest to avoid implementing a separate ISO-8601 parser solely for this purpose. 

Hardly anyone needs support for all ISO-8601 formats, and the full spec is pretty challenging to implement (so no one in the python eco-system got it right as far as I know).

I would suggest offering support for 2-3 common formats (e.g., YYYY-MM-DDTHH:MM:SS and a few other variants), using arrow to parse those formats and checking that the length of the string being parsed matches the length of the format.

eyurtsev: Didn't see the thread above -- if ciso8601 works than that's great

branen: I agree that we don't need all of ISO-8601.

I'd be fine committing to `ciso8601`'s behavior, which has been stable for more than two years, and dropping the admonition about undefined behavior.  If @obi1kenobi or anyone else is uncomfortable about dropping the admonition when parsing is delegated to third-party code (which I think is a reasonable concern), then I suggest we implement a parser for RFC 3339's `date-time` and `full-date` rules, possibly with the following variance to [the RFC 3339 grammar](https://tools.ietf.org/html/rfc3339#section-5.6):
```
partial-time = time-hour ":" time-minute [time-second]
time-second  = ":" 2DIGIT [time-secfrac]
```

This wouldn't take much to implement, and I'd prefer it strongly to undefined behavior.  (Though, again, I'm fine just making `ciso8601`'s behavior part of our major-version contract.)

obi1kenobi: `ciso8601` is probably both faster and more likely to be correct than a parser we write 😄 

How about we explicitly state which of the many ISO 8601 formats we support:
- `YYYY-MM-DD` for dates
- `YYYY-MM-DDThh:mm:ss`, optionally suffixed with up to 6 decimal places of fractional seconds `.ssssss`, for datetimes

And then use `ciso8601` instead of `arrow` to parse them and call it a day? 😄

obi1kenobi: I don't think we can prevent `ciso8601` from parsing the other formats it supports, and this is where the undefined behavior in the spec is useful. We can guarantee parsing works for the formats we want to support, and everything else is left up to the implementation or completely undefined. It communicates that it's not a breaking change if we decide in the future to switch from `ciso8601` to another library that parses other "extras" in addition to the formats we promise to support.

I really don't think there's value in implementing our own parser, and I really don't think we stand to gain anything by making our spec tighter than "here are the formats we definitely support, we don't lose precision within them, everything else is undefined or implementation-defined."

graphql_compiler/schema/__init__.py
```diff
@@ -340,36 +340,47 @@ def _serialize_date(value: Any) -> str:
...
     """Deserialize a DateTime object from its proper ISO-8601 representation."""
-    # attempt to parse with microsecond information
-    try:
-        arrow_result = arrow.get(value, "YYYY-MM-DDTHH:mm:ss")
-    except arrow.parser.ParserMatchError:
-        arrow_result = arrow.get(value, "YYYY-MM-DDTHH:mm:ss.S")
-
-    # arrow parses datetime naive strings into Arrow objects with a UTC timezone.
-    return arrow_result.naive
+    if isinstance(value, datetime) and value.tzinfo is None:
```
LWprogramming, line 371: Should this also be checking for exact type equality like with `date` or do we expect subclasses of `datetime` to appear at some point? If the latter, does that mean we're trusting whatever library we choose (`arrow`, `ciso8601`, etc) to parse the subclasses correctly?

obi1kenobi: I removed our dependency on `arrow` and switched everything over to `ciso8601` which returns "real" Python `datetime` objects, so while I don't expect subclasses of `datetime` to appear, I don't think we necessarily need to prevent them either.

The reason we check for exact type equality in the `date` case is to prevent implicit loss of precision: since `datetime` is technically a subclass of `date`, `isinstance(some_datetime, date)` would return true but when it's used as a date, the time and time zone components would be implicitly lost. In the opposite direction, there is no associated loss of precision (the conversion is widening, not narrowing), so this is not really a concern. Additionally, `isinstance(some_date, datetime)` returns `False`, which is nice and as expected 😄

obi1kenobi:
Big thanks 🙏  to our past selves for having _extremely_ thorough unit tests on all the parsing and serialization stuff. This PR would have been 1000x more difficult to pull off safely without the exceptional test coverage we have.

TL;DR of the current state:
- Implicit widening conversions (no loss of precision) from date to datetime are allowed, both in their string representations and as Python objects.
- Datetime to date, or timezone-aware to timezone-naive datetime are narrowing conversions, and they remain disallowed since we do not want to implicitly lose precision.
- We switched to `ciso8601` for all our ISO 8601 parsing, and moved away from `arrow`. Since `pytz` was only required for Python 2, we were able to remove that dependency as well.
- We clarified the schema spec for the date and datetime types, explicitly promising support for particular string formats while allowing implementations to support a broader range of date or datetime representations. Our implementation supports all of `ciso8601`'s supported formats, but in principle if another implementation of our query protocol were defined in the future (e.g. in Rust or TypeScript), we'd consider it spec-compliant if it simply supported the guaranteed serialization formats — it would not have to support everything `ciso8601` supports.

obi1kenobi:
I had to upper-bound `bandit` due to this regression in `bandit==1.7.0`: https://github.com/PyCQA/bandit/issues/658

It seems like it's slated to be fixed in v1.7.1, so we should probably upgrade to it when that happens.

branen (approved):

PR #992, opened 2021-01-11
Fix bug involving cascading suppression error

LWprogramming:
Suppose a GraphQL schema contains the following types:

```
type Human  {
  name: String
  pet: Dog
}

type Dog {
  id: String
}
```

It would be illegal to suppress the type `Dog` because the type `Human` still has a field `pet` of type `Dog`, but if the field `pet` was also suppressed, this would be a legal schema transformation. Previously, the code would incorrectly raise a `CascadingSuppressionError` regardless of whether or not `pet` was suppressed-- this PR fixes that bug.


LWprogramming:
Something seems to have gone wrong with the pull request (frozen tests, no webhooks working etc) while creating it so I'm going to close and re-open to see if that fixes it

obi1kenobi (approved):

PR #994, opened 2021-01-15
Move builtin scalar renaming check to code rather than separate validation step

LWprogramming:
Schema renaming currently has some technical debt where there is a separate validation step before actually attempting to use the renamings to modify the schema. For this to work, it requires us to keep the validation code and the processing code in sync, which  we'd like to avoid in favor of attempting to modify the schema directly and gathering errors if something goes wrong.

In this PR, I move a validation step into the code that actually performs the renaming.

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -626,8 +626,8 @@ class RenameSchemaTypesVisitor(Visitor):
     type_renamed_to_builtin_scalar_conflicts: Dict[str, str]
 
     # Collects naming errors that arise from attempting to rename a builtin scalar. If
-    # type_renamings["String"] == "Foo" when there is a String field in the schema,
-    # illegal_builtin_scalar_renamings will map "String" to "Foo"
+    # type_renamings["String"] == "Foo" schema, illegal_builtin_scalar_renamings will contain
+    # "String"
     illegal_builtin_scalar_renamings: Set[str]
 
     # reverse_name_map maps renamed type name to original type name, containing all non-suppressed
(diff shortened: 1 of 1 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -642,6 +625,11 @@ class RenameSchemaTypesVisitor(Visitor):
     # "Foo" to "String"
     type_renamed_to_builtin_scalar_conflicts: Dict[str, str]
 
+    # Collects naming errors that arise from attempting to rename a builtin scalar. If
+    # type_renamings["String"] == "Foo" when there is a String field in the schema,
```
obi1kenobi, line 629: Should this be
```suggestion
    # type_renamings["String"] == "Foo" when there is a Foo field in the schema,
```
instead? As currently described, this should be a legal renaming since there's no stated conflict on `Foo` and the `String` type exists.

Consider clarifying the comment here and above at `type_renamed_to_builtin_scalar_conflicts` to more explicitly and precisely describe what the conflict is: is the issue that `String` is a built-in and cannot be renamed, or is the problem that the renaming is illegal since it's attempting to rename to a name that already exists, or something else entirely.

obi1kenobi (changes requested):

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: Originally when I made this change I was thinking that if `type_renamings` attempted to rename `String` when there were no `String` fields in the schema, it'd be covered by the no-op renaming error case but after re-reading I think it makes more sense just to check all the renamings up-front as before (instead of collecting them in the process of traversing the schema) since this is a rule that is really unlikely to change anytime soon and it'd be nice to know that this is illegal immediately instead of only when a schema contains a field of that builtin type

obi1kenobi (approved):

PR #995, opened 2021-02-03
Refactor cascading validation

LWprogramming:
As described in PR #994:

> Schema renaming currently has some technical debt where there is a separate validation step before actually attempting to use the renamings to modify the schema. For this to work, it requires us to keep the validation code and the processing code in sync, which we'd like to avoid in favor of attempting to modify the schema directly and gathering errors if something goes wrong.

This PR continues cleaning up this technical debt. It takes cascading suppression errors and moves the checks into the renaming code rather than having a separate validation step, simplifying the renaming process.

This PR lets us remove the entire `CascadingSuppressionCheckVisitor` class. If it'd be significantly easier to understand the changes here if I tried moving the functionality out of the `CascadingSuppressionCheckVisitor` piecemeal, just let me know and I can do that-- here I decided not to do that initially since all the code in the visitor follows the same pattern of "`foo` in the schema depends on `bar` where `bar` is to be suppressed".

### DIFF
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -787,9 +787,14 @@ class RenameSchemaTypesVisitor(Visitor):
         for field_node in node.fields:
             field_name = field_node.name.value
             field_type_name = get_ast_with_non_null_and_list_stripped(field_node.type).name.value
-            field_type_suppressed = self.type_renamings.get(field_type_name, field_type_name) is None
-            field_node_suppressed = type_name in self.field_renamings and self.field_renamings[type_name].get(field_name, {field_name}) == set()
-            if field_type_suppressed and not(field_node_suppressed):
+            field_type_suppressed = (
+                self.type_renamings.get(field_type_name, field_type_name) is None
+            )
+            field_node_suppressed = (
+                type_name in self.field_renamings
+                and self.field_renamings[type_name].get(field_name, {field_name}) == set()
+            )
+            if field_type_suppressed and not (field_node_suppressed):
                 # If the type of the field is suppressed but the field itself is not, it's invalid.
                 current_type_fields_to_suppress[field_name] = field_type_name
         if current_type_fields_to_suppress != {}:
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -368,39 +368,32 @@ def _rename_and_suppress_types_and_fields(
         )
         raise CascadingSuppressionError("\n".join(error_message_components))
     if visitor.invalid_type_names or visitor.invalid_field_names:
-        explanation = (
+        error_message_components = [
             "Applying the renaming would involve names that are not valid, non-reserved "
             "GraphQL names. Valid, non-reserved GraphQL names must consist of only alphanumeric "
             "characters and underscores, must not start with a numeric character, and must not "
             "start with double underscores."
-        )
-        invalid_type_names_message = None
+        ]
         if visitor.invalid_type_names:
             sorted_invalid_type_names = sorted(visitor.invalid_type_names.items())
-            invalid_type_names_message = (
+            error_message_components.append(
                 f"The following is a list of tuples that describes what needs to be fixed for type "
                 f"renamings. Each tuple is of the form (original_name, invalid_new_name) where "
                 f"original_name is the name in the original schema and invalid_new_name is what "
                 f"original_name would be renamed to: {sorted_invalid_type_names}"
             )
-        invalid_field_names_message = None
         if visitor.invalid_field_names:
             sorted_invalid_field_names = [
                 (type_name, sorted(field_renamings.items()))
                 for type_name, field_renamings in sorted(visitor.invalid_field_names.items())
             ]
-            invalid_field_names_message = (
+            error_message_components.append(
                 f"The following is a list of tuples that describes what needs to be fixed for "
                 f"field renamings. Each tuple is of the form (type_name, field_renamings) "
                 f"where type_name is the name of the type in the original schema and "
                 f"field_renamings is a list of tuples mapping the original field name to the "
                 f"invalid GraphQL name it would be renamed to: {sorted_invalid_field_names}"
             )
-        error_message_components = [
-            explanation,
-            invalid_type_names_message,
-            invalid_field_names_message,
-        ]
         raise InvalidNameError("\n".join([i for i in error_message_components if i is not None]))
     if (
         visitor.type_name_conflicts
--- a/graphql_compiler/schema_transformation/rename_schema.py
+++ b/graphql_compiler/schema_transformation/rename_schema.py
@@ -235,7 +235,7 @@ def rename_schema(
     schema = build_ast_schema(schema_ast)
     query_type = get_query_type_name(schema)
 
-    _validate_renamings(schema_ast, type_renamings, field_renamings, query_type)
+    _ensure_no_unsupported_suppressions(schema_ast, type_renamings)
 
     # Rename types, interfaces, enums, unions and suppress types, unions
     schema_ast, reverse_name_map, reverse_field_name_map = _rename_and_suppress_types_and_fields(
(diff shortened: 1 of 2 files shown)


graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -235,7 +235,7 @@ def rename_schema(
     schema = build_ast_schema(schema_ast)
     query_type = get_query_type_name(schema)
 
-    _validate_renamings(schema_ast, type_renamings, field_renamings, query_type)
+    _ensure_no_unsupported_suppressions(schema_ast, type_renamings)
```
LWprogramming, line 238: After this PR, the only separate validation step left would be this one, so there's no need for a special `_validate_renamings` function.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -415,40 +327,73 @@ def _rename_and_suppress_types_and_fields(
...
             "GraphQL names. Valid, non-reserved GraphQL names must consist of only alphanumeric "
             "characters and underscores, must not start with a numeric character, and must not "
             "start with double underscores."
-        )
-        invalid_type_names_message = None
+        ]
         if visitor.invalid_type_names:
             sorted_invalid_type_names = sorted(visitor.invalid_type_names.items())
-            invalid_type_names_message = (
+            error_message_components.append(
```
LWprogramming, line 379: Needing to make the changes here surprised me. For some reason, before adding the `CascadingSuppressionError` error message above, `mypy` was fine with setting these temporary strings like `invalid_type_names_message` to `None`.

However, after adding the seemingly-unrelated `CascadingSuppressionError` code, `mypy` raised an error here because we call `"\n".join()` on a list of strings and it pointed out that initializing them to `None` meant trying to join items of `Optional[str]` type rather than `str`. Is there a known reason why it didn't catch this earlier, or is this likely a bug in `mypy` to check out?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -793,8 +756,44 @@ def _rename_or_suppress_or_ignore_name_and_add_to_record(
...
+
+        Args:
+            node: type node with fields to be renamed
+
+        Returns:
+            node with fields renamed if there are no cascading suppression issues.
+        """
         type_name = node.name.value
+        current_type_fields_to_suppress = {}
+        for field_node in node.fields:
```
LWprogramming, line 777: This loop strongly resembles the one below on line 806, but I'm not sure of how to organize it best. The problem here is with the function structure. Let's say we're at a type `T` with a field `foo`, and `foo` is of type Bar`. Then this function is essentially doing the following:

1. For each field `foo` of type `Bar`: Is the type `Bar` going to be suppressed? If so, and if `foo` _isn't_ going to be suppressed, the renamings are invalid and we should do the bare minimum necessary to continue the schema renaming and collect any other independent errors while recording the problem. Return from this node-- no more to be done here.
2. Do we expect to renaming `T`'s fields at all? If not, we're done and can return.
3. For each field `foo`: rename `foo` if necessary.

I considered trying something along the lines of implementing step 2 via "if `T`'s fields are not to be renamed, this is equivalent to using the identity mapping for all of `T`'s fields with the reasoning that this means the loops in steps 1 and 3 can probably combined into one single loop, but it's not clear to me that this is necessarily easier to understand. Would definitely invite any feedback!

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: Likely shortcoming of mypy of some sort. Likely not worth tracking down, in terms of cost-benefit.

graphql_compiler/schema_transformation/rename_schema.py
obi1kenobi: This PR comment would be useful to integrate into the code itself. The logic around line 780 is quite complex and it's not really obvious from the source code itself what is being done and why.

In particular, I'd love to see complex expressions like `self.type_renamings.get(field_type_name, field_type_name) is None` get assigned to well-named local variables, because they are complex and surprising — in this case, that's because it's unusual to see a `.get()` check with an explicit non-`None` default value be checked against a `None` result.

What I'd recommend is writing something similar to your PR comment as a code comment in a continuous block just above this for loop, and then reference the individual parts of the algorithm the comment describes from the appropriate place inside the for loop. Together with good naming of intermediate expression values, I think this code will become much clearer.

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -844,10 +843,24 @@ def _rename_fields(self, node: ObjectTypeDefinitionNode) -> ObjectTypeDefinition
...
 
+    def leave_union_type_definition(
+        self,
+        node: UnionTypeDefinitionNode,
+        key: Any,
+        parent: Any,
+        path: List[Any],
+        ancestors: List[Any],
+    ) -> None:
+        """Check that each union still has at least one non-suppressed member."""
```
obi1kenobi, line 860: nit: This docstring doesn't really capture what's going on. The implementation does an action if that check fails, but the docstring doesn't make any mention of that. 

Consider instead:
```suggestion
        """Suppress unions that do not have any remaining non-suppressed members."""
```

obi1kenobi (changes requested):

graphql_compiler/schema_transformation/rename_schema.py
LWprogramming: The action that happens if the check fails is actually raising an error instead of automatically suppressing the union (since renaming is defined so that unions in such situations have to get explicitly suppressed) and the error handling code is outside of the visitor itself, so I'm inclined to keep the docstring as-is. Worth adding a bit to mention what the `union_types_to_suppress` field is going to be used for, even if it doesn't happen in this function itself?

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -772,15 +772,29 @@ def _rename_fields(self, node: ObjectTypeDefinitionNode) -> ObjectTypeDefinition
...
         """
+        # Field renaming takes place in three steps:
+        # 1. For each field foo of type Bar: if Bar is going to be suppressed but foo isn't going to
+        #    be suppressed, the renamings are invalid. In that case, we should do the bare minimum
+        #    necessary to continue the schema renaming and collect any other errors that occur
+        #    independently while recording the problem. Once done, we return from the current type
+        #    because there is no more to be done for this particular type's fields.
+        # 2. If the current type has no field renamings at all, we return immediately after doing
+        #    nothing.
+        # 3. If the current type does have field renamings, apply those field renamings.
```
obi1kenobi, line 783: Consider making the step enumeration visually exactly match the references to the steps. That way, the steps are numbered and succinctly described, with a more detailed description up front and with a very clear structure that shows that the later `Step N` comments are linked to the steps as described earlier.
```suggestion
        # Step 1: detect fields that depend on types that were suppressed.
        #    For each field foo of type Bar: if Bar is going to be suppressed but foo isn't going to
        #    be suppressed, the renamings are invalid. In that case, we should do the bare minimum
        #    necessary to continue the schema renaming and collect any other errors that occur
        #    independently while recording the problem. Once done, we return from the current type
        #    because there is no more to be done for this particular type's fields.
        # Step 2: return immediately if we're not going to be renaming any fields.
        #    If the current type has no field renamings at all, we return immediately after doing
        #    nothing.
        # Step 3: apply the field renamings to each field.
        #    If the current type does have field renamings, apply those field renamings.
```

graphql_compiler/schema_transformation/rename_schema.py
```diff
@@ -772,15 +772,29 @@ def _rename_fields(self, node: ObjectTypeDefinitionNode) -> ObjectTypeDefinition
...
-            if self.type_renamings.get(field_type_name, field_type_name) is None and not (
+            field_type_suppressed = (
+                self.type_renamings.get(field_type_name, field_type_name) is None
+            )
+            field_node_suppressed = (
                 type_name in self.field_renamings
                 and self.field_renamings[type_name].get(field_name, {field_name}) == set()
-            ):
+            )
+            if field_type_suppressed and not (field_node_suppressed):
```
obi1kenobi, line 797:
```suggestion
            if field_type_suppressed and not field_node_suppressed:
```

obi1kenobi (commented):
It isn't clear to me whether you were done tweaking this PR — when you're done, hit that "re-request review" button next to my name in the reviewers list to put this PR back on my review queue.

I have two minor suggestions, and otherwise this seems fine to merge.

obi1kenobi (approved):
Looks great!

PR #997, opened 2021-02-09
Update python-mysqldb package name for Ubuntu 20.04

LWprogramming:
Our scripts use `ubuntu-latest`, which is [migrating](https://github.com/actions/virtual-environments/issues/1816) from 18.04 to 20.04. The package [python-mysqldb](https://packages.ubuntu.com/bionic/python-mysqldb) in 18.04 has been renamed to [python3-mysqldb](https://packages.ubuntu.com/focal/python3-mysqldb) in 20.04. This PR makes that update for our CI scripts.

### DIFF
--- a/scripts/install_ubuntu_ci_core_dependencies.sh
+++ b/scripts/install_ubuntu_ci_core_dependencies.sh
@@ -18,7 +18,7 @@ set -e
 bash -c "wget -qO- https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -"
 sudo add-apt-repository "$(wget -qO- https://packages.microsoft.com/config/ubuntu/"$(lsb_release -r -s)"/prod.list)"
 sudo apt-get update
-sudo apt-get install unixodbc-dev python-mysqldb libmysqlclient-dev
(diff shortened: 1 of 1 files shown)


LWprogramming:
I've looked into the Codecov issue and it looks like the combination of several things happening in a really unfortunate way:

1. PR #994 was fine while in its separate branch but then CI ran into an intermittent (possibly network-related) issue after it was merged into main. The failed CI means that Codecov never got a report for the main branch after #994 was merged.
2. Opening this PR means Codecov has no [base commit](https://docs.codecov.io/docs/comparing-commits) report to compare this PR to, so Codecov is reporting incorrect results because to Codecov this PR (#997) appears _two_ PRs ahead of main (because it includes #994 and #997).
3. I tried fixing this by re-running the CI on main, but that doesn't work for the same reason this PR (#997) exists-- because the CI environment has gone from Ubuntu 18.04 to 20.04, so the package installation scripts that currently exist need to get changed.

It looks fine to me though, despite what Codecov says. Just didn't want to cause any confusion about the seemingly-failing test. :)

bojanserafimov (approved):

PR #998, opened 2021-02-16
Add edges defined by a composite join clause

bojanserafimov:
Allows us to define edges towards vertices with composite primary key and traverse those edges. Fold and recurse for composite traversals will not be implemented in this PR.

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -906,7 +906,7 @@ class CompilationState(object):
 
         on_clause = sqlalchemy.and_(*(
             parent_alias.c[from_column] == self._current_alias.c[to_column]
-            for from_column, to_column in matching_column_pairs
+            for from_column, to_column in sorted(matching_column_pairs)
         ))
 
         # Join to where we came from.
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -21,7 +21,12 @@ from sqlalchemy.sql.selectable import FromClause, Join, Select
 from . import blocks
 from ..global_utils import VertexPath
 from ..schema import COUNT_META_FIELD_NAME
-from ..schema.schema_info import DirectJoinDescriptor, SQLAlchemySchemaInfo, CompositeJoinDescriptor, JoinDescriptor
+from ..schema.schema_info import (
+    CompositeJoinDescriptor,
+    DirectJoinDescriptor,
+    JoinDescriptor,
+    SQLAlchemySchemaInfo,
+)
 from .compiler_entities import BasicBlock
 from .compiler_frontend import IrAndMetadata
 from .expressions import ContextField, Expression
@@ -147,14 +152,8 @@ def _find_used_columns(
                 columns_at_location = {edge.from_column}
                 columns_at_child = {edge.to_column}
             elif isinstance(edge, CompositeJoinDescriptor):
-                columns_at_location = {
-                    column_pair[0]
-                    for column_pair in edge.column_pairs
-                }
-                columns_at_child = {
-                    column_pair[1]
-                    for column_pair in edge.column_pairs
-                }
+                columns_at_location = {column_pair[0] for column_pair in edge.column_pairs}
+                columns_at_child = {column_pair[1] for column_pair in edge.column_pairs}
 
             used_columns.setdefault(get_vertex_path(location), set()).update(columns_at_location)
             used_columns.setdefault(get_vertex_path(child_location), set()).update(columns_at_child)
@@ -873,7 +872,7 @@ class CompilationState(object):
         elif isinstance(join_descriptor, CompositeJoinDescriptor):
             matching_column_pairs = join_descriptor.column_pairs
         else:
-            raise AssertionError(u"TODO")
+            raise AssertionError("TODO")
 
         non_null_column = sorted(matching_column_pairs)[0][1]
         self._came_from[self._current_alias] = self._current_alias.c[non_null_column]
@@ -904,10 +903,12 @@ class CompilationState(object):
                 )
             )
 
-        on_clause = sqlalchemy.and_(*(
-            parent_alias.c[from_column] == self._current_alias.c[to_column]
-            for from_column, to_column in sorted(matching_column_pairs)
-        ))
+        on_clause = sqlalchemy.and_(
+            *(
+                parent_alias.c[from_column] == self._current_alias.c[to_column]
+                for from_column, to_column in sorted(matching_column_pairs)
+            )
+        )
 
         # Join to where we came from.
         self._from_clause = self._from_clause.join(
@@ -1045,7 +1046,6 @@ class CompilationState(object):
                 f"Cannot recurse when _current_location is not a Location. _current_location "
                 f"was set to {self._current_location}."
             )
-
 
         edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
         if not isinstance(edge, DirectJoinDescriptor):
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -155,7 +155,7 @@ def _find_used_columns(
                 columns_at_location = {column_pair[0] for column_pair in edge.column_pairs}
                 columns_at_child = {column_pair[1] for column_pair in edge.column_pairs}
             else:
-                raise AssertionError("TODO")
+                raise AssertionError(f"Unknown join descriptor type {edge}: {type(edge)}")
 
             used_columns.setdefault(get_vertex_path(location), set()).update(columns_at_location)
             used_columns.setdefault(get_vertex_path(child_location), set()).update(columns_at_child)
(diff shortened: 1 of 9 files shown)


graphql_compiler/compiler/emit_sql.py
```diff
@@ -143,21 +143,34 @@ def _find_used_columns(
...
+                columns_at_child = {edge.to_column}
+            elif isinstance(edge, CompositeJoinDescriptor):
+                columns_at_location = {
+                    column_pair[0]
+                    for column_pair in edge.column_pairs
+                }
+                columns_at_child = {
+                    column_pair[1]
+                    for column_pair in edge.column_pairs
+                }
```
obi1kenobi, line 157: Is the `else` here a `pass` or a `raise AssertionError` kind of `else`? Not sure what the inferred type of `edge` is on line 145, but hoping we can make that explicit.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -879,10 +904,15 @@ def _join_to_parent_location(
                 )
             )
 
+        on_clause = sqlalchemy.and_(*(
+            parent_alias.c[from_column] == self._current_alias.c[to_column]
+            for from_column, to_column in matching_column_pairs
+        ))
```
obi1kenobi, line 910: You've probably tested this already, so please excuse my ignorance — what does `sqlalchemy.and_` do when the list has only one element? Presumably it outputs it as-is?

But what about having no elements? And is that ever possible?

graphql_compiler/compiler/emit_sql.py
```diff
@@ -851,7 +865,18 @@ def _join_to_parent_location(
...
+        if isinstance(join_descriptor, DirectJoinDescriptor):
+            matching_column_pairs = {
+                (join_descriptor.from_column, join_descriptor.to_column),
+            }
+        elif isinstance(join_descriptor, CompositeJoinDescriptor):
+            matching_column_pairs = join_descriptor.column_pairs
+        else:
+            raise AssertionError(u"TODO")
+
+        non_null_column = sorted(matching_column_pairs)[0][1]
```
obi1kenobi, line 878: The `[0][1]` seems a bit less than ideal, care to add a comment to it?

Is it ever possible that `matching_column_pairs` is empty? I'm assuming its type is something like `AbstractSet[Tuple[str, str]]` so the `[1]` should always be safe.

obi1kenobi (commented):
Seems like a very reasonable approach. Left a couple of nitpicks that I'm sure you're already working on, just wanted to flag them before I forgot them.

graphql_compiler/compiler/emit_sql.py
bojanserafimov: It's the identity function for 1 argument. Old tests would fail if it did something else.

I'll add an assertion to check for 0 elements.

graphql_compiler/compiler/emit_sql.py
bojanserafimov: Yes! Ty for catching it.

graphql_compiler/tests/test_helpers.py
```diff
@@ -842,12 +846,22 @@ def get_sqlalchemy_schema_info(dialect: str = "mssql") -> SQLAlchemySchemaInfo:
...
-            "in_{}".format(edge["name"])
-        ] = DirectJoinDescriptor(edge["to_column"], edge["from_column"])
+        if "column_pairs" in edge:
+            join_descriptors.setdefault(edge["from_table"], {})[
+                "out_{}".format(edge["name"])
+            ] = CompositeJoinDescriptor(edge["column_pairs"])
+            join_descriptors.setdefault(edge["to_table"], {})[
+                "in_{}".format(edge["name"])
+            ] = CompositeJoinDescriptor(
+                {(to_column, from_column) for from_column, to_column in edge["column_pairs"]}
```
LWprogramming, line 856: Just to check, should this be `(to_column, from_column)` or `(from_column, to_column)`?

bojanserafimov: It's correct. This is the edge in the backwards direction. I'll add a comment to explain here.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1017,6 +1060,11 @@ def recurse(self, vertex_field: str, depth: int) -> None:
             )
 
         edge = self._sql_schema_info.join_descriptors[self._current_classname][vertex_field]
+        if not isinstance(edge, DirectJoinDescriptor):
+            raise NotImplementedError(
+                f"Edge {vertex_field} is backed by a CompositeJoinDescriptor, "
+                "so it can't be used with @recurse."
```
obi1kenobi, line 1066: I believe this is a user-facing error message, so let's try to avoid the parts of our terminology that are not suitable for querying users. How about something like this:
```suggestion
                f"Edge {vertex_field} requires a JOIN across a composite key, which is currently "
                f"not supported for use with @recurse."
```

graphql_compiler/compiler/emit_sql.py
```diff
@@ -1131,6 +1179,11 @@ def fold(self, fold_scope_location: FoldScopeLocation) -> None:
         join_descriptor = self._sql_schema_info.join_descriptors[self._current_classname][
             full_edge_name
         ]
+        if not isinstance(join_descriptor, DirectJoinDescriptor):
+            raise NotImplementedError(
+                f"Edge {full_edge_name} is backed by a CompositeJoinDescriptor, "
+                "so it can't be used with @fold."
```
obi1kenobi, line 1185: Same reasoning as above, here's the equivalent error message here that follows my template from above:
```suggestion
                f"Edge {vertex_field} requires a JOIN across a composite key, which is currently "
                f"not supported for use with @fold."
```

graphql_compiler/schema/schema_info.py
```diff
@@ -20,20 +20,46 @@
...
+        origin_table.from_column_2 == destination_table.to_column_2 AND
+        origin_table.from_column_3 == destination_table.to_column_3
+
+    The type of join (inner vs left, etc.) is not specified.
+    The tables are not specified.
+    """
+
+    # (from_column, to_column) pairs, where from_column is on the origin table
+    # and to_column is on the destination table of the join.
+    column_pairs: Set[Tuple[str, str]]
```
obi1kenobi, line 54: Consider `FrozenSet` or `AbstractSet` instead here. `Set` is literally `set()`, the explicitly-mutable kind. `AbstractSet` is sufficient to make it immutable-in-practice because abstract sets don't have mutability functions defined and `mypy` will prevent mutations in typed code unless you literally `isinstance` or `cast` to a mutable set.

graphql_compiler/schema/schema_info.py
```diff
@@ -20,20 +20,46 @@
...
+
+    def __post_init__(self) -> None:
+        """Validate fields."""
+        if not self.column_pairs:
+            raise AssertionError("The column_pairs field is expected to be non-empty.")
```
obi1kenobi, line 59: Very nice!

graphql_compiler/tests/test_helpers.py
```diff
@@ -772,82 +775,91 @@ def get_sqlalchemy_schema_info(dialect: str = "mssql") -> SQLAlchemySchemaInfo:
...
+    # in and out directions.
+    join_descriptors: Dict[str, Dict[str, JoinDescriptor]] = {}
     for edge in edges:
-        join_descriptors.setdefault(edge["from_table"], {})[
-            "out_{}".format(edge["name"])
-        ] = DirectJoinDescriptor(edge["from_column"], edge["to_column"])
-        join_descriptors.setdefault(edge["to_table"], {})[
-            "in_{}".format(edge["name"])
-        ] = DirectJoinDescriptor(edge["to_column"], edge["from_column"])
+        column_pairs = cast(Set[Tuple[str, str]], edge["column_pairs"])
```
obi1kenobi, line 843: This cast here is needed because the `edges` type is not specified and is probably inferred incorrectly. This is test code and is correct, so I don't have strong feelings one way or the other — just wanted to flag it in case you wanted to reconsider the alternative.

graphql_compiler/tests/test_ir_generation.py
```diff
@@ -6259,3 +6259,7 @@ def test_recursive_field_type_is_subtype_of_parent_field(self):
         }
 
         check_test_data(self, test_data, expected_blocks, expected_location_types)
+
+    def test_animal_born_at_traversal(self):
+        """Ensure that sql composite key traversals work."""
+        # skipping ir generation testing since this is a sql-specific test
```
obi1kenobi, line 6265: Do you mind adding the expected blocks anyway? This feels like subverting the "all common test data inputs have matching IR and code generation tests" decision we made, and which is also there to help us debug code generation issues because we immediately know whether the issue is due to changes in IR generation or after it.

graphql_compiler/tests/test_helpers.py
bojanserafimov: Yeah there's no way to add an inline type hint that will fix this, and I don't want to blow up this PR

obi1kenobi (approved):

PR #1000, opened 2021-02-25
Fix MSSQL Decimal bug

bojanserafimov:
We interpret `sqlalchemy.DECIMAL` as `GraphQLDecimal`, and `decimal.Decimal` is an allowed type for input. The problem is that `sqlalchemy.DECIMAL` raises an error when the input argument is has too many decimals, which `decimal.Decimal` allows. My test case in this PR reproduces this bug.

A bigger problem is that, `decimal.Decimal(float(1.1))` is `Decimal('1.100000000000000088817841970012523233890533447265625')`, so deserializing a float value as `GraphQLDecimal` can inflate its precision and cause errors where they shouldn't exist.

TODO:
- fix the problem


obi1kenobi:
Is the plan to cap the precision of our `GraphQLDecimal` type, or is there a better solution?

bojanserafimov:
> Is the plan to cap the precision of our `GraphQLDecimal` type, or is there a better solution?

Or make sure "1.1" goes back to "1.1" when serialized and deserialized. IMO the second part of this problem is bad, and the first is maybe not even a problem. If someone intentionally queries with a super precise float maybe it's better to get an error than to get a result.

graphql_compiler/schema/__init__.py
```diff
@@ -415,6 +415,17 @@ def _parse_datetime_value(value: Any) -> datetime:
...
 
 
+def _parse_decimal_value(value: Any) -> Decimal:
+    """Parse decimal value."""
+    if isinstance(value, float):
+        # Decimal(float(1.1)) is Decimal('1.100000000000000088817841970012523233890533447265625'),
+        # which is probably not what the user intended. Inflating the intended level of precision
+        # like this can cause errors with database types that are not configured with precision
+        # that high. Just use Decimal or str values if you're serious about precision.
+        return Decimal(str(value))
```
bojanserafimov, line 425: I'm not yet 100% sure on this approach. An alternative is to disallow floats for GraphQLDecimal.

bojanserafimov:
Best solution for now is to use `str` values for `GraphQLDecimal` fields.

PR #1008, opened 2021-04-16
[Query Planning Refactor]: SimpleExecute

chewselene:
This is part of a larger refactor to streamline the way we generate query plans. This is the simplest form of query plan where the given query is executed against the database.

Future work includes adding paginated queries and cross database queries.

There are some outstanding TODOs mainly around how to best utilize provider_id to look up both the corresponding schema and corresponding database query execution function.

### DIFF
--- a/graphql_compiler/query_planning/make_query_plan.py
+++ b/graphql_compiler/query_planning/make_query_plan.py
@@ -18,13 +18,9 @@ from graphql.language.ast import (
     StringValueNode,
 )
 from graphql.pyutils import FrozenList
-from sqlalchemy.dialects.mssql.base import MSDialect
 
 from ..ast_manipulation import get_only_query_definition
 from ..compiler.common import (
-    CYPHER_LANGUAGE,
-    GREMLIN_LANGUAGE,
-    MATCH_LANGUAGE,
     compile_graphql_to_cypher,
     compile_graphql_to_gremlin,
     compile_graphql_to_match,
@@ -36,7 +32,7 @@ from ..global_utils import QueryStringWithParameters
 from ..schema import FilterDirective, OutputDirective
 from ..schema.schema_info import CommonSchemaInfo, SQLAlchemySchemaInfo
 from ..schema_transformation.split_query import AstType, SubQueryNode
-from .typedefs import ProviderMetadata, QueryPlan, SimpleExecute
+from .typedefs import BackendType, ProviderMetadata, QueryPlan, SimpleExecute
 
 
 @dataclass
@@ -85,7 +81,7 @@ def _make_simple_execute_node(
     query_and_parameters: QueryStringWithParameters,
     schema_info: Union[CommonSchemaInfo, SQLAlchemySchemaInfo],
     provider_id: str,
-    backend_type: Optional[str],
+    backend_type: BackendType,
 ) -> SimpleExecute:
     """Generate a SimpleExecuteNode.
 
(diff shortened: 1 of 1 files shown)


graphql_compiler/query_planning_and_execution/make_query_plan.py
```diff
@@ -291,6 +327,118 @@ def _get_in_collection_filter_directive(input_filter_name: str) -> DirectiveNode
...
+######
+# Public API
+######
+
+
+def generate_query_plan(
+    query_and_parameters: QueryStringWithParameters,
+    provider_id: str,
+    *,
+    desired_page_size: Optional[int] = 5000,
```
obi1kenobi, line 355: I'm kind of on the fence about setting a default here. The right number will depend on each user's statistics collection configuration — aside from some special cases, we can't generate pages that are more fine-grained than the statistics backing them. Within Kensho, we always know and control what that configuration might be. But in general, we might not.

On the flip side, not setting a number would make pagination disabled by default, and that's also a default value — and not a good one. Even if the statistics aren't fine-grained enough, the compiler will emit advisories when paginating which should be an obvious and "soft" way to communicate the issue. And 5000 is a pretty decent compromise between large page sizes (for high throughput) and not blowing up the server or client.

What do you think?

graphql_compiler/query_planning_and_execution/__init__.py
```diff
@@ -0,0 +1 @@
+# Copyright 2021-present Kensho Technologies, LLC.
```
obi1kenobi, line 1: This module name is a bit of a mouthful :) Would it be worth trying to come up with something shorter?

obi1kenobi (commented):
This looks quite good, thank you for putting it together!

Since this is still a draft and we'll want a couple more iterations to finalize everything, it might be good to try splitting it up into a few parts. For example, the following parts have straightforward mutual dependencies and could in principle merge separately, shrinking the code size and making iteration and reviews easier:
- the `namedtuple -> dataclass` part
- the "move existing planning functionality to a new module" part
- the new typedefs
- the code that uses the new typedefs to do query planning

graphql_compiler/query_planning_and_execution/make_query_plan.py
```diff
@@ -66,44 +81,65 @@ class QueryPlanDescriptor:
...
-    _make_query_plan_recursive(root_sub_query_node, root_sub_query_plan, output_join_descriptors)
+    """
+    if isinstance(schema_info, SQLAlchemySchemaInfo):
+        # Compiled SQL query.
+        compilation_result = compile_graphql_to_sql(schema_info, query_and_parameters.query_string)
+        query = print_sqlalchemy_query_string(compilation_result.query, schema_info.dialect)
+
+        provider_metadata = ProviderMetadata(
+            backend_type=schema_info.dialect.name,
+            requires_fold_postprocessing=isinstance(schema_info.dialect, MSDialect),
```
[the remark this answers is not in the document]
chewselene: I like this idea! I will work on a separate PR to incorporate.

graphql_compiler/query_planning_and_execution/make_query_plan.py
```diff
@@ -66,44 +81,65 @@ class QueryPlanDescriptor:
...
-    equal to the child's out name as specified in the QueryConnection. The newly added @filter
-    will be a 'in_collection' type filter, and the name of the local variable is guaranteed to
-    be the same as the out_name of the @output on the parent.
-
-    ASTs contained in the input node and its children nodes will not be modified.
+def _make_simple_execute_node(
+    query_and_parameters: QueryStringWithParameters,
+    schema_info: Union[CommonSchemaInfo, SQLAlchemySchemaInfo],
+    provider_id: str,
+    backend_type: Optional[str],
```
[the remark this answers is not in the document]
chewselene: I switched this to use the `BackendType` enum so I think it makes sense for it to not be optional. It is still a bit weird because I currently check if `schema_info` is of type `SQLAlchemySchemaInfo` that `backend_type` is either MSSQL or PostgreSQL, but I don't check if it matches the dialect in the schema info. If I add the check, we could run into the problem that you mentioned

>  SQLAlchemy version adds a new dialect for MSSQL that doesn't inherit from MSDialect — our code immediately becomes wrong in yet another place

so I'm not sure what the best solution is...

graphql_compiler/query_planning_and_execution/make_query_plan.py
chewselene: I guess the alternative is that we make this a required param? Do you think it would be better to force the user to make a conscious decision here since it does depend on user stats?

graphql_compiler/query_planning/make_query_plan.py
```diff
@@ -66,6 +77,80 @@ class QueryPlanDescriptor:
...
+            )
+        elif backend_type == BackendType.gremlin:
+            compilation_result = compile_graphql_to_gremlin(
+                schema_info, query_and_parameters.query_string
+            )
+        elif backend_type == BackendType.match:
+            compilation_result = compile_graphql_to_match(
+                schema_info, query_and_parameters.query_string
+            )
+        # TODO(selene): add InterpreterAdapter based backends
```
chewselene, line 133: I'm unsure of how to deal with InterpreterAdapters. The [public methods](https://github.com/kensho-technologies/graphql-compiler/blob/interpreted_mode_v3/graphql_compiler/interpreter/api.py#L117) are `interpret_ir` and `interpret_query`. `SimpleExecute` is set up to contain a query string, which is the input of `interpret_query`, but that function also requires the schema to be passed. It seems like to want to use the schema to compile the query in this function, but then our output (the equivalent of the `query` in `SimpleExecute`) would be `IrAndMetadata` rather than a query string. Making the type of `query` be `Union[str, IrAndMetadata]` doesn't really seem to make sense, but maybe it would be more tolerable if we renamed `query` to `backend_input` or something similar? Thoughts?

kopetsss (approved):

PR #1009, opened 2021-04-20
SQLalchemy version 1.4 update

LWprogramming:
The update to SQLalchemy version 1.4 would lead to some parts of compiler functionality breaking, and this is a draft PR to try to fix that. No need to look at this right now, just putting it up so that it lives somewhere besides locally.


graphql_compiler/compiler/emit_sql.py
```diff
@@ -1187,7 +1191,12 @@ def fold(self, fold_scope_location: FoldScopeLocation) -> None:
 
         # 1. Get fold metadata.
         # Location of vertex that is folded on.
-        outer_alias = self._current_alias.alias()
+        if isinstance(self._current_alias, CTE):
```
chewselene, line 1194: `self._current_alias` is typed `Optional[Alias]`, but I don't think `CTE` is a type of `Alias`. Seems weird that that doesn't cause a mypy error, but perhaps the type of `self._current_alias` ought to updated?

LWprogramming: Based on `reveal_type`, somewhere along the line the type of `self._current_alias` is turning into `Any`-- looking into why now.

LWprogramming:
A short summary of this issue for any future readers:
* SQLAlchemy update to version 1.4 led to changes in alias generation (as shown by failing compiler and integration tests).
* Alias generation for SQL queries is difficult to get right when various compiler directives enter the picture, particularly fold and recurse. There's a tangle of complications, e.g. type hints getting confused as described earlier in this PR.
* Active development on the SQL backend support here is winding down in the very near future, reducing the priority of SQLAlchemy-based development.
* We could attempt to separate the parts of the compiler that require SQLAlchemy from those that don't, but there are fairly foundational parts of the compiler codebase that rely on SQLAlchemy (for example, definition of [`Expression`](https://github.com/kensho-technologies/graphql-compiler/blob/main/graphql_compiler/compiler/expressions.py)), so we'll be keeping the SQLAlchemy version to < 1.4.0 for simplicity.

PR #1010, opened 2021-04-20
Convert SQLAlchemySchemaInfo to a dataclass

chewselene:
In addition to converting the namedtuple to a dataclass, this PR also:

-  adds a comment about the difference between `SQLAlchemySchemaInfo` and `SQLSchemaInfo`
-  updates a test file to make it mypy compliant

This is a part of the refactor in #1008

### DIFF
--- a/graphql_compiler/compiler/emit_sql.py
+++ b/graphql_compiler/compiler/emit_sql.py
@@ -632,7 +632,7 @@ class FoldSubqueryBuilder(object):
 
     def add_traversal(
         self,
-        join_descriptor: DirectJoinDescriptor,
+        join_descriptor: JoinDescriptor,
         from_table: Alias,
         to_table: Alias,
     ) -> None:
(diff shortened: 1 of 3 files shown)


graphql_compiler/schema/schema_info.py
```diff
@@ -226,56 +226,62 @@ def _create_sql_schema_info(
...
+
+    It describes the tables that correspond to each type (object type or interface type),
+    and gives instructions on how to perform joins for each vertex field. The property fields on
+    each type are implicitly mapped to columns with the same name on the corresponding table.
+
+    Notes:
+    - RootSchemaQuery is a special type that does not need a corresponding table.
+    - Builtin types like __Schema, __Type, etc. don't need corresponding tables.
+    - Builtin fields like _x_count do not need corresponding columns.
+    TODO: This class is essentially the same as SQLSchemaInfo. SQLSchemaInfo is part of an
```
bojanserafimov, line 241:
```suggestion

    TODO: This class is essentially the same as SQLSchemaInfo. SQLSchemaInfo is part of an
```

bojanserafimov (dismissed):
Looks good, one nit.

graphql_compiler/compiler/emit_sql.py
```diff
@@ -632,7 +632,7 @@ def _get_fold_outputs(self) -> List[Label]:
 
     def add_traversal(
         self,
-        join_descriptor: DirectJoinDescriptor,
+        join_descriptor: JoinDescriptor,
```
chewselene, line 635: I thought this seemed more reasonable than adding a type check in the test that was failing, but I'm not very familiar with the composite edge stuff so @bojanserafimov let me know if this doesn't actually make sense!

bojanserafimov: This is not a user-friendly message (which edge uses composite keys?) so we should either make it user friendly, and remove the similar error check in `CompilationState.traverse`, or turn it into an assertion error.

bojanserafimov: Alternatively, just add a comment at the other message explaining why its necessary even though the internal function does the same check: that the external check returns a friendlier message

bojanserafimov: Ideally we'd have all our errors return a text span in the original query pointing at where the error is (so the UI can underline in red), but that's some polish work I don't think we'll get to

bojanserafimov (approved):

Aytac998 (commented):
A

PR #1011, opened 2021-04-21
reorganize query planning and execution

chewselene:
Moving query planning and execution to its own directory (name is lengthy - open to shorter suggestions!). Reorganize `make_query_plan` so that public methods live in the `Public API` section.

Part of the refactor in #1008


obi1kenobi:
You can run `typing_copilot init --overwrite` to bypass the mypy and typing copilot CI blocks, otherwise they'll want you to type-hint all the code you moved, which is getting significantly refactored anyway in an immediate-future PR.

obi1kenobi (dismissed):

obi1kenobi (approved):

PR #1012, opened 2021-04-21
Update Pipfile as temporary measure to unblock work involving compiler

LWprogramming:
The SQLalchemy version bump from 1.3 to 1.4 introduced breaking changes for the compiler and it's turning out to be a trickier problem than expected, so this PR will limit the version number until I figure out how to deal with the changes.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -16,7 +16,6 @@
 
 
 # -- Project information -----------------------------------------------------
-from typing import List
 
 
 project = "GraphQL Compiler"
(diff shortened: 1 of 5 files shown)


docs/source/conf.py
```diff
@@ -17,6 +17,7 @@
 
 # -- Project information -----------------------------------------------------
 
+
```
obi1kenobi, line 20: This seems perhaps unintentional — I think this file is autogenerated by some tool, so let's not edit it spuriously if we can avoid it.
```suggestion
```

graphql_compiler/tests/snapshot_tests/test_cost_estimation.py
```diff
@@ -1786,7 +1786,7 @@ def test_int_value_conversion_uuid(self):
             "80000000-0000-",
         ]
         for uuid_value in invalid_uuid_values:
-            with self.assertRaises(Exception):
+            with self.assertRaises(Union[AssertionError, NotImplementedError]):
```
obi1kenobi, line 1789: I think this should be a tuple rather than a `Union` type: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

obi1kenobi (approved):

PR #1013, opened 2021-04-22
Add query planning typedefs

chewselene:
Adds basic query planning type definitions including `SimpleExecute`. 

Part of the refactor in #1008

### DIFF
--- a/graphql_compiler/query_planning/typedefs.py
+++ b/graphql_compiler/query_planning/typedefs.py
@@ -216,7 +216,7 @@ class BackendType(Enum):
     #       Don't assume they are immutable!
     cypher = "Cypher"
     gremlin = "Gremlin"
-    interpreter = "InterpreterAdapter"
+    interpreter = "interpreter"
     match = "OrientDB MATCH"
     mssql = "MSSQL"
     postgresql = "PostgreSQL"
--- a/graphql_compiler/query_planning/typedefs.py
+++ b/graphql_compiler/query_planning/typedefs.py
@@ -217,7 +217,7 @@ class BackendType(Enum):
     cypher = "Cypher"
     gremlin = "Gremlin"
     interpreter = "InterpreterAdapter"
-    match = "Match"
+    match = "OrientDB MATCH"
     mssql = "MSSQL"
     postgresql = "PostgreSQL"
 
--- a/graphql_compiler/query_planning/typedefs.py
+++ b/graphql_compiler/query_planning/typedefs.py
@@ -1,13 +1,14 @@
 # Copyright 2021-present Kensho Technologies, LLC.
 from abc import ABCMeta
-import copy
 from dataclasses import dataclass, field
-from typing import Any, Callable, Dict, Iterable, Mapping, Optional
+from enum import Enum, unique
+from typing import Any, Callable, Dict, Iterable, Mapping, Optional, cast
 
 from dataclasses_json import DataClassJsonMixin, config
 from graphql import (
     GraphQLList,
     GraphQLNonNull,
+    GraphQLNullableType,
     GraphQLScalarType,
     GraphQLType,
     ListTypeNode,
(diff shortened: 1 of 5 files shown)


graphql_compiler/query_planning/typedefs.py
```diff
@@ -0,0 +1,295 @@
...
+        if inner_type:
+            return GraphQLList(inner_type)
+        else:
+            raise AssertionError(
+                f"Invalid type node. ListTypeNode contained inner type {inner_type}."
+            )
+    elif isinstance(type_node, NonNullTypeNode):
+        inner_type = _type_from_scalar_type_dictionary(scalar_types, type_node.type)
+        if inner_type:
+            return GraphQLNonNull(inner_type)
```
chewselene, line 88: There's a mypy error here that I'm not sure how to fix.

`graphql_compiler/query_planning/typedefs.py:90: error: Value of type variable "GNT" of "GraphQLNonNull" cannot be "GraphQLType"`

Here's the relevant `GraphQLNonNull` [code](https://github.com/graphql-python/graphql-core/blob/main/src/graphql/type/definition.py#L1505). It is expecting a [GraphQLNullableType](https://github.com/graphql-python/graphql-core/blob/main/src/graphql/type/definition.py#L1559), which is a Union of other GraphQL types. I know we have [is_same_type function](https://github.com/kensho-technologies/graphql-compiler/blob/c137fa06c7f583889202068302405e5edb3e681b/graphql_compiler/global_utils.py#L88) for comparing GraphQL types, but I'm not sure how type comparison should work when there is a `Union` of multiple GraphQL types.

graphql_compiler/query_planning/typedefs.py
```diff
@@ -0,0 +1,310 @@
...
+
+    Given a scalar type dictionary and an AST node describing a type, return a GraphQLType
+    definition, which applies to that type. For example, if provided the parsed AST node for
+    `[Date]`, a GraphQLList instance will be returned, containing the type called
+    "Date" found in the scalar type dictionary. If a type called "Date" is not found in the scalar
+    type dictionary, then None will be returned.
+
+    Note: this is very similar to GraphQL's type_from_ast. However, instead of requiring a GraphQL
+    schema this function requires a dictionary of the scalar types. This simplifies deserialization
+    and allows for custom scalar types without constructing an entire schema.
```
chewselene, line 66: copying over this comment on #1008 from @obi1kenobi 
> Are we going to have issues with this approach if a provider wants to add some custom scalar types that the compiler doesn't natively support?
> 
> Not necessarily a deal-breaker, just want to make sure we consider the costs and benefits and explicitly choose the tradeoffs that are best here.

I think that yes, we will have an issue for custom scalar types not in the compiler. However, I'm not sure what the alternative is since we don't want deserialization to be dependent on the schema. Open to suggestions!

graphql_compiler/query_planning/typedefs.py
```diff
@@ -0,0 +1,310 @@
...
+
+@unique
+class BackendType(Enum):
+    # N.B.: The values of the enums are the "human-friendly" display names. Since they are shown
+    #       to humans, they are subject to change if said humans find a friendlier name.
+    #       Don't assume they are immutable!
+    cypher = "Cypher"
+    gremlin = "Gremlin"
+    interpreter = "InterpreterAdapter"
+    match = "Match"
```
obi1kenobi, line 220:
```suggestion
    match = "OrientDB MATCH"
```

graphql_compiler/query_planning/typedefs.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+@unique
+class BackendType(Enum):
+    # N.B.: The values of the enums are the "human-friendly" display names. Since they are shown
+    #       to humans, they are subject to change if said humans find a friendlier name.
+    #       Don't assume they are immutable!
+    cypher = "Cypher"
+    gremlin = "Gremlin"
+    interpreter = "InterpreterAdapter"
```
obi1kenobi, line 219: I feel like the `InterpreterAdapter` is a bit too much of an implementation detail to highlight here. The backend is an interpreter, so maybe let's just use that?
```suggestion
    interpreter = "interpreter"
```

graphql_compiler/query_planning/typedefs.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+    )
+
+    def __eq__(self, other: Any) -> bool:
+        """Check equality between an object and this SimpleExecute."""
+        if not isinstance(other, SimpleExecute):
+            return False
+
+        # Perform special check for input_metadata since GraphQLTypes don't have equality, and
+        # check all other fields in a straight forward manner.
```
obi1kenobi, line 273: Perhaps one word instead of two?
```suggestion
        # check all other fields in a straightforward manner.
```

obi1kenobi (commented):
Looks good, just a couple of nits.

graphql_compiler/query_planning/typedefs.py
obi1kenobi: Let's just document it for now as a limitation of the current approach, and revisit when we need more custom scalar types?

obi1kenobi (approved):
LGTM!

PR #1015, opened 2021-04-27
Add fold post-processing bool to SQLAlchemySchemaInfo

chewselene:
Adding whether or not post-processing should be applied to `SQLAlchemySchemaInfo` based on @obi1kenobi's comment on #1008.

> Another (and perhaps even better) alternative would be to add requires_fold_postprocessing or something like it to the SQLAlchemySchemaInfo. After all, that object knows the dialect (and therefore the emitted SQL) better than anything else, so realistically it should be the one specifying whether post-processing is required or not.

I opted to default to automatic inference of whether or not fold post-processing is required based on `isinstance(dialect, MSDialect)`, which could create a bug if SQLAlchemy adds a MSSQL dialect that does not inherit from `MSDialect`. This can be overruled with the keyword arg `requires_fold_postprocessing` in both `make_sqlalchemy_schema_info` and `get_sqlalchemy_schema_info`. I thought inferring would be more user friendly, but could make explicitly setting `requires_fold_postprocessing` required if there's an argument for that approach.

### DIFF
--- a/graphql_compiler/schema/schema_info.py
+++ b/graphql_compiler/schema/schema_info.py
@@ -11,6 +11,7 @@ from graphql.type.definition import GraphQLInterfaceType, GraphQLObjectType
 import six
 import sqlalchemy
 from sqlalchemy.dialects.mssql import dialect as mssql_dialect
+from sqlalchemy.dialects.mssql.base import MSDialect
 from sqlalchemy.dialects.mysql import dialect as mysql_dialect
 from sqlalchemy.dialects.postgresql import dialect as postgresql_dialect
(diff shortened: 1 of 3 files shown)


LWprogramming:
Is it possible for there to be a case where the dialect is `MSDialect` but we don't want post-processing anyways? If not, it seems like we wouldn't need to have the extra parameter in the function signature and instead derive it inside the function, preventing illegal combinations of parameters from showing up. 

Also, what does 

> if SQLAlchemy adds a `MSSQL` dialect that does not inherit from `MSDialect`

refer to? I thought we explicitly define which dialects are supported and then anything else errors out immediately anyways.

